PR #1, opened 2020-06-19
Various monorepo project tasks

Invertisment:
Contains dummy TS sources for testing purposes.


.editorconfig
```diff
@@ -0,0 +1,10 @@
+# EditorConfig is awesome: https://EditorConfig.org
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_style = space
+indent_size = 2
+insert_final_newline = true
+trim_trailing_whitespace = false
```
robert-zaremba, line 10: - [ ] Please add here:

```
max_line_length = 96
```

package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.0.1",
   "description": "Framework to automate development of Algorand Assets and Smart Contracts.",
   "repository": "https://github.com/scale-it/algorand-builder.git",
-  "author": "Robert Zaremba <https://github.com/robert-zaremba>",
+  "author": "Robert Zaremba <https://github.com/robert-zaremba>, Martynas Maciulevičius <https://github.com/Invertisment>",
```
robert-zaremba, line 6: I think we need to move this things to `contirbutors` https://flaviocopes.com/package-json/#contributors

package.json
```diff
@@ -15,18 +15,27 @@
...
     "chai": "^4.2.0",
     "eslint-plugin-simple-import-sort": "^5.0.3",
     "mocha": "^8.0.1",
+    "nyc": "^14.1.0",
+    "prettier": "2.0.5",
+    "ts-node": "^8.1.0",
+    "tslint": "^5.16.0",
+    "tslint-config-prettier": "^1.18.0",
+    "tslint-consistent-codestyle": "^1.15.1",
+    "tslint-plugin-prettier": "^2.0.1",
```
robert-zaremba, line 38: we don't need `prettier`

Also, later on we will need to remove `tslint`

packages/builder-core/.mocharc.json
```diff
@@ -0,0 +1,13 @@
+{
+  "diff": true,
+  "extension": ["js", "ts"],
+  "package": "./package.json",
+  "reporter": "spec",
+  "slow": 75,
+  "timeout": 2000,
+  "ui": "bdd",
+  "watch-files": ["lib/**/*.js", "test/**/*.js", "lib/**/*.ts", "test/**/*.ts"],
```
robert-zaremba, line 9: shouldn't we have here `src` instead of `lib`?

packages/builder-core/package.json
```diff
@@ -0,0 +1,136 @@
+{
+  "name": "builder-core",
+  "version": "0.0.1",
+  "author": "Robert Zaremba <https://github.com/robert-zaremba>, Martynas Maciulevičius <https://github.com/Invertisment>",
+  "license": "SEE LICENSE IN LICENSE",
```
robert-zaremba, line 5: it should say that license is in root

packages/builder-core/package.json
```diff
@@ -0,0 +1,136 @@
+{
+  "name": "builder-core",
+  "version": "0.0.1",
+  "author": "Robert Zaremba <https://github.com/robert-zaremba>, Martynas Maciulevičius <https://github.com/Invertisment>",
```
robert-zaremba, line 4: let's remove

packages/builder-core/package.json
```diff
@@ -0,0 +1,136 @@
...
+  "license": "SEE LICENSE IN LICENSE",
+  "homepage": "",
+  "repository": "github:scale-it/algorand-builder",
+  "main": "internal/lib/buidler-lib.js",
```
robert-zaremba, line 8: why `internal`?

packages/builder-core/package.json
```diff
@@ -0,0 +1,136 @@
...
+    "smart-contracts",
+    "buidler",
+    "blockchain",
+    "dapps",
+    "javascript",
+    "tooling",
+    "task-runner"
+  ],
+  "bin": {
+    "buidler": "internal/cli/cli.js"
```
robert-zaremba, line 21: this should be called `algob`

packages/builder-core/package.json
robert-zaremba: Also filename is wrong. should be `algob.js` (or .ts)

packages/builder-core/package.json
```diff
@@ -0,0 +1,136 @@
...
+    "raw-body": "^2.4.1",
+    "semver": "^6.3.0",
+    "slash": "^3.0.0",
+    "solc": "0.6.8",
+    "source-map-support": "^0.5.13",
+    "ts-essentials": "^2.0.7",
+    "tsort": "0.0.1",
+    "uuid": "^3.3.2",
+    "ws": "^7.2.1"
+  },
```
robert-zaremba, line 112: we need to remove all ethereum specific things (ethereum, solc, solidity, ws...)

packages/builder-core/src/tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
+  "compilerOptions": {
+    "outDir": "../build/",
+    "rootDirs": ["."],
+    "composite": true
+  },
+  "include": ["./**/*.ts"],
+  "exclude": []
+}
```
robert-zaremba, line 9: let's move this file one level up

packages/builder-core/tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "outDir": "./build-test",
+    "rootDirs": ["./test"]
+  },
+  "include": ["./test/**/*.ts"],
+  "exclude": ["./test/**/buidler.config.ts"],
```
robert-zaremba, line 8: wrong name

tsconfig.json
```diff
@@ -8,7 +8,7 @@
     "allowJs": true,
     "skipLibCheck": true,
     "esModuleInterop": true,
-    "module": "esnext",
+    "module": "commonjs",
```
robert-zaremba, line 11: why changing it?

packages/builder-core/package.json
```diff
@@ -0,0 +1,136 @@
...
+    "src/",
+    "*.d.ts",
+    "*.d.ts.map",
+    "*.js",
+    "*.js.map",
+    "LICENSE",
+    "README.md",
+    "recommended-gitignore.txt",
+    "console.sol"
+  ],
```
robert-zaremba, line 50: - [ ] let's remove the whole `files` section.

packages/builder-core/src/tsconfig.json
Invertisment: I found that linting setup uses these both configs separately. It's possible to run the linter for src and for tests separately. I'm not sure if it's useful right away.

tsconfig.json
Invertisment: This causes these kinds of errors:
```
import { assert } from "chai";
^^^^^^

SyntaxError: Cannot use import statement outside a module```

File: packages/builder-core/package.json [hunk 14]
```diff
@@ -0,0 +1,70 @@
...
+  "dependencies": {
+    "ansi-escapes": "^4.3.0",
+    "chalk": "^2.4.2",
+    "cross-env": "5.0.5",
+    "fs-extra": "^7.0.1",
+    "glob": "^7.1.3",
+    "raw-body": "^2.4.1",
+    "semver": "^6.3.0",
+    "source-map-support": "^0.5.13",
+    "ts-essentials": "^2.0.7"
```
robert-zaremba, line 52: I think 3 above dependencies should be in `devDependencies`

packages/builder-core/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "outDir": "./build",
+    "rootDirs": ["./src"]
```
robert-zaremba, line 5: why do we need to specify `rootDirs`? I think `baseUrl` is preferred.

tsconfig.json
```diff
@@ -8,15 +8,15 @@
...
     "resolveJsonModule": true,
-    "isolatedModules": true,
     "allowSyntheticDefaultImports": true,
     "noImplicitReturns": true,
     "strict": true,
     "forceConsistentCasingInFileNames": true,
     "sourceMap": true,
-    "baseUrl": "src"
+    "baseUrl": "src",
```
robert-zaremba, line 18: maybe removing `baseUrl` from here and moving it to subproject will do the 'job'

packages/builder-core/package.json
Invertisment:
```
  "devDependencies": {
    "@types/chai": "^4.2.0",
    "chai": "^4.2.0",
    "time-require": "^0.1.2",
    "cross-env": "5.0.5",
    "source-map-support": "^0.5.13"
  },
  "dependencies": {
  },
```

packages/builder-core/tsconfig.json
Invertisment: changing to baseUrl: src

package.json
Invertisment: Removed

packages/builder-core/package.json
Invertisment: apache

packages/builder-core/tsconfig.json
Invertisment: Removed. Root project config is extended.

packages/builder-core/src/tsconfig.json
Invertisment: builder-core/tsconfig.json compilation file was changed to compile everything instead of tests

PR #2, opened 2020-06-23
Version command

Invertisment:
`--version` works:
`node build/internal/cli/cli.js --version`


packages/builder/package.json
```diff
@@ -29,16 +29,32 @@
     "test": "node ../../node_modules/mocha/bin/mocha",
     "coverage": "node ../../node_modules/nyc/bin/nyc.js ../../node_modules/mocha/bin/mocha",
     "build": "node ../../node_modules/typescript/bin/tsc --build .",
-    "clean": "node ../../node_modules/rimraf/bin.js builtin-tasks internal *.d.ts *.map *.js tsconfig.tsbuildinfo build"
+    "clean": "node ../../node_modules/rimraf/bin.js builtin-tasks internal *.d.ts *.map *.js tsconfig.tsbuildinfo build",
```
robert-zaremba, line 32: why using a `rimraf`? Can we use `rm`?

packages/builder/src/internal/cli/arguments-parser.ts
```diff
@@ -0,0 +1,278 @@
+import {
```
robert-zaremba, line 1: Why we are not using a package to handle CMD arguments parsing? 
Example: https://www.npmjs.com/package/cli-params

packages/builder/src/internal/context.ts
```diff
@@ -0,0 +1,65 @@
+import { BuidlerRuntimeEnvironment, ConfigExtender } from "../types";
```
robert-zaremba, line 1: lets grep through the project and replace buidler -> builder (with all case combinations).

robert-zaremba: Usually I'm using `sed` for that.

packages/builder/src/internal/constants.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export const SOLC_INPUT_FILENAME = "solc-input.json";
+export const SOLC_OUTPUT_FILENAME = "solc-output.json";
+
+export const BUIDLEREVM_SUPPORTED_HARDFORKS = [
+  "byzantium",
+  "constantinople",
+  "petersburg",
+  "istanbul",
+];
```
robert-zaremba, line 14: everything in this file should be removed or changed

packages/builder/src/internal/lib/buidler-lib.ts
```diff
@@ -0,0 +1,55 @@
...
+    debug.enable("buidler*");
+  }
+
+  const config = loadConfigAndTasks(buidlerArguments);
+
+  // TODO: This is here for backwards compatibility.
+  // There are very few projects using this.
+  if (buidlerArguments.network === undefined) {
+    buidlerArguments.network = config.defaultNetwork;
+  }
```
robert-zaremba, line 43: let's remove this condition. we don't need to keep backwards compatibility

robert-zaremba:
There is lot of solidity / EVM related code, especially in tests. It should be removed.

packages/builder/src/internal/cli/arguments-parser.ts
Invertisment: This is copy paste, not development from scratch.

packages/builder/package.json
Invertisment: It was used by the original project. I had no intention to clean up all of the parts of the project.

packages/builder/src/internal/context.ts
robert-zaremba: Why later? It should be very easy to do. `sed` can handle it perfectly and it probably takes max 15min. No reason to do it later.
Otherwise there are refactoring tools for typescript which should also work with this.

Invertisment: Because it will be easier to review the diff and not a single squashed commit. 
The change ss included in this PR.

packages/builder/src/internal/constants.ts
Invertisment: Removed.

packages/builder/src/internal/lib/buidler-lib.ts
Invertisment: I prefer to leave small fixes like these for later. I think there should be loads of them. And I thought we won't be removing all network-related source code, only EVM.

PR #3, opened 2020-06-25
Rename fixture configuration files, delete eth contract fixtures

Invertisment:


packages/builder/package.json
```diff
@@ -2,16 +2,15 @@
   "name": "@scale-it/builder",
   "version": "0.0.1",
   "author": "SEE package.json from PROJECT ROOT",
-  "contributors": "SEE package.json from PROJECT ROOT",
```
Invertisment, line 5: It expects array. So if author is present contributors are not needed.

PR #4, opened 2020-06-25
Init project in current working directory

Invertisment:


packages/builder/src/internal/cli/project-creation.ts
```diff
@@ -9,9 +9,6 @@ import { getPackageJson, getPackageRoot } from "../util/package-info";
 
 import { emoji } from "./emoji";
 
-const CREATE_SAMPLE_PROJECT_ACTION = "Create a sample project";
-const CREATE_EMPTY_BUILDER_CONFIG_ACTION = "Create an empty builder.config.js";
-
 const SAMPLE_PROJECT_DEPENDENCIES = [
   "@nomiclabs/builder-waffle",
   "ethereum-waffle",
```
robert-zaremba, line 14: `ethereum-waffle` and `builder-waffle` this should be removed

Invertisment: These are printed on startup.

PR #5, opened 2020-06-25
New tasks: Console and help

Invertisment:


packages/builder/src/register.ts
```diff
@@ -0,0 +1,53 @@
+import debug from "debug";
+
```
robert-zaremba, line 2: what is this file for?

packages/builder/src/builtin-tasks/clean.ts
```diff
@@ -1,16 +0,0 @@
...
-export default function () {
-  task(
-    TASK_CLEAN,
-    "Clears the cache and deletes all artifacts",
-    async (_, { config }) => {
-      await fsExtra.remove(config.paths.cache);
-      await fsExtra.remove(config.paths.artifacts);
-    }
-  );
-}
```
robert-zaremba, line 16: why removing this file?

Invertisment: I added this file as a task and currently we don't know what the project structure will be so command `algob cleanup` doesn't make sense.

robert-zaremba: It might be useful. We defined a project structure last week during a meeting. The artifacts are in `/artifacts/
https://paper.dropbox.com/doc/Algorand-builder-architecture--A2ESMswpL3US1ZltRo1W~J7wAg-Vcdp0XNngizChyUWvFXfs

Invertisment: - [ ] Restore

PR #6, opened 2020-06-26
package: update copyright and simplify package.json scripts

robert-zaremba:


packages/builder/package.json
```diff
@@ -23,12 +23,12 @@
...
   "scripts": {
-    "lint:fix": "npm run lint -- --fix",
-    "lint": "npm run lint",
-    "test": "node ../../node_modules/mocha/bin/mocha",
-    "coverage": "node ../../node_modules/nyc/bin/nyc.js ../../node_modules/mocha/bin/mocha",
-    "build": "node ../../node_modules/typescript/bin/tsc --build .",
-    "clean": "node ../../node_modules/rimraf/bin.js builtin-tasks internal *.d.ts *.map *.js tsconfig.tsbuildinfo build",
+    "lint:fix": "lint --fix",
+    "lint": "lint",
+    "test": "mocha",
```
Invertisment, line 28: You force the developer to have all of them in the $PATH

robert-zaremba: No, because this is run in `yarn` / `npm` context.
This means that yarn prioritize the bins in the `node_modules/.bin` which are automatically added to PATH when you use `yarn ...`

Invertisment: If so then it's fine.

PR #8, opened 2020-06-26
Help is default task; Explicit init

Invertisment:


packages/builder/src/builtin-tasks/clean.ts
```diff
@@ -9,6 +9,9 @@ export default function () {
     TASK_CLEAN,
     "Clears the cache and deletes all artifacts",
     async (_, { config }) => {
+      if (!config.paths) {
```
Invertisment, line 12: Unfortunately these checks either have to be performed for each task or there is a need for one more type of a task. Current tasks were designed to be run only with a present config. This allows to run them without the config and a project.

packages/builder/src/types.ts
```diff
@@ -102,29 +102,29 @@ export interface BuilderConfig {
...
+//export interface SolcInput {
+//  settings: {
+//    metadata: { useLiteralContent: boolean };
+//    optimizer: SolcOptimizerConfig;
+//    outputSelection: { "*": { "": string[]; "*": string[] } };
+//    //evmVersion?: string;
+//  };
+//  sources: { [p: string]: { content: string } };
+//  language: string;
+//}
```
robert-zaremba, line 127: let's remove it.

PR #9, opened 2020-06-30
Framework config definition and cleaning.

robert-zaremba:


docs/algob-config.md
```diff
@@ -0,0 +1,51 @@
...
+  /**
+   * Networks define how you connect to your Algorand node. By default the
+   * `development` network is used.
+   */
+  networks: {
+    // You can run an a algorand node a separate terminal, or use `algob` to run and
+    // configure for you a devnet. If `development` network is not specified, `algob`
+    // will use the internal `devnet` configuration for development network.
+    development: mydevnet,
+    devnet: devnet,
```
Invertisment, line 37: Is it a good idea to move out the config in the example? Advanced users will know for themselves.

packages/builder/src/internal/cli/cli.ts
```diff
@@ -132,10 +127,9 @@ async function main() {
       );
     }
 
-    // TODO: This is here for backwards compatibility
-    // There are very few projects using this.
-    if (builderArguments.network === undefined) {
-      builderArguments.network = config.defaultNetwork;
+    if (!builderArguments.network) {
+      // TODO:RZ
+      throw new Error("INTERNAL ERROR. Default network should be registered in `register.ts` module")
```
Invertisment, line 132: You should reuse the pattern of `BuilderError`. It is printed out in console. You can add a message and provide the parameters for formatting.

packages/builder/src/internal/core/config/config-resolution.ts
```diff
@@ -48,9 +48,6 @@ export function resolveConfig(
     ...config,
     paths,
     networks: config.networks!,
-    //solc: config.solc!,
-    defaultNetwork: config.defaultNetwork!,
-    //analytics: config.analytics!,
```
Invertisment, line 53: Analytics is no more?

packages/builder/src/internal/core/config/config-validation.ts
```diff
@@ -102,31 +95,23 @@ const HDAccountsConfig = t.type({
...
-  gasPrice: optional(t.union([t.literal("auto"), t.number])),
-  gasMultiplier: optional(t.number),
   url: optional(t.string),
-  accounts: optional(NetworkConfigAccounts),
+  accounts: optional(NetworkAccounts),
   httpHeaders: optional(HttpHeaders),
 });
 
-const NetworkConfig = t.union([BuilderNetworkConfig, HttpNetworkConfig]);
+const NetworkConfig = t.union([AlgoDevChainConfig, HttpNetworkConfig]);
```
Invertisment, line 114: How is `AlgoDevChainConfig` different from non `*Dev*` config? Previously it was more general.

packages/builder/src/internal/core/runtime-environment.ts
```diff
@@ -58,37 +50,17 @@ export class Environment implements BuilderRuntimeEnvironment {
...
-    //});
-
-    //this.network = {
-    //  name: networkName,
-    //  config: config.networks[networkName],
-    //  provider,
-    //};
-
-    //this.ethereum = provider;
+    // this.provider .... -- creation goes here
```
Invertisment, line 62: I left the comments because it wasn't evident what creation was here. At least we have history.

packages/builder/src/internal/lib/lib.ts
```diff
@@ -36,10 +36,9 @@ if (BuilderContext.isCreated()) {
 
   const config = loadConfigAndTasks(builderArguments);
 
-  // TODO: This is here for backwards compatibility.
-  // There are very few projects using this.
-  if (builderArguments.network === undefined) {
-    builderArguments.network = config.defaultNetwork;
+  if (!builderArguments.network) {
+    // TODO:RZ
+    throw new Error("INTERNAL ERROR. Default network should be registered in `register.ts` module")
```
Invertisment, line 41: Use `BuilderError`

packages/builder/src/types.ts
```diff
@@ -7,55 +7,37 @@ import * as types from "./internal/core/params/argument-types";
...
+  chainName?: string;
+  //from?: string;
 }
 
-export type NetworkConfigAccounts =
-  | "remote"
-  | string[]
-  | HDAccountsConfig
-  | OtherAccountsConfig;
+export interface AlgoDevChainConfig extends CommonNetworkConfig {
```
Invertisment, line 27: Will we have non-Algo and non-Dev configs?

packages/builder/package.json
```diff
@@ -29,7 +29,7 @@
     "coverage": "nyc.js ../../node_modules/mocha/bin/mocha",
     "build": "tsc --build .",
     "clean": "rimraf builtin-tasks internal *.d.ts *.map *.js tsconfig.tsbuildinfo build",
-    "algob": "mkdir -p project && cd project && node ../build/internal/cli/cli.js"
+    "algob:test": "rm -rf project-test && mkdir project-test && cd project-test && node ../build/internal/cli/cli.js init"
```
Invertisment, line 32: You added `init` and decided to force-remove the directory. I made it a general task so that you could run other tasks too. The old command was non-intrusive. I.e. you could run the same task any times you want. If you want to remove the directory you can use `rm -rf && yarn algob init .` -- This will be the same as you did here. And it will be searchable in terminal (`Ctrl+r`), I use a system terminal to do it.

Can you restore the old command and rename `algob:test` to `algob:init`?
We can have two commands if you want to use `init` a lot. I use console command history and I don't have this issue.

docs/algob-config.md
robert-zaremba: Are you talking about proper documentation? We will do a whole docs in the last two sprints.

packages/builder/src/internal/cli/cli.ts
robert-zaremba: I intentionally didn't want it - I want to remove this Errors later: https://www.pivotaltracker.com/story/show/173584552

packages/builder/src/internal/core/config/config-resolution.ts
robert-zaremba: we don't have it.

packages/builder/src/internal/core/config/config-validation.ts
robert-zaremba: BuilderNetwork is a config for a internal chain implementation. We also have this as an extra milestone at the end of the project. So it will require different arguments. It's not a `devnet`

packages/builder/package.json
robert-zaremba: OK, updated.

docs/algob-config.md
Invertisment: I thought that it's simpler not to have variables `mydevnet` and `devnet` and move the config into one blob.

packages/builder/src/types.ts
robert-zaremba: maybe - this is for milestone 7 (optional)
> Lightweight blockchain (simulator) to quickly run and test dapps and smart-contracts without blockchain overhead (network / consensus). This will be a stripped Algorand blockchain written in Go using the Algorand codebase.

packages/builder/src/internal/core/config/config-validation.ts
Invertisment: At the end of the project -- so currently it's not needed to have `Dev` in it's name. It could be `AlgorandNetworkConfig` or name shouldn't change.

packages/builder/src/internal/cli/cli.ts
Invertisment: Ah, it's the `TODO:RZ` comment. We also have the default config somewhere in the project. This config already has some details and you should simply replace the arguments and crash if no network is found.

The user will not know how to recover from deleted config if he accidentally deletes the `default: {}` section.
Do we want the user to recover from this?

robert-zaremba:
BuilderNetworkConfig is even more meaningless.
I can comment it.

robert-zaremba:
No, this is something different - it's about an internal assignment
which I still need to test. I'm moving it to a separate task.

docs/algob-config.md
robert-zaremba: well, it's just a variable. Feel free to rename it in an example.

packages/builder/src/internal/core/config/config-validation.ts
robert-zaremba: The Dev means that it's a not a normal `algorand` node. But  a stripped node. If you find a better name then let me know on telegram and we can rename it

PR #10, opened 2020-06-30
Some test fixes

Invertisment:


packages/builder/.mocharc.json
```diff
@@ -4,7 +4,7 @@
   "package": "./package.json",
   "reporter": "spec",
   "slow": 75,
-  "timeout": 2000,
+  "timeout": 25000,
```
robert-zaremba, line 7: is this 25 sec? This is a lot.

packages/builder/src/builtin-tasks/test.ts
```diff
@@ -0,0 +1,119 @@
...
+  //        chalk.yellow(
+  //          `Failed to generate ${failures} ${pluralize(
+  //            failures,
+  //            "stack trace"
+  //          )}. Run Buidler with --verbose to learn more.`
+  //        )
+  //      );
+  //    }
+  //  );
+}
```
robert-zaremba, line 119: why adding this commented code?

Invertisment: Will remove from this PR.

PR #15, opened 2020-07-01
All tests pass (except from config)

Invertisment:
This passes:
`yarn test  --ignore "test/internal/core/config/*"`

### DIFF
--- a/packages/builder/package.json
+++ b/packages/builder/package.json
@@ -26,7 +26,7 @@
     "lint:fix": "lint --fix",
     "lint": "lint",
     "test": "mocha",
-    "coverage": "nyc.js ../../node_modules/mocha/bin/mocha",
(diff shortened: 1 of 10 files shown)


packages/builder/package.json
```diff
@@ -26,7 +26,7 @@
     "lint:fix": "lint --fix",
     "lint": "lint",
     "test": "mocha",
-    "coverage": "nyc.js ../../node_modules/mocha/bin/mocha",
+    "coverage": "../../node_modules/nyc/bin/nyc.js ../../node_modules/mocha/bin/mocha",
```
robert-zaremba, line 29: remove the `.js` from `nyc.js` then it will work. The original command didn't work, because there is no `nyc.js` in bin. The correct form should be: 

```
"coverage": "nyc ../../node_modules/mocha/bin/mocha",
```

PR #16, opened 2020-07-01
EVM keyword removal

Invertisment:


packages/builder/src/internal/constants.ts
```diff
@@ -1,4 +1,4 @@
 export const ALGOB_NAME = "algob";
 
 export const ALGOB_BIN_NAME = "algob";
-export const ALGOB_CHAIN_NAME = "builderevm";
+export const ALGOB_CHAIN_NAME = "builder_chain_name";
```
robert-zaremba, line 4: rename to `algobnet`

packages/builder/src/internal/core/config/config-validation.ts
```diff
@@ -148,7 +148,7 @@ export function getValidationErrors(config: any): CfgErrors {
   }
 
   // io-ts can get confused if there are errors that it can't understand.
-  // Especially around BuilderEVM's config. It will treat it as an HTTPConfig,
+  // Especially around Builder's network config. It will treat it as an HTTPConfig,
```
robert-zaremba, line 151: Builder -> algob

PR #17, opened 2020-07-01
tests: fix config tests

robert-zaremba:
fixing `test/internal/core/config/config-validation.ts` test.

### DIFF
--- a/packages/builder/test/internal/core/config/config-validation.ts
+++ b/packages/builder/test/internal/core/config/config-validation.ts
@@ -7,96 +7,34 @@ import {
 } from "../../../../src/internal/core/config/config-validation";
 import { ERRORS } from "../../../../src/internal/core/errors-list";
 import { expectBuilderError } from "../../../helpers/errors";
-import CfgErrors from "../../../../src/internal/core/config/config-errors";
 
 describe("Config validation", function () {
-  describe("default network config", function () {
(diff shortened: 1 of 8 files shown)


packages/builder/test/internal/core/config/config-validation.ts
```diff
@@ -10,93 +10,32 @@ import { expectBuilderError } from "../../../helpers/errors";
...
-      expectBuilderError(
-        () => validateConfig(invalidSourcesType),
-        ERRORS.GENERAL.INVALID_CONFIG
-      );
-
-      expectBuilderError(
-        () => validateConfig(invalidTestsType),
-        ERRORS.GENERAL.INVALID_CONFIG
-      );
+    it("Should fail with invalid types (paths)", function () {
```
Invertisment, line 25: So you prefer to not get a stack trace line in the error but get an error from a loop... Is this better?

robert-zaremba: I can add test-case to the error message. too much copy-paste is not good. Have a look at tabular tests. It basically creates a set of test-cases and then do the execution and checks without copy-pasting.

robert-zaremba: I added a failing config to the message.

PR #20, opened 2020-07-02
Config2

robert-zaremba:
Cleaning tasks and updating the sample project.


package.json
```diff
@@ -52,7 +52,7 @@
     "typescript": "^3.9.5"
   },
   "lint-staged": {
-    "*.src.*{ts,js}": [
+    ".*src.*{ts,js}": [
```
Invertisment, line 55: What does this do?

packages/builder/sample-project/algob.config.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+  networks: {
+    localhost: {
```
Invertisment, line 3: No `default`?

package.json
robert-zaremba: there was a bug (dot was in a wrong place).

packages/builder/sample-project/algob.config.js
robert-zaremba: No need for `default`  -- if user is not providing it, it will come from the default config (this is also an original behaviour)

robert-zaremba: The default config is hardcoded in a runtime.

PR #21, opened 2020-07-02
remove ts-node support

robert-zaremba:
I found why the startup is so long of the command. It's due to `ts-node`.
Here is the task: https://www.pivotaltracker.com/story/show/173629836


Invertisment:
I think performance optimization shouldn't be our priority. We don't even know if this will work for the user.
If you want performance we should've chosen C instead of JS.

robert-zaremba:
> I think performance optimization shouldn't be our priority. We don't even know if this will work for the user.
> If you want performance we should've chosen C instead of JS.
> And the user should spend most of his time in smart contract, not in scripts (with which we haven't even started).

It's not simply a performance optimization, but usability, and better user experience. I (and many others) don't like when a tool takes some seconds just to start. Even Java optimizes it.

robert-zaremba:
this is in this branch:
```
yarn run test  20.06s user 0.66s system 169% cpu 12.245 total
```
this is is master:
```
Done in 49.21s.
yarn run test  94.18s user 2.09s system 194% cpu 49.439 total
```
this is in master after removing typescript-support.ts tests:
```
yarn run test  80.86s user 1.75s system 200% cpu 41.247 total
```

There is no test with ts-node in fixtures.

robert-zaremba:
to wrap up:
* there is no evidence about better user experience with ts-node,  (BTW - if user want to use it he can always start with ts-node instead of node and he can run his TS directly!)
* the startup and tests are running considerably slower with ts-node in the tool.
* to help user a setup a TS project I propose to use a TS package template, instead of integrating `ts-node` here (which again, is a huge overhead for everyone).

robert-zaremba:
Also, if needed we can very easily add this support in the future by reverting this commits and do real tests (no we can't do that tests).

packages/builder/src/internal/core/project-structure.ts
```diff
@@ -1,22 +1,14 @@
...
+  return Boolean(findUp.sync(JS_CONFIG_FILENAME));
 }
 
 export function getUserConfigPath(): string | undefined {
-  const tsConfigPath = findUp.sync(TS_CONFIG_FILENAME);
-  if (tsConfigPath) {
-    if (isTypescriptSupported()) {
-      return tsConfigPath;
-    }
+  if (findUp.sync(TS_CONFIG_FILENAME)) {
```
Invertisment, line 11: You don't only disable TS support but crash if the user wants to actually use TS using `ts-node` as executor.

And you'll also crash if the user will want to have an unrelated file `algob.config.ts` that he won't want to load.

I would prefer to: 
1. check if `ts-node` is loaded.
2. look for JS config if it's not loaded. Fail if no config is found.
3. look for TS config if it's loaded.

- [ ]  Fix the crash that will occur when using TS files.

robert-zaremba: We will see if we need a `ts-node` support. For the moment there is no benefit of it (and as already noted it significantly increases the runtime, making the too usage a bit clunky) so we will DROP it. We can simply enable it in the future if needed and users will be able to compare 3 versions.

robert-zaremba: I removed the TS config check with the `throw` statement.

robert-zaremba:
Closing note:
After the release we CAN simply enable TS support if needed in a separate branch and see if it's useful or not.

PR #22, opened 2020-07-03
Cmd network info

robert-zaremba:
Here is the `node-info` command, but it's build on top of `ts-node` branch, so you will see some old commits.


robert-zaremba:
look at the files without conflicts only.

package.json
```diff
@@ -52,7 +52,7 @@
     "typescript": "^3.9.5"
   },
   "lint-staged": {
-    "*.src.*{ts,js}": [
+    ".*src.*{ts,js}": [
```
Invertisment, line 55: How is this different from `*.`?
Will we have `test-src` too? Or something else?
What if we'd have a path `test/mysrcfolder/noRun.js`? It should match.

packages/builder/src/builtin-tasks/node-info.ts
```diff
@@ -0,0 +1,21 @@
...
+export default function (): void {
+  task(TASK_NODE_INFO, "Prints node info and status")
+    .setAction(nodeInfo);
+}
+
+async function nodeInfo(_taskArgs: TaskArguments, env: AlgobRuntimeEnv) {
+  console.log("IN THE TASK", env.network);
+  const algocl = new algosdk.Algod("55e1b3b85b7ca6a755a6a01509fca40bdb52b5dc120da07b9c196ab7d364ff66", "localhost", 8080);
+  const st = await algocl.status();
+  console.log("NETWORK NAME", env.network.name);
```
Invertisment, line 16: Why capitals?

packages/builder/src/builtin-tasks/node-info.ts
```diff
@@ -0,0 +1,21 @@
...
+
+const algosdk = require("algosdk");
+
+export default function (): void {
+  task(TASK_NODE_INFO, "Prints node info and status")
+    .setAction(nodeInfo);
+}
+
+async function nodeInfo(_taskArgs: TaskArguments, env: AlgobRuntimeEnv) {
+  console.log("IN THE TASK", env.network);
```
Invertisment, line 13: Is this a debug comment?

packages/builder/src/builtin-tasks/task-names.ts
```diff
@@ -4,10 +4,12 @@ export const TASK_HELP = "help";
 export const TASK_CONSOLE = "console";
 export const TASK_CLEAN = "clean";
 export const TASK_INIT = "init";
-
+export const TASK_NODE_INFO = "node-info";
```
Invertisment, line 7: Can we simply have `info`? It's a long task name

packages/builder/src/internal/cli/cli.ts
```diff
@@ -140,15 +135,15 @@ async function main() {
...
+    await env.run(taskName, taskArguments);
 
     const tAfterRun = new Date().getTime();  // eslint-disable-line @typescript-eslint/no-unused-vars
     //if (tAfterRun - tBeforeRun > ANALYTICS_SLOW_TASK_THRESHOLD) {
     //  await hitPromise;
     //} else {
     //  abortAnalytics();
     //}
-    log(`Killing Builder after successfully running task ${taskName}`);
+    log(`Quitting algob after successfully running task ${taskName}`);
```
Invertisment, line 146: Exiting.

packages/builder/src/internal/cli/project-creation.ts
```diff
@@ -251,21 +248,21 @@ async function installDependencies(
...
-async function getRecommendedDependenciesInstallationCommand(): Promise<
-  string[]
-> {
+async function npmInstallCmd(): Promise<string[]> {
   const isGlobal =
     getExecutionMode() === ExecutionMode.EXECUTION_MODE_GLOBAL_INSTALLATION;
 
-  if (!isGlobal && (await isYarnProject())) {
-    return ["yarn", "add", "--dev", ...SAMPLE_PROJECT_DEPENDENCIES];
+  if (await isYarnProject()) {
```
Invertisment, line 255: So suddenly we're coupling to yarn?

package.json
robert-zaremba: `*.` is a wrong expression (it doesn't work).

There is a task to enable linting in `tests` in the next sprint (I added it last week).

packages/builder/src/builtin-tasks/node-info.ts
robert-zaremba: yes, the task is not fully done, but the functionality is already there.

packages/builder/src/builtin-tasks/task-names.ts
robert-zaremba: What would `info` mean? Here I want to print the connection and node information.

packages/builder/src/internal/cli/project-creation.ts
robert-zaremba: No, it's still in `if`. I added a support to global installation using `yarn`

packages/builder/src/builtin-tasks/node-info.ts
robert-zaremba: to make it as a section header.

PR #24, opened 2020-07-09
Migration task

Invertisment:
No script file for ordering yet.


packages/builder/src/builtin-tasks/migrate.ts
```diff
@@ -0,0 +1,59 @@
...
+      "directory",
+      "A directory that contains js files to be run within builder's environment",
+      "scripts"
+    )
+    .addFlag("noCompile", "Don't compile before running this task")
+    .setAction(
+      async (
+        { directory, noCompile }: { directory: string; noCompile: boolean },
+        { run, runtimeArgs }
+      ) => {
```
robert-zaremba, line 29: move the content outside a closure. Check out `gen-accounts` in `master` for example.

packages/builder/src/builtin-tasks/migrate.ts
```diff
@@ -0,0 +1,59 @@
...
+
+        if (!fs.existsSync(directory)) {
+          throw new BuilderError(ERRORS.BUILTIN_TASKS.MIGRATE_DIRECTORY_NOT_FOUND, {
+            directory,
+          });
+        }
+
+        const scriptNames = (await glob(path.join(directory, "*.js"), {})).sort()
```
robert-zaremba, line 37: if you move this outside, then you can test it.

packages/builder/src/internal/core/errors-list.ts
```diff
@@ -459,8 +459,24 @@ Please double check your script's path`,
       title: "Error running script",
       description: `Running a script resulted in an error.
 
-Please check Buidler's output for more details.`,
-    }
+Please check Builder's output for more details.`,
+    },
+    MIGRATE_DIRECTORY_NOT_FOUND: {
```
robert-zaremba, line 464: Let's rename: migrate -> scripts

packages/builder/src/internal/core/errors-list.ts
```diff
@@ -449,7 +449,7 @@ Please double check how you invoked algob or run your task.`,
       number: 601,
       message: "Script %script% doesn't exist.",
       title: "Script doesn't exist",
-      description: `Tried to use \`buidler run\` to execut a non-existing script.
+      description: `Tried to use \`algob run\` to execute a non-existing script.
```
robert-zaremba, line 452: :+1:

packages/builder/test/builtin-tasks/migrate.ts
```diff
@@ -0,0 +1,73 @@
...
+import { useFixtureProject } from "../helpers/project";
+import { TASK_CLEAN, TASK_MIGRATE } from "../../src/builtin-tasks/task-names";
+import { AlgobRuntimeEnv } from "../../src/types";
+
+const outputFile = "output.txt"
+
+describe("migration task", function () {
+  useFixtureProject("scripts-dir");
+  useEnvironmentWithBeforeEach(async (algobEnv: AlgobRuntimeEnv) => {
+    await algobEnv.run(TASK_CLEAN, {});
```
robert-zaremba, line 17: async - await is not needed here.

packages/builder/test/helpers/environment.ts
```diff
@@ -1,18 +1,25 @@
...
 }
 
-export function useEnvironment() {
-  beforeEach("Load environment", function () {
+export function useEnvironmentWithBeforeEach(beforeEachFn?: (algobRuntimeEnv: AlgobRuntimeEnv) => PromiseAny) {
+  beforeEach("Load environment", async function () {
     this.env = require("../../src/internal/lib/lib");
+    if (beforeEachFn) {
+      await beforeEachFn(this.env)
+    }
```
robert-zaremba, line 15: here as well, async - await is not needed. Just `return beforeEachFn(...)` - it's a promise already.

packages/builder/test/helpers/environment.ts
```diff
@@ -1,18 +1,25 @@
...
   });
 
   afterEach("reset builder context", function () {
     resetBuilderContext();
   });
 }
+
+export function useEnvironment() {
+  useEnvironmentWithBeforeEach()
+}
```
robert-zaremba, line 25: Is this function needed?

packages/builder/src/builtin-tasks/migrate.ts
Invertisment: But there is no way to test it because the filesystem always sorts them in my case. So either I have to omit the `glob` part because it loads filenames directly or I have to only pass filenames to sorter which is known to work fine.

packages/builder/test/builtin-tasks/migrate.ts
Invertisment: It is needed because `run` is async already. And I want to execute only after `clean` finishes. 
What would be your way of doing it?

packages/builder/test/helpers/environment.ts
Invertisment: It may be removed but I wanted to document what I do with the complex one.
`useEnvironment` is used a lot.

packages/builder/src/builtin-tasks/migrate.ts
Invertisment: A preference. I did what was faster.

robert-zaremba: Let's move it please. It takes less than a minute to move the code to a normal function, and it increases the code readability.

packages/builder/src/builtin-tasks/migrate.ts
robert-zaremba: I don't understand the claim about the filesystem. It doesn't matter if filesystem sorts it or not - because you call a JS sort function at the end. 

So the test is simple - to check if indeed files are sorted with JS

packages/builder/src/internal/core/errors-list.ts
robert-zaremba: - [ ] use `scritps` instead of `migrate`

packages/builder/test/builtin-tasks/migrate.ts
robert-zaremba: Async is a syntactic sugar in JS.
Use this and you will see that it works:
```
  useEnvironmentWithBeforeEach((algobEnv: AlgobRuntimeEnv) => {
    return algobEnv.run(TASK_CLEAN, {});})
```

- [ ] remove async/await when it's not needed.

packages/builder/test/helpers/environment.ts
robert-zaremba: `useEnvironment` and `useEnvironmentWithBeforeEach` is the same thing. let's use one. If you prefer to use `useEnvironment`, then let's use that name and rename useEnvironmentWithBeforeEach to useEnvironment.

packages/builder/src/builtin-tasks/migrate.ts
```diff
@@ -0,0 +1,59 @@
...
+export default function () : void {
+  const log = debug("builder:core:tasks:migrate");
+
+  task(TASK_MIGRATE, "Compiles and runs user-defined scripts from scripts directory")
+    .addPositionalParam(
+      "directory",
+      "A directory that contains js files to be run within builder's environment",
+      "scripts"
+    )
+    .addFlag("noCompile", "Don't compile before running this task")
```
robert-zaremba, line 24: let's remove this flag (and the exception below). It may only produce some bad behaviour. `truffle` for example dosn't have such functionality.

robert-zaremba:
There is a difference between `run` and `migrate`. Also, according to the spec, we don't have a `migrate` command. Please review the spec in the dropbox document.

packages/builder/src/builtin-tasks/run.ts
```diff
@@ -35,23 +56,7 @@ export default function () : void {
...
-          throw new BuilderError(
-            ERRORS.BUILTIN_TASKS.RUN_SCRIPT_ERROR,
-            {
-              script,
-              error: error.message,
-            },
-            error
-          );
-        }
+        await runSingleScript(runtimeArgs, script, log)
```
robert-zaremba, line 59: same here, `noCompile` can be removed.
Also, it should be able to run multiple scripts (please see the spec).

packages/builder/src/builtin-tasks/migrate.ts
```diff
@@ -0,0 +1,59 @@
...
+import { runSingleScript } from "./run";
+
+export default function () : void {
+  const log = debug("builder:core:tasks:migrate");
+
+  task(TASK_MIGRATE, "Compiles and runs user-defined scripts from scripts directory")
+    .addPositionalParam(
+      "directory",
+      "A directory that contains js files to be run within builder's environment",
+      "scripts"
```
robert-zaremba, line 22: According to the spec, the task should be called `deploy` and as an optional parameter, it should take list of files.

packages/builder/src/builtin-tasks/migrate.ts
Invertisment: This is not useful:
```
export type TaskArguments = any;
```

Invertisment: We have TypeScript to support us and this is not useful:
```
 mkAccounts(taskArgs: TaskArguments
```
used as:
```
  const n = taskArgs.n as number;
```
where:
```
export type TaskArguments = any;  // eslint-disable-line @typescript-eslint/no-explicit-any
```

We could code in plain JS then if we throw away the type system.

robert-zaremba: Yes, it's not ideal. This is how the argument parser was orignally designed. I only added the comment to make the linter happy. Changing it would require refactoring.
Look at the source code, they made a comment about it somewhere.

The way how the arguments are handled is by the argument builder pattern: when adding an argument you also mention a type. But this is only checked in the runtime.

packages/builder/src/builtin-tasks/migrate.ts
Invertisment: Sorting order is already checked in the other test because the `output.txt` is populated in a particular way and it's contents are checked.
I added a unit test to check this particular logic. I don't think it was needed because the output test covers this but I added it.

packages/builder/src/builtin-tasks/run.ts
Invertisment: This is `run`.

packages/builder/src/builtin-tasks/migrate.ts
Invertisment: You're trying to make it work as `run`. There is already a task like that.

packages/builder/src/builtin-tasks/migrate.ts
Invertisment: But it's JavaScript. There are not that many choices except that we could parse this by hand or change the args parser itself (which you complained about).

And even if it would be written in Java or Go you'd still check for the type at runtime and parse it from String. So there is no compile-time check in the end.

robert-zaremba: What's your point here? This is not part of this task. This behaviour is inherited from the previous project.

If you want to change how the argument parser is working then you are welcome to add a task in icebox and discuss it there (this could somehow relate to my proposition to use `yargs`). But this is not a priority now, and let me mention that last week (or even 2 weeks ago) you were against changing the parser.

packages/builder/src/builtin-tasks/migrate.ts
Invertisment: Moved to it's own method in the `builtin-tasks/util.ts` file.

packages/builder/src/builtin-tasks/run.ts
Invertisment: `noCompile` removed. `run` currently doesn't support multiple scripts.

packages/builder/src/builtin-tasks/migrate.ts
Invertisment: I was against because it will not have too much benefit. If there would be a clear benefit we could migrate. 
For me it was fine to have strings without parsing. If there will be a lot of number inputs then we'll have to use yargs. 
If it supports complex custom formats then we can have key inputs as key objects directly. Which could be useful.

packages/builder/src/builtin-tasks/run.ts
robert-zaremba: Running multiple scripts sounds already simple to add. If you don't add it here then please add a new task (thought if it's easy let's do it here)

packages/builder/src/builtin-tasks/util.ts
```diff
@@ -0,0 +1,18 @@
...
+export async function getSortedScriptsNoGlob(
+  directory: string,
+  globFn: (pattern: string, params?: any) => Promise<string[]>
+): Promise<string[]> {
+  return (await globFn(directory, {})).sort()
+}
+
+export function getSortedScripts(directory: string): Promise<string[]> {
+  return getSortedScriptsNoGlob(directory, glob)
+}
```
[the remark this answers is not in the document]
Invertisment: You wanted fast tests but what you asked was to use file system. So I didn't.

PR #25, opened 2020-07-13
Run task accepts varargs

Invertisment:


packages/builder/src/internal/core/errors-list.ts
```diff
@@ -452,21 +452,21 @@ Please double check how you invoked algob or run your task.`,
...
   BUILTIN_TASKS: {
-    RUN_FILE_NOT_FOUND: {
+    RUN_FILES_NOT_FOUND: {
       number: 601,
-      message: "Script %script% doesn't exist.",
-      title: "Script doesn't exist",
-      description: `Tried to use \`algob run\` to execute a non-existing script.
+      message: "Scripts don't exist: %scripts%.",
+      title: "Scripts don't exist.",
+      description: `Tried to use \`algob run\` to execute a non-existing script(s).
```
robert-zaremba, line 459: this message doesn't make much sense. A user is already running `algob run`. 
Also, why we are removing a filename from the message?

packages/builder/test/helpers/project.ts
```diff
@@ -54,3 +60,22 @@ export function useFixtureProjectCopy(srcProjectName: string) {
...
+  });
+
+  beforeEach(function () {
+    try {
+      fs.unlinkSync(testFixtureOutputFile)
+    } catch (err) {
+      // ignored
+    }
+  })
+}
```
robert-zaremba, line 81: Isn't this a copy-paste from some task test? (eg TASK_CLEAN)? If yes, then let's remove the function form that test.

Invertisment: I moved this to here from deploy tests.

packages/builder/src/internal/core/errors-list.ts
Invertisment: This message appears when user specifies the scripts that he wants to execute:
`algob run 1 2 3`.
When any of those don't exist they are displayed in this message.
If scripts 2 and 3 don't exist they will be displayed this way:
`Scripts don't exist: 1,2`

robert-zaremba: OK. Good.
We shall abort running all scripts, when any of them doesn't exists.

Invertisment:
> Looks good. Please confirm that we first verify if all scripts exist before starting the execution.

File checks are done only for run. And they are done only before the execution.
There is no point to do it for deploy because user doesn't specify them.

PR #26, opened 2020-07-14
Short parameter for help

Invertisment:
It's not printed in help menu.

### DIFF
--- a/packages/builder/src/internal/cli/arguments-parser.ts
+++ b/packages/builder/src/internal/cli/arguments-parser.ts
@@ -24,6 +24,13 @@ export class ArgumentsParser {
     );
   }
 
+  public static shortParamNameToCLA(paramName?: string): string {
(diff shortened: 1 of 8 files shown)


packages/builder/src/internal/cli/arguments-parser.ts
```diff
@@ -40,8 +42,19 @@ export class ArgumentsParser {
     );
   }
 
+  private _substituteShortParam(arg: string, algobShortParamSubs: AlgobShortParamSubstitutions): string {
```
robert-zaremba, line 45: Why do you use `algob` prefix? It can be use with anything, not only algob. Le's remove it.

Invertisment: I kept what I saw in 
`algobParamDefs: AlgobParamDefinitions,`
Will remove from that one too.

PR #29, opened 2020-07-14
cleaning: removing util module from builtin-tasks

robert-zaremba:
https://www.pivotaltracker.com/n/projects/2452320#


robert-zaremba:
Few notes: 
+ For imports with types only we should use `import type ...`
+ we should prefer `interface A {... }`  over `type A = { ... }`

packages/builder/test/builtin-tasks/deploy.ts
```diff
@@ -1,17 +1,24 @@
...
+import { TASK_DEPLOY } from "../../src/builtin-tasks/task-names";
+import { loadFilenames } from "../../src/builtin-tasks/deploy";
 
 describe("Deploy task", function () {
   useCleanFixtureProject("scripts-dir")
 
+  it("Should load scripts in the right order", function(){
+    const ls = loadFilenames("scripts");
+    let e = ls[0];
+    for (let i = 1; i<ls.length; ++i) {
```
Invertisment, line 16: Wasn't it enought to run `assert.deepEquals`?

robert-zaremba: Amount of lines would be similar, because I would need to copy, sort and then do deep comparison

robert-zaremba:
Thanks for review

PR #31, opened 2020-07-16
Script exec using require

Invertisment:


packages/builder/src/builtin-tasks/console.ts
```diff
@@ -15,36 +16,39 @@ export default function (): void {
...
+        // );
 
-        const nodeArgs = [];
-        if (semver.gte(process.version, "10.0.0")) {
-          nodeArgs.push("--experimental-repl-await");
-        }
+        // const nodeArgs = [];
+        // if (semver.gte(process.version, "10.0.0")) {
+        //  nodeArgs.push("--experimental-repl-await");
+        // }
```
robert-zaremba, line 39: Why this code is commented?
Please use `/*...*/` for commenting out multiple lines of code.
Maybe it's worth to add a note if we need this code or not.

packages/builder/src/builtin-tasks/console.ts
```diff
@@ -15,36 +16,39 @@ export default function (): void {
     .setAction(
       async (
         { noCompile }: { noCompile: boolean }, // eslint-disable-line
-        { config, run, runtimeArgs }           // eslint-disable-line
-
+        runtimeEnv: AlgobRuntimeEnv            // eslint-disable-line
```
robert-zaremba, line 19: `// eslint-disable-line` can be removed here

packages/builder/src/internal/core/errors-list.ts
```diff
@@ -133,6 +133,22 @@ Please [report it](https://github.com/scale-it/algorand-builder/issues/new) to h
       message: "Builder project file was detected: '%clashingFile%'. Move the file or use an empty directory.",
       title: "Directory contains a builder file",
       description: `You are trying to run Builder in a directory that contains builder project file.`
+    },
+    NO_DEFAULT_EXPORT_IN_SCRIPT: {
+      number: 12,
+      message: "Error. '%script%' doesn't have an exported default method.",
```
robert-zaremba, line 139: change method -> function in message

packages/builder/src/internal/core/errors-list.ts
```diff
@@ -484,11 +500,11 @@ Please check your directory`
 
 Please double check your script's path`
     },
-    EXECUTION_ERROR: {
+    SCRIPT_NON_ZERO_RETURN_STATUS: {
       number: 605,
-      message: "Error. Process of '%script%' returned an error status code: %errorStatus%",
-      title: "Error status code while executing a script",
-      description: `Script process returned an error status code.
+      message: "Error. Default method of '%script%' returned an error status code: %errorStatus%",
```
robert-zaremba, line 505: same as above

packages/builder/test/fixture-projects/project-with-scripts/failing-status-script.js
```diff
@@ -0,0 +1,4 @@
+
+export default function (runtimeEnv, deployer, accounts) {
+  return 123;
+}
```
robert-zaremba, line 4: linter doesn't complain about unused arguments? (it should). 
If it will, just add `_` prefix.

packages/builder/test/fixture-projects/project-with-scripts/no-default-method-script.js
```diff
@@ -0,0 +1,3 @@
+
+const fs = require("fs")
+fs.appendFileSync("output.txt", "script with no default method has been loaded");
```
robert-zaremba, line 3: for this kind of scripts let's do something simple (to speed up the tests) instead of loading packages. Eg:
```
1+2
```

packages/builder/test/fixture-projects/project-with-scripts/params-script.js
```diff
@@ -1,8 +1,6 @@
...
-  process.argv[2] !== "a" ||
-  process.argv[3] !== "b" ||
-  process.argv[4] !== "c"
-) {
-  process.exit(12);
+
+const fs = require("fs");
+
+export default async function (runtimeEnv, deployer, accounts) {
+  fs.appendFileSync("output.txt", runtimeEnv.network.name);
```
robert-zaremba, line 5: as above

packages/builder/test/fixture-projects/project-with-scripts/successful-script.js
```diff
@@ -1,3 +1,9 @@
-if (global.config === undefined || global.network === undefined) {
-  process.exit(123123);
+
+const fs = require("fs");
+
+export default async function (runtimeEnv, deployer, accounts) {
+  if (runtimeEnv.config === undefined || runtimeEnv.network === undefined) {
+    throw new Error("Config was not provided");
+  }
+  fs.appendFileSync("output.txt", runtimeEnv.network.name);
```
robert-zaremba, line 8: we can have one script with IO. This one is good.

packages/builder/test/fixture-projects/scripts-dir/failing.js
```diff
@@ -1,3 +1,7 @@
+
 const fs = require('fs')
-fs.appendFileSync("output.txt", "failing scripts: script failed\n");
-process.exit(123);
+
+export default async function (runtimeEnv, deployer, accounts) {
+  fs.appendFileSync("output.txt", "failing script: script executed\n");
+  return 123;
+}
```
robert-zaremba, line 7: why this script is failing?

packages/builder/test/fixture-projects/scripts-dir/scripts/2.js
```diff
@@ -1,6 +1,6 @@
...
-    const fs = require('fs')
-    fs.appendFileSync("output.txt", "scripts directory: script 2 executed\n");
-  }
-}, 100);
+const fs = require('fs')
+
+export default async function (runtimeEnv, deployer, accounts) {
+  await new Promise(resolve => setTimeout(resolve, 100));
+  fs.appendFileSync("output.txt", "scripts directory: script 2 executed\n");
+}
```
robert-zaremba, line 6: as above, do we need to use `fs`?

packages/builder/test/helpers/params.ts
```diff
@@ -0,0 +1,26 @@
+import { AlgobRuntimeEnv, PromiseAny } from "../../src/types";
+
+export function mkAlgoEnv(): AlgobRuntimeEnv {
```
robert-zaremba, line 3: let's rename to `mkAlgobEnv` (notice `b`)

packages/builder/src/builtin-tasks/console.ts
Invertisment: Linter wasn't happy but I thought I should keep the code.
What do you think?

packages/builder/src/builtin-tasks/console.ts
Invertisment: Will remove

packages/builder/src/builtin-tasks/console.ts
robert-zaremba: Not sure if we need it (we can always roll-back). but for commenting section of code we should use `/*...*/` because otherwise we need to think twice if there are other changes or not.

packages/builder/test/fixture-projects/project-with-scripts/failing-status-script.js
Invertisment: It didn't say anything. Also this is in tests, so they may not be checked.

packages/builder/test/fixture-projects/scripts-dir/failing.js
Invertisment: This script returns a status 123 so I don't use `process.exit(123)` but instead expect the function to return a number.
This will allow the developers to have custom statuses in their CI builds.

packages/builder/test/fixture-projects/scripts-dir/scripts/2.js
Invertisment: I check the output of the script this way. 
There is no easy way to capture stdout.

packages/builder/test/helpers/params.ts
Invertisment: To what?

packages/builder/test/fixture-projects/scripts-dir/scripts/2.js
robert-zaremba: It's good to have one tests with `fs`. 
For other tests - isn't it easier to check the return value?

packages/builder/src/builtin-tasks/console.ts
Invertisment: I think about removing `console.ts` completely

packages/builder/test/helpers/params.ts
robert-zaremba: you don't have `b` in `mkAlgoEnv` name

packages/builder/src/builtin-tasks/console.ts
robert-zaremba: I don't think we should spend much time on it now. If it passes tests and compiles then let's leave.

packages/builder/test/fixture-projects/project-with-scripts/successful-script.js
Invertisment: It's easier to ensure script execution order through IO.
I'm not sure what can be a different good way to do it.

packages/builder/test/fixture-projects/scripts-dir/scripts/2.js
Invertisment: When `run.ts` executes multiple scripts there are no return values.
Scripts either pass or fail -- all or some of them.
So there is no way to ensure if by requesting to run scripts 1,2,3 it ran 1st or only the 3rd one.

packages/builder/test/helpers/params.ts
Invertisment: ok

packages/builder/test/fixture-projects/scripts-dir/scripts/2.js
robert-zaremba: id you run them in the same process, then they can return a value (eg `7`)

Invertisment: But this reduces ability to reason about what they do.

PR #32, opened 2020-07-17
Migrations with snapshots

Invertisment:


README.md
```diff
@@ -26,6 +26,12 @@ In algob, scripts are JS files stored in the `scripts` directory. They are used
 
 Please see the [architecture document](https://paper.dropbox.com/doc/Algorand-builder-architecture--A3aVSVEt3HIRGIiCnTMbn64DAg-Vcdp0XNngizChyUWvFXfs#:uid=213683005476107006060621) to see how the scripts are organized and working.
 
+### Script exception
+
+Scripts can be executed in two ways:
+- `algob run` -- Runs provided scripts by not saving any script state snapshots.
+- `algob deploy` -- Conditionally runs scripts by checking whether previous runs succeeded. Checkpoints can be overridden by using `--force` to run the scripts.
```
robert-zaremba, line 33: This is not a correct explanation. You only explained here one of the outcomes.

`deploy` provides deployer and should be used to deploy assets. Moreover it creates checkpoints when the script is run and prohibits running the script second time if the checkpoint exists in the given network.

packages/builder/sample-project/scripts/0-sampleScript.js
```diff
@@ -1 +1,6 @@
-console.log("Hello world!")
+
+function run(runtimeEnv, deployer, accounts) {
+  console.log("Sample script has been executed!")
+}
```
robert-zaremba, line 4: if this is an `algob run` script, than deployer is not provided (this has been written in the spec more than a week ago)

packages/builder/src/builtin-tasks/run.ts
```diff
@@ -17,17 +22,29 @@ function filterNonExistent (scripts: string[]): string[] {
 }
 
 export async function runMultipleScripts (runtimeEnv: AlgobRuntimeEnv,
-  scriptNames: string[],
-  log: (...args: unknown[]) => unknown): Promise<void> {
+                                          scriptNames: string[],
+                                          log: (...args: unknown[]) => unknown,
```
robert-zaremba, line 26: Why passing `log` here? If it's for tests, than probably it's better to use `debug` module.

Also, the function should return `void`. I prefer to use `any` instead of `unknown` in this context, because we know that it can be any argument.

packages/builder/src/builtin-tasks/run.ts
```diff
@@ -17,17 +22,29 @@ function filterNonExistent (scripts: string[]): string[] {
...
+                                          persistSnapshots: boolean,
```
robert-zaremba, line 27: we are not doing a `snaphost` --> snapshot means that we serialize the current state. We do a checkpoint here.
 So let's rename to `doCheckpoint`

packages/builder/src/builtin-tasks/run.ts
```diff
@@ -17,17 +22,29 @@ function filterNonExistent (scripts: string[]): string[] {
...
       scriptLocation,
       runtimeEnv
     );
+    if (persistSnapshots) {
+      persistCheckpoint(
+        scriptLocation,
+        appendEnv(
+          currentCP,
+          runtimeEnv))
+    }
```
robert-zaremba, line 47: We need to have a checkpoint class - and use it's object here. We will need to pass it to the deployed `script` as an argument - because user have to be able to add custom metadata.

robert-zaremba: ah, now I see that we are in `algob run`. It would be still good to have same functionality in both `run` and `deploy`. In `run` we will not pass the checkopoint object to  the user script as an argument.

packages/builder/src/internal/util/script-checkpoints.ts
```diff
@@ -0,0 +1,53 @@
+
```
robert-zaremba, line 1: Please move this file to `src/lib/`

packages/builder/src/internal/util/script-checkpoints.ts
```diff
@@ -0,0 +1,53 @@
...
+  )
+}
+
+export function loadCheckpoint(scriptName: string): ScriptCheckpoint {
+  const scriptPath = toCheckpointFileName(scriptName)
+  if (!fs.existsSync(scriptPath)) {
+    return {}
+  }
+  return YAML.parse(fs.readFileSync(scriptPath).toString())
+}
```
robert-zaremba, line 53: As explained above, we should have a class. I propose that the class has following methods:

* `constructor` -- will get information about the script file etc... 
* `_assertNew`  -- will verify that the checkpoint doesn't exist.
* `addMetadata` -- will allow user to add more info (feel free to propose other name)
* `_save` -- will save all information to the checkpoint file.
* `_addTransactionID`
* .... maybe some other methods

packages/builder/src/builtin-tasks/run.ts
Invertisment: This is only because linter complains.

packages/builder/src/builtin-tasks/run.ts
Invertisment: Is this simply a rename of `persistSnapshots` to `doCheckpoint`?

packages/builder/src/builtin-tasks/run.ts
Invertisment: Do you still want to have a class to handle this?

README.md
Invertisment: put into docs.

packages/builder/sample-project/scripts/0-sampleScript.js
Invertisment: Deployer is last param and can be `null`

Invertisment:
Add a constraint to deploy only from scripts

packages/builder/src/builtin-tasks/run.ts
robert-zaremba: for `any` - there is a reason for that. You can mute it with appropriate comment (eslint-disable ....)

robert-zaremba: So why we are using `log` as a function argument?

packages/builder/src/builtin-tasks/run.ts
robert-zaremba: yes

packages/builder/src/builtin-tasks/run.ts
robert-zaremba: I think it's better to have a class.

packages/builder/src/builtin-tasks/run.ts
Invertisment: This is because `deploy` and `run` tasks need to log during verbose mode. It's a very minor thing, log has a tag so it preserves that tag.

packages/builder/src/internal/util/script-checkpoints.ts
Invertisment: Um.. ok. I also added `Deployer` class to this file.

packages/builder/src/builtin-tasks/run.ts
Invertisment: Tasks create the method themselves. Log doesn't need to be passed as a parameter anymore.

packages/builder/src/builtin-tasks/run.ts
Invertisment: Created a `Deployer` class and this logic gets handled inside of each task.

packages/builder/src/builtin-tasks/run.ts
Invertisment: Created a `Deployer` class, this class manages how checkpoints are mutated through multiple executions.

packages/builder/src/internal/util/script-checkpoints.ts
Invertisment: Deployer class will be able to store these fields. As we don't need to use them I only added methods `putMetadata` and `getMetadata`.

README.md
```diff
@@ -20,11 +20,7 @@ To attract more web developers we plan to build a JavaScript DSL for TEAL with T
 
 # Usage
 
-## Scripts
-
-In algob, scripts are JS files stored in the `scripts` directory. They are used to manage migrations and deployments.
-
-Please see the [architecture document](https://paper.dropbox.com/doc/Algorand-builder-architecture--A3aVSVEt3HIRGIiCnTMbn64DAg-Vcdp0XNngizChyUWvFXfs#:uid=213683005476107006060621) to see how the scripts are organized and working.
+See [user script documentation](docs/user-script-execution.md) for info about custom scripts and migrations.
```
robert-zaremba, line 23: let's change migrations -> deployments.

docs/user-script-execution.md
```diff
@@ -0,0 +1,28 @@
...
+Please see the [architecture document](https://paper.dropbox.com/doc/Algorand-builder-architecture--A3aVSVEt3HIRGIiCnTMbn64DAg-Vcdp0XNngizChyUWvFXfs#:uid=213683005476107006060621) to see how the scripts are organized and how to use them.
+
+### Script execution
+
+Scripts can be executed in two ways:
+- `algob run`
+- `algob deploy`
+
+#### Run
+Runs provided scripts by not saving any script state snapshots. Useful to query the current state of blockchain.
```
robert-zaremba, line 18: we don't do snapshot. We are doing checkpoints.
Snapshot is a saved state of the system.

docs/user-script-execution.md
```diff
@@ -0,0 +1,28 @@
...
+
+#### Deploy
+Runs scripts and stops if any of them fails.
+Any script that throws an exception will stop the execution and script state won't be saved.
+During next execution previous scripts will be skipped (results will be cached) and the task will start with the failed script.
```
robert-zaremba, line 23: and the task will start FROM (change "with" -> "from")

packages/builder/src/types.ts
```diff
@@ -207,6 +207,15 @@ export interface TaskDefinition extends ConfigurableTaskDefinition {
...
 }
 
+export interface ScriptCheckpoint {
+  [network: string]: ScriptNetCheckpoint
+};
+
+export interface ScriptNetCheckpoint {
+  timestamp: number
+  metadata: {[key: string]: string}
+};
```
robert-zaremba, line 217: No need to create `interfaces`. Let's define a `class` then, in typescripts, this will automatically becomes a type

packages/builder/test/fixture-projects/project-with-scripts/scripts/successful-script.js
```diff
@@ -1,9 +1,11 @@
...
 
-export default async function (runtimeEnv, deployer, accounts) {
+async function run(runtimeEnv, accounts, deployer) {
   if (runtimeEnv.config === undefined || runtimeEnv.network === undefined) {
     throw new Error("Config was not provided");
   }
   fs.appendFileSync("output.txt", runtimeEnv.network.name);
 }
+
+module.exports = { default: run }
```
robert-zaremba, line 11: Why using `module.exports` rather than `export default`?

packages/builder/src/builtin-tasks/deploy.ts
```diff
@@ -28,24 +35,43 @@ export function loadFilenames (directory: string): string[] {
...
+      return;
+    }
+    log(`Running script ${relativeScriptPath}`);
+    await runScript(
+      relativeScriptPath,
+      runtimeEnv,
+      deployer.appendCheckpoints(currentCP)
+    );
+    persistCheckpoint(relativeScriptPath, deployer.checkpoints);
+  });
```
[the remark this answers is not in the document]
Invertisment: This avoids use of `log: () => unknown` completely.

README.md
Invertisment: Changed

docs/user-script-execution.md
Invertisment: updated

docs/user-script-execution.md
Invertisment: changed

packages/builder/src/lib/files.ts
```diff
@@ -0,0 +1,21 @@
+import path from "path";
+
+import { BuilderError } from "../internal/core/errors";
+import { ERRORS } from "../internal/core/errors-list";
+import { scriptsDirectory } from "../lib/script-checkpoints";
+
+export function checkRelativePaths (scriptNames: string[]): string[] {
+  const nonScriptPaths = [];
+  for (let i = 0; i < scriptNames.length; i++) {
+    const scriptName = scriptNames[i];
```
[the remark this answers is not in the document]
Invertisment: Updated

packages/builder/src/lib/script-checkpoints.ts
```diff
@@ -0,0 +1,105 @@
...
+    YAML.stringify(checkpoint)
+  );
+}
+
+export function loadCheckpoint (scriptName: string): ScriptCheckpoint {
+  const scriptPath = toCheckpointFileName(scriptName);
+  if (!fs.existsSync(scriptPath)) {
+    return {};
+  }
+  return YAML.parse(fs.readFileSync(scriptPath).toString());
```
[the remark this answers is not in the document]
Invertisment: updated

packages/builder/test/fixture-projects/project-with-scripts/scripts/successful-script.js
Invertisment: It doesn't work the other way. This is plain JS, without any syntax additions.

packages/builder/src/lib/script-checkpoints.ts
```diff
@@ -0,0 +1,105 @@
...
+
+export function mergeCheckpoints (
+  prev: ScriptCheckpoint,
+  curr: ScriptCheckpoint
+): ScriptCheckpoint {
+  const keys: string[] = Object.keys(curr);
+  return keys.reduce((out: ScriptCheckpoint, key: string) => {
+    return appendToCheckpoint(out, key, curr[key]);
+  }, prev);
+}
```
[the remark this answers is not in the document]
Invertisment: It says that if I use an indexed type I can't have a member that returns a different kind of value.
i.e. I can't do this in the same type:
```
  [network: string]: ScriptNetCheckpoint

  mergeCheckpoints (
    prev: ScriptCheckpoint,
    curr: ScriptCheckpoint
  ) => ScriptCheckpoint
```

packages/builder/src/types.ts
Invertisment: This one can have a class but the main type can't

packages/builder/src/lib/script-checkpoints.ts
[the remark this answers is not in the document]
Invertisment: Will not change into class

packages/builder/src/lib/script-checkpoints.ts
```diff
@@ -0,0 +1,105 @@
...
+  const scriptPath = toCheckpointFileName(scriptName);
+  const scriptDir = path.dirname(scriptPath);
+  if (!fs.existsSync(scriptDir)) {
+    fs.mkdirSync(scriptDir, { recursive: true });
+  }
+  fs.writeFileSync(
+    scriptPath,
+    YAML.stringify(checkpoint)
+  );
+}
```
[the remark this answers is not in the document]
Invertisment: Will not change into class

PR #34, opened 2020-07-27
Load keys from account files.

robert-zaremba:
+ update config file
+ add account file parser
+ updated sample project
+ export package lib functions (now we are able to easily import algob functions
  in external packages)


docs/algob-config.md
```diff
@@ -13,13 +14,15 @@ A network object can specify following entries:
 A special network named `algobchain` can specify the `AlgobChain` configuration:
 
 ```
-  accounts?: NetworkAccounts;
+  accounts_file?: string;
```
Invertisment, line 17: We already have dashes and camelCaseNames. Do we want to have one more?

packages/builder/package.json
```diff
@@ -16,6 +15,7 @@
     "tooling",
     "task-runner"
   ],
+  "main": "build/index.js",
```
Invertisment, line 18: You export a single function -- `mkAccounts` whereas `lib.ts` exoprts many things. Could you move `mkAccounts` to `lib.ts`?

packages/builder/sample-project/algob.config.js
```diff
@@ -1,9 +1,20 @@
+import { mkAccounts } from "algob";
+
+let accounts = mkAccounts([{
+  addr: "KFMPC5QWM3SC54X7UWUW6OSDOIT3H3YA5UOCUAE2ABERXYSKZS5Q3X5IZY",
```
Invertisment, line 4: I don't think any accounts should be hardcoded in the sample project. 
If you think there should already exist an account we can crash the tool if there are no accounts during the run.

packages/builder/sample-project/algob.config.js
```diff
@@ -1,9 +1,20 @@
...
+  mnemonic: "call boy rubber fashion arch day capable one sweet skate outside purse six early learn tuition eagle love breeze pizza loud today popular able divide"
+}]);
+
+let defaultCfg = {
+  host: "http://localhost",
+  port: 8080,
+  token: "$ALGORAND_DATA/algod.token",
```
Invertisment, line 11: Do you expect that every developer will have this environment variable?

docs/algob-config.md
Invertisment: Or you may want to use `ACCOUNTS_FILE` instead.

packages/builder/src/index.ts
```diff
@@ -0,0 +1,3 @@
+import { mkAccounts } from "./lib/account";
+
+export { mkAccounts };
```
Invertisment, line 3: `lib.ts` already exports several variables. Why a new entry point?

packages/builder/src/internal/core/config/default-config.ts
```diff
@@ -1,22 +1,16 @@
...
 
 const defaultConfig: AlgobConfig = {
   networks: {
-    default: DEFAULT_NETWORK_CONFIG,
-    [ALGOB_CHAIN_NAME]: {
-      chainName: ALGOB_CHAIN_NAME,
-      throwOnTransactionFailures: true,
-      throwOnCallFailures: true
-    }
+    [ALGOB_CHAIN_NAME]: cfg
```
Invertisment, line 13: This looks like it's an array. Can you remove square brackets?
It's not a type.

packages/builder/src/lib/account.ts
```diff
@@ -0,0 +1,46 @@
...
+
+import { BuilderError } from "../internal/core/errors";
+import { ERRORS } from "../internal/core/errors-list";
+import type { AccountDef, HDAccount, MnemonicAccount } from "../types";
+
+export function mkAccounts (input: AccountDef[]): Account[] {
+  const accounts: Account[] = [];
+  for (const i of input) {
+    if ((i as HDAccount).path) {
+      throw new BuilderError(ERRORS.ACCOUNT.HD_ACCOUNT, {});
```
Invertisment, line 13: Can you add a pubkey/path to the error object and message?

packages/builder/src/lib/account.ts
```diff
@@ -0,0 +1,46 @@
...
+    } else if ((i as Account).sk) {
+      accounts.push(i as Account);
+    } else {
+      const ia = i as MnemonicAccount;
+      const a = parseMnemonic(ia.mnemonic);
+      if (a.addr !== ia.addr && ia.addr !== "") {
```
Invertisment, line 19: Two layers of braces are already a limit for me. Can we remove the `else { __ }` wrapper or split into methods?
You simply match for object type.
Isn't it possible to use `switch` here?
Or `if+continue` without `else`? This would result in less braces.

packages/builder/test/internal/core/config/fixtures-account.ts
```diff
@@ -0,0 +1,6 @@
+import { Account, mnemonicToSecretKey } from "algosdk";
```
Invertisment, line 1: We have our test projects in a separate directory. I think it's a good idea to separate mock objects too.

tsconfig-project.json
```diff
@@ -4,7 +4,7 @@
     "outDir": "./build"
   },
   "include": [
-    "src",
+    "src", "test",
```
Invertisment, line 7: I tried this. It overwrites the sources if they are named the same way.

docs/algob-config.md
robert-zaremba: This is old, we are not using `accounts_file` any more. I will need to change it.

packages/builder/package.json
robert-zaremba: `internal/lib/lib.ts`?  - it doesn't export any function. It exports only `env`. This is not what we want to be imported by external applications.

packages/builder/sample-project/algob.config.js
robert-zaremba: I was thinking about it. It helps for testing and bootstraping, and also shows an example. 
In any case we will have examples, and some less savy users can copy paste an account from example and use it.

I propose to add the following comment:
```
// NOTE: this is only an example. DON'T this account in any working environment because
// everyone can check it and use the private key (this account is visible to everyone).
```

packages/builder/sample-project/algob.config.js
robert-zaremba: no, but coding any other path makes less sense.

packages/builder/src/index.ts
robert-zaremba: duplicated, I responded above. Please re-respond there.

packages/builder/src/lib/account.ts
robert-zaremba: OK

packages/builder/src/lib/account.ts
robert-zaremba: `switch` doesn't work here. I think `if` is the idiomatic way. 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch

robert-zaremba: Check out this files for example, `if else ...` chain is used pretty often: 
+ `algosdk/src/transaction.js`
+ lodash.js
....

packages/builder/test/internal/core/config/fixtures-account.ts
robert-zaremba: OK, I will create a `mock` subdirectory.

tsconfig-project.json
robert-zaremba: Yes, I've already removed it but didn't commit ( I used it to opimize some tests)

packages/builder/package.json
Invertisment: Then it would be best to remove or refactor it.

packages/builder/sample-project/algob.config.js
Invertisment: What about a relative path to a file? What does this file contain? Isn't it possible to have a variable as for `accounts` array?

packages/builder/src/lib/account.ts
Invertisment: It could help to have `return method( __ )` for nested methods like these.
If you think it's a good way I simply complained about the nested `if`. If you think it's ok then we can have it.

packages/builder/sample-project/algob.config.js
Invertisment: I also don't like the idea of having these accounts in the main config because it will end up in code.
For me it looks like a good design would be to have `algob.config.local.js` to hold this kind of data.

We can have the comment for now. We may need to create a separate file for local user config.

Tool crash will also work if the accounts will get deleted. Helpful info should be added there anyway.

packages/builder/package.json
robert-zaremba: No, because it's used in scripts and in the `console`.

Moreover there is another task to clean up the `lib.ts` and we even discussed yesterday.

packages/builder/sample-project/algob.config.js
robert-zaremba: I want to make it possible for a user to have a quick start. that's why we also have a dev-net in the backlog.

robert-zaremba: Re `....local.js` - what's the benefit of it, the normal config will be there anyway. And there is other function - which I created - to load accounts from a file. And this is what users should use to not commit config.

robert-zaremba: Or accounts from `ENV`

packages/builder/sample-project/algob.config.js
robert-zaremba: this is something different. It's not related to accounts. It's needed for the client authorization to the node. 
PS: How did you connect to a private node without accessing / looking at the token file (that was one of the first tasks).

other solution is `<path to your token file>/algod.token` - which is not better.

packages/builder/package.json
Invertisment: ok

packages/builder/sample-project/algob.config.js
Invertisment: ok

packages/builder/src/internal/core/config/default-config.ts
robert-zaremba: No, it's not an array, and it won't work without the brackes. It's a language feature, special syntax to get a name from a variable.

packages/builder/sample-project/algob.config.js
Invertisment: I don't remember, I simply used `goal` to run everything. So it should've handled that itself.

packages/builder/src/lib/account.ts
robert-zaremba: I agree - if the function will be bigger, currently it's only few lines of code only for the one case.

robert-zaremba: anyway - adding new method to and using it to do simple push:

```
    } else {
      accounts.push(fromMnemonic(i as MnemonicAccount));
    }
```

PR #36, opened 2020-07-29
Run lint+tests on push, block merge

Invertisment:


.github/workflows/test-after-push.yaml
```diff
@@ -0,0 +1,19 @@
+name: Run tests on push
+on: push
+jobs:
+  tests:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
```
robert-zaremba, line 8: what is `checkout@v1`? And why there is `@1`?

Invertisment: This checks out the repository. It didn't work without this line. 
This is what happens when this checkout is not executed:
https://github.com/scale-it/algorand-builder/runs/922469012?check_suite_focus=true

Invertisment: This is what it does:
https://github.com/marketplace/actions/checkout

robert-zaremba: let's update to `v2` if possible.

Invertisment: updated

PR #37, opened 2020-07-29
Linter config for tests

Invertisment:
This is only the config. Actual change will be huge.
`eslint` differentiates between `src` and `test` directories and it fixes sources from both of them. They have to have a `tsconfig.json` file to do it.


packages/builder/test/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "../tsconfig.json",
+  "include": [
+    "../src",
```
Invertisment, line 4: Eslint does check both sources. So this config works as long as it isn't used to compile the files.

packages/builder/test/tsconfig.json
```diff
@@ -0,0 +1,8 @@
...
+    "../../@types",
+    "../test"
```
Invertisment, line 6: Eslint does only check that sources are included into the project.
So this config works as long as it isn't used to compile the files.

robert-zaremba: so, maybe it will be easier to mute that check in `eslintrc`? Because now you are adding all files, so obviously all are included in the project, so the rule doesn't make sense any more.

Invertisment: This is not what I wanted to address.
Eslint complains if `test` is not added to `tsconfig.json`. 
And compiler overrides the sources if it finds the same ones in `test/`.
So if we have this separate `tsconfig.json` file and it's only used by eslint then everything works:
- Compiler compiles as before
- Eslint can check both: `src` and `tests`.

robert-zaremba: OK, but probably we will rollback it in the future once we will support `tests` compilation.

robert-zaremba: BTW, instead of writing `../test` we should write `./`

PR #38, opened 2020-08-03
Extended deployer

Invertisment:


packages/builder/src/builtin-tasks/deploy.ts
```diff
@@ -35,8 +32,18 @@ export function loadFilenames (directory: string): string[] {
...
+    try {
+      fs.unlinkSync(toCheckpointFileName(scriptName));
+    } catch (e) {
+      // ignored
+    }
+  }
+}
+
 async function doDeploy ({ fileNames, force }: TaskArgs, runtimeEnv: AlgobRuntimeEnv): Promise<void> {
-  const log = debug("builder:core:tasks:deploy");
```
robert-zaremba, line 39: let's have something shorter. Not sure why we need `core` and `tasks`. Also, the tool name is `algob`.
Let's use: `algob:deploy`

packages/builder/src/builtin-tasks/run.ts
```diff
@@ -17,23 +23,40 @@ function filterNonExistent (scripts: string[]): string[] {
...
-  runScriptFn: (
-    relativeScriptPath: string,
-    runtimeEnv: AlgobRuntimeEnv
-  ) => Promise<void>): Promise<void> {
+  onSuccessFn: (cpData: CheckpointData, relativeScriptPath: string) => void,
+  force: boolean,
+  logTag: string,
+  wrapDeployer: (orig: AlgobDeployer) => AlgobDeployer): Promise<void> {
+  const log = debug(logTag);
+  const cpData: CheckpointData = loadCheckpointsRecursive();
```
robert-zaremba, line 34: no need to add a type here. The function has a clear return type

packages/builder/src/lib/script-checkpoints.ts
```diff
@@ -1,105 +1,310 @@
...
+};
+
+function findCheckpointsRecursive (): string[] {
+  const checkpointsPath = path.join(".", artifactsPath, scriptsDirectory);
+  fs.mkdirSync(checkpointsPath, { recursive: true });
+  return walk(checkpointsPath)
+    .filter(filename => filename.endsWith(checkpointFileSuffix));
+}
+
+export function loadCheckpointsRecursive (): CheckpointData {
```
robert-zaremba, line 196: I'm not sure if `recuriseve` is a right suffix here (and in the function above).

We don't do an arbitrary recursion. We load all checkpoints. So, how about using suffix `All`?

packages/builder/src/builtin-tasks/run.ts
```diff
@@ -17,23 +23,40 @@ function filterNonExistent (scripts: string[]): string[] {
...
+  const deployer: AlgobDeployer = wrapDeployer(new AlgobDeployerImpl(runtimeEnv, cpData));
```
robert-zaremba, line 35: why we need a function here? This looks very hacky.

robert-zaremba: I think that instead of wrapping, we should have a builder method which will construct right `Deployer`, Example:
```
mkDeployer(runtimeEnv, checkpoints, readOnly)
```

packages/builder/src/types.ts
```diff
@@ -284,12 +275,70 @@ export interface LinkReferences {
...
+};
+
+export interface ScriptNetCheckpoint {
+  timestamp: number
+  metadata: {[key: string]: string}
+  asa: {[name: string]: ASAInfo}
+  asc: {[name: string]: ASCInfo}
+};
+
 export interface AlgobDeployer {
```
robert-zaremba, line 328: let's remove `Algob` prefix.

packages/builder/src/lib/script-checkpoints.ts
```diff
@@ -1,105 +1,310 @@
...
+        ERRORS.BUILTIN_TASKS.CHECKPOINT_ERROR_DUPLICATE_ASSET_DEFINITION,
+        { assetName: assetName });
+    }
+  }
+  orig.asa = Object.assign(
+    {}, orig.asa, append.asa
+  );
+  orig.asc = Object.assign(
+    {}, orig.asc, append.asc
+  );
```
robert-zaremba, line 79: how about doing it in one line?

packages/builder/src/lib/script-checkpoints.ts
```diff
@@ -1,105 +1,310 @@
...
+  const allAssetNames = Object.keys(append.asa).concat(Object.keys(append.asc));
+  for (const assetName of allAssetNames) {
+    if ((orig.asa[assetName] && !deepEqual(orig.asa[assetName], append.asa[assetName])) ||
+      (orig.asc[assetName] && !deepEqual(orig.asc[assetName], append.asc[assetName]))) {
+      throw new BuilderError(
+        ERRORS.BUILTIN_TASKS.CHECKPOINT_ERROR_DUPLICATE_ASSET_DEFINITION,
+        { assetName: assetName });
+    }
+  }
```
robert-zaremba, line 73: The asset duplicate condition should be only with name -- we shouldn't do `deepEqual` check (we can remove that dependency). Only name must be unique.

packages/builder/src/lib/script-checkpoints.ts
```diff
@@ -1,105 +1,310 @@
...
+    this.asc = {};
+  }
 }
 
-export function appendEnv (
-  prev: ScriptCheckpoint,
-  runtimeEnv: AlgobRuntimeEnv
-): ScriptCheckpoint {
-  return appendToCheckpoint(prev, runtimeEnv.network.name, createNetCheckpoint());
+export function appendToCheckpoint (
```
robert-zaremba, line 54: can you make this functions as a class method, please?

packages/builder/src/lib/script-checkpoints.ts
```diff
@@ -1,105 +1,310 @@
...
+export class CheckpointDataImpl implements CheckpointData {
+  strippedCP: ScriptCheckpoints = {};
+  visibleCP: ScriptCheckpoints = {};
+  globalCP: ScriptCheckpoints = {};
+
+  private _mergeTo (target: ScriptCheckpoints, cp: ScriptCheckpoints): ScriptCheckpoints {
+    const keys: string[] = Object.keys(cp);
+    return keys.reduce((out: ScriptCheckpoints, key: string) => {
+      return appendToCheckpoint(out, key, cp[key]);
+    }, target);
```
robert-zaremba, line 92: It's shorter and more efficient to do it with for:

```
for (const k of Object.keys(cp) 
    appendToCheckpoint(target, key, cp[key])
return target
```

packages/builder/src/lib/script-checkpoints.ts
```diff
@@ -1,105 +1,310 @@
...
-    prev[networkName].metadata = Object.assign({}, prev[networkName].metadata, append.metadata);
-    return prev;
-  }
-  prev[networkName] = append;
-  return prev;
+export function registerASA (
+  cp: ScriptNetCheckpoint, name: string, creator: string): ScriptNetCheckpoint {
+  cp.asa[name] = { creator: creator };
+  return cp;
+}
```
robert-zaremba, line 32: This should be a method on `ScriptNetCheckpoint`

packages/builder/src/lib/script-checkpoints.ts
```diff
@@ -1,105 +1,310 @@
...
+
+export function registerASC (
+  cp: ScriptNetCheckpoint, name: string, creator: string): ScriptNetCheckpoint {
+  cp.asc[name] = { creator: creator };
+  return cp;
 }
```
robert-zaremba, line 38: This should be a method on ScriptNetCheckpoint

packages/builder/src/lib/script-checkpoints.ts
```diff
@@ -1,105 +1,310 @@
...
+  );
+  return checkpoints;
 }
 
-export function createNetCheckpoint (metadata?: {[key: string]: string}): ScriptNetCheckpoint {
-  return {
-    timestamp: +new Date(),
-    metadata: metadata === undefined ? {} : metadata
-  };
+export class CheckpointDataImpl implements CheckpointData {
```
robert-zaremba, line 83: Let's do the following renames:

```
ScriptNetCheckpoint -> Checkpoin
ScriptNetCheckpointImpl -> CheckpointImpl
CheckpointData -> CheckpointRepo
CheckpointDataImpl -> CheckpointRepoImpl
ScriptCheckpoints -> Checkpoints
```

packages/builder/src/types.ts
```diff
@@ -284,12 +275,70 @@ export interface LinkReferences {
...
+  // Three different checkpoint states are used for three different occasions.
+  // During run of a script their values are changed in different ways.
+
+  // Accumulates state as scripts are executed.
+  // This way it hides values generated by remaining checkpoints.
+  // It is what should be exposed to the running scripts.
+  visibleCP: ScriptCheckpoints
+  // Variables that current script introduced, short version of what was added.
+  // Used for state persistence.
+  strippedCP: ScriptCheckpoints
```
robert-zaremba, line 300: how about renaming to `currentCP`?

packages/builder/src/types.ts
```diff
@@ -284,12 +275,70 @@ export interface LinkReferences {
...
+  // All possible values that are loaded in advance.
+  // This allows to prevent asset name clashes between scripts.
+  globalCP: ScriptCheckpoints
```
robert-zaremba, line 303: why do we have globalCP and visibleCP?

packages/builder/src/builtin-tasks/deploy.ts
```diff
@@ -35,8 +32,18 @@ export function loadFilenames (directory: string): string[] {
   return files.sort(cmpStr);
 }
 
+function clearCheckpointFiles (scriptNames: string[]): void {
+  for (const scriptName of scriptNames) {
+    try {
+      fs.unlinkSync(toCheckpointFileName(scriptName));
```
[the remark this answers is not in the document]
Invertisment: This removes the files. Deletes them.

packages/builder/src/lib/script-checkpoints.ts
Invertisment: This loads all directories in `artifacts/scripts/` and does the recursion walk. I did it just in case.
So current title is more accurate.

packages/builder/src/builtin-tasks/run.ts
Invertisment: The factory method that you talk about would do the same thing. Instead of a function we'd have a boolean.
Currently this doesn't do anything else so it's fine to do it this way.

packages/builder/src/lib/script-checkpoints.ts
Invertisment: I tried it but then `assert.deepEquals` complains that my types don't have the required methods. And I didn't want to go that route.

packages/builder/src/types.ts
Invertisment: Explained in chat.

packages/builder/src/lib/script-checkpoints.ts
Invertisment: Previous comment about `assert.deepEquals`

packages/builder/src/builtin-tasks/deploy.ts
Invertisment: It was left from previous project. Then I should update all tags.

packages/builder/src/builtin-tasks/deploy.ts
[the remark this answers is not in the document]
robert-zaremba: ah, ok. let's add a comment

packages/builder/src/builtin-tasks/deploy.ts
robert-zaremba: was this `builder:core`?

Invertisment: The name is constructed this way:
`builder:core` for the package and `tasks:deploy` for deploy task. Which makes sense.

I can change it to `algob:core:...` but it doesn't make much difference since we don't use command debugging for now.

packages/builder/src/builtin-tasks/run.ts
Invertisment: It is clear but I prefer to have it in this place.

packages/builder/src/builtin-tasks/run.ts
Invertisment: Changed to factory method

packages/builder/src/lib/script-checkpoints.ts
Invertisment: Changed

packages/builder/src/builtin-tasks/deploy.ts
Invertisment: changed to `algob:task:deploy` and run too

PR #42, opened 2020-08-07
fix default config checks

robert-zaremba:


packages/algob/src/internal/core/tasks/builtin-tasks.ts
```diff
@@ -13,3 +13,10 @@ export default function (): void{
     loadPluginFile(path.join(basedir, t[1]));
   }
 }
+
+// checks if the task name is not a setup kind of task
+export function isSetupTask (taskName: string): boolean {
+  return taskName === tasks.TASK_HELP ||
```
Invertisment, line 19: You check for HELP task three times.
Once it's during the start of cli method, then in the middle and now it's here. Can you check for it once?

packages/algob/src/internal/core/tasks/builtin-tasks.ts
```diff
@@ -13,3 +13,10 @@ export default function (): void{
...
+    taskName === tasks.TASK_CLEAN ||
```
Invertisment, line 20: Also clean shouldn't be allowed to be run outside of a project so it's not a setup task

packages/algob/src/internal/core/runtime-environment.ts
```diff
@@ -40,18 +40,21 @@ export class Environment implements AlgobRuntimeEnv {
...
   ) {
     log("Creating AlgobRuntimeEnv");
 
-    const networkConfig = config.networks[runtimeArgs.network];
-
-    if (networkConfig === undefined) {
+    const ncfg = config.networks[runtimeArgs.network];
+    // network configuration is required for all tasks except few setup tasks
+    if (!ncfg && networkRequired) {
+      console.log(">>>> here")
```
Invertisment: Debug comment

packages/algob/src/internal/cli/cli.ts
```diff
@@ -82,6 +84,11 @@ async function main (): Promise<void> {
     const ctx = BuilderContext.createBuilderContext();
     const config = loadConfigAndTasks(runtimeArgs);
 
+    let taskName = parsedTaskName ?? TASK_HELP;
+    if (runtimeArgs.help && taskName !== TASK_HELP) {
+      taskName = TASK_HELP;
```
Invertisment, line 89: This can break how help task is used for help of a specific command:
`algob help init` -> should print help about `init` task.
And this should work too:
`algob -h init` -> should print help about `init` task.
And this (but it's covered by 2nd case):
`algob init -h` -> should print help about `init` task.

packages/algob/src/internal/core/tasks/builtin-tasks.ts
robert-zaremba: No, I can't check it once because the conditions are different. Here we are also checking for init. The method is not called `isHelpTask`.

packages/algob/src/internal/core/tasks/builtin-tasks.ts
robert-zaremba: Yes. I will remove it.

robert-zaremba: thanks for catching this

packages/algob/src/internal/cli/cli.ts
robert-zaremba: you are right. will fix it.

PR #43, opened 2020-08-07
runtime params: fix help checks

robert-zaremba:


packages/algob/src/internal/cli/cli.ts
```diff
@@ -99,24 +94,37 @@ async function main (): Promise<void> {
...
 
+    let taskName = parsedTaskName ?? TASK_HELP;
     if (taskDefinitions[taskName] == null) {
       throw new BuilderError(ERRORS.ARGUMENTS.UNRECOGNIZED_TASK, {
         task: taskName
       });
     }
 
+    // Being inside of a project is non-mandatory for help and init
+    if ((taskName !== TASK_HELP && taskName !== TASK_INIT && !runtimeArgs.help) &&
```
Invertisment, line 105: So you have the same statuses returned from isSetupTask. isSetupTask also doesn't check runtimeArgs.help parameter which should allow execution.
And you throw the same error on line 130. I think that code should be removed if it only does that check.

robert-zaremba: correct. I will add `origTaskName` to handle the Exception correctly, and will remove the if in this comment

PR #44, opened 2020-08-08
Compile module

as04:
Created compile functionality to compile incoming smart contracts in `assets/` only if they have not been compiled already and store the compiled info in the yaml in `artifacts/cache/`.


packages/builder/src/builtin-tasks/compile.ts
```diff
@@ -0,0 +1,72 @@
...
+
+let algodclient = new algosdk.Algodv2(token, host, port);
+
+export default function (): void {
+  task(TASK_COMPILE, "Compilation task")
+    .setAction(
+      async () => {
+        console.log("compiling......");
+        /*File paths for assets/ and artifacts/cache/ */
+        var srcPath = path.join(__dirname, '../../sample-project/assets/');
```
Invertisment, line 23: Nope.
In short: you shouldn't change the default project, only the one that is in current directory. i.e. we check that we're in a project.

This should be current directory, like this: `path.join(__dirname, 'assets/')`.
The command is prevented to run out of a project folder.
So first of all you have to create a project with `init` and then you'll be able to use the command `compile` in that directory.
The config variables should be available from `algobEnv` variable which you didn't specify in the `setAction` async function input (line 20).

Invertisment: Also you may want to create that directory so that it will exist. It belongs to you, it's not part of the sample project (why would sample project ship with compiled code? it shouldn't, compile task should create a directory for it.)
(think about command `mkdir -p dir/dir/dir` -- it will create all intermediate directories.

packages/builder/src/builtin-tasks/compile.ts
```diff
@@ -0,0 +1,72 @@
...
+        var destPath = path.join(__dirname, '../../sample-project/artifacts/cache');
+        /*Read files from assets/ and artifacts/cache/ */
+        const assetsFiles = fs.readdirSync(srcPath);
+        const compiledFiles = fs.readdirSync(destPath);
+        const compiledSrcHashes = [];
+        /* Store all source file hash from already compiled files in an array */
+    	for(let cfile in compiledFiles) {
```
Invertisment, line 30: Can you create a function for this loop?

packages/builder/src/builtin-tasks/compile.ts
```diff
@@ -0,0 +1,72 @@
...
+    		let pathCFile = path.join(destPath, compiledFiles[cfile]);
+    		let cdata = YAML.parse(fs.readFileSync(pathCFile).toString());
+    		console.log("compiled file data", cdata.srcHash);
+    		compiledSrcHashes.push(cdata.srcHash);
+    	}
+    	/* Iterate through contracts in assets/ and compile new contracts */
+		for (let file in assetsFiles) {
```
Invertisment, line 37: A function for this one would be nice too.
Also indentation is off. Don't use `commit -n` or run `yarn lint:fix` before commiting.

packages/builder/src/builtin-tasks/compile.ts
```diff
@@ -0,0 +1,72 @@
...
+			var flag = 0;
+			var src = path.join(srcPath, assetsFiles[file]); //absolute path for contract in assets/
```
Invertisment, line 39: This should be a separate function in itself -- `getFileName`

packages/builder/src/builtin-tasks/compile.ts
```diff
@@ -0,0 +1,72 @@
...
+			var name = path.basename(assetsFiles[file], '.teal'); //name of file without extension
+			var destFile = path.join(destPath, name + '.yaml'); //destination yaml file name
+	    	let data = fs.readFileSync(src); //read the contact data
```
Invertisment, line 42: Random magic indentation change. Not readable.

packages/builder/src/builtin-tasks/compile.ts
```diff
@@ -0,0 +1,72 @@
...
+const port = networks.localhost.port;
+
+let algodclient = new algosdk.Algodv2(token, host, port);
+
+export default function (): void {
+  task(TASK_COMPILE, "Compilation task")
+    .setAction(
+      async () => {
+        console.log("compiling......");
+        /*File paths for assets/ and artifacts/cache/ */
```
Invertisment, line 22: Do not add these comments. Instead create multiple methods and their names will be actual documentation.
These comments don't add value:

// I do this
this
// I do that
that
// I do the rest
1
2
// some random file IO
echo hi > output.txt
3
4
// Hi mom
mom

packages/builder/src/builtin-tasks/compile.ts
```diff
@@ -0,0 +1,72 @@
...
+const token = networks.localhost.token;
+const host = networks.localhost.host;
+const port = networks.localhost.port;
+
+let algodclient = new algosdk.Algodv2(token, host, port);
+
+export default function (): void {
+  task(TASK_COMPILE, "Compilation task")
+    .setAction(
+      async () => {
```
Invertisment, line 20: Create a new method before the call to `setAction` and name it. Only call it here.

packages/builder/src/builtin-tasks/compile.ts
```diff
@@ -0,0 +1,72 @@
...
+        var srcPath = path.join(__dirname, '../../sample-project/assets/');
+        var destPath = path.join(__dirname, '../../sample-project/artifacts/cache');
+        /*Read files from assets/ and artifacts/cache/ */
+        const assetsFiles = fs.readdirSync(srcPath);
+        const compiledFiles = fs.readdirSync(destPath);
+        const compiledSrcHashes = [];
+        /* Store all source file hash from already compiled files in an array */
+    	for(let cfile in compiledFiles) {
+    		let pathCFile = path.join(destPath, compiledFiles[cfile]);
+    		let cdata = YAML.parse(fs.readFileSync(pathCFile).toString());
```
Invertisment, line 32: Change to a new function: `readCompiledFile`

packages/builder/src/builtin-tasks/compile.ts
```diff
@@ -0,0 +1,72 @@
...
+    		console.log("compiled file data", cdata.srcHash);
+    		compiledSrcHashes.push(cdata.srcHash);
```
Invertisment, line 34: You rely on array indices. It works but it's not readable/understandable.
Can you have a method that returns an object of:
```
{
  hash: "murmurhash",
  compiledFile: "data",
  maybeAssetFile: "data" | undefined
  ...
}
```
This will reduce the amount of `for` loops and will be more readable -- also it could be refactored to a lazy version later on.

packages/builder/src/builtin-tasks/compile.ts
```diff
@@ -0,0 +1,72 @@
...
+		for (let file in assetsFiles) {
+			var flag = 0;
+			var src = path.join(srcPath, assetsFiles[file]); //absolute path for contract in assets/
+			var name = path.basename(assetsFiles[file], '.teal'); //name of file without extension
+			var destFile = path.join(destPath, name + '.yaml'); //destination yaml file name
+	    	let data = fs.readFileSync(src); //read the contact data
+	    	/* Skip already compiled files (if the srcHash is already stored in artifacts/cache/)*/
+	    	for(let hash_i in compiledSrcHashes) {
+	    		if(compiledSrcHashes[hash_i] == murmurhash(data)){
+	    			flag = 1;
```
Invertisment, line 46: This has to be changed to an Error object this way: `throw new BuilderError()`
Flag doesn't show what happened. And where. And to which file. It's magic and magic is not debuggable. The error can mean anything.

You use `flag` to skip the loop. But the `throw` statement will do exactly that. And also show a crash instead of silently omitting the error for the user.

packages/builder/src/builtin-tasks/compile.ts
```diff
@@ -0,0 +1,72 @@
...
+    	/* Iterate through contracts in assets/ and compile new contracts */
+		for (let file in assetsFiles) {
+			var flag = 0;
+			var src = path.join(srcPath, assetsFiles[file]); //absolute path for contract in assets/
+			var name = path.basename(assetsFiles[file], '.teal'); //name of file without extension
+			var destFile = path.join(destPath, name + '.yaml'); //destination yaml file name
+	    	let data = fs.readFileSync(src); //read the contact data
+	    	/* Skip already compiled files (if the srcHash is already stored in artifacts/cache/)*/
+	    	for(let hash_i in compiledSrcHashes) {
+	    		if(compiledSrcHashes[hash_i] == murmurhash(data)){
```
Invertisment, line 45: `hash_i` -> move to an object so that you could call:
`compileSrc(filename)` -> `{compiled: "data", hash: "hash123"}`

packages/builder/src/builtin-tasks/compile.ts
```diff
@@ -0,0 +1,72 @@
...
+	    			break;
+	    		}
+	    	}
+	    	if(flag == 1)
+	    		continue;
+
+		    let results = await algodclient.compile(data).do(); //Compile file
+		    console.log("Hash = " + results.hash);
+		    console.log("Result = " + results.result);
+		    let date: any;
```
Invertisment, line 57: No `any` please. 
Can you replace to `let date = new Date();`? This will have the correct type by default.

packages/builder/src/builtin-tasks/compile.ts
```diff
@@ -0,0 +1,72 @@
...
+		    date = new Date();
+		    /* Create yaml */
```
Invertisment, line 59: You should have a top-level method that returns this object. You'll then simply send it to `save` method with it's file name.

packages/builder/src/builtin-tasks/compile.ts
```diff
@@ -0,0 +1,72 @@
...
+		    let fields = { 
+		    				timestamp: Math.floor(date / 1000),
+		    				srcHash: murmurhash(data),
+		    				address: results.hash,
+		    				compiled: results.result,
+		    				compiledHash: murmurhash(results.result)
+		    			}
+		    let yamlStr = YAML.stringify(fields);
```
Invertisment, line 67: Create a separate method:
`saveCompiled(fields, filename)`

packages/builder/src/builtin-tasks/compile.ts
```diff
@@ -0,0 +1,72 @@
+import { task } from "../internal/core/config/config-env";
+import { networks } from "../../sample-project/algob.config";
+import { TASK_COMPILE } from "./task-names";
+import YAML from "yaml";
+const algosdk = require('algosdk');
+const fs = require('fs');
+const path = require('path');
+const murmurhash = require('murmurhash');
+
+/* Credentials */
```
Invertisment, line 10: I told you that if you can't use the config you should hardcode the variables. You shouldn't import `sample-project` in any circumstance. This is not the use of the `sample-project`. It mustn't be used in `compile` task. You have a local project that you have to change, not the default one.

packages/builder/src/builtin-tasks/compile.ts
```diff
@@ -0,0 +1,72 @@
...
+	    	let data = fs.readFileSync(src); //read the contact data
+	    	/* Skip already compiled files (if the srcHash is already stored in artifacts/cache/)*/
+	    	for(let hash_i in compiledSrcHashes) {
+	    		if(compiledSrcHashes[hash_i] == murmurhash(data)){
+	    			flag = 1;
+	    			console.log("matched", compiledSrcHashes[hash_i], murmurhash(data));
+	    			break;
+	    		}
+	    	}
+	    	if(flag == 1)
```
Invertisment, line 51: Remove this. Crash with `Error` instead.

packages/builder/src/builtin-tasks/compile.ts
robert-zaremba: @as04  - are the git hooks work in your projects? `eslint` should start and fix all this syntax issues whenever you do a commit.

robert-zaremba:
This code is poorly done. The environment is not integrated, the paths are not checked, tests are missing, and in fact it doesn't work with the CMD.

PR #47, opened 2020-08-10
config: load accounts from file

robert-zaremba:
Example on how to load accounts from a file

### DIFF
--- a/packages/algob/sample-project/algob.config.js
+++ b/packages/algob/sample-project/algob.config.js
@@ -1,16 +1,24 @@
-// import { mkAccounts } from "algob";
+// NOTE: below we provide some example accounts.
+// DON'T this account in any working environment because everyone can check it and use
+// the private keys (this accounts are visible to everyone).
 
-// NOTE: this is only an example. DON'T this account in any working environment because
-// everyone can check it and use the private key (this account is visible to everyone).
-
+// Example: accounts constructed using mnemonic. `addr` is optional.
+// const { mkAccounts } = require("algob");
 // let accounts = mkAccounts([{
 //   addr: "KFMPC5QWM3SC54X7UWUW6OSDOIT3H3YA5UOCUAE2ABERXYSKZS5Q3X5IZY",
 //   mnemonic: "call boy rubber fashion arch day capable one sweet skate outside purse six early learn tuition eagle love breeze pizza loud today popular able divide"
 // }]);
+
+// Example: accounts constructed using the Account object
 let accounts = [{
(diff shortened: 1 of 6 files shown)


packages/algob/sample-project/algob.config.js
```diff
@@ -1,17 +1,25 @@
...
 //   mnemonic: "call boy rubber fashion arch day capable one sweet skate outside purse six early learn tuition eagle love breeze pizza loud today popular able divide"
 // }]);
+
+// Example: accounts constructed using the Account object
 let accounts = [{
   addr: 'UDF7DS5QXECBUEDF3GZVHHLXDRJOVTGR7EORYGDBPJ2FNB5D5T636QMWZY',
   sk: new Uint8Array([28,  45,  45,  15,  70, 188,  57, 228,  18,  21,  42, 228,  33, 187, 222, 162,  89,  15,  22,  52, 143, 171, 182,  17, 168, 238,  96, 177,  12, 163, 243, 231, 160, 203, 241, 203, 176, 185,   4,  26,  16, 101, 217, 179, 83, 157, 119,  28,  82, 234, 204, 209, 249,  29,  28, 24,  97, 122, 116,  86, 135, 163, 236, 253])
 }]
 
+// const { loadAccountsFromFileSync } = require("algob");
```
robert-zaremba, line 18: We don't have a full test with file - but the most important step is tested (parsing accounts with different formats) - so the only test which might be missing is just loading the file content. Not sure how important it is (it's just reading a file) - it's a one function call from standard library (`fs`).

robert-zaremba:
We don't have a full test with file - but the most important step is tested (parsing accounts with different formats) - so the only test which might be missing is just loading the file content. Not sure how important it is (it's just reading a file) - it's a one function call from standard library (fs).

packages/algob/sample-project/algob.config.js
```diff
@@ -1,17 +1,25 @@
-// import { mkAccounts } from "algob";
-
-// NOTE: this is only an example. DON'T this account in any working environment because
-// everyone can check it and use the private key (this account is visible to everyone).
+// NOTE: below we provide some example accounts.
+// DON'T this account in any working environment because everyone can check it and use
+// the private keys (this accounts are visible to everyone).
 
+// Example: accounts constructed using mnemonic. `addr` is optional.
+// const { mkAccounts } = require("algob");
```
robert-zaremba, line 6: using it requites that `algob` is in node_modules (so it's "installed")

PR #48, opened 2020-08-10
algorand driver: handle unathorized exception

robert-zaremba:
This PR handles a cryptic runtime exception when Algorand client is initialized with wrong credentials

```
% yarn algob:test node-info
yarn run v1.22.4
$ mkdir -p project-test && cd project-test && node ../build/internal/cli/cli.js node-info
An unexpected error occurred: Unauthorized

Error: Unauthorized
    at Request.callback (/home/robert/projects/algorand/algorand-builder/node_modules/superagent/lib/node/index.js:804:15)
    at /home/robert/projects/algorand/algorand-builder/node_modules/superagent/lib/node/index.js:1036:18
    at IncomingMessage.<anonymous> (/home/robert/projects/algorand/algorand-builder/node_modules/superagent/lib/node/parsers/json.js:19:7)
    at IncomingMessage.emit (events.js:326:22)
    at endReadableNT (_stream_readable.js:1244:12)
    at processTicksAndRejections (internal/process/task_queues.js:80:21)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
[1]    35648 exit 1     yarn algob:test node-info
```

After this PR, the we will see more "useful" error message:

```
% yarn algob:test node-info              
yarn run v1.22.4
$ mkdir -p project-test && cd project-test && node ../build/internal/cli/cli.js node-info
Wrong credentials to access the Algorand Node. Please verify the access token for the "default" network.
error Command failed with exit code 1.
```

### DIFF
--- a/packages/algob/src/builtin-tasks/run.ts
+++ b/packages/algob/src/builtin-tasks/run.ts
@@ -94,7 +94,7 @@ export async function runMultipleScripts (
   try {
     return await _runMultipleScripts(runtimeEnv, scriptNames, onSuccessFn, force, logTag, allowWrite);
   } catch (e) {
-    if (!checkAlgorandUnauthorized(e, runtimeEnv.network)) { throw e; } else { throw new Error("Algorand Node: Unauthorized"); }
+    if (!checkAlgorandUnauthorized(e, runtimeEnv.network)) { throw e; }
   }
 }
 
--- a/packages/algob/src/builtin-tasks/run.ts
+++ b/packages/algob/src/builtin-tasks/run.ts
@@ -5,6 +5,7 @@ import { task } from "../internal/core/config/config-env";
 import { BuilderError } from "../internal/core/errors";
 import { ERRORS } from "../internal/core/errors-list";
 import { runScript } from "../internal/util/scripts-runner";
+import { checkAlgorandUnauthorized } from "../lib/exceptions";
 import { assertDirChildren } from "../lib/files";
 import {
   AlgobDeployerImpl,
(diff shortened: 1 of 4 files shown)


packages/algob/src/lib/exceptions.ts
```diff
@@ -0,0 +1,9 @@
+import type { Network } from "../types";
+
+export function checkAlgorandUnauthorized (e: any, n: Network): boolean { // eslint-disable-line @typescript-eslint/no-explicit-any
+  if (e instanceof Error && (e.message === 'Unauthorized')) {
```
Invertisment, line 4: This can match any exception that is cryptic

packages/algob/src/builtin-tasks/run.ts
```diff
@@ -84,6 +85,20 @@ export async function runMultipleScriptsOneByOne (
...
+  runtimeEnv: AlgobRuntimeEnv,
+  scriptNames: string[],
+  onSuccessFn: (cpData: CheckpointRepo, relativeScriptPath: string) => void,
+  force: boolean,
+  logTag: string,
+  allowWrite: boolean): Promise<void> {
+  try {
+    return await _runMultipleScripts(runtimeEnv, scriptNames, onSuccessFn, force, logTag, allowWrite);
+  } catch (e) {
+    if (!checkAlgorandUnauthorized(e, runtimeEnv.network)) { throw e; }
```
Invertisment, line 97: If this exception occurs in multiple places then let's handle it in `cli.ts`.

packages/algob/src/lib/exceptions.ts
robert-zaremba: Yes, I know. But I don't know any better way how to do it. I can add more `e` checks, but there always will be a risk.
On the other hand, I don't know any other dependency which can rise this exception, because we are only connecting to algroand.
Do you have better idea?

packages/algob/src/builtin-tasks/run.ts
robert-zaremba: Wrapping SDK is too much code to do. I was thinking about moving it to `cli.ts`. If you think it's a better idea then I can do it.

Invertisment: I think it's not a better idea because we'll lose the context of exception.
Wrapping the SDK is a task that we may think about.

robert-zaremba: moved to `cli.ts`

packages/algob/src/lib/exceptions.ts
robert-zaremba: I've added more "error decoding"

PR #50, opened 2020-08-15
ASA validation; Deployer logic moved into separate file

Invertisment:


packages/algob/src/lib/validation-errors.ts
```diff
@@ -0,0 +1,26 @@
+import * as z from 'zod';
+import { ZodErrorCode, ZodSuberror } from "zod/lib/src/ZodError";
+
+function parseZodSubError (e: ZodSuberror, indent: number): string {
```
robert-zaremba, line 4: Let's put this file next to the `errors-list.ts` file (in same for)

Invertisment: Moved.

packages/algob/src/types-input.ts
```diff
@@ -0,0 +1,35 @@
+import * as z from 'zod';
+
+export const AddressSchema = z.string();
```
robert-zaremba, line 3: Do we need this?

packages/algob/src/types-input.ts
```diff
@@ -0,0 +1,35 @@
...
+    path: ['unitName']
+  })
+  .refine(o => (!o.url || (o.url && (o.url.length <= 32))), {
+    message: "URL must not be longer than 32 bytes",
+    path: ['url']
+  })
+  .refine(o => (!o.metadataHash || (o.metadataHash && (o.metadataHash.length <= 32))), {
+    message: "Metadata Hash must not be longer than 32 bytes",
+    path: ['metadataHash']
+  });
```
robert-zaremba, line 34: Maybe we can have one define with many `if` inside. Same amount big code and maybe even better readability.

Invertisment: I can have one large blob but they will return the same objects anyway.

packages/algob/src/types-input.ts
Invertisment: This is how address is decoded. 
It's used four times in the ASA definition:
```
  manager: AddressSchema.optional(),
  reserve: AddressSchema.optional(),
  freeze: AddressSchema.optional(),
  clawback: AddressSchema.optional()
```

packages/algob/src/types-input.ts
Invertisment: It's not possible with this method.
It accepts two parameters and it doesn't do whole object check.

https://github.com/vriad/zod/#custom-validation

packages/algob/src/lib/asa.ts
```diff
@@ -0,0 +1,30 @@
...
+import { BuilderError } from "../internal/core/errors";
+import { ERRORS } from "../internal/core/errors-list";
+import { parseZodError } from "../internal/core/validation-errors";
+import { ASADescription } from "../types";
+import { ASADescriptionSchema } from "../types-input";
+
+export function parseASADef (obj: Object, filename: string): ASADescription {
+  try {
+    const parsed = ASADescriptionSchema.parse(obj);
+    if (parsed.defaultFrozen === undefined) {
```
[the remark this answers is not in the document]
Invertisment: Renamed

packages/algob/src/types-input.ts
robert-zaremba: OK, thanks.

PR #53, opened 2020-08-24
Compile command

robert-zaremba:


packages/algob/src/internal/core/errors.ts
```diff
@@ -185,3 +186,14 @@ function _applyErrorMessageTemplate (
 
   return template;
 }
+
+export function parseAlgorandError (e: RequestError, ctx: AnyMap): Error {
```
Invertisment, line 190: I handle this error in a little different way in my PR. Can you look at it and choose what you prefer?
I check `error.clientError` variable (something similar to this one) and then I slightly change the message.
I also didn't use JSON functions.

packages/algob/src/builtin-tasks/compile.ts
```diff
@@ -1,13 +1,112 @@
...
 export default function (): void {
-  task(TASK_COMPILE, "Compilation task")
-    .setAction(
-      async () => {
-        console.log("compiling......");
-      }
-    );
+  task(TASK_COMPILE, "Compile all TEAL smart contracts")
+    .addFlag("force", "recompile even if the source file didn't change")
+    .setAction(_compile);
```
Invertisment, line 19: Let's not have both `_compile` and `compile` if they do different things and operate in different contexts.
I'd prefer `runCompile` for this one.

packages/algob/src/builtin-tasks/compile.ts
```diff
@@ -1,13 +1,112 @@
...
+  await assertDir(CACHE_DIR);
+  const cache = readArtifacts(CACHE_DIR);
+
+  for (const f of readdirSync(ASSETS_DIR)) {
+    if (!f.endsWith(tealExt)) { continue; }
+
+    let c = cache.get(f);
+    const [teal, thash] = readTealAndHash(path.join(ASSETS_DIR, f));
+    if (!force && c !== undefined && c.srcHash === thash) {
+      console.log(`smart-contract source "${f}" didn't change, skipping.`);
```
Invertisment, line 41: Let's unify how we use logs. 
Either change the old ones or these ones. We already have two differents methods for it and one of them (the old one) doesn't output anything.
I think we should have a task for that.

packages/algob/src/builtin-tasks/compile.ts
```diff
@@ -1,13 +1,112 @@
...
+
+  constructor (n: Network) {
+    this.algocl = createClient(n);
+  }
+
+  async compile (code: string): Promise<CompileOut> {
+    return await this.algocl.compile(code).do();
+  }
+
+  writeFile (filename: string, content: string): void {
```
Invertisment, line 63: This function doesn't depend on `this.algocl` in any way. I think it's not needed to be in this class.
And then it means that this wrapping class may not be needed. Is it planned to be mocked? I don't see much use for it.

packages/algob/src/builtin-tasks/compile.ts
```diff
@@ -1,13 +1,112 @@
...
+  constructor (filename: string, tealCode: string, tealHash: number) {
+    this.filename = filename;
+    this.tealCode = tealCode;
+    this.tealHash = tealHash;
+  }
+
+  mkASCCache (co: CompileOut): ASCCache {
+    return {
+      filename: this.filename,
+      timestamp: Math.floor(Date.now() / 1000),
```
Invertisment, line 97: Most of the time I use `+ new Date()`. Can we have a function for timestamps?

packages/algob/src/builtin-tasks/compile.ts
```diff
@@ -1,13 +1,112 @@
...
+
+    let c = cache.get(f);
+    const [teal, thash] = readTealAndHash(path.join(ASSETS_DIR, f));
+    if (!force && c !== undefined && c.srcHash === thash) {
+      console.log(`smart-contract source "${f}" didn't change, skipping.`);
+      continue;
+    }
+    console.log("compiling", f);
+    c = await compileAndSave(op, new ASCCachePartial(f, teal, thash));
+    cache.set(f, c);
```
Invertisment, line 46: You only use `c.srcHash` from this type in `c.srcHash === thash`. So all of other data is not needed in the type.
Timestamp is not checked as well so either it should be checked or removed.

I also think that `const [teal, thash] = readTealAndHash(...)` already directly returns all that you care about: the hash.
So you should only index by `filename -> hash` or `filename -> {hash, timestamp}` and then remove this type and its interface.

packages/algob/src/builtin-tasks/compile.ts
```diff
@@ -1,13 +1,112 @@
...
+  return cache;
+}
+
+function readTealAndHash (filename: string): [string, number] {
+  const content = readFileSync(filename, 'utf8');
+  return [content, murmurhash.v3(content)];
+}
+
+class ASCCachePartial {
+  filename: string;
```
Invertisment, line 84: You have a cache indexed by filename and you have it here again `{filename -> {filename}}`.
The only things needed are hash and maybe a timestamp for further filtering which you don't currently do.

I would construct this data type only after the compilation and only leave the index to have
`{filename -> {hash, timestamp}}`.

packages/algob/src/internal/core/errors-list.ts
```diff
@@ -626,6 +627,23 @@ const accountErrors = {
 
 };
 
+const algorandNode = {
+  BAD_REQUEST: {
+    number: 410,
+    message: `status_code=%status%,
```
Invertisment, line 633: Looks cryptic. Can we have a human readable error like in other places?

yarn.lock
```diff
@@ -417,10 +417,9 @@ ajv@^6.10.0, ajv@^6.10.2:
     json-schema-traverse "^0.4.1"
     uri-js "^4.2.2"
 
-algosdk@^1.6.2:
-  version "1.6.2"
-  resolved "https://registry.yarnpkg.com/algosdk/-/algosdk-1.6.2.tgz#dc7b861a62bb40036a2744f69d2595318806fdd3"
-  integrity sha512-THz5I0x7h1tLo5rNbEtC9sZ+b5XBf+H3CCu5Kw2d2u9p84Rx6sEPsg9Ae+G1AfkiJoiMnr3+kSXXrkM5Lc0cMQ==
+"algosdk@github:algorand/js-algorand-sdk":
+  version "1.7.0"
```
Invertisment, line 421: :+1:

packages/algob/src/builtin-tasks/compile.ts
robert-zaremba: how about `compileCommand` and `compile`? `do` vs `run` is not very informative

packages/algob/src/internal/core/errors.ts
robert-zaremba: I prefer my version because:
1. We should be able to pass more context
2. We should have Algorand Node error type (something more specific than IO)
3. It's more Algorand Node specific, and doesn't got through many checks (ifs / `? ... :`)

That being said we can update it and take some of your ideas (client / server). I use the StatusCode for that - 4** error is Client error and passes all details. So if you will find useful to add more things there, you can do it in a PR.

packages/algob/src/builtin-tasks/compile.ts
robert-zaremba: What's the other way to print info messages to the user? `debug` is used for debugging.
What I see, in other places (eg the `init` command and `project-creation.ts`) `console.log` is used for info logs.  If you want to change it then you can create a task in the icebox.

packages/algob/src/builtin-tasks/compile.ts
robert-zaremba: `new Date()` creates a Date object. Here we want a number. `Date.now()` static method is proper way to do it.

I will move it to an util function.

packages/algob/src/builtin-tasks/compile.ts
robert-zaremba: It is used in the next line - we serialize the whole `c: ASCCache` structure.

packages/algob/src/builtin-tasks/compile.ts
robert-zaremba: OK, I can remove the type. I needed it more in the early version of the code.

packages/algob/src/builtin-tasks/compile.ts
robert-zaremba: Why it should depend on the `algocl` ? The class is a compiler operator and it's responsible for IO. I'm using it to mock it in tests. It's not a `algocl` wrapper. That's why I use the name "compiler operator"

packages/algob/src/builtin-tasks/compile.ts
Invertisment: I know but you only used the hash and you didn't use any of the other variables. So they are not needed and can be added when we'll decide to use them.
Also it's strange that you create `ASCCachePartial` when instead you could return it from `readTealAndHash` and not create a tuple there. So either the tuple `[teal, thash]` or this type should be changed.

Invertisment (approved):

packages/algob/src/builtin-tasks/compile.ts
Invertisment: You replaced the type. So it doesn't matter.

packages/algob/src/builtin-tasks/compile.ts
Invertisment: If you wouldn't have mocked it in tests then this could be replaced by a static function.

packages/algob/src/builtin-tasks/compile.ts
Invertisment: I didn't use debug logging and this statement looks like it's too verbose for normal operation where user simply reruns the compilation.
So I started thinking whether we should do something about it.
I don't know if I want a task about it.

packages/algob/src/builtin-tasks/compile.ts
Invertisment: Looks fine.

packages/algob/src/internal/core/errors.ts
Invertisment: I think that a check for `4**` vs `5**` is the same thing that they state in their variables. 
The part that I wanted to be reused is this:
```
  // TX execution error
  if (error?.response?.error !== undefined) {
```
It's not exactly the same error that you have here.
Also it contains a human-readable message:
```
        message: error.response?.body?.message
          ? error.response.body.message
          : error.response.error.message
```

packages/algob/src/internal/core/errors-list.ts
robert-zaremba: it's in the next line

packages/algob/src/internal/core/errors.ts
robert-zaremba: OK, then let's add it in a  separate PR. You can do it.

PR #54, opened 2020-08-25
test: script-runner test - change time check to have tests passing on…

robert-zaremba:
… faster runtimes


packages/algob/test/internal/util/scripts-runner.ts
```diff
@@ -22,7 +22,7 @@ describe("Scripts runner", function () {
     const before = new Date();
     await runScript("./scripts/async-script.js", mkAlgobEnv());
     const after = new Date();
-    assert.isAtLeast(after.getTime() - before.getTime(), 100);
+    assert.isAtLeast(after.getTime() - before.getTime(), 20);
```
robert-zaremba, line 25: if the value is too big, the tests can fail sometimes.

amityadav0 (approved):
looks good.

PR #55, opened 2020-08-25
Compile tests

robert-zaremba:
Add tests to the compile command


packages/algob/test/fixture-projects/default-config-project/assets/README.md
```diff
@@ -0,0 +1 @@
+This is a sample README file.
```
Invertisment, line 1: Can we have info here?

packages/algob/src/builtin-tasks/compile.ts
```diff
@@ -25,15 +26,15 @@ export interface TaskArgs {
 }
 
 function _compile ({ force }: TaskArgs, env: AlgobRuntimeEnv): Promise<void> {
```
Invertisment, line 28: It would be best to rename this function.

Invertisment (approved):

packages/algob/test/fixture-projects/default-config-project/assets/README.md
robert-zaremba: what kind of info? I can add whatever you want.

My intention was just to create some other file (which is not `.teal`) so I can test that only teal files are handled and compiled.

packages/algob/src/builtin-tasks/compile.ts
robert-zaremba: ah, I forgot it. OK. will do as discussed in the previous PR.

packages/algob/test/fixture-projects/default-config-project/assets/README.md
Invertisment: Oh. Then this could be the contents of this file.

amityadav0 (approved):

Invertisment (approved):

PR #57, opened 2020-08-26
Clash errors

amityadav0:


packages/algob/src/lib/script-checkpoints.ts
```diff
@@ -81,23 +82,43 @@ export class CheckpointRepoImpl implements CheckpointRepo {
...
     return this;
   }
 
-  mergeToGlobal (cp: Checkpoints): CheckpointRepo {
-    this.allCPs = this._mergeTo(this.allCPs, cp);
+  mergeToGlobal (cp: Checkpoints, scriptName: string): CheckpointRepo {
+    this.allCPs = this._mergeTo(this.allCPs, cp, this.scriptMap);
+
+    if (scriptName.charAt(0) === 'a') {
+      scriptName = scriptName.slice(10, scriptName.length - 8);
```
Invertisment, line 105: You either have to pass script names into this method or checkpoint names. Not both. You don't need this check.

`10` is hardcoded. 
`=== 'a'` is only the first letter, it will be true for `a_scripts1` too which shouldn't be matched.
8 is hardcoded. It will crash if we'll change our checkpoint ending from `.cp.yaml` to anything that is not of length 8.

Extract this into a method.

packages/algob/src/lib/script-checkpoints.ts
```diff
@@ -81,23 +82,43 @@ export class CheckpointRepoImpl implements CheckpointRepo {
...
+    }
+
+    const keys: string[] = Object.keys(cp);
+    for (const k of keys) {
+      const orig = cp[k];
+      const allAssetNames = Object.keys(orig.asa).concat(Object.keys(orig.asc));
+      for (const assetName of allAssetNames) {
+        if (!(this.scriptMap[assetName])) { this.scriptMap[assetName] = scriptName; } else {
```
Invertisment, line 113: Can you expand this statement?
This already has three levels of nesting so custom formatting is even worse.
Also it creates a long line.

amityadav0: yes

packages/algob/src/lib/script-checkpoints.ts
amityadav0: will change the input from parent methods

packages/algob/src/lib/script-checkpoints.ts
```diff
@@ -81,23 +82,43 @@ export class CheckpointRepoImpl implements CheckpointRepo {
...
-    this.mergeToGlobal(cp);
+    this.precedingCP = this._mergeTo(this.precedingCP, cp, this.scriptMap);
+    this.mergeToGlobal(cp, scriptName);
     return this;
   }
 
-  mergeToGlobal (cp: Checkpoints): CheckpointRepo {
-    this.allCPs = this._mergeTo(this.allCPs, cp);
+  mergeToGlobal (cp: Checkpoints, scriptName: string): CheckpointRepo {
+    this.allCPs = this._mergeTo(this.allCPs, cp, this.scriptMap);
```
Invertisment, line 102: Can you move this line after you check `cp` for clashes? It's bad to mutate all the things and then show that you have an error. It's better to have clean state during the crash.

packages/algob/src/lib/script-checkpoints.ts
Invertisment: I have this `.cp.yaml` saved somewhere in the code. Or it should be extracted and ending should be checked.
`string.endsWith(...)`

packages/algob/src/lib/script-checkpoints.ts
amityadav0: yes

packages/algob/src/lib/script-checkpoints.ts
```diff
@@ -22,6 +23,15 @@ export function toCheckpointFileName (scriptName: string): string {
   return path.join(artifactsPath, scriptName + checkpointFileSuffix);
 }
 
+export function toScriptFileName (filename: string): string {
+  filename = filename.replace(artifactsPath + path.sep, '');
+  if (filename.endsWith(checkpointFileSuffix)) {
```
Invertisment, line 28: Actually it's not even needed to check this one. It's always true.

packages/algob/src/lib/script-checkpoints.ts
```diff
@@ -81,23 +91,40 @@ export class CheckpointRepoImpl implements CheckpointRepo {
...
+        } else {
+          if (this.scriptMap[assetName] !== scriptName) {
+            throw new BuilderError(
+              ERRORS.BUILTIN_TASKS.CHECKPOINT_ERROR_DUPLICATE_ASSET_DEFINITION,
+              { assetName: [this.scriptMap[assetName], scriptName] });
+          }
+        }
+      }
+    }
+    this.allCPs = this._mergeTo(this.allCPs, cp, this.scriptMap);
```
Invertisment, line 127: :ok_hand:

packages/algob/src/types.ts
```diff
@@ -334,6 +334,10 @@ export interface Checkpoint {
   asc: {[name: string]: ASCInfo}
 };
 
+export interface assetScriptMap {
```
Invertisment, line 337: Use a capital letter here, it's a type.

Invertisment (approved):

packages/algob/src/builtin-tasks/run.ts
```diff
@@ -1,4 +1,5 @@
 import debug from "debug";
+import { constVoid } from "fp-ts/lib/function";
```
Invertisment, line 2: Remove this one. Not used.

packages/algob/src/types.ts
amityadav0: yes

packages/algob/src/builtin-tasks/run.ts
amityadav0: yes

packages/algob/src/lib/script-checkpoints.ts
amityadav0: yes

PR #61, opened 2020-08-28
Deployer.algod & Deployer.waitForConfirmation & Deployer.accountsByName

Invertisment:


robert-zaremba:
1. OK for `algodClient`, thought IHO `algocl` is also good, the `cl` suffix is common in many projects, so it's not creating confusins and it's shorter / easier.
2. `accounts` should be a mapping from name -> `algosd.Account`

packages/algob/src/lib/algo-client.ts
```diff
@@ -17,39 +17,41 @@ export function createDeployClient (network: Network): AlgoDeployClient {
...
+  waitForConfirmation: (txId: string) => Promise<algosdk.ConfirmedTxInfo>
 }
 
 export class AlgoClientImpl implements AlgoDeployClient {
-  private readonly algoClient: algosdk.Algodv2;
+  algodClient: algosdk.Algodv2;
 
-  constructor (algoClient: algosdk.Algodv2) {
-    this.algoClient = algoClient;
+  constructor (algod: algosdk.Algodv2) {
```
robert-zaremba, line 28: rename `algod` to `algocl`

packages/algob/src/lib/deployer.ts
```diff
@@ -20,12 +21,12 @@ export class AlgobDeployerImpl implements AlgobDeployer {
   private readonly algoClient: AlgoDeployClient;
 
   constructor (
-    runtimeEnv: AlgobRuntimeEnv, cpData: CheckpointRepo, asaDefs: ASADefs, algoSdk: AlgoDeployClient
+    runtimeEnv: AlgobRuntimeEnv, cpData: CheckpointRepo, asaDefs: ASADefs, algoClient: AlgoDeployClient
```
robert-zaremba, line 24: reanme `algoClient`. It makes confusion with a client for the node.

Why you didn't use Operator as suggested?

packages/algob/src/lib/deployer.ts
```diff
@@ -20,12 +21,12 @@ export class AlgobDeployerImpl implements AlgobDeployer {
...
   ) {
     this.runtimeEnv = runtimeEnv;
     this.cpData = cpData;
     this.loadedAsaDefs = asaDefs;
-    this.algoClient = algoSdk;
+    this.algoClient = algoClient;
```
robert-zaremba, line 29: ditto, also for `this.algoClient`

packages/algob/src/lib/deployer.ts
```diff
@@ -137,4 +146,12 @@ export class AlgobDeployerReadOnlyImpl implements AlgobDeployer {
   isDefined (name: string): boolean {
     return this._internal.isDefined(name);
   }
+
+  get algodClient() {
+    return this._internal.algodClient
+  }
```
robert-zaremba, line 152: This is good,

But seconding the comment above `this.algodClient` vs `this.algoClient`

packages/algob/src/lib/deployer.ts
```diff
@@ -137,4 +146,12 @@ export class AlgobDeployerReadOnlyImpl implements AlgobDeployer {
...
+
+  async waitForConfirmation (txId: string): Promise<algosdk.ConfirmedTxInfo> {
+    return this._internal.waitForConfirmation(txId)
+  }
```
robert-zaremba, line 156: There should be some inheritance to avoid code duplication.

packages/types-algosdk/index.d.ts
```diff
@@ -222,7 +222,7 @@ declare module 'algosdk' {
     txId: string
   }
 
-  interface PendingTransactionInformation {
+  interface ConfirmedTxInfo {
```
robert-zaremba, line 225: :+1:

robert-zaremba:
Why the class name `AlgobDeployerReadOnlyImpl` didn't change? We were already discussing several times that `ReadOnly` is not the right adjective here, because it can make changes (eg it still allows arbitrary transactions)

packages/algob/src/lib/deployer.ts
Invertisment: I wanted to explicitly not inherit in this class so that compiler would fail when interface is not implemented.

packages/algob/src/lib/deployer.ts
Invertisment: Because I want to have all the methods in this class so that Deployer class would be dumb -- have mostly oneliner functions. 
And simply would offload logic to this one.

Invertisment: Renaming to `algoActions` and `AlgoActions`, and `AlgoActionsImpl`

packages/algob/src/lib/deployer.ts
Invertisment: algoActions

packages/algob/src/lib/deployer.ts
Invertisment: algoActions

packages/algob/src/lib/deployer.ts
Invertisment: We can have some complex inheritance where only a part of it is inherited but I don't think it's worth it.

Invertisment (commented):
Renamed `AlgoClientImpl` and related things to `AlgoActions` and similar
Renames fixed.
`Deployer.accounts(name)` implemented

packages/algob/src/lib/deployer.ts
```diff
@@ -18,18 +22,16 @@ export class AlgobDeployerImpl implements AlgobDeployer {
...
     this.runtimeEnv = runtimeEnv;
     this.cpData = cpData;
     this.loadedAsaDefs = asaDefs;
-    this.algoClient = algoSdk;
-  }
-
-  get accounts (): Account[] {
-    return this.runtimeEnv.network.config.accounts;
+    this.algoClient = algoClient;
+    this.accounts = mkAccountIndex(runtimeEnv.network.config.accounts);
```
robert-zaremba, line 34: this has a wrong type. Please check the spec doc. I intentionally mentioned in the chat to make sure we will not go again in the loops because we are diverging from spec without previous agreement. The spec is there for a reason to make sure that everyone is on the same page.

Invertisment:
> Why the class name `AlgobDeployerReadOnlyImpl` didn't change? We were already discussing several times that `ReadOnly` is not the right adjective here, because it can make changes (eg it still allows arbitrary transactions)

Updated the names.

packages/algob/src/lib/deployer.ts
Invertisment: Ok, didn't know that it was needed to have both, not only one.

Invertisment: accounts: Account[];
  accountsByName: Accounts;

Invertisment: Updated. Now it contains both vars.

packages/algob/src/lib/deployer.ts
robert-zaremba: but this is not related to the `algosdk.Action` type? correct?

packages/algob/src/lib/account.ts
```diff
@@ -63,3 +63,12 @@ export function validateAccount (a: Account, errs: ErrorPutter): boolean {
...
   return errs.isEmpty;
 }
+
+export function mkAccountIndex (accountList: Account[]): Accounts {
+  return accountList.reduce(
+    (out: Accounts, a: Account) => {
+      out[a.name] = a;
+      return out;
+    },
+    {});
```
robert-zaremba, line 73: The type is wrong. It should be `Map<>`

Side note: using `for..of` loop would be shorter and easier for newcomers.

Invertisment: Updated to Map.

packages/algob/src/lib/deployer.ts
Invertisment: It is not related.

packages/algob/src/lib/account.ts
robert-zaremba: `map` should we use for homomoprhic operations between containers. Moreover it's not more readable. So, I would prefer to stay with `reduce` then with `map`.

robert-zaremba (approved):
Approving, but there are two things which must be changed:

+ renaming which we discussed

PR #62, opened 2020-08-28
algob: move compile to lib

robert-zaremba:


amityadav0 (approved):
Yes, it is helpful for deploying ASC task

packages/algob/src/lib/compile.ts
```diff
@@ -0,0 +1,85 @@
...
+
+export class CompileOp {
+  algocl: Algodv2;
+  cacheAssured = false;
+
+  constructor (algocl: Algodv2) {
+    this.algocl = algocl;
+  }
+
+  // Get's the TEAL compiled result from artifacts cache and compiles the code if necessary.
```
Invertisment, line 22: gets

robert-zaremba: thanks.

PR #64, opened 2020-08-28
Checkpoint members use maps

Invertisment:


packages/algob/test/builtin-tasks/deploy.ts
```diff
@@ -36,14 +37,14 @@ scripts directory: script 2 executed
       noCompile: true
     });
     const snapshot1 = loadCheckpoint("./scripts/1.js");
-    assert.deepEqual(snapshot1.default.metadata, {
+    assert.deepEqual(snapshot1.default.metadata, toMap({
       "script 1 key": "script 1 value"
-    });
+    }));
```
robert-zaremba, line 42: the way how to create Map directly is:
```
let first = new Map([
  [1, 'one'],
  [2, 'two'],
  [3, 'three'],
])
```

robert-zaremba: Let's use a direct syntax.

packages/algob/src/lib/script-checkpoints.ts
```diff
@@ -182,8 +176,22 @@ export function persistCheckpoint (scriptName: string, checkpoint: Checkpoints):
...
+  // Some structures are objects, some others are maps. Oh why.
+  const loaded = loadFromYamlFileSilent(checkpointName, { mapAsMap: true });
+  const obj: Checkpoints = {};
+  for (const [key, checkpointMap] of loaded.entries()) {
+    const cp: any = {};
+    for (const [cpKey, cpVal] of checkpointMap.entries()) {
+      cp[cpKey] = cpVal;
+    }
+    obj[key] = cp;
+  }
```
robert-zaremba, line 189: so, why do we need to copy the `loaded` object?

robert-zaremba (approved):
Approving, but there are 2 change requests.

packages/algob/src/lib/script-checkpoints.ts
Invertisment: Because it's a map, not an object. YAML parser can either make everything an object or a map.
So `mapAsMap: true` sets this option to maps. And then I have to construct objects where needed.

packages/algob/test/builtin-tasks/deploy.ts
Invertisment: It is not also a direct syntax.
It creates a map from a list of lists which probably is more objects than simply using one object and a function.

packages/algob/src/lib/script-checkpoints.ts
```diff
@@ -182,8 +176,22 @@ export function persistCheckpoint (scriptName: string, checkpoint: Checkpoints):
   );
 }
 
+export function loadCheckpointByCPName (checkpointName: string): Checkpoints {
+  // Some structures are objects, some others are maps. Oh why.
+  const loaded = loadFromYamlFileSilent(checkpointName, { mapAsMap: true });
+  const obj: Checkpoints = {};
+  for (const [key, checkpointMap] of loaded.entries()) {
```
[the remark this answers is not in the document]
Invertisment: changed to `loaded`

packages/algob/test/builtin-tasks/deploy.ts
Invertisment: Also this:
Screenshot at 2020-08-30 21-35-32

PR #66, opened 2020-08-30
Deploy ASC 

amityadav0:


packages/algob/src/lib/deployer.ts
```diff
@@ -85,13 +89,14 @@ export class AlgobDeployerImpl implements AlgobDeployer {
     return this.cpData.precedingCP[this.networkName].asa[name];
   }
 
-  async deployASC (name: string, source: string, account: Account): Promise<ASCInfo> {
+  async deployASC (name: string, scParams: Object, flags: ASCDeploymentFlags): Promise<ASCInfo> {
+    const result: ASCCache = await this.algoOp.ensuredCompiled(name, false);
+    const programb64 = result.compiled;
```
Invertisment, line 94: Move console statements and programb64 to operator class.

Leave minimum amount of code here.

Make `this.algoOp.deployASC` accept `ASCCache` directly.

packages/algob/src/lib/deployer.ts
```diff
@@ -85,13 +89,14 @@ export class AlgobDeployerImpl implements AlgobDeployer {
     return this.cpData.precedingCP[this.networkName].asa[name];
   }
 
-  async deployASC (name: string, source: string, account: Account): Promise<ASCInfo> {
+  async deployASC (name: string, scParams: Object, flags: ASCDeploymentFlags): Promise<ASCInfo> {
```
Invertisment, line 92: scParams should have a type, don't leave `Object` here.

packages/algob/src/types.ts
```diff
@@ -349,6 +351,18 @@ export interface ASADeploymentFlags {
   validRounds?: number
 }
 
+export interface ASCDeploymentFlags {
+  creator: Account
```
Invertisment, line 355: This field is the same as in `ASADeploymentFlags`. Other fields are similar too. Can you have a parent class for these two?

Invertisment: Also make sure Robert agrees with this parameter list. 
All of these params were not defined by the spec

packages/algob/src/types.ts
```diff
@@ -363,7 +377,7 @@ export interface AlgobDeployer {
   putMetadata: (key: string, value: string) => void
   getMetadata: (key: string) => string | undefined
   deployASA: (name: string, flags: ASADeploymentFlags) => Promise<ASAInfo>
-  deployASC: (name: string, source: string, account: Account) => Promise<ASCInfo>
+  deployASC: (name: string, sc_params: Object, flags: ASCDeploymentFlags) => Promise<ASCInfo>
```
Invertisment, line 380: How does `sc_params` differ from `flags`?
In spec there is only one object. You have to remove one parameter.

packages/algob/test/fixture-projects/scripts-dir-recursive-cp/scripts/2.js
```diff
@@ -1,7 +1,7 @@
 
 async function run (runtimeEnv, accounts, deployer) {
   if (deployer.isDeployMode) {
-    await deployer.deployASC('ASC from second', {}, deployer.accountsByName.get('acc-name-1'))
+    await deployer.deployASC('fee-check.teal', {}, { creator: deployer.accounts[0], microAlgo: 100000000 })
```
Invertisment, line 4: Remove the middle dummy `{}` which is not used.

packages/algob/test/mocks/deployer.ts
```diff
@@ -17,7 +17,7 @@ export class FakeDeployer implements AlgobDeployer {
     throw new Error("Not implemented");
   };
 
-  async deployASC (name: string, source: string, account: AccountDef): Promise<ASCInfo> {
+  async deployASC (name: string, scParams: object, flags: ASCDeploymentFlags): Promise<ASCInfo> {
```
Invertisment, line 20: remove this param

packages/algob/test/stubs/algo-operator.ts
```diff
@@ -1,12 +1,18 @@
 
-import { Algodv2 } from "algosdk";
+import algosdk, { Algodv2 } from "algosdk";
 
 import { AlgoOperator } from "../../src/lib/algo-operator";
-import { Account, ASADef, ASADeploymentFlags, ASAInfo } from "../../src/types";
+import {
+  Account, ASADef, ASADeploymentFlags, ASAInfo, ASCCache, ASCDeploymentFlags,
+  ASCInfo
```
Invertisment, line 7: Either everything has to be in new lines or nothing.

packages/algob/test/stubs/algo-operator.ts
```diff
@@ -1,12 +1,18 @@
...
+} from "../../src/types";
 
 export class AlgoOperatorDryRunImpl implements AlgoOperator {
   get algodClient (): Algodv2 {
-    throw new Error("Not implemented");
+    const token = "adsadasd";
+    const host = "127.0.0.1";
+    const port = 42981;
+    return new algosdk.Algodv2(token, host, port);
```
Invertisment: Remove this config. Throw exception or mock the methods using `sinon` library.
What uses this? This shouldn't be used from Deployer. It's only provided for the user. You have your `CompileOp` which takes care of `Algodv2` and Deployer must not use it directly.

packages/algob/src/lib/deployer.ts
```diff
@@ -32,6 +35,7 @@ export class AlgobDeployerImpl implements AlgobDeployer {
     this.cpData = cpData;
     this.loadedAsaDefs = asaDefs;
     this.algoOp = algoOp;
+    // this.compileOp = new CompileOp(this.algoOp.algodClient);
```
Invertisment, line 38: If you see on line 37 I already pass `algoOp` that is created before the `Deployer`. This means that Deployer doesn't need to create `CompileOp` itself.

packages/algob/src/lib/asc.ts
```diff
@@ -0,0 +1,33 @@
...
+  if (params.firstRound === 0) {
+    throw new Error("Suggested params returned 0 as firstRound. Ensure that your node progresses.");
+    // params.firstRound = 1
+  }
+  return params;
+}
+
+export async function getSuggestedParamsWithUserDefaults (
+  algocl: algosdk.Algodv2, userDefaults: ASCDeploymentFlags): Promise<algosdk.SuggestedParams> {
+  const suggested = await getSuggestedParams(algocl);
```
Invertisment, line 19: Are these params from the spec? You simply copied my code for ASA.

packages/algob/src/lib/asc.ts
```diff
@@ -0,0 +1,33 @@
+import algosdk from "algosdk";
+
+import {
+  ASCDeploymentFlags
+} from "../types";
+
+async function getSuggestedParams (algocl: algosdk.Algodv2): Promise<algosdk.SuggestedParams> {
```
Invertisment, line 7: Do not duplicate the functions. This already exists in `asa.ts`. Why would you duplicate the logic? Reuse. 
Move logic from `asa.ts` to a different file and use it from both places.

packages/algob/src/lib/algo-operator.ts
```diff
@@ -60,4 +70,56 @@ export class AlgoOperatorImpl implements AlgoOperator {
       confirmedRound: txConfirmation[confirmedRound]
     };
   }
+
+  async deployASC (programb64: string, scParams: object, flags: ASCDeploymentFlags
+  ): Promise<ASCInfo> {
+    const program = new Uint8Array(Buffer.from(programb64, "base64"));
+    const encoder = new TextEncoder();
```
Invertisment, line 77: Create where it's used, not at the top. You don't use it in every line.

packages/algob/src/lib/algo-operator.ts
```diff
@@ -60,4 +70,56 @@ export class AlgoOperatorImpl implements AlgoOperator {
...
+    const lsig = algosdk.makeLogicSig(program, scParams);
+
+    const params = await getSuggestedParamsWithUserDefaults(this.algodClient, flags);
+
+    // ASC1 signed by funder
+    lsig.sign(flags.creator.sk);
+    const funder = flags.creator.addr;
+    const contractAddress = lsig.address();
+    // const receiver = funder;
```
Invertisment, line 86: remove

packages/algob/src/lib/algo-operator.ts
```diff
@@ -60,4 +70,56 @@ export class AlgoOperatorImpl implements AlgoOperator {
...
+    const tran = algosdk.makePaymentTxnWithSuggestedParams(funder, contractAddress,
+      flags.microAlgo, flags.closeToRemainder,
+      flags.note ? encoder.encode(flags.note) : undefined,
+      params);
+
+    const signedTxn = tran.signTxn(flags.creator.sk);
+
+    const tranInfo = await this.algodClient.sendRawTransaction(signedTxn).do();
+
+    const confirmedTxn = await this.algodClient.pendingTransactionInformation(tranInfo.txId).do();
```
Invertisment, line 100: I think this doesn't wait for tx confirmation.
Does it call `waitForConfirmation`?

packages/algob/src/lib/deployer.ts
```diff
@@ -85,13 +89,14 @@ export class AlgobDeployerImpl implements AlgobDeployer {
     return this.cpData.precedingCP[this.networkName].asa[name];
   }
 
-  async deployASC (name: string, source: string, account: Account): Promise<ASCInfo> {
+  async deployASC (name: string, scParams: Object, flags: ASCDeploymentFlags): Promise<ASCInfo> {
+    const result: ASCCache = await this.algoOp.ensuredCompiled(name, false);
```
Invertisment, line 93: You compile the file before you check that the code is deployed (`assertNoAsset`). Why would you recompile when it's deployed?
Move this into algoOp and remove `ensuredCompiled` from `algoOp`

packages/algob/src/types.ts
amityadav0: scParams are the initialization parameters for contract that will be defined by user,

and flags are transaction parameters that are required to fund and makeLogic signature for contract.

packages/algob/src/lib/deployer.ts
amityadav0: Ok

packages/algob/src/lib/deployer.ts
amityadav0: scParams will be defined by user, how do i classify it in a type ?

packages/algob/src/lib/algo-operator.ts
amityadav0: No, this is what i found in algorand docs for making a payment transaction

packages/algob/src/lib/algo-operator.ts
amityadav0: ok

packages/algob/src/lib/asc.ts
amityadav0: I was getting fail transaction when i was using the tx logic, but using functions directly from algosdk works, but by calling tx.function() transaction was failing

packages/algob/src/lib/deployer.ts
Invertisment: Ok. Then it's fine.

packages/algob/src/types.ts
Invertisment: Let's then not have underscores

packages/algob/test/fixture-projects/scripts-dir-recursive-cp/scripts/2.js
Invertisment: It's needed for scParams

packages/algob/test/mocks/deployer.ts
Invertisment: needed

packages/algob/src/types.ts
amityadav0: ok

packages/algob/src/lib/algo-operator.ts
Invertisment: But the pending transaction will still have to be waited to be confirmed. It should be the same for all transactions.

packages/algob/test/stubs/algo-operator.ts
amityadav0: ok

packages/algob/src/lib/asc.ts
amityadav0: yes, These parameters are for payment transaction which is same for all,
- for funding smart contract from creator account.

packages/algob/src/lib/asc.ts
Invertisment: Then there is some kind of a difference between these two code chunks. What is it?

amityadav0: asa.ts code calls function from tx.ts file, and asc.ts calls function from algosdk.
and tx file is in lib folder

packages/algob/src/lib/algo-operator.ts
amityadav0: ok

packages/algob/src/lib/deployer.ts
amityadav0: yes

packages/algob/src/lib/asc.ts
Invertisment: So why couldn't you reuse `getSuggestedParams` and added your own? Why do you need two implementations that are exactly the same?

amityadav0: asa.ts calls function from tx.ts and using same for ASC was giving me genesis hash error for this transaction, so i checked the algorand docs, they were using algocl.getsuggestedparams() instead of tx.getsuggestedparams() and that worked.

amityadav0: these are not exactly same

amityadav0: the parameters that is passed one is tx.AlgoV2, and other one is algosdk.AlgoV2

Invertisment: Really...?
`import tx from "algosdk";`
`import algosdk from "algosdk";`
You import the whole library and name it `algosdk` so they are the exact same thing

Invertisment: It it would be this:
```
import { tx } from "algosdk";
import { algosdk } from "algosdk";
```
then they would be different. But they are not imported this way.

amityadav0: ` async function getSuggestedParams (algocl: tx.Algodv2): Promise<tx.SuggestedParams> {
  const params = await algocl.getTransactionParams().do();
  // Private chains may have an issue with firstRound
  if (params.firstRound === 0) {
    throw new Error("Suggested params returned 0 as firstRound. Ensure that your node progresses.");
    // params.firstRound = 1
  }
  return params;
} `

Yes, but
I was getting gensis hash error from this function when i debugged, 
I will check again ..

packages/algob/src/lib/algo-operator.ts
robert-zaremba, line 86: let's remove this comment.

packages/algob/src/lib/algo-operator.ts
robert-zaremba, line 77: are you sure we should use `TextEncoder` ( thought we need to it as base64, but I'm not sure).

packages/algob/src/lib/asc.ts
amityadav0: got it, removed the double implementation

robert-zaremba: Sorry, i'm not getting too much into this comments, but I agree with Martynas, that if possible, would be great to no duplicate the code. PS: You can always update other code ;)

packages/algob/src/lib/deployer.ts
robert-zaremba: this is exactly what this code is doing.

robert-zaremba: BTW: the `force` argument should be provided by user in a command line (don't assume it's `false`)

packages/algob/src/types.ts
robert-zaremba: Martynas is correct. You merged 2 parameters into one:

Everything below `microAlgo` should go into the `payTxParams`:
https://paper.dropbox.com/doc/Algorand-builder-specs-Vcdp0XNngizChyUWvFXfs#:uid=855054373483415843735838&h2=Smart-Contracts-deployment

packages/algob/src/lib/algo-operator.ts
amityadav0: Yes , added

packages/algob/src/types.ts
robert-zaremba, line 380: referring to the comment above, we are missing one Object parameter here.

packages/algob/src/lib/algo-operator.ts
amityadav0: this is for the "note"  parameter

packages/algob/src/lib/algo-operator.ts
amityadav0: ok

packages/algob/src/lib/deployer.ts
amityadav0: I moved this part to algoOp, that is after assertNoAsset

packages/algob/src/types.ts
amityadav0: ok, i will make two parameters

packages/algob/src/types.ts
amityadav0: Done

packages/algob/src/lib/algo-operator.ts
amityadav0: program is need as base64, but if a user gives notes then we need to convert it first then only we can pass it to transaction function

packages/algob/src/lib/algo-operator.ts
```diff
@@ -19,14 +25,18 @@ export function createAlgoOperator (network: Network): AlgoOperator {
 export interface AlgoOperator {
   algodClient: algosdk.Algodv2
   deployASA: (name: string, asaDesc: ASADef, flags: ASADeploymentFlags, account: Account) => Promise<ASAInfo>
+  deployASC: (programb64: string, scParams: object, flags: ASCDeploymentFlags, payFlags: ASCPaymentFlags,
+    account: Account) => Promise<ASCInfo>
   waitForConfirmation: (txId: string) => Promise<algosdk.ConfirmedTxInfo>
+  ensuredCompiled: (name: string, force: boolean) => Promise<ASCCache>
```
Invertisment, line 31: This is not used anywhere except your own code when you can use private method and not have it in interface.

packages/algob/src/lib/algo-operator.ts
```diff
@@ -60,4 +70,57 @@ export class AlgoOperatorImpl implements AlgoOperator {
...
+      params);
+
+    const signedTxn = tran.signTxn(flags.funder.sk);
+
+    const tranInfo = await this.algodClient.sendRawTransaction(signedTxn).do();
+
+    const confirmedTxn = await this.waitForConfirmation(tranInfo.txId);
+    // console.log("Transaction information: %o", confirmedTxn);
+
+    /* const amount = 100;
```
Invertisment, line 104: All of these comments are not needed anymore

packages/algob/src/lib/deployer.ts
```diff
@@ -85,13 +87,12 @@ export class AlgobDeployerImpl implements AlgobDeployer {
     return this.cpData.precedingCP[this.networkName].asa[name];
   }
 
-  async deployASC (name: string, source: string, account: Account): Promise<ASCInfo> {
+  async deployASC (name: string, scParams: Object, flags: ASCDeploymentFlags,
+    payFlags: ASCPaymentFlags): Promise<ASCInfo> {
+    const creator = flags.funder;
```
Invertisment, line 92: You define a variable just to use it in next line and then you also pass `flags` too which has this parameter.

packages/algob/src/lib/tx.ts
```diff
@@ -34,6 +35,21 @@ async function getSuggestedParamsWithUserDefaults (
   return suggested;
 }
 
+export async function getSuggestedParamsWithUserDefaultsASC (
+  algocl: tx.Algodv2, userDefaults: ASCPaymentFlags): Promise<tx.SuggestedParams> {
+  const suggested = await getSuggestedParams(algocl);
+  suggested.fee = userDefaults.totalFee === undefined
```
Invertisment, line 41: Please reuse the method from `asa.ts`. As the fields are almost all the same you can use all of it or most of it.
If you don't need a `creator` field then create a type that omits this field (typescript can create this type).
You also have `DeploymentFlags` so you can change asa.ts method to use it and remove this method.

packages/algob/src/types.ts
```diff
@@ -341,14 +343,29 @@ export interface Checkpoint {
...
+export interface DeploymentFlags {
   totalFee?: number
-  feePerByte?: number
   firstValid?: number
   validRounds?: number
 }
 
+export interface ASADeploymentFlags extends DeploymentFlags {
+  creator: Account
+  feePerByte?: number
```
Invertisment, line 354: Consult with Robert to use this parameter.

packages/algob/src/lib/algo-operator.ts
amityadav0: This is also used in AlgoOperatorDryRunImpl to return mock file while testing

packages/algob/src/lib/algo-operator.ts
amityadav0: ok

packages/algob/src/lib/deployer.ts
amityadav0: deployASA was used for reference , that has same structure

packages/algob/src/lib/tx.ts
amityadav0: do you mean tx.ts ?
I did that and used getSuggestedParams common in both, but this function can't be common as flag types are different for ASC and ASA

amityadav0: and deleted asc.ts

packages/algob/src/lib/deployer.ts
Invertisment: But you pass the params twice. You don't need creator parameter in `this.algoOp.deployASC(name, scParams, flags, creator)`

packages/algob/src/lib/tx.ts
Invertisment: I know they are a little different. But you should then create a third function that has same flag processing logic and have three functions in total.
This will reduce code duplication.

packages/algob/src/lib/deployer.ts
amityadav0: ok

packages/algob/src/lib/algo-operator.ts
Invertisment: But it's not useful outside of this. It's not useful code. Nothing new happens when it exists.

packages/algob/src/lib/tx.ts
amityadav0: I think that would still increase the code as the code will have third type with same parameters , and will have to initialize this third with some parameters from both of them, this all for only four parameters.

I can do that if you think that is good.

packages/algob/src/lib/algo-operator.ts
amityadav0: But then how will I pass the mock files and test the code , the initial problem (the hardcoded network) will occur.

amityadav0: Done

packages/algob/src/lib/tx.ts
Invertisment: You already have that third type: `DeploymentFlags`. So you can use it here.

packages/algob/src/types.ts
Invertisment: This parameter should be used in the same way as in ASA flags. And then you have to move it to `DeploymentFlags`.

packages/algob/src/lib/tx.ts
amityadav0: Yes, understood, this is better

amityadav0: Done

packages/algob/src/types.ts
amityadav0: This was not in specs , there is only one fee/flat  fee in specs

packages/algob/src/lib/algo-operator.ts
```diff
@@ -60,4 +69,45 @@ export class AlgoOperatorImpl implements AlgoOperator {
       confirmedRound: txConfirmation[confirmedRound]
     };
   }
+
+  async deployASC (name: string, scParams: object, flags: ASCDeploymentFlags, payFlags: ASCPaymentFlags
+  ): Promise<ASCInfo> {
+    console.log("Deploying ASC:", name);
+    const result: ASCCache = await this.ensuredCompiled(name, false);
```
Invertisment, line 76: Rename `ensuredCompiled` to `ensureCompiled`. This is a typo.

packages/algob/src/lib/algo-operator.ts
```diff
@@ -60,4 +69,45 @@ export class AlgoOperatorImpl implements AlgoOperator {
...
+      creator: flags.funder.addr,
+      contractAddress: contractAddress,
+      txId: tranInfo.txId,
+      logicSignature: lsig,
+      confirmedRound: confirmedTxn[confirmedRound]
+    };
+  }
+
+  private async ensuredCompiled (name: string, force: boolean): Promise<ASCCache> {
+    return await this.compileOp.ensureCompiled(name, false);
```
Invertisment, line 111: change `false` to `force`. Let's pass here as method expects.

packages/algob/src/types.ts
amityadav0: what do i do ? include it or not ?

packages/algob/src/lib/algo-operator.ts
Invertisment, line 76: `force` parameter is `false` but it should be taken from input. We don't have a global `--force` parameter but we may need to have it later. I think we should add a task to add the parameter.

packages/algob/src/types.ts
amityadav0: Done

Invertisment:
Fix `force` parameter and Rename `ensuredCompiled` to `ensureCompiled`. Then you can merge.

packages/algob/src/lib/algo-operator.ts
amityadav0: ok

Invertisment (approved):

packages/algob/src/lib/algo-operator.ts
amityadav0: ok

packages/algob/src/types.ts
robert-zaremba: In the spec this is under:
```
  payTxParams: Object // other raw payment transaction parameters 
```

This has to be compatible as with all other transactions.

amityadav0: Yes, it is same for both for paymentParams

PR #74, opened 2020-09-04
Store output of transaction

amityadav0:


packages/algob/src/lib/files.ts
```diff
@@ -64,3 +65,10 @@ export function loadFromYamlFileSilentWithMessage (
     return defaultYamlValue(options);
   }
 }
+
+export function writeTransaction (transactionConfirmation: Object): void {
+  fs.appendFileSync(
+    path.join(artifactsPath, "transactionInfo.txt"),
+    YAML.stringify(transactionConfirmation)
+  );
+}
```
robert-zaremba, line 74: let's remove this function. We need more things and this is scheduled for other task: https://www.pivotaltracker.com/story/show/174085665

FYI: we will write the output of all transactions in a file with filename related to the script filename. 
So we need more abstraction here (probably a `TxLogWriter` class, which will  be part of the deployer and can be used by a user as well.
Here let's only do printing to a console

packages/algob/src/lib/script-checkpoints.ts
```diff
@@ -17,7 +17,7 @@ import {
 import { loadFromYamlFileSilent } from "./files";
 
 export const scriptsDirectory = "scripts";
-const artifactsPath = "artifacts";
+export const artifactsPath = "artifacts";
```
robert-zaremba, line 20: Please import path from `src/internal/core/project-structure`

robert-zaremba (changes requested):

packages/algob/src/lib/files.ts
amityadav0: ok

robert-zaremba (approved):

PR #75, opened 2020-09-04
Asa opt-in with balance check

Invertisment:


packages/algob/src/builtin-tasks/run.ts
```diff
@@ -69,11 +75,13 @@ function loadCheckpointsIntoCPData (cpData: CheckpointRepo, scriptPaths: string[
...
-// TODO: Reduce file IO:
-// Function only accepts sorted scripts -- only this way it loads the state correctly.
-// Optimization: Split the scripts into sorted array chunks and run those chunks
-// This will save some disk reads because sub-arrays will be sorted
-export async function runMultipleScriptsOneByOne (
+function batchScriptNames (unbatched: string[]): string[][] {
+  return partitionByFn(
+    (a: string, b: string) => cmpStr(a, b) === 1,
+    unbatched);
+}
```
robert-zaremba, line 82: Please rename:

* batchScriptNames -> groupSameStrings
* partitionByFn -> groupByFn

packages/algob/src/builtin-tasks/run.ts
```diff
@@ -82,31 +90,32 @@ export async function runMultipleScriptsOneByOne (
...
+    await runScriptsBatch(
       runtimeEnv,
-      [script],
+      scriptsBatch,
       onSuccessFn,
       force,
       logDebugTag,
-      allowWrite,
-      algoOp
+      (cpData: CheckpointRepo) => mkDeployer(runtimeEnv, cpData, allowWrite, algoOp, asaDefs, accounts)
```
robert-zaremba, line 102: 1. The function doesn't change, so no point to redefine the function in the loop.
2. Can we get rid of this function? It seams that we don't need it - we can pass other arguments here (we can even use object to group them)

packages/algob/src/internal/util/lists.ts
```diff
@@ -0,0 +1,27 @@
+
+function getLast<T> (ts: T[]): T {
+  return ts[ts.length - 1];
+}
+
+function getLastSubchild<T> (ts: T[][]): T {
```
robert-zaremba, line 6: Please add a comment or rename.
It's not clear without reading the code what the function is doing (what is a subchild)?

packages/algob/src/internal/util/lists.ts
```diff
@@ -0,0 +1,27 @@
...
+  for (const current of input.slice(1)) {
+    const last = getLastSubchild(out);
+    if (f(last, current)) {
+      out.push([current]);
+    } else {
+      out[out.length - 1].push(current);
+    }
+  }
+  return out;
+}
```
robert-zaremba, line 27: this is not tested, please add tests.

packages/algob/src/lib/algo-operator.ts
```diff
@@ -18,13 +21,18 @@ import * as tx from "./tx";
 
 const confirmedRound = "confirmed-round";
 
+// This was not exported in algosdk
+const ALGORAND_MIN_TX_FEE = 1000;
+// Extracted from interaction with Algorand node
+const ALGORAND_ASA_OWNERSHIP_COST = 100000;
```
robert-zaremba, line 27: please add a comment `// 100k microAlgos`

packages/algob/src/lib/algo-operator.ts
```diff
@@ -54,18 +62,103 @@ export class AlgoOperatorImpl implements AlgoOperator {
...
+
+  async optInToASA (
+    asaName: string, assetIndex: number, account: Account, params: algosdk.SuggestedParams
+  ): Promise<void> {
+    console.log(`ASA ${account.name} opt-in for for ASA ${asaName}`);
+    const sampleASAOptInTX = tx.makeASAOptInTx(account.addr, assetIndex, params);
+    const rawSignedTxn = sampleASAOptInTX.signTxn(account.sk);
+    const txInfo = await this.algodClient.sendRawTransaction(rawSignedTxn).do();
+    await this.waitForConfirmation(txInfo.txId);
+  }
```
robert-zaremba, line 99: This function must be also exported to user through deployer, however the deployer version should take as a parameter `TxParams` instead of `algosdk.SuggestedParams`.

packages/algob/src/lib/algo-operator.ts
```diff
@@ -54,18 +62,103 @@ export class AlgoOperatorImpl implements AlgoOperator {
...
+
+  async optInToASAMultiple (
+    asaName: string, assetIndex: number, accounts: Account[], params: algosdk.SuggestedParams
+  ): Promise<void> {
+    for (const account of accounts) {
+      await this.optInToASA(asaName, assetIndex, account, params);
+    }
+  }
+
+  async checkASAOptInAccountBalances (
```
robert-zaremba, line 109: let's rename to `checkBalanceForOptInTx`

packages/algob/src/lib/asa.ts
```diff
@@ -5,25 +5,45 @@ import { BuilderError } from "../internal/core/errors";
...
+      throw new BuilderError(
+        ERRORS.SCRIPT.ASA_PARAM_ERROR_NO_NAMED_OPT_IN_ACCOUNT, {
+          filename: filename,
+          optInAccName: accName
+        });
+    }
+  }
+}
+
+function validateASADefsOneByOne (accounts: Accounts, asaDefs: ASADefs, filename: string): void {
```
robert-zaremba, line 27: We don't need to be that verbose, let's rename function  to `validateASADefs` (or feel free to propose alternative)

packages/types-algosdk/index.d.ts
```diff
@@ -242,4 +243,29 @@ declare module 'algosdk' {
 
   type TxnBytes = Uint8Array
 
+  type Address = string
```
robert-zaremba, line 246: do we need this? We already use `string` everywhere.

packages/types-algosdk/index.d.ts
```diff
@@ -242,4 +243,29 @@ declare module 'algosdk' {
...
+    'pending-rewards': number
+    'reward-base': number
+    rewards: number
+    round: number
+    status: string
+    'apps-local-state': any
+    'apps-total-schema': any
+    'created-apps': any
+    'created-assets': any
+  }
```
robert-zaremba, line 269: :+1:

robert-zaremba (changes requested):

packages/algob/src/internal/util/lists.ts
```diff
@@ -0,0 +1,27 @@
...
+function getLast<T> (ts: T[]): T {
+  return ts[ts.length - 1];
+}
+
+function getLastSubchild<T> (ts: T[][]): T {
+  return getLast(getLast(ts));
+}
+
+export function partitionByFn<T> (
+  f: (s: T, s1: T) => boolean,
```
amityadav0, line 11: What does this function do ?

robert-zaremba:
It groups by a predicate function

packages/algob/src/internal/util/lists.ts
Invertisment: It takes a list which has some kind of order
then it also takes a function that checks whether the list should be split there. If it returns true then it should split.
then this:
`f: >` and `[c, d, a, b]`
will return:
`[[c d], [a b]]`

packages/algob/src/lib/algo-operator.ts
Invertisment: This function also takes assetID which is too verbose as well.
I'll make a new function that will prepare the parameters for this one and won't touch this function.

Invertisment: I'll use this as a signature for `AlgoOperator`:
`optInToASA (asaName: string, assetIndex: number, account: Account, params: DeploymentFlags): Promise<void>`
And this for `Deployer`:
`optInToASA (asaName: string, account: Account, params: DeploymentFlags): Promise<void>`

It will also need to check that `asaName` has been saved in CP

Invertisment: Deployer's signature changed to
`optInToASA (name: string, accountName: string, flags: ASADeploymentFlags): Promise<void>`
`accountName` is better than `Account`

Invertisment: Example updated. `tesla` will do the manual opt-in.

packages/algob/src/builtin-tasks/run.ts
robert-zaremba: Re 1 I would prefer `groupIntoSorted`

Re 2. You are using a `partitionByFn` function name, which takes an argument. The function itself doesn't know anything about the order. If you would use as an argument `(a, b) => cmpStr(a, b) === 0)` then it wouldn't imply any order.

packages/algob/src/builtin-tasks/run.ts
robert-zaremba: My point is that we don't even need to use a function here.

packages/algob/src/internal/util/lists.ts
robert-zaremba: To be honest I don't have good proposal here ;)

packages/algob/src/lib/algo-operator.ts
robert-zaremba: Why `DeploymentFlags` instead of `TxParams` (isn't the base interface enough)?

packages/algob/src/lib/algo-operator.ts
robert-zaremba: OK, but we don't need `ASA` infix (there no other thing we can opt-it).

robert-zaremba: So we could have it a bit shorter ;)

packages/algob/src/builtin-tasks/run.ts
Invertisment: Then the function would split when it will encounter not a smaller element but equal one:
`[1 2 2 3 4 3 2 1]`
would return
`[[1 2] [2 3 4 3 2 1]]`.
And as it runs sequentially from left to right it preserves the order which is important for me. This is not grouping and more of a splitting of a list. There is no sorting so sorted doesn't belong here. The order is not changed.

packages/algob/src/builtin-tasks/run.ts
Invertisment: What else would you want to use to preserve loaded `runtimeEnv` and other arguments when `cpData` is reloaded inside of `runScript`?

packages/algob/src/lib/algo-operator.ts
Invertisment: It looks like I have an old code. `DeploymentFlags` was the old name of `TxParams`

packages/algob/src/lib/algo-operator.ts
Invertisment: `checkBalanceForOptInTx`

packages/algob/src/builtin-tasks/run.ts
robert-zaremba: OK, clear. Please add a comment to the function with example.

packages/algob/src/builtin-tasks/run.ts
robert-zaremba: We can pass arguments (`env`, `allowWrite`) and construct the `deployer` in `runScriptsBatch`.

Invertisment: Ok.
Added 4 parameters to the method to not create a function:
```
  allowWrite: boolean,
  algoOp: AlgoOperator,
  asaDefs: ASADefs,
  accounts: Accounts
```
Now the method takes 9 parameters of which 4 is exclusively used for `mkDeployer` function.

robert-zaremba: Why you don't put this parameters for `mkDeployer` in an object as I suggested above? (in the first post)

packages/algob/src/internal/util/lists.ts
Invertisment: This is then later passed to the CP loading method which loads all preceding checkpoints:
For instance I have CPs `1 2 3 4`.
Then I want to execute scripts `2 3 1` which will produce an order of: 
`[[2 3] [1]]`
Then `[2 3]` will:
load CPs `1, 2` then execute script `2`. Then load `3` and execute `3`.
The last `[1]` part will load CP `1` and execute script `1`

packages/algob/src/builtin-tasks/run.ts
Invertisment: Added this:
```
/** Partitions an unsorted string list into sorted parts:
    `[1 2 2 3 4 3 4 2 1]` returns `[[1 2 2 3 4] [3 4] [2] [1]]` */
function partitionIntoSorted (unsorted: string[]): string[][] {
  return partitionByFn(
    (a: string, b: string) => cmpStr(a, b) === 1, // split when a > b
    unsorted);
}
```

packages/algob/src/builtin-tasks/run.ts
Invertisment: Then we'll need a type and I preferred to use functions so I used them. `runSortedScripts` doesn't need to know everything about internals of deployer so I prefer to remove what is not needed by it and make `runSortedScripts` not care instead of having huge objects that can be interacted from every place.
It's a viable strategy both ways but I preferred to know less where I don't need to know.

Invertisment: And even when we'll add an object then either:
- Deployer.constructor will take it as a parameter (probably even add it to `types.ts` and make it public to everything)
- Destructure the object inside of the method but then everything is still visible to the method.
- Curried function (I used this way to have only one param instead of several)
- Pass everything -- Equivalent to object but without a type and without showing it to the whole app.
These ways are fine depending on the style that you prefer to use.

robert-zaremba: With function you need both: a function and a types. 

Let's group arguments in an object: `DeployerConfig`, and store it `src/lib/deployer.ts`. This way the function will not need to know details, and will just pass the config, without having a responsibility to call external functions.

packages/algob/src/internal/util/lists.ts
amityadav0: Okay, understood,

amityadav0: what if I have CPs ` 1 2 3 4 5 6 7 ` and 

I want execute ` 6 7 5 4 2 3 1 ` 

will this be splited recursively ?

amityadav0: Got it , ` partitionByFn `

packages/algob/src/builtin-tasks/run.ts
Invertisment: You didn't understand what I meant by "know types".
There are two functions:
Outer one:`runMultipleScripts` and inner one: `runSortedScripts`.
The outer one does some IO and creates some variables.
The inner one does only use CPRepo and has to load it inside. It also needs Deployer to send it to the script method. That's it. It doesn't need to care about any other variables so I encapsulated it in a function.

The function type is simpler than the type which you propose because it takes only CPRepo and produces Deployer.
What you propose is to have a type with 4 or more fields that are still visible to the inner function.
It could be that this type signals that we have too much logic in Deployer already and we should limit it.

Now what you want is not only to show the variables to the inner function but to create a larger type in a completely different file to solve the issue with first file. If we will end up having 20 variables in DeployerImpl's constructor then it's probably an issue with DeployerImpl class and not with this function.

And the object will still need to be created so my function solution would still have worked in the exact same way.
Except instead of having a type with four or more fields I had a type with input and output. 
And my top-level `mkDeployer` function has more fields too.

Invertisment: I alread restored the code to the method with many variables. And for now the DeployerImpl's constructor doesn't have many variables so the type there is not needed.

packages/algob/src/internal/util/lists.ts
amityadav0:
```

export function partitionByFn<T> (
  f: (s: T, s1: T) => boolean,
  input: T[]
): T[][] {
  if (input.length === 0) {
    return [];
  }
  var out = [[input[0]]];
  let last = input[0];
  for (const current of input.slice(1)) {
    if (f(last, current)) {
      out.push([current]);
    } else {
      out[out.length - 1].push(current);
    }
    last = current;
  }
  return out;
}
```

amityadav0: can we change it to something like this ?
and remove the getLast and getLastSubChild functions ?

amityadav0: will this work ?

packages/algob/src/internal/util/lists.ts
Invertisment: `6 7 5 2 3 1` ->
[[6 7] [5] [2 3] [1]]

packages/algob/src/internal/util/lists.ts
Invertisment: It will work but it will be even harder to understand what it does.
And it will be faster.

packages/algob/src/builtin-tasks/run.ts
robert-zaremba: >  I preferred to know less where I didn't need to know.

I agree

robert-zaremba: Merge it and I will update it later, because it is taking too long.

packages/types-algosdk/index.d.ts
```diff
@@ -242,4 +243,29 @@ declare module 'algosdk' {
 
   type TxnBytes = Uint8Array
 
+  type Address = string
+
+  interface AccountAssetInfo {
+    amount: number,
+    'asset-id': number,
+    creator: Address,
+    'is-frozen': false
```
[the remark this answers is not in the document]
robert-zaremba: Did you change this?

robert-zaremba (approved):

robert-zaremba:
Make sure to check not resolved comments before merging.

packages/types-algosdk/index.d.ts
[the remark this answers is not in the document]
Invertisment: 'is-frozen': boolean

PR #81, opened 2020-09-08
Sphinx text-based docs

Invertisment:


robert-zaremba:
What are we going to do with this PR?

robert-zaremba:
If we will need it, then let's move sphinx docs to `/docs-sphinx`

PR #84, opened 2020-09-10
Unify deployASC and deployASA as per SPEC

amityadav0:


packages/algob/src/lib/tx.ts
```diff
@@ -36,11 +36,25 @@ export async function mkSuggestedParams (
 export function makeAssetCreateTxn (
   name: string, asaDef: ASADef, flags: ASADeploymentFlags, txSuggestedParams: tx.SuggestedParams
 ): tx.Transaction {
+  // Load Note
+  // Load payFlags.note (ignored if payFlags.noteb64 is present)
+  // undefined if none of them is present.
+  // If TxParams has noteb64 or note , it gets precedence
```
robert-zaremba, line 42: let's keep only the last one (and remove lines 39-41)

packages/algob/src/lib/tx.ts
```diff
@@ -36,11 +36,25 @@ export async function mkSuggestedParams (
...
   const encoder = new TextEncoder();
+  let note;
+  if (flags.noteb64 ?? flags.note) {
+    // TxParams note
+    note = flags.noteb64 ? encoder.encode(flags.noteb64) : encoder.encode(flags.note);
+  } else {
+    // ASA definition note
+    note = asaDef.noteb64 ? encoder.encode(asaDef.noteb64)
+      : (asaDef.note ? encoder.encode(asaDef.note) : undefined);
```
robert-zaremba, line 51: let's create and use a function (because we have a repetition )
```
note = decodeNote(note, note64)
```

robert-zaremba: Also, please add a test to that function.

robert-zaremba (approved):
Approving, but left 2 change requests.

packages/algob/src/lib/tx.ts
amityadav0: ok

packages/algob/src/lib/tx.ts
amityadav0: did you mean encodeNote ?

amityadav0: Done created encodeNote function with test

PR #86, opened 2020-09-15
docs update

robert-zaremba:
Small documentation updates. 

This is to assure the right order and cleaner organization.


amityadav0 (approved):
Looks Good

PR #90, opened 2020-10-07
remove fp-ts and io-ts from package.json

amityadav0:


packages/algob/src/internal/core/config/config-validation.ts
```diff
@@ -155,26 +99,30 @@ export function getValidationErrors(config: any): CfgErrors {  // eslint-disable
...
 
-      const netConfigResult = HttpNetworkType.decode(hcfg);
-      if (isLeft(netConfigResult)) {
-        errors.push(net, "", hcfg, "HttpNetworkConfig");
+      try {
+        HttpNetworkType.parse(hcfg);
+      } 
+      catch(e) {
+        if (e instanceof z.ZodError) {
+          console.log(parseZodError(e));
```
robert-zaremba, line 107: please use `console.error` here

robert-zaremba (approved):
One change request.

packages/algob/src/internal/core/config/config-validation.ts
amityadav0: ok , this line is not needed , deleting it

PR #91, opened 2020-10-08
rename 'deployer' classes

amityadav0:


robert-zaremba (approved):
Approving, but when reading I've got an idea: how about removing `Algob` prefix? We are in the algob context, so maybe it's not necessary to prefix it.

amityadav0:
yes that would be good

PR #95, opened 2020-10-12
update name

amityadav0:


packages/algob/src/lib/status.ts
```diff
@@ -14,7 +14,7 @@ export async function balanceOf (
   }
 };
 
-export async function printAssets (deployer: AlgobDeployerImpl, account: string): Promise<void> {
+export async function printAssets (deployer: DeployerDeployMode, account: string): Promise<void> {
```
robert-zaremba, line 17: This can be any deployer, so you should use the interface here.

robert-zaremba (changes requested):

packages/algob/src/lib/status.ts
amityadav0: ok

robert-zaremba (approved):

PR #97, opened 2020-10-12
Fix dependency

ratik21:
- move / remove duplicate devDependencies 
- updating cleanup.sh


packages/algob/package.json
```diff
@@ -56,9 +56,11 @@
     "@types/glob": "^7.1.2",
     "@types/semver": "^7.2.0",
     "@types/sinon": "^9.0.4",
+    "@types/chai": "^4.2.12",
     "chai": "^4.2.0",
     "cross-env": "5.0.5",
```
robert-zaremba, line 61: cross-env should be moved to root.

robert-zaremba (approved):
Pre-approving, but left one change request.

packages/algob/package.json
ratik21: okay, fixing.

PR #98, opened 2020-10-13
remove TODO

amityadav0:


packages/algob/test/internal/core/runtime-environment.ts
```diff
@@ -313,23 +313,6 @@ describe("Environment", () => {
         );
       }, ERRORS.NETWORK.CONFIG_NOT_FOUND);
     });
-
-    // TODO:RZ -- network may need to be nullable
-    // or default parameter should be provided as default in CLI
```
amityadav0, line 318: does the network needs to be nullable ?

if not, these lines are removed

amityadav0:
Other subtasks were already done

packages/algob/test/internal/core/runtime-environment.ts
robert-zaremba: We should have a test to check if a default network is selected, when a user doesn't specify any network.

robert-zaremba (changes requested):

packages/algob/test/internal/core/runtime-environment.ts
amityadav0: we cannot test this here, as the code where it happens is in `gatherArguments`
https://github.com/scale-it/algorand-builder/blob/4aa277d437bb70aa947d81c43cac0610b7e0fda2/packages/algob/src/internal/cli/cli.ts#L80

i am adding the test for the same in `arguments-parser.ts` test

amityadav0: this code firsts load default 
https://github.com/scale-it/algorand-builder/blob/4aa277d437bb70aa947d81c43cac0610b7e0fda2/packages/algob/src/internal/core/params/builder-params.ts#L4

and then change Arguments as per cli input from user

robert-zaremba (approved):

robert-zaremba:
Thanks for adding a new test. :+1:

PR #103, opened 2020-10-19
Rename as per SPEC

amityadav0:
- resolve msgpack build:watch error
- rename `delegatedSig` to `mkDelegatedSig` as per SPEC

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -2,7 +2,8 @@
   "compilerOptions": {
     "target": "es2017",
     "lib": [
-      "esnext"
+      "esnext",
+      "dom"
     ],
     "experimentalDecorators": true,
     "esModuleInterop": true,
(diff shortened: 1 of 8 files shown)


tsconfig.json
```diff
@@ -2,7 +2,8 @@
   "compilerOptions": {
     "target": "es2017",
     "lib": [
-      "esnext"
+      "esnext",
+      "dom"
```
robert-zaremba, line 6: This should be removed.

robert-zaremba (approved):
Approving, but please remove `dom` from tsconfig.

tsconfig.json
amityadav0: chaneg
This was added because of `msgpack` build:watch error

amityadav0: If i don't add this i get this error

robert-zaremba: Strange, it seams that `msgpack` has this dependency. OK. Feel free to merge.

PR #104, opened 2020-10-21
move ts-essentials

amityadav0:


amityadav0:
moved `ts-essentials` to devDependencies

`type-is` is not part of dependencies

robert-zaremba:
Where do we use `ts-essentials` ?

amityadav0:
we are using two functions from ts-essentials `DeepReadOnly` and `StrictOmit`

amityadav0:
https://github.com/scale-it/algorand-builder/blob/e8d43ff72fcd744a4bc90b06da6557a497dad404/packages/algob/src/types.ts#L110

https://github.com/scale-it/algorand-builder/blob/e8d43ff72fcd744a4bc90b06da6557a497dad404/packages/algob/src/types.ts#L87

amityadav0:
we can remove this dependency as instead of `DeepReadOnly` we can use `ReadOnly` because we don't need recursive one  for current use case

and same is for `StrictOmit` - we can use `Omit` for current use case

as the only difference is that `StrictOmit` checks strict filters , but for current code `Omit` is enough

amityadav0:
but if in future if used some more complex types , then this will be useful

robert-zaremba:
If you are sure that it's safe to remove it then we can do it. I'm approving the PR.

robert-zaremba (approved):

amityadav0:
OK, removing

PR #105, opened 2020-10-21
Add deployer base class

ratik21:
- Added `DeployerBasicMode` as a parent class for `deployerDeployMode` and `deployerRunMode` to encorporate common functions.

- Updated unit tests (`deployerRunMode` class initialization)

- I have kept the `FakeDeployer` mock, though it's not being used anymore, just for future purposes.


robert-zaremba (approved):
Approving, but I don't see where did you create the base class.

robert-zaremba (approved):

PR #106, opened 2020-10-22
Support PyTeal 

amityadav0:


packages/algob/src/lib/pyteal.ts
```diff
@@ -0,0 +1,36 @@
...
+ */
+export function compilePyTeal (filename: string): string {
+  if (!filename.endsWith(pythonExt)) {
+    throw new Error(`filename "${filename}" must end with "${pythonExt}"`);
+  }
+  const subprocess: SpawnSyncReturns<string> = runPythonScript(filename);
+
+  if (subprocess.stderr) {
+    console.error(subprocess.stderr);
+    throw Error(subprocess.stderr);
```
ratik21, line 32: shouldn't this be `throw new Error(..)` ?

amityadav0: yes

amityadav0:
- As a developer I'm able to run and compile the pyteal code (.py)
- Store the compiled PyTEAL code as a teal in cache
- Add a support to PyCompileOp in Deployer

packages/algob/src/lib/pyteal.ts
```diff
@@ -0,0 +1,36 @@
+import { spawnSync, SpawnSyncReturns } from "child_process";
+import path from "path";
+
+import { ASSETS_DIR } from "../internal/core/project-structure";
+
+const pythonExt = ".py";
```
ratik21, line 6: Above file uses `pyExt`. Let's keep the same here as well.

amityadav0: this file is only for task purpose , i don't see any use of this, this file will be removed,(all code is moved to PyCompileOp)

packages/algob/src/lib/compile.ts
```diff
@@ -85,3 +92,87 @@ export class CompileOp {
...
+  /**
+   * Description: returns TEAL code using pyTeal compiler
+   * @param filename : python filename in assets folder
+   */
+  compilePyTeal (filename: string): string {
+    const subprocess: SpawnSyncReturns<string> = this.runPythonScript(filename);
+
+    if (subprocess.stderr) {
+      console.error(subprocess.stderr);
+      throw new Error(subprocess.stderr);
```
robert-zaremba, line 174: Please use Algob errors. Maybe you will need to add a new error to `src/internal/core/errors-list.ts`

packages/algob/src/lib/compile.ts
```diff
@@ -23,9 +25,14 @@ export class CompileOp {
   // Will throw an exception if the source file doesn't exists.
   // @param filename: name of the TEAL code in `/assets` directory.
   //   (Examples: `mysc.teal, security/rbac.teal`)
-  //   MUST have a .teal extension
+  //   MUST have a .teal or .py extension
   // @param force: if true it will force recompilation even if the cache is up to date.
   async ensureCompiled (filename: string, force: boolean): Promise<ASCCache> {
+    if (filename.endsWith(pyExt)) {
+      const pyCompile = new PyCompileOp(this);
```
robert-zaremba, line 32: Please move this to constructor and store the `pyCompiler` as a class field.

packages/algob/test/builtin-tasks/compile.ts
```diff
@@ -53,18 +53,22 @@ describe("Compile task", () => {
   const cacheDir = path.join("artifacts", "cache");
   const f1 = "asc-fee-check.copy.teal";
   const f2 = "asc-fee-check.teal";
+  const f3 = "gold-asa.py";
+  const f4 = "gold-asa.teal";
```
robert-zaremba, line 57: Please add a `py` suffix to py files, eg: `f3-py`

packages/algob/test/builtin-tasks/compile.ts
```diff
@@ -53,18 +53,22 @@ describe("Compile task", () => {
...
   const fhash = 2374470440; // murmur3 hash for f1 file
 
   it("on first run it should compile all .teal sources", async () => {
     await op.resetAndCompile(false);
 
-    assert.equal(op.timestamp, 2);
+    assert.equal(op.timestamp, 4);
```
robert-zaremba, line 63: why the timestamp changed?

packages/algob/test/builtin-tasks/compile.ts
```diff
@@ -95,8 +99,15 @@ describe("Compile task", () => {
...
+    assert.lengthOf(op.compiledFiles, 4);
+    assert.lengthOf(op.writtenFiles, 4);
+  });
+
+  it("should return TEAL code", async () => {
+    const pyOp = new PyCompileOp(op);
+    const content = fs.readFileSync(path.join(ASSETS_DIR, f4), 'utf8');
+    const res = pyOp.compilePyTeal(f3) + '\n';
+    assert.deepEqual(content.toString(), res);
   });
```
robert-zaremba, line 112: Please add a test for `compileOp.ensureCompiled(somePythonFile)` and check the returned `PyASCCache`.

packages/algob/test/fixture-projects/config-project/assets/gold-asa.py
```diff
@@ -0,0 +1,31 @@
+
+from pyteal import *
+
+"""GOLD Transfer"""
+''' Accepts only if (transaction type is OPT-IN OR (transaction type is asset transfer, 
+    sender is goldOwnerAccount and asset transfer amount is less than equal to 1000 ))'''
+
+asset_amt = Int(1000)
+tmpl_sen = Addr("M7VR2MGHI35EG2NMYOF3X337636PIOFVSP2HNIFUKAG7WW6BDWDCA3E2DA")
```
robert-zaremba, line 9: what is `sen` is not really clear. Let's rename that variable to `tmpl_sender`

robert-zaremba (changes requested):

packages/algob/test/builtin-tasks/compile.ts
amityadav0: ok

packages/algob/test/builtin-tasks/compile.ts
amityadav0: because , instead of 2 files, now 2 more files are passed in `compile`
check123

packages/algob/src/lib/compile.ts
amityadav0: ok

packages/algob/src/lib/compile.ts
amityadav0: ok

packages/algob/test/fixture-projects/config-project/assets/gold-asa.py
amityadav0: ok

packages/algob/test/builtin-tasks/compile.ts
amityadav0: Ok, but for that `Algodv2` will be needed to compile the TEAL code, as `PyCompileOp` uses same `CompileOp` compiler functions for compiling Teal Files
should i create it ?

robert-zaremba: Yes.

packages/algob/test/builtin-tasks/compile.ts
robert-zaremba: ok

packages/algob/test/builtin-tasks/compile.ts
amityadav0: ok

amityadav0: This has one problem, we will need network config for creating `Algodv2`. 
we can take it from `AlgobRuntimeEnv` but we will need to have a network started before running tests else tests will fail.
is this ok ?
algob

amityadav0: network config will be different for different systems, will have to change it everytime we do test in different system

packages/algob/src/internal/core/errors-list.ts
```diff
@@ -29,7 +29,8 @@ export const ERROR_RANGES = {
   BUILTIN_TASKS: { min: 600, max: 699, title: "Built-in tasks errors" },
   PLUGINS: { min: 800, max: 899, title: "Plugin system errors" },
 
-  SCRIPT: { min: 900, max: 999, title: "Script related errors" }
+  SCRIPT: { min: 900, max: 999, title: "Script related errors" },
+  PYTHON: { min: 1000, max: 1099, title: "Python (PyTeal) related errors" }
```
robert-zaremba, line 33:
```suggestion
  PyTEAL: { min: 450, max: 460, title: "Python (PyTeal) related errors" }
```

And you will have to update the numbers in the new errors accordingly.

packages/algob/src/internal/core/errors-list.ts
```diff
@@ -741,5 +752,6 @@ export const ERRORS: {
   ALGORAND: algorandNode,
   BUILTIN_TASKS: taskErrors,
   PLUGINS: pluginErrors,
-  SCRIPT: scriptErrors
+  SCRIPT: scriptErrors,
+  PYTHON: pyTealErrors
```
robert-zaremba, line 756:
```suggestion
  PyTEAL: pyTealErrors
```

packages/algob/test/builtin-tasks/compile.ts
```diff
@@ -95,8 +100,22 @@ describe("Compile task", () => {
...
+    const content = fs.readFileSync(path.join(ASSETS_DIR, f4), 'utf8');
+    const res = pyOp.compilePyTeal(f3PY) + '\n';
+    assert.deepEqual(content.toString(), res);
+  });
+
+  it("should ", async () => {
+    const pyOp = new PyCompileOp(op);
+    const result: PyASCCache = await pyOp.ensureCompiled(f3PY, false);
+    const expected = fs.readFileSync(path.join(ASSETS_DIR, 'gold-asa-py-check.yaml'), 'utf8');
+    assert.deepEqual(YAML.stringify(result), expected);
```
robert-zaremba, line 119: add a check that `op.compile` was called. You have examples how to do it in previous tests.

robert-zaremba (changes requested):
Few more things to change.

packages/algob/test/builtin-tasks/compile.ts
amityadav0: adding

packages/algob/src/internal/core/errors-list.ts
amityadav0: ok

packages/algob/test/builtin-tasks/compile.ts
amityadav0: Done, added `true` in force and updated timestamp to check that `op.compiled` was called

packages/algob/src/internal/core/errors-list.ts
```diff
@@ -29,7 +29,8 @@ export const ERROR_RANGES = {
   BUILTIN_TASKS: { min: 600, max: 699, title: "Built-in tasks errors" },
   PLUGINS: { min: 800, max: 899, title: "Plugin system errors" },
 
-  SCRIPT: { min: 900, max: 999, title: "Script related errors" }
+  SCRIPT: { min: 900, max: 999, title: "Script related errors" },
+  PyTEAL: { min: 450, max: 460, title: "Python (PyTeal) related errors" }
```
robert-zaremba, line 33: Please move it to the place to keep the order based on the number (min  / max).

packages/algob/test/builtin-tasks/compile.ts
```diff
@@ -95,8 +100,28 @@ describe("Compile task", () => {
...
+  it("should return correct PyASCCache from CompileOp", async () => {
+    const result = await op.ensureCompiled(f3PY, true);
+    const expected = fs.readFileSync(path.join(ASSETS_DIR, 'gold-asa-py-check.yaml'), 'utf8');
+    assert.deepEqual(YAML.stringify(result), expected);
+  });
+
+  it("should return correct PyASCCache from PyCompileOp", async () => {
+    const pyOp = new PyCompileOp(op);
+    const result = await pyOp.ensureCompiled(f3PY, false);
+    const expected = fs.readFileSync(path.join(ASSETS_DIR, 'gold-asa-py-check.yaml'), 'utf8');
```
robert-zaremba, line 124: we should remove the file before starting the test. Otherwise we don't know if the file was there because of the previous test, or this test.

robert-zaremba (approved):
conditional approve. Please solve the comments before merging.

packages/algob/test/builtin-tasks/compile.ts
amityadav0: this file is for comparing the result, we will get the result in previous line, this is expected file

packages/algob/src/internal/core/errors-list.ts
amityadav0: ok

PR #108, opened 2020-10-26
Algob console

ratik21:
Support for `algob console`. Available objects :- 
- `deployer` : deployer in `Run` Mode
-  `algodClient` 
-  `algob`
-  `algosdk`

(tests to be updated)


docs/algob-console.md
```diff
@@ -0,0 +1,17 @@
...
+
+`algob` provides you an easy way to do this via an interactive console, with your contracts available and ready to use.
+
+## Usage
+
+To open console session run `yarn run algob console`
+
+## Globals
+
+Following globals can be accessed during `algob console` session
```
robert-zaremba, line 13:
```suggestion
Following globals are available in an `algob console` REPL:
```

docs/algob-console.md
```diff
@@ -0,0 +1,17 @@
...
+* `deployer` : algob deployer in run mode. User can access checkpoints, get logic signature, transferAlgos and all other functions supported by `algob deployer`.
+* `algodClient` : `algosdk.Algodv2`- an instance of algorand driver based on the current network.
+* `algosdk` : User can access `algosdk` package functions using this object. (eg. `algosdk.encodeAddress(..)`)  
+* `algob` : User has access to `algob` functions as well. (eg. `algob.mkAccounts(..)`, `algob.balanceOf(..) etc)`
```
robert-zaremba, line 17: Let's also make all function from `src/index.ts` available. What do you think?

ratik21: image
@robert-zaremba do you mean these functions? I have already added them in `repl` context within the `algob` gloabal object.

robert-zaremba: cool. Thanks!

robert-zaremba:
```suggestion
* `algob` : All`algob` [exported](https://github.com/scale-it/algorand-builder/blob/master/packages/algob/src/index.ts) functions as well. (eg. `algob.mkAccounts(..)`, `algob.balanceOf(..) etc)`
```

robert-zaremba (approved):
Approved, but please add this commit: https://github.com/scale-it/algorand-builder/pull/108/files/fbd4d788499454b6090120bf050d80f3df709d99#r512157459

docs/algob-console.md
ratik21: done

packages/algob/test/builtin-tasks/console.ts
```diff
@@ -0,0 +1,11 @@
+import { TASK_CONSOLE } from "../../src/builtin-tasks/task-names";
+import { useEnvironment } from "../helpers/environment";
+
+describe("Compile task", function () {
+  useEnvironment();
+  it("Should open repl session", async function () {
+    await this.env.run(TASK_CONSOLE, {
+      noCompile: true
+    });
```
amityadav0, line 9: is this complete ?

ratik21: Yes, will add more tests later.

packages/algob/src/builtin-tasks/console.ts
```diff
@@ -1,42 +1,90 @@
...
+import { loadASAFile } from "../lib/asa";
+import { createClient } from "../lib/driver";
+import { loadCheckpointsRecursive } from "../lib/script-checkpoints";
+import { AlgobDeployer, AlgobRuntimeEnv, CheckpointRepo } from "../types";
 import { TASK_CONSOLE } from "./task-names";
 
-export default function (): void {
-  // const log = debug("algob:core:tasks:console");
+// colorize text to yellow
+function colorize (message: string): string {
```
amityadav0, line 17: yellow color is generally used for warning, can we change it to something else ?
or maybe white is good ?

packages/algob/src/builtin-tasks/console.ts
```diff
@@ -1,42 +1,90 @@
...
+    });
 
+    // assign repl context
+    server.context.deployer = initializeDeployer(runtimeEnv);
+    server.context.algodClient = createClient(runtimeEnv.network);
+    server.context.algob = require('algob');
+    server.context.algosdk = require('algosdk');
+
+    server.on('exit', () => {
+      resolve();
```
amityadav0, line 71: exit
`exit` command doesn't work, instead `.exit` works

amityadav0:
Have you checked algob console for different networks ?
It doesn't give error for a network that is not even started

network

packages/algob/src/builtin-tasks/console.ts
ratik21: This is expected behaviour. You can exit a REPL session using `.exit`, `ctrl + c` or `ctrl + d`.

ratik21: If not clear, i will update the docs

packages/algob/src/builtin-tasks/console.ts
ratik21: I think yellow color is used as `info` rather than a warning. I don't think using white is good. What other color do you suggest?

ratik21:
> Have you checked algob console for different networks ?
> It doesn't give error for a network that is not even started
> 
> network

@amityadav0 as per discussed on discord :-
**This is expected behaviour**.
+ As in here `algob` is just creating and returing a new instance of `algodClient` based on the network config passed using the `--network` flag. (check `createClient` function in the image below).
+ So, an instance is created even if the network params are invalid, but you won't be able to interact with the blockchain. (if host/port is incorrect you'll get `ERRCONNECTION REFUSED`, if `token` is incorrect then `unauthorized`).
image

packages/algob/src/builtin-tasks/console.ts
ratik21: Changing to green.

amityadav0:
yes

packages/algob/test/builtin-tasks/console.ts
ratik21: removed.

PR #112, opened 2020-10-29
Clean deployer

ratik21:
- add `deployerConfig` to pass a single object to `mkdeployer`.
- fix linter warnings


packages/algob/src/builtin-tasks/run.ts
```diff
@@ -34,30 +33,25 @@ function filterNonExistent (scripts: string[]): string[] {
 }
 
 export function mkDeployer (
```
robert-zaremba, line 35: let's move this function somewhere to `src/internal`

packages/algob/src/builtin-tasks/run.ts
```diff
@@ -34,30 +33,25 @@ function filterNonExistent (scripts: string[]): string[] {
...
-  runtimeEnv: AlgobRuntimeEnv,
-  cpData: CheckpointRepo,
   allowWrite: boolean,
-  algoOp: AlgoOperator,
-  asaDefs: ASADefs,
-  accounts: Accounts,
-  txWriter: txWriter
+  deployerConfig: DeployerConfig
```
robert-zaremba, line 37: :+1:

packages/algob/src/lib/deployer.ts
```diff
@@ -0,0 +1,34 @@
...
+    algoOp: AlgoOperator
+  ) {
+    this.runtimeEnv = runtimeEnv;
+    this.cpData = loadCheckpointsRecursive();
+    this.algoOp = algoOp;
+    this.accounts = mkAccountIndex(runtimeEnv.network.config.accounts); ;
+    this.txWriter = new TxWriterImpl('');
+    this.asaDefs = loadASAFile(this.accounts);
+  }
+}
```
robert-zaremba, line 34: this should be moved to `src/internal`.  The idea is following:

* everything in `src/lib` can be accessible to the user
* everything in `src/internal` shouldn't be presented to the user

I don't think you should be able to create a `DeployerConfig`.

packages/algob/src/types.ts
```diff
@@ -347,6 +348,15 @@ export interface LsigInfo {
...
 }
 
+export interface DeployerConfig {
+  runtimeEnv: AlgobRuntimeEnv
+  cpData: CheckpointRepo
+  asaDefs: ASADefs
+  algoOp: AlgoOperator
+  accounts: Accounts
+  txWriter: txWriter
+}
```
robert-zaremba, line 358: I think we can keep this type next to the `MkDeployerConfig` - user will not use it.

robert-zaremba (changes requested):

packages/algob/src/types.ts
ratik21: Okay

packages/algob/src/lib/deployer.ts
ratik21: Ok, understood.

packages/algob/src/builtin-tasks/run.ts
ratik21: Alright

packages/algob/src/lib/deployer.ts
ratik21: Moving to `src/internal/util/deployer.ts`

packages/algob/src/builtin-tasks/run.ts
ratik21: Moving to `src/internal/util/deployer.ts`

packages/algob/src/lib/deployer.ts
robert-zaremba: Please don't move it to `util`.
`util` should be reserved for small common things. In general we should avoid it in fact. we can have it in `src/internal/deployer_cfg.ts`

packages/algob/src/builtin-tasks/run.ts
robert-zaremba: like in other comment. Please don't move it to `util`. Sorry. I should have note about it earlier.

packages/algob/src/internal/util/deployer.ts
```diff
@@ -0,0 +1,69 @@
...
+): AlgobDeployer {
+  if (allowWrite) {
+    return new DeployerDeployMode(
+      deployerConfig.runtimeEnv,
+      deployerConfig.cpData,
+      deployerConfig.asaDefs,
+      deployerConfig.algoOp,
+      deployerConfig.accounts,
+      deployerConfig.txWriter);
+  }
```
amityadav0, line 39: can we pass it as `deployerConfig` ? and use `this.runtimeEnv = deployerConfig.runtimeEnv` in deployer constructor.
this will reduce the repetitve code.
also i suggest to rename it as `deployerCfg`.

packages/algob/src/internal/util/deployer.ts
```diff
@@ -0,0 +1,69 @@
...
+  return new DeployerRunMode(
+    deployerConfig.runtimeEnv,
+    deployerConfig.cpData,
+    deployerConfig.asaDefs,
+    deployerConfig.algoOp,
+    deployerConfig.accounts,
+    deployerConfig.txWriter);
+}
```
amityadav0, line 47: same as above.

packages/algob/src/internal/util/deployer.ts
```diff
@@ -0,0 +1,69 @@
...
+  cpData: CheckpointRepo;
+  asaDefs: ASADefs;
+  algoOp: AlgoOperator;
+  txWriter: txWriter;
+  accounts: Accounts;
+
+  constructor (
+    runtimeEnv: AlgobRuntimeEnv,
+    algoOp: AlgoOperator
+  ) {
```
amityadav0, line 61: this can be put in a single line as there are only two variables. `constructor (runtimeEnv: AlgobRuntimeEnv, algoOp: AlgoOperator) {`

packages/algob/src/internal/util/deployer.ts
```diff
@@ -0,0 +1,69 @@
...
+  Accounts,
+  AlgobDeployer,
+  AlgobRuntimeEnv,
+  ASADefs,
+  CheckpointRepo
+} from "../../types";
+import {
+  DeployerDeployMode,
+  DeployerRunMode
+} from "../deployer";
```
amityadav0, line 15: this can be put in a single line as there are only two imports

packages/algob/src/lib/deployer.ts
ratik21: Ok, noted.

packages/algob/src/builtin-tasks/run.ts
ratik21: Alright

packages/algob/src/internal/util/deployer.ts
ratik21: Ok let's change to `deployerCfg`. 
Yes we can pass the whole object to deployer and destructure it in the deployer constructer. With that the test would also need to be updated and in some of them, `asaDefs` are passed separately (some other value). So to accomodate that we would have to pass `cpData` explicitely to `MkDeploy...` and also create it before passing. I think the scope of this PR was to clean up the construction by inroducing an object. Please let me know if this makes sense. 
I will update it if you say otherwise.

packages/algob/src/internal/util/deployer.ts
ratik21: Changing to `deployerCfg`

packages/algob/src/internal/util/deployer.ts
ratik21: Let's see if we pass entire object to deployer. If yes then we'll need to pass `asaDefs` (maybe, not sure) as well. Will update accordingly.

packages/algob/src/internal/util/deployer.ts
ratik21: Alright

packages/algob/src/internal/util/deployer.ts
amityadav0: i think that's why we had `fakeDeployer` , so we don't have to do changes everywhere. but that was removed

amityadav0: makes sense

packages/algob/src/internal/util/deployer.ts
ratik21: Updated.

packages/algob/src/internal/deployer_cfg.ts
```diff
@@ -0,0 +1,63 @@
...
+    deployerCfg.runtimeEnv,
+    deployerCfg.cpData,
+    deployerCfg.asaDefs,
+    deployerCfg.algoOp,
+    deployerCfg.accounts,
+    deployerCfg.txWriter);
+}
+
+// intialize deployer config obj
+export class MkDeployerConfig implements DeployerConfig {
```
robert-zaremba, line 47: Last comment. 
This class is  a structure without methods. We don't need an interface. Let's remove `DeployerConfig` interface and rename `MkDeployerConfig` to `DeployerConfig`. Class names should be nouns or noun expression

robert-zaremba (changes requested):

packages/algob/src/internal/util/deployer.ts
robert-zaremba: You can also overwrite the `asaDef` later on (if that helps): `deployerCfg =...` .
I also like the idea to passing the `deployerCfg` to the `DeployerDeployMode` constructor.

packages/algob/src/internal/deployer_cfg.ts
ratik21: Makes sense. Updated

robert-zaremba (approved):
Good job.

PR #116, opened 2020-11-02
Naming convention

amityadav0:


amityadav0:
name

I suggest we should resolve them first. But it has to be done manually because there may be some variables which needs to be changes manually.
what do you suggest ?

.eslintrc.js
```diff
@@ -30,7 +30,13 @@ module.exports = {
     "@typescript-eslint/quotes": "off",
     "@typescript-eslint/semi": ["error", "always"],
     "@typescript-eslint/strict-boolean-expressions": "off",
-
+    "@typescript-eslint/naming-convention": [
+      "warn",
```
robert-zaremba, line 34: let's change it to error

robert-zaremba (approved):

.eslintrc.js
amityadav0: ok

PR #117, opened 2020-11-03
KmdCfg config validation and KmdCfg config tests

amityadav0:


packages/algob/test/internal/core/config/config-validation.ts
```diff
@@ -608,4 +608,148 @@ describe("Config validation", function () {
...
+              [ALGOB_CHAIN_NAME]: {
+                asdasd: "3"
+              }
+            }
+          }),
+        ERRORS.GENERAL.INVALID_CONFIG
+      );
+    });
+
+    it("Shouldn't fail with valid KMD config", function () {
```
robert-zaremba, line 704:
```suggestion
    it("Should work with valid KMD config", function () {
```

robert-zaremba: Let's don't use a double negations.

packages/algob/test/internal/core/config/config-validation.ts
```diff
@@ -608,4 +608,148 @@ describe("Config validation", function () {
...
+          [ALGOB_CHAIN_NAME]: {
+            asdasd: "1234"
+          }
+        }
+      });
+
+      assert.isEmpty(errors.errors, errors.toString());
+    });
+
+    it("Shouldn't fail with unrecognized params", function () {
```
robert-zaremba, line 729:
```suggestion
    it("Should work with unrecognized params", function () {
```

packages/algob/test/internal/core/config/config-validation.ts
```diff
@@ -608,4 +608,148 @@ describe("Config validation", function () {
...
+            wallets: [
+              {
+                name: "Wallet",
+                password: "",
+                accounts: [
+                  { name: "Account1", address: "addr-4" }]
+              }
+            ],
+            otherParam: "some_other_detail"
+          },
```
robert-zaremba, line 745: you can store a valid config in variable and reuse it in 2 tests instead of copying lot of lines.
Here, in this test you can do:
```
cfg.networks.localhost.kmdCfg.otherParam = "some_other_detail"
```

robert-zaremba: In other tests you can copy the object using `deepmerge` (we already have it in dependencies):
```
let invalidCfg = deepmerge({}, cfg)
```

packages/algob/test/internal/core/config/config-validation.ts
amityadav0: ok

packages/algob/test/internal/core/config/config-validation.ts
amityadav0: ok

amityadav0: asd

If I do it using `deepmerge`, it doesn't let me assign wrong types to values, but user can put wrong types in `algob.config.js`

robert-zaremba: Ah, so maybe using a function will be a good solution?

```
let cfg = mkCfg()
```

amityadav0: work

this works if i use any in tests.

amityadav0: works

this works if i use object.assign()

amityadav0: work

this works if i use `any` with `object.assign()`

amityadav0: work1

works when i use `any`

robert-zaremba (approved):
The configs in tests looks a bit cryptic. I though that `deepmerge` would make it easy and reduce the code to create a cfg to few lines. Feel free to finalize it as you like

amityadav0:
understood, i have added a valid config at top of the tests and i am resusing this config using, `deepmerge` and `any` (as algob.config.js is in javascript). updated in https://github.com/scale-it/algorand-builder/pull/117/commits/30d11bac2a13bc1f8ccb549cebebf815243fc154 this commmit.

PR #118, opened 2020-11-03
remove murmurhash and lsig require

ratik21:


packages/algob/src/lib/compile.ts
```diff
@@ -1,6 +1,7 @@
 import type { Algodv2, CompileOut } from "algosdk";
 import { spawnSync, SpawnSyncReturns } from "child_process";
 import * as fs from 'fs';
+import * as murmurhash from 'murmurhash';
```
robert-zaremba, line 4: :+1:

packages/algob/src/lib/lsig.ts
```diff
@@ -17,3 +16,12 @@ export async function getLsig (name: string, scParams: Object, algodClient: algo
...
 }
+
+/**
+ * Description: this function creates and returns a dummy logic signature
+ */
+export function getDummyLsig (): LogicSig {
+  const dummyProgram = new Uint8Array(56);
+  dummyProgram.fill(0);
+  return algosdk.makeLogicSig(dummyProgram, []);
+}
```
robert-zaremba, line 27: This can be a `const`:
```
export DUMMY_LSIG = algosdk.makeLogicSig(new Uint8Array(56), []);
```

The `  dummyProgram.fill(0);` is not needed, because 0 is a default value already.

robert-zaremba (changes requested):
Let's use `DUMMY_LSIG` instead of `getDummyLsig`

packages/algob/src/lib/lsig.ts
ratik21: Okay

packages/algob/src/internal/deployer.ts
```diff
@@ -131,14 +129,10 @@ class DeployerBasicMode {
     const data = await readBinaryMultiSig(name);
     const program = new Uint8Array(Buffer.from(data as string, 'base64'));
     const logicSignature = decode(program) as RawLsig;
-
-    // dummy logic signature
-    const dummyProgram = new Uint8Array(56);
-    dummyProgram.fill(0);
-    const lsig = new logicsig.LogicSig(dummyProgram, []);
+    const lsig = DUMMY_LSIG; // dummy logic signature
```
robert-zaremba, line 132: ah, you are modifying it in line 135. Sorry, I didn't notice it. This might be not secure, because you change the constant. 
Sorry, could you rollback?
```
function dummyLsig() {
    return algosdk.makeLogicSig(new Uint8Array(56), []);
}
```

robert-zaremba (approved):
Preapproving. Please look at the comment.

packages/algob/src/internal/deployer.ts
ratik21: Okay will rollback. Sorry, i should have noticed it too :/

PR #120, opened 2020-11-05
Add kmd tests

ratik21:


packages/algob/test/internal/core/config/config-loading.ts
```diff
@@ -169,4 +190,79 @@ describe("config loading", function () {
...
+      assert.isUndefined(result);
+    });
+
+    it("should not connect to kmd if config is invalid", async function () {
+      const invalidKmdCfg = net.kmdCfg as KmdCfg;
+      invalidKmdCfg.port = 123; // invalid port
+
+      expectBuilderErrorAsync(
+        () => _loadKMDAccounts(net, new KMDOperator(createKmdClient(invalidKmdCfg))),
+        ERRORS.KMD.CONNECTION
```
amityadav0, line 222: I don't think we should createKmdClient in tests. 
as this is not mock and we cannot test the correct tests for it (in test environment)
`this will always fail`

amityadav0, line 222: this is not implemented as per task, i think we need to have mock client.
also if we are using createKmdClient. this test will always fail 
therefore we cannot test it correctly

amityadav0: + accounts are loaded only when kmdCfg for given network is specified and only if it can connect to the KMD -task

can you please add loading of accounts from a particular network ?

packages/algob/test/internal/core/config/config-loading.ts
```diff
@@ -169,4 +190,79 @@ describe("config loading", function () {
...
+      ).catch((err) => console.log(err)); ;
+    });
+
+    it("network accounts should take precedence over KMD accounts", async function () {
```
amityadav0, line 226: + should detect conflict of account names 

can you please add this one too ?

packages/algob/test/internal/core/config/config-loading.ts
```diff
@@ -169,4 +190,79 @@ describe("config loading", function () {
...
+
+    it("kmd accounts with different names should be successfully merged", async function () {
+      kmdOp.addKmdAccount({ name: "some-different-name-1", addr: "some-addr-1", sk: new Uint8Array(kmdOp.skArray) });
+      kmdOp.addKmdAccount({ name: "some-different-name-2", addr: "some-addr-2", sk: new Uint8Array(kmdOp.skArray) });
+
+      await _loadKMDAccounts(net, kmdOp);
+
+      for (const a of net.accounts) {
+        assert.include(net.accounts, a); // assert if network accounts are merged
+      }
```
amityadav0, line 251: `an array will always have it's own element in it`
this will always be true. right ?

packages/algob/test/internal/core/config/config-loading.ts
ratik21: I wanted to. But i wasn't able to assert `console.warn(..)`. Do you have any suggestion?

packages/algob/test/internal/core/config/config-loading.ts
ratik21: Yes, that's correct. Just checked here for sanity purpose. Looked more clear to check for all accounts. Let's wait for @robert-zaremba to see this.

packages/algob/test/internal/core/config/config-loading.ts
amityadav0: https://medium.com/@the_teacher/how-to-test-console-output-console-log-console-warn-with-rtl-react-testing-library-and-jest-6df367736cf0
this maybe helpful

packages/algob/test/internal/core/config/config-loading.ts
amityadav0: did you mean sanity testing ? https://www.edureka.co/blog/what-is-sanity-testing/
or i have misunderstood

packages/algob/test/internal/core/config/config-loading.ts
ratik21: (reply to previous comment) I think i am successfully testing the kmd connection here. We are creating an invalid kmdClient (which we can. we discussed this, will reference the discussion here). And passing this client to an instance of parent kmdOp (not base - which is the mock). `Algosdk` allows us to make clients (`algodv2`, `kmd`), but it won't allow us to execute chain operations (since i am adding an invalid port, it should and does give `ECONNREFUSED`, which we are asserting here). Please let me know if we still need to change it.
Reference -  https://github.com/scale-it/algorand-builder/pull/108#issuecomment-717117508

packages/algob/test/internal/core/config/config-loading.ts
ratik21: I checked, but it adds other dependencies. Didn't wanted to do that just for a single test.

packages/algob/test/internal/core/config/config-loading.ts
ratik21: Yes, since `_loadKMDAccounts` is merging kmd accounts into network.accounts, and some accounts are ignored based on name conflict, i wanted to do a check for all accounts. But yes, i know network.accounts have higher precendence and won't be ignored. Maybe we can remove this.

packages/algob/test/internal/core/config/config-loading.ts
ratik21: checking for a better solution

amityadav0: I checked, we have `sinon` installed already. we can use that

packages/algob/test/internal/core/config/config-loading.ts
amityadav0: yes, understood, we have `mock operator` but not `mock client`.
i was looking at `AlgoOperatorDryRunImpl` for reference , this has different way of testing.
let's keep it same.

packages/algob/test/internal/core/config/config-loading.ts
ratik21: ok cool! Thanks

packages/algob/test/internal/core/config/config-loading.ts
amityadav0: yes, But In these lines:
consider an example
```
for (const element of array) {
        assert.include(array, element);
}
```
No matter what operations are done previously, this will always be true as `we are picking element from array and checking if it is in the array`

ratik21: Ok, let's remove

packages/algob/src/internal/core/config/config-loading.ts
```diff
@@ -56,16 +57,19 @@ export async function loadConfigAndTasks (
...
+      await _loadKMDAccounts(net, kmdOp);
+    }
   }
 
   return cfg;
 }
 
 // loads KMD accounts if the net.kmdCfg is specified and merges them into net.accounts
-async function _loadKMDAccounts (net: NetworkConfig): Promise<void> {
+export async function _loadKMDAccounts (net: NetworkConfig, kmdOp: KMDOperator): Promise<void> {
```
robert-zaremba, line 70: If you export this function, then please rename it. Exported functions should have `_` prefix. You can just remove the `_`.

Maybe we can add this function to the KMDOperator?

packages/algob/src/lib/account.ts
```diff
@@ -172,3 +127,55 @@ function validateAlgobAccounts (algobAccounts: AlgobAccount[]): void {
...
+    this.kmdcl = kmdcl;
+  }
+
+  kmdWalletAddrNames (kwallet: KmdWallet): StrMap {
+    const m: StrMap = {};
+    for (const a of kwallet.accounts) {
+      m[a.address] = a.name;
+    }
+    return m;
+  }
```
robert-zaremba, line 144: :+1:

robert-zaremba (changes requested):

packages/algob/src/internal/core/config/config-loading.ts
ratik21: Okay, will rename. Also, i assumed that this function is for internal use only. Right?
If we can move this, then i suggest to move to the `KmdOperator` class to a new file in `src/lib/kmd-operator.ts` (similar to `algo-operator.ts`).
Currently i am updating the function. Will move according to furthur discussion.

robert-zaremba: yes.

robert-zaremba (approved):

PR #121, opened 2020-11-06
Deployer config

ratik21:
- Pass entire `deployerCfg` to create new instance of deployerBasic class.
- update and clean tests
- remove encode-decode lsig and show decoded lsig in console logs, checkpoint.

### DIFF
--- a/packages/algob/src/builtin-tasks/console.ts
+++ b/packages/algob/src/builtin-tasks/console.ts
@@ -13,11 +13,6 @@ import { loadCheckpointsIntoCPData, lsScriptsDir } from "../lib/script-checkpoin
 import { AlgobDeployer, AlgobRuntimeEnv } from "../types";
 import { TASK_CONSOLE } from "./task-names";
 
-// colorize text to yellow
-function colorize (message: string): string {
-  return `\x1b[32m${message}\x1b[0m`;
-}
-
 function initializeDeployer (runtimeEnv: AlgobRuntimeEnv): AlgobDeployer {
   const algoOp = createAlgoOperator(runtimeEnv.network);
   const deployerCfg = new DeployerConfig(runtimeEnv, algoOp);
@@ -47,7 +42,7 @@ async function startConsole (runtimeEnv: AlgobRuntimeEnv): Promise<void> {
   const algodClient = createClient(runtimeEnv.network);
   await new Promise<void>((resolve, reject) => {
     console.log("★", chalk.cyan(" Welcome to algob console"), "★");
-    console.log(colorize('Try typing: config'));
+    console.log(chalk.green('Try typing: config'));
 
     const server = repl.start({
       prompt: 'algob> ',
--- a/packages/algob/src/builtin-tasks/console.ts
+++ b/packages/algob/src/builtin-tasks/console.ts
@@ -1,4 +1,5 @@
 import * as algosdk from "algosdk";
+import chalk from "chalk";
 import repl from "repl";
 import { runInNewContext } from "vm";
 
@@ -45,7 +46,7 @@ async function startConsole (runtimeEnv: AlgobRuntimeEnv): Promise<void> {
   const deployer = initializeDeployer(runtimeEnv);
   const algodClient = createClient(runtimeEnv.network);
   await new Promise<void>((resolve, reject) => {
-    console.log(colorize('Welcome to algob console'));
+    console.log("★", chalk.cyan(" Welcome to algob console"), "★");
     console.log(colorize('Try typing: config'));
 
     const server = repl.start({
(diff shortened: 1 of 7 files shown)


packages/algob/src/builtin-tasks/console.ts
```diff
@@ -45,7 +46,7 @@ async function startConsole (runtimeEnv: AlgobRuntimeEnv): Promise<void> {
   const deployer = initializeDeployer(runtimeEnv);
   const algodClient = createClient(runtimeEnv.network);
   await new Promise<void>((resolve, reject) => {
-    console.log(colorize('Welcome to algob console'));
+    console.log("★", chalk.cyan(" Welcome to algob console"), "★");
     console.log(colorize('Try typing: config'));
```
amityadav0, line 50: If we can use `chalk`, let's remove colorize function and use chalk everywhere

packages/algob/src/internal/deployer_cfg.ts
```diff
@@ -1,7 +1,7 @@
 import { mkAccountIndex } from "../lib/account";
 import { AlgoOperator } from "../lib/algo-operator";
 import { loadASAFile } from "../lib/asa";
-import { loadCheckpointsRecursive } from "../lib/script-checkpoints";
+import { CheckpointRepoImpl } from "../lib/script-checkpoints";
 import type {
```
amityadav0, line 5: what is the difference between `loadCheckpointsRecursive ` and `CheckpointRepoImpl` ?

packages/algob/src/internal/deployer_cfg.ts
```diff
@@ -45,9 +33,9 @@ export class DeployerConfig {
 
   constructor (runtimeEnv: AlgobRuntimeEnv, algoOp: AlgoOperator) {
     this.runtimeEnv = runtimeEnv;
-    this.cpData = loadCheckpointsRecursive();
+    this.cpData = new CheckpointRepoImpl();
```
amityadav0, line 36: why are we using  `CheckpointRepoImpl` instead of `loadCheckpointsRecursive` ? are these same ?

packages/algob/src/internal/deployer_cfg.ts
ratik21: `CheckpointRepoImpl` is a raw checkpoint implementation with empty data. `loadCheckpointsRecursive` is returning the cp impl with loaded cp data(`asa`, `lsig`..) of the previous deployed scripts.

packages/algob/src/internal/deployer_cfg.ts
ratik21: Since we are initializing a new deployer, i was initializing with a fresh cp data (empty). But lets keep `loadCheckpointsRecursive` as user may deploy multiple scripts one by one. (Reverting).

amityadav0: yes

packages/algob/src/internal/deployer_cfg.ts
amityadav0: ok

ratik21:
We can remove encoding(while saving to checkpoint) and decoding(while reading from checkpoint) of `logic signature`. This will show the decoded lsig in console logs and cp as well. We previously assumed that loading it from a file without `algosdk.makesig(..)` was giving error and serialization solved this. But we don't need to do that, i think since we loaded an object (lsig in this case) directly from checkpoint, it is treated as an object only and not a logic signature (which is an instance of `logicsig` class in `algosdk`). So we just create a dummylsig (using `algosdk.makesig(..)`) and assign loaded lsig to the dummy (which restores it's properties). 
Discussion - (https://discordapp.com/channels/754286650540032030/754286650540032033/765871411255574568)[]
image

amityadav0:
yes, makes sense

packages/algob/src/internal/deployer.ts
```diff
@@ -257,7 +256,7 @@ export class DeployerDeployMode extends DeployerBasicMode implements AlgobDeploy
       lsigInfo = {
         creator: signer.addr,
         contractAddress: lsig.address(),
-        lsig: encode(lsig)
+        lsig: (lsig)
       };
```
amityadav0, line 260: we can remove brackets as well

packages/algob/src/internal/deployer.ts
```diff
@@ -85,9 +85,8 @@ class DeployerBasicMode {
     const resultMap = this.cpData.precedingCP[this.networkName]?.dLsig ?? new Map(); ;
```
amityadav0, line 85: please change `Object | undefined` to `LogicSig | undefined`

robert-zaremba (commented):
LGTM, I will leave the final approval to @amityadav0

amityadav0 (approved):

PR #127, opened 2020-11-11
Port "HTLC" to /examples/ref-templates

amityadav0:
- Port "HTLC" to /examples/ref-templates
- Add `closeRemainderTo` in `TxParams`
- update Readme.md


examples/ref-templates/assets/htlc.py
```diff
@@ -0,0 +1,100 @@
...
+fee = 10000
+pre_image = "QzYhq9JlYbn2QdOMrhyxVlNtNjeyvyJc/I8d8VAGfGc="
+timeout = 2000
+
+def htlc(TMPL_RCV = john,
+         TMPL_OWN = master,
+         TMPL_FEE = fee,
+         TMPL_HASHIMG = pre_image,
+         TMPL_HASHFN = Sha256,
+         TMPL_TIMEOUT = timeout):
```
robert-zaremba, line 34: Please move the comments to function doc string and remove default values. We will use the arguments in the function call:
```
def htlc(TMPL_RCV, TMPL_FEE, ... ):
    """HTLC template
    Parameters: 
      TMPL_RCV (string): the address to send funds to when the preimage is supplied
     ...
   """
```

and in the call:
```
htlc(john, ...)
```

examples/ref-templates/assets/htlc.py
```diff
@@ -0,0 +1,100 @@
...
+        fee_check,
+        pay_check,
+        rec_field_check,
+        amount_check
+    )
+
+    """Payout scenarios : At this point in the execution, there is one boolean variable on the 
+    stack that must be true in order for the transaction to be valid. The checks we have done 
+    above apply to any transaction that may be approved by this script.We will now check if we
+     are in one of the two payment scenarios described in the functionality section."""
```
robert-zaremba, line 63: this is a doc string, it doesn't make sens to use it here. We use it only to function, classes, globals etc...
Let's use normal coment:
```
# 
# 
```

examples/ref-templates/assets/htlc.py
```diff
@@ -0,0 +1,100 @@
...
+    # stack is a boolean holding the results of the initial transaction validity checks. 
+    # This is followed by two booleans indicating the results of the scenario 1 and 2 checks.
+
+    # We want to approve this transaction if we are in scenario 1 or 2. 
+    # So we logically OR the results of those checks together.
+    # Finally, we logically AND the scenario checks with the initial checks.
+    return And(Or(scenario_1, scenario_2), required_condition)
+
+    # At this point, the stack contains just one value: a boolean indicating 
+    # whether or not it has been approved by this contract.
```
robert-zaremba, line 97: please move the last 2 lines just above return (combine it with the previous comments)

examples/ref-templates/scripts/htlc.js
```diff
@@ -0,0 +1,40 @@
...
+    } catch (e) {
+        console.error('Transaction Failed', e);
+    }
+}
+
+async function run(runtimeEnv, deployer) {
+
+  const masterAccount = deployer.accountsByName.get("master-account")
+  const johnAccount = deployer.accountsByName.get("john-account");
+  const globalZeroAddress = "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAY5HFKQ";
```
robert-zaremba, line 16: Le's move this to algob to `src/lib/constants.ts` and import it here.

examples/ref-templates/scripts/htlc.js
```diff
@@ -0,0 +1,40 @@
...
+  const secret = "hero wisdom green split loop element vote belt";
+  const wrongSecret = "hero wisdom red split loop element vote belt";
+
+  // setup a contract account and send 1 ALGO from master
+  await deployer.fundLsig("htlc.py", [], { funder: masterAccount, fundingMicroAlgo: 1000000 }, { closeRemainderTo: johnAccount.addr }); 
+
+  let contract = await deployer.loadLogic("htlc.py", [ wrongSecret ]);
+  let contractAddress = contract.address();
+  
+  // Fails as wrong secret is passed
```
robert-zaremba, line 29:
```suggestion
  // Fails because wrong secret is passed
```

examples/ref-templates/scripts/htlc.js
```diff
@@ -0,0 +1,40 @@
...
+  await transferAlgo(deployer, { addr: contractAddress}, globalZeroAddress, 0, contract, { totalFee: 1000, closeRemainderTo: johnAccount.addr});
+
+  contract = await deployer.loadLogic("htlc.py", [ secret ]);
+  contractAddress = contract.address();
+
+  // Passes as right secret is passed
```
robert-zaremba, line 35:
```suggestion
  // Passes because right secret is passed
```

packages/algob/src/types.ts
```diff
@@ -397,6 +397,7 @@ export interface TxParams {
   lease?: string
   note?: string
   noteb64?: string
+  closeRemainderTo?: AccountAddress
```
robert-zaremba, line 400: :+1:

robert-zaremba (changes requested):

examples/ref-templates/scripts/htlc.js
```diff
@@ -0,0 +1,40 @@
+const { transferMicroAlgos, transferMicroAlgosLsig } = require("algob");
+
+async function transferAlgo(deployer, senderAddr, receiverAddr, amount, lsig, payFlags){
+    try {
+        const details = await transferMicroAlgosLsig(deployer, senderAddr, receiverAddr, amount, lsig, payFlags);
+        console.log(details);
+    } catch (e) {
+        console.error('Transaction Failed', e);
```
ratik21, line 8: I think we should print the error here and not the entire response - `e.response.error`

packages/algob/src/lib/tx.ts
```diff
@@ -191,14 +191,15 @@ export async function transferMicroAlgosLsig (
...
   amountMicroAlgos: number,
-  lsig: LogicSig): Promise<tx.ConfirmedTxInfo> {
-  const params = await deployer.algodClient.getTransactionParams().do();
+  lsig: LogicSig,
+  payFlags: TxParams): Promise<tx.ConfirmedTxInfo> {
+  const params = await mkSuggestedParams(deployer.algodClient, payFlags);
 
   const receiver = toAccountAddr;
-  const note = tx.encodeObj("ALGO PAID");
+  const note = tx.encodeObj(payFlags.note);
```
ratik21, line 199: Since note is an optional param, what happens when note is not passed?

examples/ref-templates/assets/htlc.py
amityadav0: ok

examples/ref-templates/assets/htlc.py
amityadav0: ok

examples/ref-templates/assets/htlc.py
amityadav0: ok

examples/ref-templates/scripts/htlc.js
amityadav0: ok

examples/ref-templates/scripts/htlc.js
amityadav0: ok

packages/algob/src/lib/tx.ts
amityadav0: yes, this needs to be changed

examples/ref-templates/assets/htlc.py
```diff
@@ -0,0 +1,99 @@
...
+pre_image = "QzYhq9JlYbn2QdOMrhyxVlNtNjeyvyJc/I8d8VAGfGc="
+timeout = 2000
+
+def htlc(TMPL_RCV,
+         TMPL_OWN,
+         TMPL_FEE,
+         TMPL_HASHIMG,
+         TMPL_HASHFN,
+         TMPL_TIMEOUT):
+
```
robert-zaremba, line 15:
```suggestion
```

robert-zaremba: I think the comment has to be right below the function to qualify as a docstring

robert-zaremba (approved):
I added one code suggestion to remove the line.

PR #128, opened 2020-11-13
Support Group transactions for algo transfer

ratik21:
- Add function to support group transactions for - transferring microAlgo's signed by logic.
- update example


ratik21:
We can also merge `transferMicroAlgosLsig` with `transferMicroAlgosLsigAtomic`. The difference would be in case of a single txn, we will assign a groupId for that transaction also. So, a single transaction now would mean - a group of 1 transaction.

examples/asa/scripts/transfer/gold-contract-sc.js
```diff
@@ -22,6 +21,22 @@ async function run(runtimeEnv, deployer) {
...
+  // Transaction PASS - for all transactions amount <= 100 and receiver is john
+  await transferAlgoAtomic(deployer, transactions);
+  
+  transactions = [
+    mkTxnParams({addr: sender}, johnAccount.addr, 200, lsig, { note: "ALGO PAID" }),
+    mkTxnParams({addr: sender}, johnAccount.addr, 20, lsig, { note: "ALGO PAID" })]
+  
+  // Transaction FAIL - since for the first transaction amount = 200 (which is > 100, so both transactions will fail)
+  await transferAlgoAtomic(deployer, transactions);
 }
```
amityadav0, line 40: I don't think we should merge group transaction example in our current examples. It may be confusing to the user.
also we have different ref-template for group transactions only. what do you think ?

ratik21: yes, these will be moved to ref-templates only. Just logged it here if someone wanted to test before rolling out the next PR (as functions will be identical).

examples/asa/scripts/transfer/common.js
```diff
@@ -9,11 +9,30 @@ exports.transferAlgo = async function (deployer, senderAddr, receiverAddr, amoun
...
+exports.transferASA = async function (deployer, senderAccount, receiverAddr, amount, assetID, lsig) {
     try {
         const details = await transferASALsig(deployer, senderAccount, receiverAddr, amount, assetID, lsig);
         console.log(details);
     } catch (e) {
         console.error('Transaction Failed', e.response.error);
     }
 }
+
+exports.transferAlgoAtomic = async function (deployer, txnParams) {
```
robert-zaremba, line 21: Let's rename this function and add `Micro` to the name: 

    transferMicroAlgoAtomic

examples/asa/scripts/transfer/gold-contract-sc.js
```diff
@@ -22,6 +21,22 @@ async function run(runtimeEnv, deployer) {
 
   // Transaction FAIL as Elon tried to receive instead of John
   await transferAlgo(deployer, { addr: sender}, elonMuskAccount.addr, 50, lsig);
+
+  //group transactions
+  let transactions = [
+    mkTxnParams({addr: sender}, johnAccount.addr, 30, lsig, {}),
+    mkTxnParams({addr: sender}, johnAccount.addr, 40, lsig, {}),
+    mkTxnParams({addr: sender}, johnAccount.addr, 50, lsig, {})]
```
robert-zaremba, line 29: can we add a normal transaction here? Eg sending Algo from `john` to `elon`?

examples/asa/scripts/transfer/gold-contract-sc.js
```diff
@@ -22,6 +21,22 @@ async function run(runtimeEnv, deployer) {
...
+  
+  // Transaction PASS - for all transactions amount <= 100 and receiver is john
+  await transferAlgoAtomic(deployer, transactions);
+  
+  transactions = [
+    mkTxnParams({addr: sender}, johnAccount.addr, 200, lsig, { note: "ALGO PAID" }),
+    mkTxnParams({addr: sender}, johnAccount.addr, 20, lsig, { note: "ALGO PAID" })]
+  
+  // Transaction FAIL - since for the first transaction amount = 200 (which is > 100, so both transactions will fail)
```
robert-zaremba, line 38:
```suggestion
  // Group transaction FAIL - one of the transaction amount is = 200 but the smart contract requires amount to be <= 100.
```

examples/asa/scripts/transfer/gold-contract-sc.js
robert-zaremba: OK. If you want to move it in other PR, then please add a task (or a subtask) in pivotal tracker.

packages/algob/src/lib/tx.ts
```diff
@@ -6,10 +6,13 @@ import {
   AlgobDeployer,
   ASADef,
   ASADeploymentFlags,
+  GrpTxnParams,
   TxParams
 } from "../types";
 import { ALGORAND_MIN_TX_FEE } from "./algo-operator";
 
+const ALGO_MSG = "Transferring algo (in micro algos):";
```
robert-zaremba, line 14:
```suggestion
const ALGO_MSG = "Transferring micro Algo:";
```

packages/algob/src/lib/tx.ts
```diff
@@ -248,3 +251,54 @@ export async function transferASALsig (
...
+
+/**
+ * Description:
+ * This function executes multiple transactions atomically where Algos are
+ * transferred from one account to another using logic signature (for signing)
+
+ * Returns:
+ * Transaction details
+*/
+
```
robert-zaremba, line 263:
```suggestion
```

packages/algob/src/lib/tx.ts
```diff
@@ -248,3 +251,54 @@ export async function transferASALsig (
...
   return await deployer.waitForConfirmation(tx1.txId);
 }
+
+/**
+ * Description:
+ * This function executes multiple transactions atomically where Algos are
+ * transferred from one account to another using logic signature (for signing)
+
+ * Returns:
+ * Transaction details
```
robert-zaremba, line 261:
```suggestion
 * ConfirmedTxInfo (transaction receipt)
```

examples/asa/scripts/transfer/gold-contract-sc.js
robert-zaremba: Ah, this will work only for lsig transfers. Please make a note about this somewhere here (in this file).

packages/algob/src/lib/tx.ts
```diff
@@ -248,3 +251,54 @@ export async function transferASALsig (
...
+      txid: signedTxn.txID
+    };
+    logs.push(txnInfo);
+    idx++;
+  }
+
+  const pendingTx = await deployer.algodClient.sendRawTransaction(signed).do();
+  console.log(ALGO_MSG, logs);
+  return await deployer.waitForConfirmation(pendingTx.txId);
+}
```
robert-zaremba, line 304: I agree with @amityadav0  - this function is too specialized. Probably it's better to move it to the example `common.js`  file (unless we need it in other place).

robert-zaremba (approved):
Left few comments to be addressed before merging.

examples/asa/scripts/transfer/common.js
ratik21: ok

examples/asa/scripts/transfer/gold-contract-sc.js
ratik21: I am removing this example from here.

PR #130, opened 2020-11-14
Add generic function for atomic transfers

amityadav0:
- Add a function which can handle single or group transactions

there is slight different in calling of funtions 
To transfer some asset, user can do something like this:
```
await executeTransaction(deployer, {
    type: "asset",
    sign: "sk",
    fromAccount: goldOwnerAccount,
    toAccountAddr: johnAccount.addr,
    amount: 1,
    assetID: goldAssetID,
    payFlags: {}});
```
This may seem a little complex, than previous one.
but for more complex examples, user don't have to use different functions.

This function supports any combination of `asset` transfer, `algo` transfer, `ssc` calls(NoOp, delete, clear, close).
Now, we don't have to add specialized function for every example we write.


amityadav0:
All the example are left for update, I will update after further discussion

packages/algob/src/types.ts
```diff
@@ -385,14 +385,37 @@ export interface Checkpoint {
...
+  type: "asset"
+  fromAccount: AccountSDK
+  toAccountAddr: AccountAddress
+  amount: number
+  assetID: number
+  payFlags: TxParams
+}
+
+export interface SSCCallsParam extends SSCOptionalFlags, Sign {
+  type: "clearSSC" | "deleteSSC" | "callNoOpSSC" | "closeSSC"
```
ratik21, line 413: This should be ENUM and not a string. In general `execParams.type` should be an enum. As we shouldn't ask the user to pass strings as txn type. Eg - user can pass 'clearssc' instead of 'clearSSC'.

packages/algob/src/types.ts
```diff
@@ -385,14 +385,37 @@ export interface Checkpoint {
...
-  lsig: LogicSig
+export type execParams = AlgoTransferParam | AssetTransferParam | SSCCallsParam;
+
+export interface Sign {
+  sign: "sk" | "lsig"
+  lsig?: LogicSig
+}
+
+export interface AlgoTransferParam extends Sign {
+  type: "algo"
```
ratik21, line 396: this should be enum

packages/algob/src/types.ts
```diff
@@ -385,14 +385,37 @@ export interface Checkpoint {
...
+  fromAccount: AccountSDK
+  toAccountAddr: AccountAddress
   amountMicroAlgos: number
   payFlags: TxParams
 }
 
+export interface AssetTransferParam extends Sign {
+  type: "asset"
```
ratik21, line 404: same

packages/algob/src/lib/tx.ts
```diff
@@ -96,208 +94,109 @@ export function encodeNote (note: string | undefined, noteb64: string| undefined
...
-  return await deployer.waitForConfirmation(pendingTx.txId);
+function signTransaction (txn: Transaction, txnParam: execParams): Uint8Array {
+  switch (txnParam.sign) {
+    case "sk": {
+      return txn.signTxn(txnParam.fromAccount.sk);
+    }
+    case "lsig": {
+      const logicsig = txnParam.lsig;
+      if (logicsig === undefined) {
+        throw new Error("Lsig undefined");
```
ratik21, line 158: Can we make this error a little more descriptive?

packages/algob/src/types.ts
```diff
@@ -385,14 +385,37 @@ export interface Checkpoint {
...
 export type ASADefs = z.infer<typeof ASADefsSchema>;
 
-export interface GrpTxnParams {
-  fromAccount: Account
-  toAccountAddr: string
-  lsig: LogicSig
+export type execParams = AlgoTransferParam | AssetTransferParam | SSCCallsParam;
+
+export interface Sign {
+  sign: "sk" | "lsig"
```
ratik21, line 391: These should be enum as well. Also, should we ask the user for this? let's say a user wants to transfer some Algo's from A to B, then it makes sense for him to pass `txntype.algo`, but should we also ask him for `Sign` (as it should be understood that it will be signed by the secret key)? Can we handle this internally?

amityadav0: No, because what if,  user wants to send algo using delegated Lsig. (for example, user wants to send algo only if fee is lower than some threshold)
We cannot assume what user wants.

packages/algob/src/types.ts
amityadav0: yes, thanks

packages/algob/src/types.ts
amityadav0: ok

packages/algob/src/types.ts
amityadav0: ok

packages/algob/src/lib/tx.ts
amityadav0: can you please suggest something ?
how about `lsig for this transaction is not defined` ?

ratik21: `logic signature for this transaction was not passed or - is not defined`. How about this?

ratik21:
- I feel that `executeTransaction(..)` can be a bit complex for a user wanting to perform a single type of transaction (eg - Algo transfer from A to B). Can you please update one example for this (you can modify a script using `transferMicroAlgo` function)?

- Also, i don't see any console logs (for algo transfer, asset transfer etc). Will we be adding them?

packages/algob/src/types.ts
ratik21: User will pass the lsig if he is using delegated lsig. Take `assetTransfer` function for example - we sign the txn using secret key, but we don't ask the user about the signing authority (passing "sk" explicitly).

amityadav0:
for algo transfer: 
```
await executeTransaction(deployer, {
    type: "algo",
    sign: "sk",
    fromAccount: goldOwnerAccount,
    toAccountAddr: johnAccount.addr,
    amountMicroAlgos: 1,
    payFlags: {}});
```
It will be this, I will update all the examples together.
we can change type to `transferAlgo` and `transferAsset`
yes, logs and examples will be updated.
also i am thinking to store these transactions in tx-writer logs and print only final transaction.

packages/algob/src/lib/tx.ts
amityadav0: yes

packages/algob/src/types.ts
amityadav0: yes, i understand, that's why i added a type, if user wants he can choose.
For `assetTransfer` user can also use delegated lsig.
There can be many examples.
This is the problem which was not covered earlier , we had to make different functions for each case. But in this function that problem is solved

packages/algob/src/lib/tx.ts
```diff
@@ -96,208 +96,109 @@ export function encodeNote (note: string | undefined, noteb64: string| undefined
...
+        txnParam.foreignAssets,
+        note,
+        txnParam.lease,
+        txnParam.rekeyTo);
+    }
+    case TransactionType.CloseSSC: {
+      return tx.makeApplicationCloseOutTxn(txnParam.fromAccount.addr, params, txnParam.appId);
+    }
+    default: {
+      throw new Error("Unknown type of transaction");
```
robert-zaremba, line 147: this has to be a `ALGOB` error. You can create a new error in `error-list`. It must take the `txnParam.type` as a parameter

robert-zaremba (changes requested):

packages/algob/src/lib/tx.ts
```diff
@@ -96,208 +96,109 @@ export function encodeNote (note: string | undefined, noteb64: string| undefined
...
-): Promise<tx.ConfirmedTxInfo> {
-  const params = await mkSuggestedParams(deployer.algodClient, param);
-
-  const receiver = to;
-  let note;
-  if (param.noteb64 ?? param.note) {
-    note = encodeNote(param.note, param.noteb64);
-  } else {
-    note = undefined;
+async function mkTransaction (deployer: AlgobDeployer, txnParam: execParams): Promise<Transaction> {
```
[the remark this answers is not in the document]
robert-zaremba: We need to have a good documentation: how this function work, what it does etc...

packages/algob/src/lib/tx.ts
```diff
@@ -96,208 +96,109 @@ export function encodeNote (note: string | undefined, noteb64: string| undefined
...
+  switch (txnParam.sign) {
+    case SignType.SecretKey: {
+      return txn.signTxn(txnParam.fromAccount.sk);
+    }
+    case SignType.LogicSignature: {
+      const logicsig = txnParam.lsig;
+      if (logicsig === undefined) {
+        throw new Error("logic signature for this transaction was not passed or - is not defined");
+      }
+      return tx.signLogicSigTransactionObject(txn, logicsig).blob;
```
ratik21, line 162: please change to
```
const signedTx = tx.signLogicSigTransactionObject(txn, logicsig);
return signedTx.blob;
```

amityadav0: this line is same as line `162`

packages/algob/src/lib/tx.ts
amityadav0: ok

packages/algob/src/lib/tx.ts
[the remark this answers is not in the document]
amityadav0: ok

packages/algob/src/lib/tx.ts
ratik21: I know, it's just a suggestion (thought it was better to write this way).

packages/algob/src/internal/core/errors-list.ts
```diff
@@ -150,6 +150,13 @@ Please check algob output for more details.`
     title: "Script can't load",
     description: `Script failed during load.
 Please check algob output for more details.`
+  },
+  TRANSACTION_TYPE_ERROR: {
+    number: 14,
+    message: "Error. `%error%` is Unknown",
+    title: "Unknown Transaction type",
+    description: `Transaction type you entered is unknown
```
ratik21, line 158:
```suggestion
    description: `Transaction type entered is unknown 
```

packages/algob/src/lib/tx.ts
```diff
@@ -97,207 +99,131 @@ export function encodeNote (note: string | undefined, noteb64: string| undefined
...
-  let note;
-  if (param.noteb64 ?? param.note) {
-    note = encodeNote(param.note, param.noteb64);
-  } else {
-    note = undefined;
+ * Description: Returns unsigned transaction as per execParams
+ * execParams can be of following types:
+ * AlgoTransferParam used for transferring algo
+ * AssetTransferParam used for transferring asset
+ * SSCCallsParam used for calling stateful smart contracts
```
robert-zaremba, line 106:
```suggestion
 * SSCCallsParam used for calling stateful smart contracts.
 For more advanced use-cases, please use `algosdk.tx` directly.
```

robert-zaremba (commented):
Updated a comment.

packages/algob/src/internal/core/errors-list.ts
```diff
@@ -150,6 +150,13 @@ Please check algob output for more details.`
     title: "Script can't load",
     description: `Script failed during load.
 Please check algob output for more details.`
+  },
+  TRANSACTION_TYPE_ERROR: {
+    number: 14,
+    message: "Error. `%error%` is Unknown",
+    title: "Unknown Transaction type",
+    description: `Transaction type entered is unknown
```
robert-zaremba, line 158:
```suggestion
    description: `Provided transaction type is unknown 
```

packages/algob/src/types.ts
```diff
@@ -413,22 +451,12 @@ export interface ASADeploymentFlags extends TxParams {
...
 
-export interface SSCDeploymentFlags {
-  sender: Account
-  localInts: number
-  localBytes: number
-  globalInts: number
-  globalBytes: number
+export interface SSCOptionalFlags {
   appArgs?: Uint8Array[]
   accounts?: string
```
ratik21, line 461: this should be an array of strings. ()[https://developer.algorand.org/docs/features/asc1/stateful/#write-to-state]

amityadav0: https://developer.algorand.org/docs/reference/transactions/#application-call-transaction
It says list of accounts, but type is mentioned as `address`.
what do you think ?

amityadav0: let's change it to array , there may be error

ratik21: I think it should be an array. I checked `js-algorand-sdk`'s `makeApplicationNoOpTxn(..)`

ratik21 (approved):

robert-zaremba (approved):

PR #131, opened 2020-11-16
Non fungible token

ratik21:
- add example to createNFT, transferNFT
- check readme


examples/nft/assets/approval_program.teal
```diff
@@ -0,0 +1,103 @@
+#pragma version 2
```
robert-zaremba, line 1: Please add more description of this smart-contract.

examples/nft/algob.config.js
```diff
@@ -0,0 +1,51 @@
...
+  token: "aade468d25a7aa48fec8082d6a847c48492066a2741f3731e613fdde086cd6e9",
+  accounts: accounts,
+};
+
+module.exports = {
+  networks: {
+    localhost: defaultCfg,
+    default: defaultCfg
+  }
+};
```
robert-zaremba, line 51: how about using sym links for the config?

examples/nft/README.md
```diff
@@ -0,0 +1,21 @@
+# Non-Fungible-Token Example using StateFul Teal
+
+In this example, we create a new non-fungible-token represented by a name and a url.
```
robert-zaremba, line 3:
```suggestion
In this example, we create a new non-fungible-token represented by a name and a url.
Compared to standard ASA, in this example we create a smart-contract to manage and store NFT balances. 

```

examples/nft/README.md
```diff
@@ -0,0 +1,21 @@
...
+`algob` provides two functions :-
+* `createNewNFT` - creates a new Non Fungible Token. Only the smart contract admin can create a new NFT.
+* `transferNFT` - transfer an NFT from one account to another. Both account should have opted-in to the smart contract and the account wishing to transfer the NFT must hold the NFT (in the account's local state).
+
+Also to be noted
+* An account can hold upto a maximum of 16 NFT's in it's local storage.
+* The system can hold a maximum of 63 NFT's (in global storage)
```
robert-zaremba, line 10: Please add explanation to this limitations.

robert-zaremba, line 10:
```suggestion
* The NFT smart contract can hold a maximum of 63 NFT's (in global storage)
```

examples/nft/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "htlc-pyteal",
+  "version": "0.0.1",
+  "main": "index.js",
+  "license": "MIT",
```
robert-zaremba, line 5: let's use Apache-2.0 license

packages/algob/src/lib/nft.ts
```diff
@@ -0,0 +1,106 @@
...
+      throw new Error(`Non Fungible Token with name ${nft.name} already exists`);
+    }
+  }
+
+  nft.url = _NFTToken();
+
+  const appArgs = [];
+  appArgs.push(new Uint8Array(Buffer.from("create")));
+  appArgs.push(new Uint8Array(Buffer.from(nft.name)));
+  appArgs.push(new Uint8Array(Buffer.from(nft.url)));
```
robert-zaremba, line 43: I think using simple Buffer will work here as well ( we don't need to convert to `Uint8Array`). Can you check that please?

packages/algob/src/lib/nft.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  const txnParam = {
+    type: 2, // TransactionType.CallNoOpSSC = 2
+    sign: SignType.SecretKey,
+    fromAccount: creater,
```
robert-zaremba, line 48: typo: should be creator (O instead of E)

packages/algob/src/lib/nft.ts
```diff
@@ -0,0 +1,106 @@
...
+    sign: SignType.SecretKey,
+    fromAccount: fromAccount,
+    appId: appId,
+    payFlags: {},
+    appArgs,
+    accounts
+  };
+
+  await executeTransaction(deployer, txnParam);
+}
```
robert-zaremba, line 106: This file (and the NFT type definition) shouldn't be part of `algob`. Let's move it to the example, or ... if you want - we can create a new package, eg: `packages/algob-nfg`

robert-zaremba (changes requested):

amityadav0:
This NFT example is logically incorrect.
Some of the cases:
- What if creator deletes the contract ? every one will lose their NFT
- what if creator updates the contract ?
- Also we cannot create same name NFTs with this contract
- Owner can create multiple NFT with same name , and same URI which will result in Fungible Token.

robert-zaremba:
> What if creator deletes the contract ? every one will lose their NFT
> what if creator updates the contract ?

These are valid concerns. But this is related to all Algorand smart-contracts (voting etc ...).
Let's add a clear note in README saying that: _**This is only a proof of concept**. In production use we need to handle properly the account which can manage the smart-contract (update or delete)_.

robert-zaremba:
BTW, I think it's possible to block the smart-contract upgradeability and possibility to remove the logic by removing the management address (setting it to zero). @amityadav0  - could you check that?

amityadav0:
ghj
yes, we can catch update or delete call from the contract code and permanently block them. 
```
int UpdateApplication
txn OnCompletion
==
bz not_update
int 0
return
```
- Also we cannot create same name NFTs with this contract
- Owner can create multiple NFT with same name , and same URI which will result in Fungible Token.
These points can also be resolved by using different map structure.

robert-zaremba:
> Also we cannot create same name NFTs with this contract
> Owner can create multiple NFT with same name , and same URI which will result in Fungible Token.

I agree with @amityadav0 . Using names is vulnerable. Instead I propose following:
* each NFT is an `(ID, ref_data, hash)` triple. We can store it as the following mapping: `{id: ref_data, id_h: hash}` (similarly as you do it now). The NFT name and other things will be encoded in `ref_data`. Ideally this will be a link to some external resources. Hash is a hash of the external resources data. 
* when creating a new NFT, we firstlty increment the `total`, and use the new `total` as an NFT id. So each NFT has a numerical ID.
* To convert total to `id` and `id_h` we need to use: `itob total` opcode. To create `id_h`, we need to use `concat` op code. Example:  `{1: "https://mytoken.com/1/ref", 1_h: "af230985a..."}`

Please update the documentation explaining how the token is constructed and how it is represented (both in the global state and local state).

robert-zaremba (changes requested):
- [ ] Let's add a not about upgradeability (with a link to the original resource) in the readme (the one provided by Amit).
- [ ] change representation of the token
- [ ] Update docs (readme, comments in the smart contract files)

examples/nft/assets/approval_program.teal
```diff
@@ -0,0 +1,103 @@
...
+app_global_put
+
+//add NFT to caller's local storage
+int 0
+txna ApplicationArgs 1
+txna ApplicationArgs 2
+app_local_put
+
+//increment global count
+byte "count"
```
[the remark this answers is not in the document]
ratik21: ok

examples/nft/assets/approval_program.teal
```diff
@@ -0,0 +1,103 @@
...
+==
+bz failed
+
+//add NFT to global storage
+txna ApplicationArgs 1
+txna ApplicationArgs 2
+app_global_put
+
+//add NFT to caller's local storage
+int 0
```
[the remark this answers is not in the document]
ratik21: int 0 represents caller's address. (Not integer - 0)

examples/nft/assets/approval_program.teal
```diff
@@ -0,0 +1,103 @@
...
+
+load 0
+return
+
+
+//transfers NFT from account A to B
+transfer_nft:
+
+//store nft value in scratch space
+int 1
```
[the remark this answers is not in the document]
ratik21: int 1 represents 1st account passed using the `--app-account` flag. (Not integer 1)

examples/nft/assets/approval_program.teal
```diff
@@ -0,0 +1,103 @@
...
+// making the call
+byte "Creator"
+app_global_get
+txn Sender
+==
+bz failed
+
+//add NFT to global storage
+txna ApplicationArgs 1
+txna ApplicationArgs 2
```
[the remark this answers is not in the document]
ratik21: Added expected arguments in description.

packages/algob/src/lib/nft.ts
ratik21: removed from algob.

packages/algob/src/lib/nft.ts
ratik21: :+1:

examples/nft/README.md
ratik21: ok

packages/algob/src/lib/nft.ts
ratik21: Simple buffer doesn't work here. The `application call` in `algosdk` expects appArgs to be an array of `Uint8Array`.
image

examples/nft/assets/approval_program.teal
ratik21: ok

ratik21:
> * [x]  Let's add a not about upgradeability (with a link to the original resource) in the readme (the one provided by Amit).
> * [x]  change representation of the token
> * [x]  Update docs (readme, comments in the smart contract files)
> * [ ]  Should we also clear user state in the clear program?

(For the 4th point) The clear call by itself clears out the user's local state. Did you mean something else?

robert-zaremba:
> (For the 4th point) The clear call by itself clears out the user's local state. Did you mean something else?

Does it clear the state for all users?

ratik21:
> Does it clear the state for all users?

No. Only the user who 'calls' the application (`--from` flag if using `goal`) using callType = ClearSSC. We can read the local data of exernal accounts (upto 4) using `app_local_get_ex`. If we want to modify/delete the data of any account other than the application caller, it must be provided in the clear call (using `--app-accounts` if using `goal`).

examples/nft/README.md
```diff
@@ -0,0 +1,30 @@
+# Non-Fungible-Token Example using StateFul Teal
```
robert-zaremba, line 1:
```suggestion
# Non-Fungible-Token Example using stateful TEAL
```

examples/nft/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+In this example, we create a new non-fungible-token represented by a name and a ref.
+Compared to standard ASA, in this example we create a smart-contract to manage and store NFT balances. 
+
+- each NFT is an `(ID, ref_data, hash)` triple, Stored in the smart contract with the mapping of `{id: ref_data, id_h: hash}`
+- `id` is the NFT ID which is calculated as - (total count of nft in asc + 1). Eg. `nft_id = 1` for the first nft in the smart contract.
+- `ref_data` is represented by encoding nft details passed to smart contract. Eg. `name: "nft-1"`, `ref: some-ref` is passed to asc, then `ref_data` = `https://nft-1.com/<nft_id>/some-ref`.
+- `id_h` is a `Sha256` hash of  `ref`.
```
robert-zaremba, line 9:
```suggestion
- Each NFT is an `(ID, ref_data, hash)` triple, stored in the smart contract as `{id: ref_data, id_h: hash}` mapping.
- `id` is the NFT ID which is calculated as: `total_count_of_nft + 1`. Eg: `id = 1` for the first nft in the smart contract.
- `ref_data` is associated reference data. Eg: `ref_data` = `https://nft.com/<nft_id>/ref`.
- `id_h` is a `sha256` hash of  the content of reference data.
```

examples/nft/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+Please check smart contract for the available commands and passing arguments to the smart contract.
```
robert-zaremba, line 11:
```suggestion
Please check smart contract for the available commands and arguments to the smart contract.
```

examples/nft/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+Also to be noted
```
robert-zaremba, line 13:
```suggestion
NOTE:
```

examples/nft/README.md
```diff
@@ -0,0 +1,30 @@
...
+* An account can hold upto a maximum of 8 NFT's in it's local storage.
+* The NFT smart contract can hold a maximum of 32 NFT's (in global storage)
+* _**This is only a proof of concept**. In production use we need to handle properly the account which can manage the smart-contract (update or delete)_. 
+* In this ASC, Update Application call is **blocked** i.e once deployed, you cannot update the smart contract by another program. Refer to [this](https://developer.algorand.org/docs/features/asc1/stateful/#update-stateful-smart-contract) for more details. 
+
+This is because the ASC can store a maximum of 16 values (`int/[]byte`) in local storage and 64 values (`int/[]byte`)  in the global storage. Read more about state storage [here](https://developer.algorand.org/docs/features/asc1/stateful/sdks/#state-storage).
```
robert-zaremba, line 19:
```suggestion
This is because the ASC can store a maximum of 16 values (`int/[]byte`) in user account local storage and 64 values (`int/[]byte`)  in the global storage. Read more about state storage [here](https://developer.algorand.org/docs/features/asc1/stateful/sdks/#state-storage).
```

examples/nft/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "htlc-pyteal",
```
robert-zaremba, line 2:
```suggestion
  "name": "nft-pyteal",
```

examples/nft/package.json
```diff
@@ -0,0 +1,12 @@
...
+  "version": "0.0.1",
```
robert-zaremba, line 3:
```suggestion
  "version": "1.0.0",
```

examples/nft/scripts/transfer/create-transfer-nft.js
```diff
@@ -0,0 +1,55 @@
...
+  const sscInfo = await deployer.getSSC("nft_approval.py", "nft_clear_state.py");
+  const appId = sscInfo.appID;
+  console.log(sscInfo);
+
+  const nft = { name: "some-nft", ref: "ref" }
+  //push arguments "create" and nft data 
+  let appArgs = [
+    new Uint8Array(Buffer.from("create")),
+    new Uint8Array(Buffer.from(nft.name)),
+    new Uint8Array(Buffer.from(nft.ref))
```
robert-zaremba, line 18: Let's create a function for this, and use it everywhere (you can put it somewhere to `alob/src/lib`):
```
function toBytes(s: string): Uint8Array {
    return new Uint8Array(Buffer.from(s))
}
```

packages/algob/src/lib/status.ts
```diff
@@ -16,6 +16,28 @@ export async function balanceOf (
...
+  creator: string,
+  appId: number): Promise<SSCStateSchema[] | undefined> {
+  const accountInfoResponse = await deployer.algodClient.accountInformation(creator).do();
+  for (const value of accountInfoResponse['created-apps']) {
+    if (value.id === appId) { return value.params['global-state']; }
+  }
+  return undefined;
+}
+
+export async function readLocalStateSSC (
```
robert-zaremba, line 30: Add documentation to both functions.

packages/algob/src/types.ts
```diff
@@ -456,6 +456,15 @@ export interface FundASCFlags {
   fundingMicroAlgo: number
 }
 
+export interface SSCStateSchema {
```
robert-zaremba, line 459:
```suggestion
Stateful Smart Contract Schema
export interface SSCStateSchema {
```

examples/nft/assets/nft_approval.py
```diff
@@ -0,0 +1,99 @@
...
+def approval_program():
+    """
+    This smart contract implements a Non Fungible Token 
+    Each NFT is an (ID, ref_data, hash) triple.
+    NFT ID is the total count of NFT's in asc.
+    ref-data is represented as "https://nft-name/<total>/nft-ref"
+    hash is an hash of the external resources data.
+
+    Commands:
+        create    Creates a new NFT. Expects two additional arguments - nft-name, nft-ref
```
robert-zaremba, line 12:
```suggestion
        create    Creates a new NFT. Expects two additional arguments: nft-name, nft-ref 
                  Only creator can create new NFTs.
```

examples/nft/assets/nft_approval.py
```diff
@@ -0,0 +1,99 @@
...
+
+    #create ref_data for NFT (https://nft-name/<total>/nft-ref)
+    ref_data = Concat(  
+        Bytes("https://"), 
+        Txn.application_args[1], # nft-name
+        Bytes(".com/"), 
+        Itob(App.globalGet(Bytes("total"))), 
+        Bytes("/"), 
+        Txn.application_args[2] # nft-ref
+    )
```
robert-zaremba, line 41: Sorry, I wasn't clear. We are not suppose to create ref data, user can put here whatever he likes. We want to suggest in the README that URL is a good idea. So, as a `ref_data` just use `Txn.application_args[1]`.

examples/nft/assets/nft_approval.py
```diff
@@ -0,0 +1,99 @@
...
+        Assert(Txn.application_args.length() == Int(0)),
+        App.globalPut(Bytes("total"), Int(0)),
+        Return(Int(1))
+    ])                  
+
+	# Checks whether the sender is creator.
+    is_creator = Txn.sender() == App.globalGet(Bytes("creator"))
+
+    # Get total count of NFT's from global storage
+    total_nft = Itob(App.globalGet(Bytes("total")))
```
robert-zaremba, line 31:
```suggestion
    nft_id = Itob(App.globalGet(Bytes("total")))
```

robert-zaremba: Don't forget to update below as well.

examples/nft/assets/nft_approval.py
```diff
@@ -0,0 +1,99 @@
...
+        Assert(is_creator),
+        Assert(Txn.application_args.length() == Int(3)),
+
+        App.globalPut(Bytes("total"), App.globalGet(Bytes("total")) + Int(1)),
+
+        App.globalPut(total_nft, ref_data),
+        App.globalPut(Concat(total_nft, Bytes("_h")) , Sha256(Txn.application_args[2])),
+
+        App.localPut(Int(0), total_nft, ref_data),  # Int(0) represents the address of caller
+        App.localPut(Int(0), Concat(total_nft, Bytes("_h")) , Sha256(Txn.application_args[2])),
```
robert-zaremba, line 59: 1. Let's store `Concat(total_nft, Bytes("_h"))` in a scratch and reuse it
2. Same for `Sha256(Txn.application_args[2])` - this operation is expensive!

examples/nft/assets/nft_approval.py
```diff
@@ -0,0 +1,99 @@
...
+    # Verifies that first argument is "vote" and jumps to on_vote.
+    program = Cond(
+        [Txn.application_id() == Int(0), on_creation],
+        [Txn.on_completion() == OnComplete.UpdateApplication, Return(Int(0))], #block update
+        [Txn.on_completion() == OnComplete.DeleteApplication, Return(is_creator)],
+        [Txn.on_completion() == OnComplete.CloseOut, Return(Int(1))],
+        [Txn.on_completion() == OnComplete.OptIn, Return(Int(1))],
+        [Txn.application_args[0] == Bytes("create"), create_nft],
+        [Txn.application_args[0] == Bytes("transfer"), transfer_nft]
+    )
```
robert-zaremba, line 94: Will the teal program fail if all conditions are false? (eg if `Txn.application_args[0] = "somethingelse"`)

robert-zaremba (changes requested):
Added few more comments. 

Thanks for updating the PR, I think the PyTEAL code is more clear and readable.

examples/nft/assets/nft_approval.py
ratik21: ok

examples/nft/assets/nft_approval.py
ratik21: :+1:

examples/nft/assets/nft_approval.py
ratik21: Yes

examples/nft/scripts/transfer/create-transfer-nft.js
ratik21: ok

packages/algob/src/lib/status.ts
ratik21: ok

examples/nft/assets/nft_approval.py
ratik21: I have cleaned up the construction a little bit. I saw that pyTEAL provides functions for scratch space (`ScratchLoad`, `ScratchStore` and `ScratchSlot`). But i coudn't find any reference or examples to use it. I am looking into it. 
[](https://pyteal.readthedocs.io/en/stable/api.html#pyteal.ScratchSlot)

examples/nft/algob.config.js
ratik21: Sorry, can you elaborate a little?

robert-zaremba: it seams we are copying the file, how about using one, even with more accounts, and instead of copying it we just make a link. So in this PR we can just create one link, other links we can do in other PR.

ratik21: Linked config to `examples-asa`.

ratik21: Also, i think we should add `algob` as a package dependency now instead of `yarn link algob`, as the user will need to do a `yarn install` to get `config`.

robert-zaremba: No, let's don't add it because we don't keep it in sync with the released package. Unless there is a way we can link using a relative path instead of npmjs.

examples/nft/assets/nft_approval.py
```diff
@@ -0,0 +1,107 @@
...
+    ])
+
+    # Always verify that the RekeyTo property of any transaction is set to the ZeroAddress 
+    # unless the contract is specifically involved ina rekeying operation.
+    no_rekey_addr = Txn.rekey_to() == Global.zero_address()
+
+	# Checks whether the sender is creator.
+    is_creator = Txn.sender() == App.globalGet(Bytes("creator"))
+
+    # Get total count of NFT's from global storage
```
robert-zaremba, line 35:
```suggestion
    # Get total amount of NFT's from global storage
```

examples/nft/assets/nft_approval.py
```diff
@@ -0,0 +1,107 @@
...
+    nft_id = Itob(App.globalGet(Bytes("total")))
+
+    # ref_data for NFT
```
robert-zaremba, line 38:
```suggestion
    # ref_data is a reference data of the NFT, usually an URL or CID.
```

examples/nft/algob.config.js
ratik21: Alright.

examples/nft/assets/nft_approval.py
```diff
@@ -0,0 +1,107 @@
...
+    # Assign hash of nft-ref to ID_h
+    # Add above two keys to global and creator's local storage
+    create_nft = Seq([
+        Assert(And(
+            Txn.application_args.length() == Int(2),
+            is_creator,
+            no_rekey_addr
+        )),
+
+        App.globalPut(Bytes("total"), App.globalGet(Bytes("total")) + Int(1)),
```
robert-zaremba, line 56: please store `Bytes("total")` in a variable and use the variable, eg:
```
var_total = Bytes("total")
```

robert-zaremba (approved):
Pre - approving. Please address the comments before merging.

examples/nft/assets/nft_approval.py
ratik21: Will update after: [https://github.com/algorand/pyteal/pull/33/files#diff-71c21b6210f42dca46ebf000fbc3425e305bea125790e986b2b6debe4c8473f8R17]()

PR #133, opened 2020-11-19
Add Deployer documentation

amityadav0:


robert-zaremba:
The task is not to copy the functions signatures, but to provide some tutorial and describe what is deploye and when it should be used. We have autogenerated API docs. I will update them on next release. So please remove the API, and you can use a link instead. I will need to update the API docs.
https://scale-it.github.io/algorand-builder/

Sorry for not being

amityadav0:
Ok, understood,  i will add description and use cases.

robert-zaremba (approved):

PR #134, opened 2020-11-19
Teal interpreter

ratik21:
TEAL Interpreter for arithmatic and arg opcodes 
- `execute(..)` to process each logic object
-  Added opcode class for `len`, `+`, `-`, `/`, `*`, `arg_0`, `arg_1`, `arg_2`, `arg_3`.
-  Added `TealError` (took reference from `BuilderError`). 
-  Added tests for `stack`, `len`, `Add`, `-`, `/`, `*`, `arg_[N]`.

Few Points (would like to hear suggestions) :-
* Max stack [capacity](https://developer.algorand.org/docs/reference/teal/specification/#the-stack) is `1000`.
* I have used `BigInt` instead of regular int/number because TEAL supports [`uint64`](https://docs.microsoft.com/en-us/dotnet/api/system.uint64.maxvalue?view=net-5.0) and max integer value for js is `2^53 - 1`. 
* [Arguments](https://developer.algorand.org/docs/features/asc1/stateless/sdks/#passing-parameters-using-the-sdks) as considered to be an array of (`string | number`). Should this be array of `Uint8Array[]`?


robert-zaremba:
> Max stack capacity is 1000.

Yes - we need to make an error if the stack will grow beyond 1000. So you need to check the stack after each operation.

> I have used BigInt instead of regular int/number because TEAL supports uint64 and max integer value for js is 2^53 - 1.

Very good. Please also check for overflows. so after each operation we need to check if the value is not bigger than `2^64`.

> Arguments as considered to be an array of (string | number). Should this be array of Uint8Array[]?

No, let's define a type for that:
```
type StackElem = string | number;
let stack: StackElem[]
```

packages/algorand-js/src/internal/interpreter/opcode-list.ts
```diff
@@ -0,0 +1,177 @@
...
+}
+
+// pushes 3rd argument from argument array to stack
+export class Arg_2 extends Op {
+  private readonly _args;
+
+  constructor (args: Array<string | number>) {
+    super();
+    this._args = args;
+  };
```
robert-zaremba, line 140: this class can inherit from `Arg_1`. Then we don't need to declare the constructor nor the `_args`,

packages/algorand-js/test/internal/interpreter/opcode-list.ts
```diff
@@ -0,0 +1,123 @@
...
+    it("should throw error if accessing arg is not defined", function () {
+      const args = [123];
+      const op = new Arg_1(args);
+      expectTealError(
+        () => op.execute(stack),
+        ERRORS.TEAL.INVALID_OPERATION
+      );
+    });
+  });
+});
```
robert-zaremba, line 123: Please add tests for all operations.

robert-zaremba (changes requested):
Looks good. Left few comments for improvements.

robert-zaremba:
I think we should never use `number` in the stack. Since we must operate on `uint64` numbers, and Node.js can't handle them, then all values must be string. So, the stack must be `Stack<string>`.

While executing operations we will need to parse stack elements and check if they are correct, example:
```
class Add {
  const op = "add"
  ...
  execute ( ... ) {
  let a = parseNum(this.op, stack.pop());  // this function will return BigInt
  let b = parseNum(this.op, stack.pop());
  ...
  }
 }
```

ratik21:
> I think we should never use `number` in the stack. Since we must operate on `uint64` numbers, and Node.js can't handle them, then all values must be string. So, the stack must be `Stack<string>`.
> 
> While executing operations we will need to parse stack elements and check if they are correct, example:
> 
> ```
> class Add {
>   const op = "add"
>   ...
>   execute ( ... ) {
>   let a = parseNum(this.op, stack.pop());  // this function will return BigInt
>   let b = parseNum(this.op, stack.pop());
>   ...
>   }
>  }
> ```

@robert-zaremba by the current design the stack stores - `stack<string | bigint>`, so we're operating on `bigint` and not number. Although i agree that keeping just string in stack would be better, but this will create some complications. Eg.
```
args = [ "123", 123 ] // arg[0] = string "123", arg[1] = number 123
```
Now if we load agrs into stack, it looks like (with string) :-
```
stack = [ "123", "123" ]
```
- consider opcode `+` on this stack (both will pe parsed and added, but it should have panicked)
- consider opcode `==`, we will compare two string and return `true`, but it should have panicked.

packages/algorand-js/src/internal/interpreter/opcode-list.ts
```diff
@@ -0,0 +1,177 @@
...
+// pops string([]byte) from stack and pushes it's length to stack
+export class Len extends Op {
+  execute (stack: IStack<string | bigint>): void {
+    this.assertStackLen(stack, 1);
+    const top = stack.pop();
+    if (typeof top !== 'string') {
+      throw new TealError(ERRORS.TEAL.INVALID_OPERATION, {
+        opcode: "len"
+      });
+    }
```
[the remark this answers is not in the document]
ratik21: ok

packages/algorand-js/src/lib/constants.ts
```diff
@@ -0,0 +1,3 @@
+// https://docs.microsoft.com/en-us/dotnet/api/system.uint64.maxvalue?view=net-5.0
+export const MAX_UINT64 = BigInt(18446744073709551615);
```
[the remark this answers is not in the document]
ratik21: ok, noted.

packages/algorand-js/src/internal/interpreter/opcode-list.ts
ratik21: Alright

packages/algorand-js/src/internal/interpreter/interpreter.ts
```diff
@@ -0,0 +1,26 @@
...
+  const stack = new Stack<string | bigint>();
+  for (const l of logic) {
+    l.execute(stack); // execute each teal opcode
+  }
+  if (stack.length() > 0) {
+    const top = stack.pop() as BigInt;
+    if (typeof top === 'string' || top >= BigInt(1)) { return true; } // Logic accept
+  }
+  return false; // Logic Reject
+}
```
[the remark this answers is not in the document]
ratik21: Okay :+1:

packages/algorand-js/test/internal/interpreter/opcode-list.ts
ratik21: :+1:

robert-zaremba:
OK, you are right. But when creating `BigInt` we should never use `number`. Only strings.

ratik21:
Ok, understood. Updating..

packages/algorand-js/src/internal/interpreter/interpreter.ts
```diff
@@ -0,0 +1,33 @@
...
+    this._stack = new Stack<StackElem>();
+  }
+
+  /**
+ * Description: this function executes set of Operator[] passed after
+ * parsing teal code
+ * @param {execParams} txn : Transaction parameters for current txn
+ * @param {Logic[]} logic : set of Logic objects
+ * @param {AppArgs} args : argument array passed to the current transaction
+ * @returns {boolean} : transaction accepted/rejected based on asc logic
```
robert-zaremba, line 20: I think this returns linter error - we don't need to do type annotation in comments, because we already have them in the function definition. If there is no linter error then leave it.

packages/algorand-js/src/internal/interpreter/interpreter.ts
```diff
@@ -0,0 +1,33 @@
+import type { execParams } from "algob/src/types";
+import { assert } from "chai";
+
+import { IStack, Stack } from "../../lib/stack";
+import type { AppArgs, Operator, StackElem } from "../types";
+
+export class Interpreter {
+  readonly _stack: IStack<StackElem>;
```
robert-zaremba, line 8: why this variable has `_` prefix?

packages/algorand-js/src/internal/interpreter/interpreter.ts
```diff
@@ -0,0 +1,33 @@
...
+ * @param {AppArgs} args : argument array passed to the current transaction
+ * @returns {boolean} : transaction accepted/rejected based on asc logic
+ */
+  execute (txn: execParams, logic: Operator[], args: AppArgs): boolean {
+    assert(Array.isArray(args));
+    for (const l of logic) {
+      l.execute(this._stack); // execute each teal opcode
+    }
+    if (this._stack.length() > 0) {
+      const top = this._stack.pop() as BigInt;
```
robert-zaremba, line 28: This type assertion doesn't make sense, because below you are using `typeof top === "string"`

packages/algorand-js/src/internal/interpreter/interpreter.ts
```diff
@@ -0,0 +1,33 @@
...
+      if (typeof top === 'string' || top >= BigInt("1")) { return true; } // Logic accept
```
robert-zaremba, line 29: 1. we can remove if and do just a return: `typeof top === 'string' || top >= BigInt("1")`
2. Why do we return true if top is a string? What if on the top we have empty string (`""`) or `"0"`?

packages/algorand-js/src/internal/interpreter/opcode-list.ts
```diff
@@ -0,0 +1,159 @@
...
+import { parseNum } from "../utils/bigint";
+import { Op } from "./opcode";
+
+const BIGINT0 = BigInt("0");
+// pops string([]byte) from stack and pushes it's length to stack
+export class Len extends Op {
+  execute (stack: IStack<StackElem>): void {
+    this.assertStackLen(stack, 1);
+    const top = stack.pop();
+    if (typeof top !== 'string') {
```
robert-zaremba, line 14: Bytes can't be represent as strings. Bytes must be represented as `Uint8Array`.

packages/algorand-js/src/internal/interpreter/opcode-list.ts
```diff
@@ -0,0 +1,159 @@
...
+export class Add extends Op {
+  execute (stack: IStack<StackElem>): void {
+    this.assertStackLen(stack, 2);
+    const a = stack.pop();
+    const b = stack.pop();
+    if (typeof a !== 'bigint' || typeof b !== 'bigint') {
+      throw new TealError(ERRORS.TEAL.INVALID_OP_ARG, {
+        opcode: "+"
+      });
+    }
```
robert-zaremba, line 34: let's create a function to handle this:
```
function assertBigInt(a: unknown)
```

robert-zaremba: And let's use that function in all operators below:
```
assertBigInt(a);
assertBigInt(b);
```

packages/algorand-js/src/internal/types.ts
```diff
@@ -0,0 +1,10 @@
+import {
+  Add, Arg_0, Arg_1, Arg_2,
+  Arg_3, Div, Len, Mul, Sub
+} from "../internal/interpreter/opcode-list";
+
+export type Operator = Len | Add | Sub |
+Mul | Div | Arg_0 | Arg_1 | Arg_2 | Arg_3;
+
+export type AppArgs = Array<string | number>;
+export type StackElem = string | bigint;
```
robert-zaremba, line 10: I've verified the opcodes.
It seams that the stack values are either `[]byte` or `uint64`. Hence we should change `string` to `Uint8Array`

robert-zaremba: Maybe we can create a shortcut type, and use it in the interpreter:
```
type TEALStack = IStack<BigInt | Uint8Array>
```

robert-zaremba (changes requested):
We need to change the type of Stack. Please see my comments. We shouldn't use `string` at all. It's dangerous because the length and indexing in string is different than in byte array.

- [x] Change stack type and review all op codes that we correctly use bytes
- [x] Don't use `internal` nor `core` directory in this : `errors` can be in `src/errors`, interpreter can be in `src/interpreter`.

packages/algorand-js/src/internal/interpreter/interpreter.ts
ratik21: Does not throw a linter error. I will remember this when writing doc string for functions in the future. Leaving it as the same for now.

packages/algorand-js/src/internal/types.ts
ratik21: Okay, makes sense.

packages/algorand-js/src/internal/interpreter/opcode-list.ts
ratik21: ok

packages/algorand-js/src/internal/interpreter/interpreter.ts
ratik21: Sorry, you are correct. Updating

robert-zaremba: Let's add tests for that.

robert-zaremba: We need to be 100% sure that all this things work correctly.

packages/algorand-js/src/internal/interpreter/opcode-list.ts
ratik21: ok

packages/algorand-js/src/internal/interpreter/interpreter.ts
ratik21: changed to `stack`.

packages/algorand-js/src/internal/interpreter/interpreter.ts
ratik21: :+1:

packages/algorand-js/src/interpreter/opcode-list.ts
```diff
@@ -0,0 +1,143 @@
...
+    this._args = args;
+  };
+
+  execute (stack: TEALStack): void {
+    const arg = this._args[0];
+    if (arg === undefined || (typeof arg !== 'string' && typeof arg !== 'number')) {
+      throw new TealError(ERRORS.TEAL.INVALID_OP_ARG, {
+        opcode: "arg_0"
+      });
+    }
```
robert-zaremba, line 89: extract `if` to a function:
```
const arg = assertStackElem(arg)
```

robert-zaremba: Also, `BigInt` is also a valid stack element.

packages/algorand-js/src/interpreter/opcode-list.ts
```diff
@@ -0,0 +1,143 @@
...
+    if (typeof arg === 'number') {
+      this.checkOverFlow(parseNum(arg));
+      stack.push(parseNum(arg));
+    } else { stack.push(toBytes(arg)); }
```
robert-zaremba, line 93: I think we should disallow `number`. `arg` should have same type as `StackElem`: only `BigInt` or `Uint8Array`

packages/algorand-js/src/interpreter/opcode.ts
```diff
@@ -0,0 +1,31 @@
...
+import type { TEALStack } from "../types";
+
+export class Op {
+  assertStackLen (stack: TEALStack, minLen: number): void {
+    if (stack.length() < minLen) {
+      throw new TealError(ERRORS.TEAL.ASSERT_STACK_LENGTH);
+    }
+  }
+
+  checkOverFlow (num: bigint): void {
```
robert-zaremba, line 13: rename to `checkOverflow`

packages/algorand-js/src/interpreter/opcode.ts
```diff
@@ -0,0 +1,31 @@
...
+    if (num > MAX_UINT64) {
+      throw new TealError(ERRORS.TEAL.UINT64_OVERFLOW);
+    }
+  }
+
+  checkUnderFlow (num: bigint): void {
```
robert-zaremba, line 19: Rename to `checkUnderflow` (no big F)

robert-zaremba (changes requested):
Still few things to change.

packages/algorand-js/src/interpreter/opcode-list.ts
ratik21: Ok. But, since the user is passing the arguments, is it good to ask the user to pass `bigint` instead of regular `int`?

robert-zaremba: The input is a TEAL code - so everything is a text there. We will read that text an parse it accordingly.

robert-zaremba: ah, important thing, I've just realized that args are ALWAYS `bytes`!

robert-zaremba: https://developer.algorand.org/docs/reference/teal/opcodes/#arg

packages/algorand-js/src/interpreter/opcode-list.ts
```diff
@@ -0,0 +1,143 @@
...
+export class Div extends Op {
+  execute (stack: TEALStack): void {
+    this.assertStackLen(stack, 2);
+    const a = this.assertBigInt(stack.pop());
+    const b = this.assertBigInt(stack.pop());
+    if (b === BIGINT0) {
+      throw new TealError(ERRORS.TEAL.INVALID_OP_ARG, {
+        opcode: "/"
+      });
+    }
```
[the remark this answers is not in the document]
ratik21: ok

packages/algorand-js/src/interpreter/opcode-list.ts
```diff
@@ -0,0 +1,143 @@
...
+
+// pops two unit64 from stack(a, b) and pushes their sum(a + b) to stack
+// panics on overflow (result > max_unit64)
+export class Add extends Op {
+  execute (stack: TEALStack): void {
+    this.assertStackLen(stack, 2);
+    const a = this.assertBigInt(stack.pop());
+    const b = this.assertBigInt(stack.pop());
+    this.checkOverFlow(a + b);
+    stack.push(a + b);
```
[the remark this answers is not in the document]
ratik21: ok

packages/algorand-js/src/interpreter/opcode-list.ts
```diff
@@ -0,0 +1,143 @@
...
+// pops string([]byte) from stack and pushes it's length to stack
+export class Len extends Op {
+  execute (stack: TEALStack): void {
+    this.assertStackLen(stack, 1);
+    const top = stack.pop();
+    if (typeof top === 'undefined' || !(top instanceof Uint8Array)) {
+      throw new TealError(ERRORS.TEAL.INVALID_OP_ARG, {
+        opcode: "len"
+      });
+    }
```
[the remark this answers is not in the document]
ratik21: :+1:

packages/algorand-js/src/interpreter/opcode-list.ts
ratik21: Yes. :+1: Updating to `bytes`

robert-zaremba (approved):

PR #136, opened 2020-11-23
Add support for external paramters to PyTEAL scritps

amityadav0:
`scInitParam` Object can be passed from scripts.

- This param is taken as Object and then converted into string using yaml.stringify.
- passed to python scripts
- decoded in python scripts

We cannot determine the type of user input for smart contract initialization paramerers, so `type` of `scInitParam` is `Object`.

dynamic-fee example is updated to show how we can pass initialization param.

### DIFF
--- a/examples/ref-templates/assets/dynamic-fee.py
+++ b/examples/ref-templates/assets/dynamic-fee.py
@@ -98,8 +98,8 @@ if __name__ == "__main__":
     "TMPL_LEASE": "023sdDE2"
   }
 
-  # decode external parameters and update current values.
-  # (if external paramters are passed)
+  # decode external parameter and update current values.
+  # (if an external paramter is passed)
   if(sys.argv[1] != "\n"):
     try:
       param = yaml.safe_load(sys.argv[1])
--- a/examples/ref-templates/assets/dynamic-fee.py
+++ b/examples/ref-templates/assets/dynamic-fee.py
@@ -1,12 +1,7 @@
-from pyteal import *
+import sys
+import yaml
 
-#replace the following params with your custom values
-john = Addr("2UBZKFR6RCZL7R24ZG327VKPTPJUPFM6WTG7PJG2ZJLU234F5RGXFLTAKA")
-master = Addr("WWYNX3TKQYVEREVSW6QQP3SXSFOCE3SKUSEIVJ7YAGUPEACNI5UGI4DZCE")
-amount = 700000
-first_valid = 10
-last_valid = 1000000
-lease = Bytes("base64", "023sdDE2")
+from pyteal import *
 
 def dynamic_fee(TMPL_TO,
                 TMPL_AMT,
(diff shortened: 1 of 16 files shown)


packages/algob/src/internal/deployer.ts
```diff
@@ -102,8 +102,8 @@ class DeployerBasicMode {
    * @param scParams parameters
    * @returns {LogicSig} loaded logic signature from assets/<file_name>.teal
    */
-  async loadLogic (name: string, scParams: LogicSigArgs): Promise<LogicSig> {
-    return await getLsig(name, scParams, this.algoOp.algodClient);
+  async loadLogic (name: string, scParams: LogicSigArgs, scInitParam?: Object): Promise<LogicSig> {
```
robert-zaremba, line 105: Instead of using an `Object`, let's use `unknown` because we don't inherit any properties from the common JS objects.

This is a good reading: https://medium.com/@KevinBGreene/a-little-theory-with-your-typescript-top-and-bottom-types-61b380f227d

examples/ref-templates/assets/dynamic-fee.py
```diff
@@ -92,4 +87,31 @@ def dynamic_fee(TMPL_TO,
...
+
+  # decode external parameter and update current values.
+  # (if an external paramter is passed)
+  if(sys.argv[1] != "\n"):
+    try:
+      param = yaml.safe_load(sys.argv[1])
+      for key, value in param.items():
+        scParam[key] = value
+    except yaml.YAMLError as exc:
+      print(exc)
```
robert-zaremba, line 109: 1. Why `sys.argv[1]` will thrown an exception if no parameter is provided. Also, why do we check against `"\n"`?
1. let's create a function for that: `parseArgs`, we can put it in `examples/algobpy` (since this will be a python package, we will need to create an empty `__init__.py` file there to make sure we can import it in other places), and use it in all PyTEAL scripts. The function can return an object or `None`. If object is returned we can overwrite a default object.
1. let's use the function above in all other PyTEAL scripts.

packages/algob/src/lib/algo-operator.ts
```diff
@@ -210,8 +211,9 @@ export class AlgoOperatorImpl implements AlgoOperator {
     scParams: LogicSigArgs,
     flags: FundASCFlags,
     payFlags: TxParams,
-    txWriter: txWriter): Promise<LsigInfo> {
-    const lsig = await getLsig(name, scParams, this.algodClient);
+    txWriter: txWriter,
+    scInitParam?: Object): Promise<LsigInfo> {
```
robert-zaremba, line 215: let's put this next to the `scParams` and please add docstring explaining all arguments.

packages/algob/src/lib/compile.ts
```diff
@@ -31,9 +31,13 @@ export class CompileOp {
   //   (Examples: `mysc.teal, security/rbac.teal`)
   //   MUST have a .teal, .lsig or .py extension
   // @param force: if true it will force recompilation even if the cache is up to date.
-  async ensureCompiled (filename: string, force: boolean): Promise<ASCCache> {
+  async ensureCompiled (filename: string, force?: boolean, scInitParam?: Object): Promise<ASCCache> {
```
robert-zaremba, line 34: Please update the docstring to explain parameters.

packages/algob/src/lib/compile.ts
```diff
@@ -117,13 +121,16 @@ export class PyCompileOp {
    *                   Examples : [ gold.py, asa.py]
    *                   MUST have .py extension
    * @param force    : if true it will force recompilation even if the cache is up to date.
+   * @param scInitParam : Smart Contract Initialization Parameters
    */
-  async ensureCompiled (filename: string, force: boolean): Promise<PyASCCache> {
+  async ensureCompiled (filename: string, force?: boolean, scInitParam?: Object): Promise<PyASCCache> {
```
robert-zaremba, line 126: Please update the docstring to explain parameters.

packages/algob/src/lib/compile.ts
```diff
@@ -165,20 +172,29 @@ export class PyCompileOp {
    * Description: Runs a subprocess to execute python script
    * @param filename : python filename in assets folder
    */
-  private runPythonScript (filename: string): SpawnSyncReturns<string> {
+  private runPythonScript (filename: string, scInitParam?: string): SpawnSyncReturns<string> {
```
robert-zaremba, line 175: docstring update

robert-zaremba (changes requested):

packages/algob/src/internal/deployer.ts
amityadav0: OK

examples/ref-templates/assets/dynamic-fee.py
amityadav0: 1. when using `spawanSync ` without arg 
`spawnSync('python3', [path.join(ASSETS_DIR, filename) ], {..} );`
argv[0] = scripts/name and argv[1] becomes `\n` line
that's why it doesn't throw exception and using "\n". I checked by running both the cases.

2., 3.  Ok

packages/algob/src/lib/algo-operator.ts
amityadav0: Ok, will have to move scParam next to scInitParam as scInitParam is optional and can't be put before required

packages/algob/src/lib/compile.ts
amityadav0: ok

packages/algob/src/lib/compile.ts
amityadav0: ok

packages/algob/src/lib/compile.ts
amityadav0: ok

examples/ref-templates/assets/dynamic-fee.py
amityadav0: For 1. I was doing YAML.stringify() before checking if scInitParam is undefined, that's why it was converted into "\n"

Adding a check.

robert-zaremba (approved):

PR #137, opened 2020-11-24
Interpreter bytec op

ratik21:
- Added Classes
  - `Bytecblock`
  - `Bytec`
  - `Intcblock`
  - `Intc`
- Added `tests`

### DIFF
--- a/packages/algorand-js/src/interpreter/opcode-list.ts
+++ b/packages/algorand-js/src/interpreter/opcode-list.ts
@@ -111,6 +111,7 @@ export class Bytec extends Op {
   }
 
   execute (stack: TEALStack): void {
+    this.checkIndexBound(this.index, this.interpreter.bytecblock);
     const bytec = this.assertBytes(this.interpreter.bytecblock[this.index]);
     stack.push(bytec);
   }
@@ -145,6 +146,7 @@ export class Intc extends Op {
   }
 
   execute (stack: TEALStack): void {
+    this.checkIndexBound(this.index, this.interpreter.intcblock);
     const intc = this.assertBigInt(this.interpreter.intcblock[this.index]);
     stack.push(intc);
   }
--- a/packages/algorand-js/src/interpreter/opcode-list.ts
+++ b/packages/algorand-js/src/interpreter/opcode-list.ts
@@ -1,6 +1,7 @@
 import { TealError } from "../errors/errors";
 import { ERRORS } from "../errors/errors-list";
 import type { TEALStack } from "../types";
+import { Interpreter } from "./interpreter";
 import { Op } from "./opcode";
 
 const BIGINT0 = BigInt("0");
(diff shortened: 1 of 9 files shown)


packages/algorand-js/src/interpreter/opcode-list.ts
```diff
@@ -80,3 +81,37 @@ export class Arg extends Op {
...
+
+// load block of byte-array constants
+export class Bytecblock extends Op {
+  readonly bytecblock: Uint8Array[];
+  readonly interpreter: Interpreter;
+
+  constructor (interpreter: Interpreter, bytecblock: Uint8Array[]) {
+    super();
+    this.interpreter = interpreter;
+    this.bytecblock = bytecblock;
```
robert-zaremba, line 93: we don't need to store `bytecblock`. It's already in the interpreter.

packages/algorand-js/src/interpreter/opcode.ts
```diff
@@ -22,6 +22,12 @@ export class Op {
     }
   }
 
+  assertArrLength (arr: Uint8Array[]): void {
+    if (arr.length > MAX_UINT8 + 1) {
+      throw new TealError(ERRORS.TEAL.ASSERT_ARR_LENGTH);
+    }
```
robert-zaremba, line 28: it must be also bigger than 0 (we need to have at least one element)

packages/algorand-js/test/src/interpreter/opcode-list.ts
```diff
@@ -266,4 +268,88 @@ describe("Teal Opcodes", function () {
...
+    });
+
+    it("should load byte block to interpreter bytecblock", function () {
+      const interpreter = new Interpreter();
+      const bytecblock = [
+        toBytes("bytec_0"),
+        toBytes("bytec_1"),
+        toBytes("bytec_2"),
+        toBytes("bytec_3")
+      ];
```
robert-zaremba, line 296: you can do:
```
const bytecblock = ["b1", "b2", ...].map(toBytes)
```

robert-zaremba (changes requested):

packages/algorand-js/test/src/interpreter/opcode-list.ts
ratik21: Good suggestion. Thanks :+1:

packages/algorand-js/src/interpreter/opcode.ts
ratik21: ok

packages/algorand-js/src/interpreter/opcode-list.ts
ratik21: Sorry, i don't follow. `this.bytecblock` is storing the `bytecblock` provided explicitly to the `new Bytecblock(..)` constructor. This is from what we are updating the interpreter `bytecblock` with.

robert-zaremba: You are right. sorry.

packages/algorand-js/src/interpreter/opcode-list.ts
```diff
@@ -80,3 +81,71 @@ export class Arg extends Op {
...
+  readonly interpreter: Interpreter;
+
+  constructor (idx: number, interpreter: Interpreter) {
+    super();
+    this.index = idx;
+    this.interpreter = interpreter;
+  }
+
+  execute (stack: TEALStack): void {
+    const bytec = this.assertBytes(this.interpreter.bytecblock[this.index]);
```
robert-zaremba, line 114: Please add a check that the index doesn't exceeds the array. We should have a clear message for it rather than falling down into `undefined` check.
Same for `Intc`

robert-zaremba (approved):
Left one comment. Otherwise looks good.

packages/algorand-js/src/interpreter/opcode-list.ts
ratik21: Alright

PR #139, opened 2020-11-25
update nft: show count

ratik21:


examples/nft/scripts/transfer/common.js
```diff
@@ -1,9 +1,38 @@
...
+}
+
+exports.printLocalNFT = async function (deployer, account, appId) {
+    try {
+        const localState = await readLocalStateSSC(deployer, account, appId);
+        let count = 0;
+
+        // length/2 because each nft is a pair of {id: ref_data, id_h: ref_hash}
+        if(localState !== undefined){ 
+            count = localState.length/2;
```
amityadav0, line 32: Or we can directly make count = 1,
what happens when this NFT is transferred to another user ? will this localState will become undefined for this user ?

examples/nft/scripts/transfer/create-transfer-nft.js
```diff
@@ -1,4 +1,8 @@
-const { executeTransaction } = require("./common");
+/**
+ * Description:
+ * This file create a new NFT and transfers 1 NFT from A to B
```
amityadav0, line 3:
```suggestion
 * This file creates a new NFT and transfers 1 NFT from A to B
```

examples/nft/scripts/transfer/common.js
ratik21: localState undefined means that there is nothing present in the local `ledger` (`accountInfo['apps-local-state']`). `goal` returns null. But if we are using sdk then it means we are trying to access value from an empty state (i.e if no nft is present, local state is empty), so we get undefined. 
So, if account A holds **one** NFT, and that nft is transferred to another account, then yes, localState becomes `undefined` (explained above). Let me know if that makes sense.

amityadav0: Okay, that means we can make count one if local state is defined and count zero if it is undefined.

ratik21: `count = 0` should be the default. If localState is `defined`, i am updating the count (as this means there are NFT's present in the `ledger`). Am i missing something here?

amityadav0: No, you are right, I thought  `localState.length/2; ` will always be one, but one application contract can have multiple NFT.
How about updating this to show NFT name and balance ? instead of showing total number of NFTs.

ratik21: We can, but this function's purpose is to display `balance` before and after the transaction. If there are multiple NFT's (say 8), before the txn we display 8 nft-names for acc_A, and after the txn we display 7. Isn't it a but cleaner the current way (just showing count)?
Also, we already have functions to print local and global state data: `printLocalStateSSC`, `printGlobalStateSSC`. We can use these functions directly to display state data. We may need to refactor to display `base64-decoded` data.

robert-zaremba (approved):

PR #140, opened 2020-11-25
Interpreter scratch op

ratik21:
- Added classes `Store`, `Load`
- Added tests

NOTE:
+ `load x` does not throw error if nothing is stored is scratch. Some default value is loaded to stack. I tested that using 
```
load 1
load 2
+
```
The logic fails. But teal is compiled and run successfully.

### DIFF
--- a/packages/algorand-js/src/interpreter/interpreter.ts
+++ b/packages/algorand-js/src/interpreter/interpreter.ts
@@ -3,7 +3,7 @@ import { assert } from "chai";
 
 import { TealError } from "../errors/errors";
 import { ERRORS } from "../errors/errors-list";
-import { DEFAULT_STACKELEM } from "../lib/constants";
+import { DEFAULT_STACK_ELEM } from "../lib/constants";
 import { Stack } from "../lib/stack";
 import type { Operator, StackElem, TEALStack } from "../types";
 
@@ -17,7 +17,7 @@ export class Interpreter {
     this.stack = new Stack<StackElem>();
     this.bytecblock = [];
     this.intcblock = [];
-    this.scratch = new Array(256).fill(DEFAULT_STACKELEM);
+    this.scratch = new Array(256).fill(DEFAULT_STACK_ELEM);
   }
 
   /**
--- a/packages/algorand-js/src/interpreter/interpreter.ts
+++ b/packages/algorand-js/src/interpreter/interpreter.ts
@@ -3,6 +3,7 @@ import { assert } from "chai";
 
 import { TealError } from "../errors/errors";
 import { ERRORS } from "../errors/errors-list";
+import { DEFAULT_STACKELEM } from "../lib/constants";
 import { Stack } from "../lib/stack";
 import type { Operator, StackElem, TEALStack } from "../types";
 
@@ -10,11 +11,13 @@ export class Interpreter {
   readonly stack: TEALStack;
   bytecblock: Uint8Array[];
   intcblock: Array<bigint>;
+  scratch: StackElem[];
 
   constructor () {
     this.stack = new Stack<StackElem>();
     this.bytecblock = [];
     this.intcblock = [];
+    this.scratch = new Array(256).fill(DEFAULT_STACKELEM);
   }
 
   /**
--- a/packages/algorand-js/src/interpreter/interpreter.ts
+++ b/packages/algorand-js/src/interpreter/interpreter.ts
@@ -10,7 +10,7 @@ import type { Operator, StackElem, TEALStack } from "../types";
 export class Interpreter {
   readonly stack: TEALStack;
   bytecblock: Uint8Array[];
-  intcblock: Array<bigint>;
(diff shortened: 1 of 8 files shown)


packages/algorand-js/src/interpreter/interpreter.ts
```diff
@@ -3,18 +3,21 @@ import { assert } from "chai";
...
 import { ERRORS } from "../errors/errors-list";
+import { DEFAULT_STACKELEM } from "../lib/constants";
 import { Stack } from "../lib/stack";
 import type { Operator, StackElem, TEALStack } from "../types";
 
 export class Interpreter {
   readonly stack: TEALStack;
   bytecblock: Uint8Array[];
   intcblock: Array<bigint>;
+  scratch: StackElem[];
```
amityadav0, line 14: does scratch offers stack storage or single storage ?

packages/algorand-js/src/interpreter/opcode-list.ts
```diff
@@ -151,3 +152,40 @@ export class Intc extends Op {
...
+  constructor (index: number, interpreter: Interpreter) { // eslint-disable-line sonarjs/no-identical-functions
+    super();
+    this.index = index;
+    this.interpreter = interpreter;
+  }
+
+  execute (stack: TEALStack): void {
+    this.checkIndexBound(this.index, this.interpreter.scratch);
+    const val = this.interpreter.scratch[this.index];
+    stack.push(val);
```
robert-zaremba, line 189:
```suggestion
    stack.push(this.interpreter.scratch[this.index]);
```

packages/algorand-js/src/interpreter/opcode-list.ts
```diff
@@ -151,3 +152,40 @@ export class Intc extends Op {
...
+  constructor (index: number, interpreter: Interpreter) { // eslint-disable-line sonarjs/no-identical-functions
+    super();
+    this.index = index;
+    this.interpreter = interpreter;
+  }
+
+  execute (stack: TEALStack): void {
+    this.checkIndexBound(this.index, this.interpreter.scratch);
+    this.assertStackLen(stack, 1);
+    const top = stack.pop() as StackElem;
```
robert-zaremba, line 170: how possible `stack.pop()` doesn't return `StackElem`?

robert-zaremba (changes requested):

packages/algorand-js/src/interpreter/opcode-list.ts
ratik21: That's because the `pop()` definition of stack returns `T | undefined`. 
image
I am changing it to `T`.

packages/algorand-js/src/interpreter/interpreter.ts
```diff
@@ -3,18 +3,21 @@ import { assert } from "chai";
...
 import { TealError } from "../errors/errors";
 import { ERRORS } from "../errors/errors-list";
+import { DEFAULT_STACKELEM } from "../lib/constants";
 import { Stack } from "../lib/stack";
 import type { Operator, StackElem, TEALStack } from "../types";
 
 export class Interpreter {
   readonly stack: TEALStack;
   bytecblock: Uint8Array[];
   intcblock: Array<bigint>;
```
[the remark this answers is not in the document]
ratik21: Sorry, i actually used `bigint[]` which was changed to `Array<..>` by `linter`. Updated.

packages/algorand-js/src/interpreter/opcode-list.ts
robert-zaremba: 1. You have typo: UnderFlow -> underflow
2. I think underflow is a wrong word here. Use `Error("pop from empty stack")`

ratik21: Ok. updated.

robert-zaremba:
>    load x does not throw error if nothing is stored is scratch. Some default value is loaded to stack. I tested that using

Compiler doesn't execute a code. I think we should fail in this scenario (where we load something which is not defined).

packages/algorand-js/src/interpreter/opcode-list.ts
```diff
@@ -151,3 +151,39 @@ export class Intc extends Op {
     stack.push(intc);
   }
 }
+
+// pop a value from the stack and store to scratch space
+export class Store extends Op {
+  readonly index: number;
+  readonly interpreter: Interpreter;
+
+  constructor (index: number, interpreter: Interpreter) { // eslint-disable-line sonarjs/no-identical-functions
```
robert-zaremba, line 160: you can put the eslint comment to in the top of the file:
```
// eslint-disable sonarjs/no-identical-functions
```

robert-zaremba (approved):

packages/algorand-js/src/interpreter/opcode-list.ts
ratik21: done.

PR #142, opened 2020-11-26
Interpreter bitwise op

ratik21:
- Added classes for `%`, `|`, `&`, `^`, `~`
- Added tests

### DIFF
--- a/packages/algorand-js/test/src/interpreter/opcode-list.ts
+++ b/packages/algorand-js/test/src/interpreter/opcode-list.ts
@@ -3,9 +3,10 @@ import { assert } from "chai";
 import { ERRORS } from "../../../src/errors/errors-list";
 import { Interpreter } from "../../../src/interpreter/interpreter";
 import {
-  Add, Arg, Bytec,
+  Add, Arg, BitwiseAnd, BitwiseNot, BitwiseOr, BitwiseXor,
+  Bytec,
   Bytecblock, Div, Intc,
-  Intcblock, Len, Mul, Sub
+  Intcblock, Len, Mod, Mul, Sub
 } from "../../../src/interpreter/opcode-list";
 import { MAX_UINT8, MAX_UINT64 } from "../../../src/lib/constants";
 import { toBytes } from "../../../src/lib/parse-data";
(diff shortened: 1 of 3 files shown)


packages/algorand-js/test/src/interpreter/opcode-list.ts
```diff
@@ -412,4 +413,187 @@ describe("Teal Opcodes", function () {
...
+    const stack = new Stack<StackElem>();
+
+    it("should return correct modulo of two unit64", function () {
+      stack.push(BigInt("2"));
+      stack.push(BigInt("5"));
+      const op = new Mod();
+      op.execute(stack);
+
+      const top = stack.pop();
+      assert.equal(top, BigInt("1"));
```
robert-zaremba, line 427: Please add tests for:
* (7, 7)


Can be in the same unit-test (same "it")

packages/algorand-js/test/src/interpreter/opcode-list.ts
```diff
@@ -412,4 +413,187 @@ describe("Teal Opcodes", function () {
...
+    });
+
+    it("should throw error if bitwise-xor is used with strings", function () {
+      stack.push(toBytes("str1"));
+      stack.push(toBytes("str2"));
+      const op = new BitwiseXor();
+      expectTealError(
+        () => op.execute(stack),
+        ERRORS.TEAL.INVALID_TYPE
+      );
```
robert-zaremba, line 565: for checking failure on strings, you can have a function and use it in tests multiple times:
```
function execExpectError(stack, strs: string[],op: Op, err) {
  return () => {
    for (let s of strs) stack.push(s)
    expectTealError(() => op.execute(stack), err)
  }
}

it("should ...", execExpectError("str1", "str2", newe BitwiseXor(), ERRORS.TEAL.INVALID_TYPE)
```

And you can reuse it in few places above.

robert-zaremba (approved):

packages/algorand-js/test/src/interpreter/opcode-list.ts
ratik21: ok

packages/algorand-js/test/src/interpreter/opcode-list.ts
ratik21: Yes. Thanks

PR #146, opened 2020-11-30
Make sure that supporting external parameters will recompile a PyTEAL program

amityadav0:
- make test that this works as expected.
   - Test both stateful and stateless smart contracts


packages/algob/test/builtin-tasks/compile.ts
```diff
@@ -125,3 +125,71 @@ describe("Compile task", () => {
...
+  it("On second run, shouln't recompile because No external parameters passed", async () => {
+    await op.resetAndCompile(false);
+
+    // timestamp didn't change because file is not recompiled
+    assert.equal(op.timestamp, 2);
+    assert.lengthOf(op.compiledFiles, 0);
+    assert.lengthOf(op.writtenFiles, 0);
+  });
+
+  it("On third run, should compile on third run because external parameters passed", async () => {
```
robert-zaremba, line 173: will the last 2 unit tests work if the first one will fail? If not then we can put all of them in a single test (`it(...)`)

packages/algob/test/fixture-projects/support-external-parameters/algob.config.js
```diff
@@ -0,0 +1,55 @@
...
+  token: 'some-fake-token',
+  wallets: [
+    {
+      name: 'some-fake-wallet',
+      password: '',
+      accounts: [
+        { name: 'name-1', address: 'some-addr' }]
+    }
+  ]
+}
```
robert-zaremba, line 27: do we need `kmdCfg` for this test project?

packages/algob/test/fixture-projects/support-external-parameters/algob.config.js
```diff
@@ -0,0 +1,55 @@
...
+      token: 'somefaketoken',
+      accounts: accounts
+    },
+    kmdNet: {
+      host: 'http://127.0.0.1',
+      port: 8080,
+      token: 'somefaketoken',
+      accounts: accounts,
+      kmdCfg: kmdCfg
+    }
```
robert-zaremba, line 52: I think one network is enough here.

packages/algob/test/fixture-projects/support-external-parameters/algob.config.js
amityadav0: No, cleaning config

packages/algob/test/fixture-projects/support-external-parameters/algob.config.js
amityadav0: ok

packages/algob/test/builtin-tasks/compile.ts
amityadav0: ok

robert-zaremba (approved):

PR #147, opened 2020-11-30
Change devnet language

ratik21:


robert-zaremba (approved):
I thought I will do it ;) Thanks.

PR #149, opened 2020-12-01
Log NFT-ID's in user's account

ratik21:


examples/nft/scripts/transfer/common.js
```diff
@@ -26,11 +27,18 @@ exports.printLocalNFT = async function (deployer, account, appId) {
     try {
         const localState = await readLocalStateSSC(deployer, account, appId);
         // each nft is stored as a one record in user store
-        let count = 0;
-        if (localState !== undefined)
-            count = localState.length
-        console.log('NFT balance of', account, ':', count);
+        let holdings = "";
+        if (localState === undefined) {
+            holdings = "null";
```
robert-zaremba, line 32:
```suggestion
            holdings = "none";
```

examples/nft/scripts/transfer/common.js
```diff
@@ -26,11 +27,18 @@ exports.printLocalNFT = async function (deployer, account, appId) {
...
+        }
+        else {
+            for(const l of localState) {
+                const key = decode(Buffer.from(l.key, 'base64'));
+                holdings += key.toString() + ' ';
+            }
+        }
+        console.log('NFT holdings of', account, ':', holdings);
```
robert-zaremba, line 40: can you show an example output please?

ratik21: Sure.
- Before Transfer
image
- After Transfer
image

robert-zaremba: thanks. I will update the message text.

examples/nft/scripts/transfer/common.js
```diff
@@ -26,10 +27,17 @@ exports.printLocalNFT = async function (deployer, account, appId) {
...
+        if (localState === undefined) {
+            holdings = "none";
+        }
+        else {
+            for(const l of localState) {
+                const key = decode(Buffer.from(l.key, 'base64'));
+                holdings += key.toString() + ' ';
+            }
+        }
+        console.log('NFT holdings of', account, ':', holdings);
```
robert-zaremba, line 40:
```suggestion
        console.log('%s account holds app(%s) NFTs: ', account, appId, holdings);
```

examples/nft/scripts/transfer/common.js
```diff
@@ -26,10 +27,17 @@ exports.printLocalNFT = async function (deployer, account, appId) {
...
-            count = localState.length
-        console.log('NFT balance of', account, ':', count);
+        let holdings = "";
+        if (localState === undefined) {
+            holdings = "none";
+        }
+        else {
+            for(const l of localState) {
+                const key = decode(Buffer.from(l.key, 'base64'));
+                holdings += key.toString() + ' ';
```
robert-zaremba, line 37: `holdings` should be an array. Avoid using string concatenation in a loop. Instead use `array.push` and after the loop you can use `array.join` to create a string.

robert-zaremba (approved):
pre-approving.

examples/nft/scripts/transfer/common.js
ratik21: ok :+1:

PR #150, opened 2020-12-02
Add more TEAL opcodes

ratik21:
- Added opcodes: `substring`, `mulw`, `Int`, `Byte`, `Addr`
- Added tests

### DIFF
--- a/packages/algorand-js/src/interpreter/opcode-list.ts
+++ b/packages/algorand-js/src/interpreter/opcode-list.ts
@@ -1,4 +1,5 @@
 /* eslint sonarjs/no-identical-functions: 0 */
+import { decodeAddress } from "algosdk";
 import { Message, sha256 } from "js-sha256";
 import { sha512_256 } from "js-sha512";
 import { decode, encode } from "uint64be";
@@ -7,7 +8,9 @@ import { TealError } from "../errors/errors";
 import { ERRORS } from "../errors/errors-list";
 import { MAX_CONCAT_SIZE, MAX_UINT64 } from "../lib/constants";
 import { compareArray } from "../lib/helpers";
+import { convertToBuffer } from "../lib/parse-data";
 import type { TEALStack } from "../types";
+import { EncodingType } from "../types";
 import { Interpreter } from "./interpreter";
 import { Op } from "./opcode";
 
@@ -595,35 +598,40 @@ export class Int extends Op {
   }
 
   execute (stack: TEALStack): void {
-    this.checkOverflow(this.uint64);
     stack.push(this.uint64);
   }
 }
 
 // push bytes to stack
 export class Byte extends Op {
-  readonly byte: Uint8Array;
+  readonly str: string;
+  readonly encoding?: EncodingType;
 
-  constructor (byte: Uint8Array) {
+  constructor (str: string, encoding?: EncodingType) {
     super();
-    this.byte = byte;
+    this.str = str;
+    if (encoding !== undefined) {
+      this.encoding = encoding;
+    }
   }
 
   execute (stack: TEALStack): void {
-    stack.push(this.byte);
+    const buffer = convertToBuffer(this.str, this.encoding);
+    stack.push(new Uint8Array(buffer));
   }
 }
 
-// push algorand address as bytes to stack
+// decodes algorand address to bytes and pushes to stack
 export class Addr extends Op {
-  readonly addr: Uint8Array;
+  readonly addr: string;
 
-  constructor (addr: Uint8Array) {
+  constructor (addr: string) {
     super();
     this.addr = addr; // parser should verify the addr first
   }
 
   execute (stack: TEALStack): void {
-    stack.push(this.addr);
+    const addr = decodeAddress(this.addr);
+    stack.push(addr.publicKey);
   }
 }
--- a/packages/algorand-js/src/interpreter/opcode-list.ts
+++ b/packages/algorand-js/src/interpreter/opcode-list.ts
@@ -587,7 +587,6 @@ export class Substring3 extends Op {
 }
 
 /** Pseudo-Ops **/
-
 // push integer to stack
 export class Int extends Op {
   readonly uint64: bigint;
--- a/packages/algorand-js/src/interpreter/opcode-list.ts
+++ b/packages/algorand-js/src/interpreter/opcode-list.ts
@@ -450,8 +450,8 @@ export class Btoi extends Op {
       throw new TealError(ERRORS.TEAL.LONG_INPUT_ERROR);
     }
     const buf = Buffer.from(bytes);
-    const uintValue = decode(buf) as bigint;
-    stack.push(uintValue);
+    const uintValue = decode(buf);
+    stack.push(BigInt(uintValue));
   }
 }
 
@@ -471,6 +471,25 @@ export class Addw extends Op {
       stack.push(BIGINT0);
       stack.push(valueC);
     }
+  }
+}
+
+// A times B out to 128-bit long result as low (top) and high uint64 values on the stack
+export class Mulw extends Op {
+  execute (stack: TEALStack): void {
+    this.assertStackLen(stack, 2);
+    const valueA = this.assertBigInt(stack.pop());
+    const valueB = this.assertBigInt(stack.pop());
+    const result = valueA * valueB;
+
+    const low = result & MAX_UINT64;
+    this.checkOverflow(low);
+
+    const high = result >> BigInt('64');
+    this.checkOverflow(high);
+
+    stack.push(high);
+    stack.push(low);
   }
 }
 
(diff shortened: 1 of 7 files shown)


packages/algorand-js/src/interpreter/opcode-list.ts
```diff
@@ -545,13 +578,52 @@ export class Substring3 extends Op {
...
+  }
+
+  execute (stack: TEALStack): void {
+    this.checkOverflow(this.uint64);
+    stack.push(this.uint64);
+  }
+}
+
+// push bytes to stack
+export class Byte extends Op {
```
amityadav0, line 604: I think this has some different function, can you please confirm ?
```
txn Lease
byte base64 TMPL_LEASE
==
&&
```
https://developer.algorand.org/docs/reference/teal/templates/dynamic_fee/

packages/algorand-js/src/interpreter/opcode-list.ts
```diff
@@ -530,8 +549,22 @@ export class Concat extends Op {
...
 // the program fails
 export class Substring extends Op {
+  readonly start: bigint;
+  readonly end: bigint;
+
+  constructor (start: bigint, end: bigint) {
+    super();
+    this.start = start;
+    this.end = end;
+  }
```
amityadav0, line 559: what is start and end in this case ? , In `substring3` we get the input from user

packages/algorand-js/src/interpreter/opcode-list.ts
```diff
@@ -545,13 +578,52 @@ export class Substring3 extends Op {
...
+}
+
+// push algorand address as bytes to stack
+export class Addr extends Op {
+  readonly addr: Uint8Array;
+
+  constructor (addr: Uint8Array) {
+    super();
+    this.addr = addr; // parser should verify the addr first
```
amityadav0, line 623:
```
addr parses an Algorand account address base32 and converts it to a regular bytes constant.
```

packages/algorand-js/test/src/interpreter/opcode-list.ts
```diff
@@ -1055,6 +1056,64 @@ describe("Teal Opcodes", function () {
...
+      const op = new Mulw();
+      op.execute(stack);
+
+      const low = stack.pop();
+      const high = stack.pop();
+      assert.equal(low, BigInt('30'));
+      assert.equal(high, BigInt('0'));
+    });
+
+    it("low and high should be 0 on a*b if a or b is 0", () => {
```
amityadav0, line 1086: can you please add tests for big numbers too ?

ratik21: adding.

packages/algorand-js/src/interpreter/opcode-list.ts
ratik21: Yes. But these should be handled by the parser. There are a lot of variations for `byte`.
Check: [https://developer.algorand.org/docs/reference/teal/specification/#constants-and-pseudo-ops]()

This pseudo-op is taking parsed byte (as `Uint8array`) and pushing it to stack.

packages/algorand-js/src/interpreter/opcode-list.ts
ratik21: Check opcode specification: [https://developer.algorand.org/docs/reference/teal/opcodes/?query=mulw#substring]()
`substring3` takes start and end from stack. But with `substring` you provide additional two params (start, end)
Eg
```
...
byte "Algorand"
substring 0 4       // pushes "Algo" to stack
...
```

packages/algorand-js/src/interpreter/opcode-list.ts
ratik21: We are already building a parser. In the opcodes we are taking parsed input in the form of `stackElem` (`bigint`/`Uint8array`). Also mentioned this in the comment.

So: Parser will read the string and decode it to algorand byte address and push `Addr(parsed_addr)` to `logic[..]`

amityadav0: Yes, addr is not an opcode, I think if parser handles this then we don't need to include pseudo codes because instead of pushing addr(..). Parser will push value to the stack.

But to calculate that value we are supposed to use this pseudo op.
Instead of giving us converted value, it only initializes value. 

I am thinking if we are writing pseudo opcodes, they should take value and return value after doing some operation , according to their respective work instead of initialization because if initialization is only thing they do, we shouldn't push addr(..), we can directly push value to stack.

ratik21: Sorry, i don't follow. Parser does not have access to stack. During parsing the execution has not started. So pushing to stack does not make sense

amityadav0: ideally, `addr(string_value)` this should push converted regular bytes to stack. right ?

amityadav0: what you are proposing here is, instead of `string_value` , parser will pass bytes value and this function will only push value and not convert it ? am i following you ?

ratik21: TEAL:
```
...
addr JAKNKAMEW63I3H3AXWINXEWW6CCLKX4MQGNTTNBSSUHNCHSLQURMMO47XM
...
```
Parser will read this string, and convert it to bytes (after verifying) and create opcode - Addr(parsed_addr_as_bytes)
During executing, `Addr(..)` will push this value to stack.

amityadav0: Correction: Parser will read this string, create opcode - Addr(parsed_addr_as_given)
During executing, Addr(..) will push converted value to stack. 
Parser will parse, but opcodes must do their respective operation
Addr(..): convert it to bytes (after verifying)  and will push this value to stack.

packages/algorand-js/src/interpreter/opcode-list.ts
amityadav0: yes

packages/algorand-js/src/interpreter/opcode-list.ts
robert-zaremba: yes, if we chose the processing in parser, then we can have a single opcode for `Addr, Int, Bytes`. That opcode will need to take as an argument `elem: StackElem` (bytes or bigint).

packages/algorand-js/src/interpreter/opcode-list.ts
ratik21: Update: Updated pseudo ops to do conversion.

packages/algorand-js/src/interpreter/opcode-list.ts
ratik21: As per our discussion i have updated the pseudo ops to do the conversion. Parser will do the verification.

packages/algorand-js/src/interpreter/opcode-list.ts
```diff
@@ -545,13 +581,57 @@ export class Substring3 extends Op {
...
+export class Addr extends Op {
+  readonly addr: string;
+
+  constructor (addr: string) {
+    super();
+    this.addr = addr; // parser should verify the addr first
+  }
+
+  execute (stack: TEALStack): void {
+    const addr = decodeAddress(this.addr);
```
amityadav0, line 634: does `decodeAddress` verifies address too ?

ratik21: No. I think we discussed parser will verify the address before execution (for better speed). Did you mean something else?

amityadav0: No, i thought we were doing all operations related to pseudo-opcodes in pseudo-opcodes

amityadav0: let's leave, will cover in parser

packages/algorand-js/src/interpreter/opcode-list.ts
```diff
@@ -545,13 +581,57 @@ export class Substring3 extends Op {
...
-    }
-    if (start > byteString.length || end > byteString.length) {
-      throw new TealError(ERRORS.TEAL.SUBSTRING_RANGE_BEYOND);
+    const subString = this.subString(start, end, byteString);
+    stack.push(subString);
+  }
+}
+
+/** Pseudo-Ops **/
+
```
amityadav0, line 590:
```suggestion
```

packages/algorand-js/test/src/interpreter/opcode-list.ts
```diff
@@ -1118,49 +1190,186 @@ describe("Teal Opcodes", function () {
...
       stack.push(BigInt('0'));
-      stack.push(BigInt('1234'));
+      stack.push(toBytes("Algorand"));
 
       const op = new Substring3();
-      expectTealError(
-        () => op.execute(stack),
+      op.execute(stack);
+
+      const top = stack.pop() as Uint8Array;
```
amityadav0, line 1278: are we supposed to use `as Uint8Array` in tests ?

packages/algorand-js/test/src/interpreter/opcode-list.ts
```diff
@@ -1118,49 +1190,186 @@ describe("Teal Opcodes", function () {
...
     it("should return correct substring", function () {
-      stack.push(BigInt('4'));
-      stack.push(BigInt('0'));
       stack.push(toBytes("Algorand"));
-
-      const op = new Substring3();
+      const op = new Substring(start, end);
       op.execute(stack);
 
       const top = stack.pop() as Uint8Array;
```
amityadav0, line 1203: same

packages/algorand-js/test/src/interpreter/opcode-list.ts
ratik21: Yes. The below functions takes `Uint8Array` as an argument. Hence the `as Uint8Array`. Do you suggest something else?

amityadav0: we can convert "Algo" to bytes

amityadav0 (approved):

PR #151, opened 2020-12-02
add rekeying checks in all examples PyTEAL code

amityadav0:
- Add rekeying, close remainder and fee checks as per TEAL guidelines
- fix indentation for some files
- change package directory for `dynamicfee` example


examples/nft/assets/nft_approval.py
```diff
@@ -1,111 +1,111 @@
...
+    [Txn.on_completion() == OnComplete.OptIn, Return(var_1)],
+    [Txn.application_args[0] == Bytes("create"), create_nft],
+    [Txn.application_args[0] == Bytes("transfer"), transfer_nft]
+  )
+
+  return program
 
 if __name__ == "__main__":
-    print(compileTeal(approval_program(), Mode.Application))
+  print(compileTeal(approval_program(), Mode.Application))
```
ratik21, line 111: Please change the tab size to 4 in all python files.

amityadav0: yes, adding this in .editorconfig also

examples/htlc-pyteal/scripts/transfer/htlc-example.js
```diff
@@ -17,7 +17,7 @@ async function run(runtimeEnv, deployer) {
     const secret = "hero wisdom green split loop element vote belt";
     const wrongSecret = "hero wisdom red split loop element vote belt";
 
-    let lsig = await deployer.loadLogic("htlc.py", [ wrongSecret ]);
+    let lsig = await deployer.loadLogic("htlc.py", [ new Uint8Array(Buffer.from(wrongSecret)) ]);
```
ratik21, line 20:
```suggestion
    let lsig = await deployer.loadLogic("htlc.py", [ toBytes(wrongSecret) ]);
```
let's use `toBytes(..)` everywhere

examples/htlc-pyteal/scripts/transfer/htlc-example.js
```diff
@@ -27,12 +27,12 @@ async function run(runtimeEnv, deployer) {
         toAccountAddr: john.addr,
         amountMicroAlgos: 200,
         lsig: lsig,
-        payFlags: {}
+        payFlags: {totalFee: 500}
```
ratik21, line 30: What happens if we don't specify the `totalFee`? Will the transaction fail?

examples/nft/assets/nft_approval.py
```diff
@@ -9,12 +9,12 @@ def approval_program():
     hash is an hash of the external resources data.
 
     Commands:
-      create    Creates a new NFT. Expects 2 additional arguments:
+    create    Creates a new NFT. Expects 2 additional arguments:
```
ratik21, line 12: can you please check the indentation here? It seems wrong

examples/htlc-pyteal/scripts/transfer/htlc-example.js
amityadav0: yes, transaction may fail and we may burn more algos than needed.
Teal Guidelines suggests to add a fee check in contracts to avoid burning extra algos.

examples/nft/assets/nft_approval.py
amityadav0: we may add a tab here, i think this is about preference only

examples/htlc-pyteal/scripts/transfer/htlc-example.js
amityadav0: yes

examples/htlc-pyteal/scripts/transfer/htlc-example.js
ratik21: We have added a fee check in the smart contract (<= 10,000). 
I was asking what happens if i pass `payFlags: {}`. (i.e do we need to pass `totalFee: 500`) 
What's the default value of totalFee? I think it's `1000`. Can you confirm?

amityadav0: yes, default value is set to 1000 by algob. But it may be higher if user wants. As per TEAL Guidelines we need to check this in contracts too. Is that what you are asking ?

ratik21:
`Dynamic-fee` example is showing parsing error.
image
Could you please validate?

examples/htlc-pyteal/scripts/transfer/htlc-example.js
ratik21: Yes. That's what i wanted to confirm. `1000` is the algorand's min default txn fee. Please leave any one txn like :
```
...
payFlags: {} // default totalFee of 1000 microalgos is used
...
```

amityadav0: yes, looks good

ratik21: Also, should we add one txn with totalFee > 10,000 and failing the transaction. (indicating the user is burning too much algo) What do think?

amityadav0: It seems there is a problem, i am getting rejected by logic when leaving payFlags empty. That means somehow fee is being set to greater than 10000.
could you please validate ?

amityadav0: sample
Look at this transaction, fee is very high, because it was left blank

ratik21: Ok let me check as well

amityadav0: I don't think minimum default fee is 1000, we are getting fee value from `const s = await getSuggestedParams(algocl);`

ratik21: With me also it's getting set to `267000`.

packages/algob/sample-project/assets/fee-check.teal
```diff
@@ -1,4 +1,13 @@
+#pragma version 2
 // Check if the provided fee is at least 10,000 microalgos
```
ratik21, line 2:
```suggestion
// Check if the provided fee is at most 10,000 microalgos
```
Right?

packages/algob/sample-project/assets/fee-check.teal
```diff
@@ -1,4 +1,13 @@
...
 txn Fee
 int 10000
 >=
```
ratik21, line 5: Shouldn't this be <= ?

packages/algob/sample-project/assets/fee-check.teal
amityadav0: I think this is good, this is just a sample example in which we require this specific condition

packages/algob/sample-project/assets/fee-check.teal
amityadav0: same as above

packages/algob/sample-project/assets/fee-check.teal
ratik21: Or it should be atleast 1000

amityadav0: But why ? this depicts an example on how TEAL works and not how fees works ? right ?

ratik21: Ok. But i think this example directly contradicts teal guidelines.

ratik21: let's leave it.

ratik21 (approved):

PR #152, opened 2020-12-03
integrate linter for examples

amityadav0:
- Fix linter errors


package.json
```diff
@@ -21,7 +21,8 @@
   "license": "Apache-2.0",
   "private": true,
   "workspaces": [
-    "packages/*"
+    "packages/*",
+    "examples/*"
```
robert-zaremba, line 25: I was thinking to do it without adding `examples` to workspace. But if this works well then it's OK

robert-zaremba (approved):

PR #153, opened 2020-12-03
Add crypto opcodes + other opcode fixes

ratik21:
- Added opcodes: `Keccak256`, `Ed25519verify`. 
- Fixed logic for all opcodes, remove ambigous naming.
- Fixed and added more tests


packages/algorand-js/src/interpreter/opcode-list.ts
```diff
@@ -20,59 +21,59 @@ const BIGINT1 = BigInt("1");
...
     stack.push(result);
   }
 }
 
-// pops two unit64 from stack(a, b) and pushes their diff(a - b) to stack
+// pops two unit64 from stack(last, prev) and pushes their diff(last - prev) to stack
 // panics on underflow (result < 0)
 export class Sub extends Op {
   execute (stack: TEALStack): void {
     this.assertStackLen(stack, 2);
```
amityadav0, line 46: can we change `assertStackLen` to something else ? because this is confusing. this seems as if it is checking if stack length is exactly equal to 2. But that's not the case here.

packages/algorand-js/src/interpreter/opcode-list.ts
```diff
@@ -20,59 +21,59 @@ const BIGINT1 = BigInt("1");
...
-    const a = this.assertBigInt(stack.pop());
-    const b = this.assertBigInt(stack.pop());
-    const result = a - b;
+    const last = this.assertBigInt(stack.pop());
+    const prev = this.assertBigInt(stack.pop());
+    const result = prev - last;
```
amityadav0, line 49: :+1:

packages/algorand-js/src/interpreter/opcode-list.ts
ratik21: yes, renaming to: `assertMinStackLen`.

amityadav0 (approved):
Looks Good

PR #155, opened 2020-12-04
Provide an example how to execute a serialized transaction

ratik21:


packages/algob/src/lib/tx.ts
```diff
@@ -231,3 +231,24 @@ export async function executeTransaction (
...
+ * @param fileName raw signed txn .tx file
+ */
+export async function executeSignedTxnFromFile (
+  deployer: AlgobDeployer,
+  fileName: string): Promise<tx.ConfirmedTxInfo> {
+  const signedTxn = loadSignedTxnFromFile(fileName);
+  if (signedTxn === undefined) {
+    throw new Error(`File ${fileName} does not exist`);
+  }
+  console.log("Decoded txn from %s: %O", fileName, decodeSignedTransaction(signedTxn));
```
robert-zaremba, line 250: do we need this log? Maybe we should use `console.debug` here?

robert-zaremba (approved):
left one comment.

packages/algob/src/lib/tx.ts
ratik21: Yes, `console.debug` would be better. Just wanted to show the use of `decodeSignedTransaction(..)` here.

PR #156, opened 2020-12-07
Interpreter - Add transaction context

ratik21:
- Added context for transaction using `mkTransaction`
- Added types for `Transaction` (`fields`, `type`, `oncomplete`), `Assets` (`holdings`, `params`) 

NOTE: Will be adding context for state related data with stateful opcodes.

### DIFF
--- a/packages/algob/src/lib/tx.ts
+++ b/packages/algob/src/lib/tx.ts
@@ -1,4 +1,4 @@
-import tx, { decodeSignedTransaction, Transaction } from "algosdk";
+import tx, { decodeSignedTransaction, SuggestedParams, Transaction } from "algosdk";
 import { TextEncoder } from "util";
 
 import { BuilderError } from "../internal/core/errors";
@@ -108,8 +108,7 @@ export function encodeNote (note: string | undefined, noteb64: string| undefined
  * @param deployer AlgobDeployer
  * @param txnParam execParam
  */
-async function mkTransaction (deployer: AlgobDeployer, txnParam: execParams): Promise<Transaction> {
-  const params = await mkSuggestedParams(deployer.algodClient, txnParam.payFlags);
+export function mkTransaction (txnParam: execParams, suggestedParams: SuggestedParams): Transaction {
   const note = encodeNote(txnParam.payFlags.note, txnParam.payFlags.noteb64);
   const transactionType = txnParam.type;
   switch (txnParam.type) {
@@ -122,7 +121,7 @@ async function mkTransaction (deployer: AlgobDeployer, txnParam: execParams): Pr
         txnParam.amount,
         note,
         txnParam.assetID,
-        params);
+        suggestedParams);
     }
     case TransactionType.TransferAlgo: {
       return tx.makePaymentTxnWithSuggestedParams(
@@ -131,20 +130,20 @@ async function mkTransaction (deployer: AlgobDeployer, txnParam: execParams): Pr
         txnParam.amountMicroAlgos,
         txnParam.payFlags.closeRemainderTo,
         note,
-        params);
+        suggestedParams);
     }
     case TransactionType.ClearSSC: {
       return tx.makeApplicationClearStateTxn(
-        txnParam.fromAccount.addr, params, txnParam.appId, txnParam.appArgs);
+        txnParam.fromAccount.addr, suggestedParams, txnParam.appId, txnParam.appArgs);
     }
     case TransactionType.DeleteSSC: {
       return tx.makeApplicationDeleteTxn(
-        txnParam.fromAccount.addr, params, txnParam.appId, txnParam.appArgs);
+        txnParam.fromAccount.addr, suggestedParams, txnParam.appId, txnParam.appArgs);
     }
     case TransactionType.CallNoOpSSC: {
       return tx.makeApplicationNoOpTxn(
         txnParam.fromAccount.addr,
-        params,
+        suggestedParams,
         txnParam.appId,
         txnParam.appArgs,
         txnParam.accounts,
@@ -156,7 +155,7 @@ async function mkTransaction (deployer: AlgobDeployer, txnParam: execParams): Pr
     }
     case TransactionType.CloseSSC: {
       return tx.makeApplicationCloseOutTxn(
-        txnParam.fromAccount.addr, params, txnParam.appId, txnParam.appArgs);
+        txnParam.fromAccount.addr, suggestedParams, txnParam.appId, txnParam.appArgs);
     }
     default: {
       throw new BuilderError(ERRORS.GENERAL.TRANSACTION_TYPE_ERROR, { error: transactionType });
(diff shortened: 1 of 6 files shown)


packages/algorand-js/src/types.ts
```diff
@@ -11,6 +11,109 @@ export type AppArgs = Array<string | number>;
 export type StackElem = bigint | Uint8Array;
 export type TEALStack = IStack<bigint | Uint8Array>;
 
+// https://developer.algorand.org/docs/reference/teal/opcodes/#txn
+export enum TxnField {
```
amityadav0, line 15: I think we already have transaction field in algosdk types. Let's use it.

packages/algorand-js/src/types.ts
```diff
@@ -11,6 +11,109 @@ export type AppArgs = Array<string | number>;
...
+  LatestTimestamp, // Last confirmed block UNIX timestamp. Fails if negative.
+  CurrentApplicationID // ID of current application executing. Fails if no such application is executing.
+}
+
+export enum AssetHolding {
+  AssetBalance, // Amount of the asset unit held by this account
+  AssetFrozen // Is the asset frozen or not
+}
+
+export enum AssetParam {
```
amityadav0, line 103: let's add a comment here `this is for global storage`
and above type will be mapped to specific account.

robert-zaremba: Yes, we should have docstrings wherever possible

packages/algorand-js/src/types.ts
ratik21: - This is an `enum` not an interface (are enums present in types/algosdk)?
- I couldn't find an interface also. Can you be more specific?

packages/algorand-js/src/types.ts
ratik21: added.

packages/algorand-js/src/interpreter/interpreter.ts
```diff
@@ -12,24 +15,53 @@ export class Interpreter {
   bytecblock: Uint8Array[];
   intcblock: BigInt[];
   scratch: StackElem[];
+  txnContext: Transaction | Transaction[];
```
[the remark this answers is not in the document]
ratik21: Ok. BDW, just to confirm: for `n`  transactions as an `atomic` group, there is only single execution of smart contract.

packages/algob/src/lib/tx.ts
```diff
@@ -108,8 +108,7 @@ export function encodeNote (note: string | undefined, noteb64: string| undefined
  * @param deployer AlgobDeployer
  * @param txnParam execParam
  */
-async function mkTransaction (deployer: AlgobDeployer, txnParam: execParams): Promise<Transaction> {
-  const params = await mkSuggestedParams(deployer.algodClient, txnParam.payFlags);
+export function mkTransaction (txnParam: execParams, suggestedParams: SuggestedParams): Transaction {
```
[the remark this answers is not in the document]
ratik21: I wanted to remove the deployer from `mkTransaction`. So that i can reuse in interpreter (without deployer client)

packages/algorand-js/src/interpreter/interpreter.ts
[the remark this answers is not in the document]
robert-zaremba: Yes

packages/algob/src/lib/tx.ts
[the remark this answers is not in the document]
robert-zaremba: ok

packages/algorand-js/src/types.ts
amityadav0: how are we planning to use this enum in context ? 
`mkTransaction` will return `transaction` interface.
User cannot pass enum, mkTransaction will not pass enum
how `txn` field will return required value ?

you can search `Transaction`. there is a interface with this name in index.d.ts in types@algosdk.

packages/algorand-js/src/types.ts
```diff
@@ -11,6 +11,109 @@ export type AppArgs = Array<string | number>;
...
+export enum TxnOnComplete {
+  NoOp,
+  OptIn,
+  CloseOut,
+  ClearState,
+  UpdateApplication,
+  DeleteApplication
+}
+
+export enum GlobalField {
```
amityadav0, line 86: I don't think we are using this enum in code(same for others). Is this for future use or i missed it ?
can you please elaborate how are we planning to use this ?

packages/algorand-js/src/types.ts
ratik21: We are already using `Transaction` interface in the `createContext` function. 
`TxnField` enum is used to represent the transaction fields. (you can check txn opcode spec)
Use Case:
 - TEAL:
```
...
txn Fee
...
```
- PARSER:
```
push `new Txn(TxnField.Fee, this.interpreter)` to logic[]
```
- INTERPRETER 
```
reads this.interpreter.tx and pushes Fee to stack (as the txfield is Fee) 
```  
Let me know if that's clear

packages/algorand-js/src/types.ts
ratik21: yes, sorry. This will be used for `global` opcode.

packages/algorand-js/src/types.ts
amityadav0: what i am trying to ask here is:
- we get transaction interface from mkTransaction
- interpreter will push Fee to stack
- how are we connecting enum with interface because currently they are disconnected, (you will get blank value for each field)
does this makes sense ?

ratik21: Yes. We will have to create a function which will return specified value from transaction. This should be done because the TEAL opcode specification for transaction fields are different (in naming) than the `Transaction` interface of sdk. This would be done while adding the txn related opcodes. 
Something like:
```
function getTxnInfo(field: TxnField, transaction: Transaction) {
  switch (TxField) {
   case Fee:
        return transaction.fee
   ... 
   ...
 }
}
```
Should this be handled some other way?

amityadav0: I was thinking of switching with strings here, this will result in less repetitive code (because we still have to switch with strings in parser)
but we have `TypeEnum`, i think handling with enum is better.

ratik21: yes, using enums is better than using strings.

amityadav0 (commented):
are we storing values of ASA, balances for a specific account ?

ratik21:
> are we storing values of ASA, balances for a specific account ?

This Pull request is for transaction context. I added a note in the description also.
Anyway, i have added ctx for state related data as well. Can you please check?

packages/algorand-js/src/interpreter/interpreter.ts
```diff
@@ -1,35 +1,85 @@
...
+        txns.push(txn);
+      }
+      assignGroupID(txns);
+      this.gtxs = txns;
+      this.tx = txns[0]; // by default current txn is the first txn
+    } else {
+      const mockParams = mockSuggestedParams(txnParams.payFlags);
+      this.tx = mkTransaction(txnParams, mockParams);
+      this.gtxs = [this.tx];
+    }
```
robert-zaremba, line 55: is the `else` part for tests? 
Hacking a normal code for tests execution is usually a pattern. Why not passing mock params through an argument to this function?

robert-zaremba (approved):
Pre-approving, but please add testing hacks to the  "production" code

packages/algorand-js/src/interpreter/interpreter.ts
ratik21: The `else` part is not for tests. It represents a single transaction (not a transaction group).
`mock Params` are not passed explicitly because some params are dynamic (eg `txn fee`). 
And for an atomic group, `mock Params` might be different for different transactions (eg - different `fee` / `firstValid` /..etc)

ratik21: If the `txnParams` is an Array then it is representing a transaction group (Multiple transactions as an atomic group). So each object in that array represents a single transaction.

Otherwise (the `else` part), we create and assign a single transaction.

robert-zaremba: I meant to include this explanation in the code next time you will do a PR, ;)

PR #157, opened 2020-12-08
Parser for TEAL interpreter PoC

amityadav0:
+ read the text
+ ignore comments / pragma
+ translate code into the list of statements / opcodes objects.


ratik21:
I think we should also calculate the total cost of the program. Each opcode has a specifc cost.
`Each TEAL op has an associated cost estimate and the program cost estimate must total less than 20000 (consensus parameter LogicSigMaxCost). Most ops have an estimated cost of 1, but a few slow crypto ops are much higher.`
[https://developer.algorand.org/docs/reference/teal/specification/#execution-environment]()

robert-zaremba:
> I think we should also calculate the total cost of the program. Each opcode has a specifc cost.
> 
> ```
> Each TEAL op has an associated cost estimate and the program cost estimate must total less than 20000 (consensus parameter LogicSigMaxCost). Most ops have an estimated cost of 1, but a few slow crypto ops are much higher.
> ```
> 
> https://developer.algorand.org/docs/reference/teal/specification/#execution-environment

Yes, we will do it in a separate task.

robert-zaremba:
Parser is only responsible for creating list of opcodes.

packages/algorand-js/src/errors/errors-list.ts
```diff
@@ -113,6 +113,18 @@ by an index that does not exist.`
     message: "Input is not uint8",
     title: "Input is outside the valid uint8 range of 0 to 255",
     description: `Input is outside the valid uint8 range of 0 to 255`
+  },
+  ASSERT_FIELD_LENGTH: {
+    number: 16,
+    message: "Input is not correct",
+    title: "Invalid Field Length",
+    description: `Invalid Field Length - Parse Error`
```
robert-zaremba, line 121:
```suggestion
    message: "Invalid Field Length - Parse Error",
    title: "Invalid Field Length",
    description: ``
```

packages/algorand-js/src/errors/errors-list.ts
```diff
@@ -113,6 +113,18 @@ by an index that does not exist.`
...
+  },
+  INVALID_ADDR: {
+    number: 17,
+    message: "Input Parse Error - Invalid Address",
+    title: "Address is invalid",
+    description: `Invalid Address`
```
robert-zaremba, line 127:
```suggestion
    message: "Input Parse Error - Invalid Address",
    title: "Invalid Address",
    description: ``
```

packages/algorand-js/src/interpreter/parser.ts
```diff
@@ -0,0 +1,180 @@
...
+    if (!(line.length === 0 || line.startsWith("//") || line.startsWith("#pragma"))) {
+      const fields = fieldsFromLine(line);
+
+      if (fields.length !== 0) {
+        opCodeList.push(opcodeFromFields(fields));
+      }
+    }
+  }
+  return opCodeList;
+}
```
robert-zaremba, line 180: Let's add more structure to the parser.
This is how I would like to have it working (pseudo-code):

Firstly let's define a helper functions. In the code below `program` is a list of lines (strings)
```
function cleanup_line(line): {
  line = remove_comment(line)
  return trim_whitespaces(line)
}

function read_pragma(program): [program, pragma || undefined] {
  for line in program:
    line = cleanup_line(line)
    if line.startsWith("#pragma"):
      return program, new Pragma(...)
    else
      program.prepend(line) // prepend back the line, because it contains an instruction
      return program, undefined
}
```

Then we have to have a map of all op codes:

```
const opCodes = {
  add = Add
}
```
And finally the parser:

```
function parser(program: string[]) {
  [program, pragma] = read_pragma(program)
  opcodes = []
  for line in program:
      line = cleanup_line(line)
      line = trim_whitespaces(line)
      if line === "": continue
    
      words = line.split(" ")
      let op = opCodes[words[0]]
      if op === undefined:
         throw UnknownOpcode(words[0], )
      opcodes.push(new op(words.slice(1)...)
  return opcodes
}
```

maybe we should also add a line counter and use it in all error messages to make it clear in which line there is an error.

robert-zaremba: It's just an idea. Let me know what do  you think. With this we can move more parsing logic to the operator constructor, and the parser implementation will be more clear.

robert-zaremba (changes requested):

packages/algorand-js/src/interpreter/parser.ts
amityadav0: Yes, adding more structure will be good.
1. I think we should keep `fieldsFromLine` function because if we use `clean_up`. There is a case where in between spaces will be more than one, Ex: `byte        base64   Kkjh/=` then line.split(" ") will give in undesired result. There maybe more corner cases.

The structure will be as above with one difference:
```
function fieldsFromLine(line): {
       trim_whitespace(line)
       ...........
       ...........
       return fields.
}

function read_pragma(program): [program, pragma || undefined] {
  for line in program:
    line = cleanup_line(line)
    if line.startsWith("#pragma"):
      return program, new Pragma(...)
    else
      program.prepend(line) // prepend back the line, because it contains an instruction
      return program, undefined
}
```
I need more clarity on usage of #pragma , currently in algorand documentation,
```
The first line may contain a special version pragma #pragma version X, which directs the assembler to generate TEAL bytecode targeting a certain version. For instance, #pragma version 2 produces bytecode targeting TEAL v2. By default, the assembler targets TEAL v1.
``` 
should we add special function for pragma or ignore it ?

2. we can use `let op = opCodes[words[0]]` this, but i have a question , we need to verify number of fields and data passed from that field. For ex: "addr some_address some_value" this will throw error, should we include a check here ?
there is one more check for address, whether the address is algorand address or not.

Similarly for `int` there are two assert functions.
If we choose to include these asserts in parser function, there will be many assert and switch or if statements in parser function.

That's why in current code they are divided in two functions, pseudo_ops and arithmetic , we can create more function according to opcodes addition but using map may result switch or if conditions.

3.Yes adding line counter is good.

robert-zaremba: Let's add a new opcode for Pragma, which will just store the TEAL version in the interpreter.
BTW: I'm not sure, but I think there can be comments before pragma

robert-zaremba: So, probably we can move argument parsing into the constructor. It could be easier.
Example.. Now we have:
```
export class Store extends Op {
  readonly index: number;
  readonly interpreter: Interpreter;
  constructor (index: number, interpreter: Interpreter) {
```

We can change the constructor to:
```
export class Store extends Op {
  constructor (interpreter: Interpreter, args: string[]) {
    super();
    assert_len(args, 1);
    this.index = parseInt(args[0]);
    this.interpreter = interpreter;
  }
```

BTW, in `let op ... ` there should be `new: `let op = new opCodes[words[0]]`

robert-zaremba: The above is just an idea, but it requires changes in all op codes.
You can try to find a better way if you like.

amityadav0: Okay, updating

amityadav0:
- Changed Structure as per comments (Moved logic to constructors, add map for opCodes, structure change)
- Checked Pragma case - we cannot put comments before Pragma (checked using teal compiler)
- Added Pragma Op which stores version
- Added Line counter for error

packages/algorand-js/src/errors/errors-list.ts
```diff
@@ -113,6 +113,18 @@ by an index that does not exist.`
     message: "Input is not uint8",
     title: "Input is outside the valid uint8 range of 0 to 255",
     description: `Input is outside the valid uint8 range of 0 to 255`
+  },
+  ASSERT_FIELD_LENGTH: {
+    number: 16,
+    message: "Invalid Field Length - Parse Error, Line : %line% ",
+    title: "Invalid Field Length",
+    description: ``
```
robert-zaremba, line 121: In the description, please add the expected and the received length

packages/algorand-js/src/errors/errors-list.ts
```diff
@@ -113,6 +113,18 @@ by an index that does not exist.`
...
+  },
+  INVALID_ADDR: {
+    number: 17,
+    message: "Input Parse Error - Invalid Address, Line: %line%",
+    title: "Invalid Address",
+    description: ``
```
robert-zaremba, line 127: In the description we should show the invalid address.

packages/algorand-js/src/interpreter/opcode-list.ts
```diff
@@ -17,8 +17,29 @@ import { Op } from "./opcode";
...
+export class Pragma extends Op {
+  readonly version;
+
+  constructor (arg: string[], line: number) {
+    super();
+    assertFieldLen(arg.length, 2, line);
+    if (arg[0] === "version") {
+      this.version = arg[1];
+    } else {
+      throw new TealError(ERRORS.TEAL.INVALID_OP_ARG, { line: line });
```
robert-zaremba, line 31: Please add a description:
```
(Expected "version", got "%got%", {got: arg[1]}) 
```

packages/algorand-js/src/interpreter/opcode-list.ts
```diff
@@ -17,8 +17,29 @@ import { Op } from "./opcode";
...
+    }
+  };
+}
+
 // pops string([]byte) from stack and pushes it's length to stack
 export class Len extends Op {
+  constructor (arg: string[], line: number) {
```
robert-zaremba, line 38: we should use `args`, not `arg` everywhere.

packages/algorand-js/src/interpreter/opcode-list.ts
```diff
@@ -17,8 +17,29 @@ import { Op } from "./opcode";
...
+    super();
+    assertFieldLen(arg.length, 0, line);
```
robert-zaremba, line 40: How about renaming: `assertFieldLen` -> `assertLen`?

packages/algorand-js/src/interpreter/parser.ts
```diff
@@ -0,0 +1,154 @@
...
+  "*": Mul
+};
+
+/**
+ * Description: Read line and split it into fields
+ * - ignore comments, keep only part that is relevant to interpreter
+ * @param line : Line read from TEAL file
+ */
+/* eslint-disable sonarjs/cognitive-complexity */
+export function fieldsFromLine (line: string): string[] {
```
robert-zaremba, line 30: In parsing context we use `word`, `lexem` or `token`. Let's don't use `fields`. We can rename to `wordsFromLine`

packages/algorand-js/src/interpreter/parser.ts
```diff
@@ -0,0 +1,154 @@
...
+    }
+  }
+
+  // add rest of the string if any
+  if (start < line.length) {
+    fields.push(line.substr(start, i - start));
+  }
+
+  return fields;
+}
```
robert-zaremba, line 109: we need to have many tests for this function. 

We can do it in a separate task next week: https://www.pivotaltracker.com/story/show/176117706

packages/algorand-js/src/interpreter/parser.ts
```diff
@@ -0,0 +1,154 @@
...
+
+/**
+ * Description: Returns Opcode object for given field
+ * @param fields : fields extracted from line
```
robert-zaremba, line 113: add comment for other arguments as well

Same for other functions.

packages/algorand-js/src/lib/helpers.ts
```diff
@@ -1,4 +1,34 @@
...
     a.every((v, i) => v === b[i]);
 }
+
+/**
+ * Description: assert if string contains digits only
+ * "123" // ok.  "12+2" // error.
+ * @param val : string
+ */
+export function assertOnlyDigits (val: string): void {
+  if (!(/^\d+$/.test(val))) {
```
robert-zaremba, line 20: let's move the regexp to a const value in the module scope instead of recreating it every time in the function call:
```
const reDigit = /^\d+$/;
```

packages/algorand-js/test/src/interpreter/parser.ts
```diff
@@ -0,0 +1,291 @@
...
+      const file = "test-file-5.teal";
+      const fPath = path.join(process.cwd(), file);
+      const res = await parser(fPath);
+      const expected = [new Pragma(["version", "2"], 1), new Int(["5"], 1),
+        new Int(["3"], 1), new Mul([], 1)];
+
+      assert.deepEqual(res, expected);
+    });
+  });
+});
```
robert-zaremba, line 291: :clap:

packages/algorand-js/src/errors/errors-list.ts
amityadav0: ok

packages/algorand-js/src/errors/errors-list.ts
amityadav0: ok

packages/algorand-js/src/interpreter/opcode-list.ts
amityadav0: ok

packages/algorand-js/src/interpreter/opcode-list.ts
amityadav0: ok

packages/algorand-js/src/interpreter/opcode-list.ts
amityadav0: yes, we renaming fields also, so `assertLen` is better

packages/algorand-js/src/interpreter/parser.ts
amityadav0: yes

packages/algorand-js/src/interpreter/parser.ts
amityadav0: ok

packages/algorand-js/src/lib/helpers.ts
amityadav0: ok

packages/algorand-js/src/errors/errors-list.ts
```diff
@@ -113,6 +113,24 @@ by an index that does not exist.`
...
+  },
+  ASSERT_FIELD_LENGTH: {
+    number: 16,
+    message: "Invalid Field Length Expected: %exp% Got: %got% - Parse Error, Line : %line% ",
+    title: "Invalid Field Length",
+    description: ``
+  },
+  INVALID_ADDR: {
+    number: 17,
+    message: "Input Parse Error - Invalid Address %addr%, Line: %line%",
```
robert-zaremba, line 125:
```suggestion
    message: "Invalid Address '%addr%', Line: %line%",
```

packages/algorand-js/src/errors/errors-list.ts
```diff
@@ -113,6 +113,24 @@ by an index that does not exist.`
...
+    title: "Invalid Address",
+    description: ``
+  },
+  PRAGMA_VERSION_ERROR: {
+    number: 18,
+    message: "Pragma version Error - Expected: version, got: %got%, Line: %line%",
+    title: "Invalid Address",
```
robert-zaremba, line 132:
```suggestion
    title: "Parse Error",
```

packages/algorand-js/src/errors/errors-list.ts
```diff
@@ -113,6 +113,24 @@ by an index that does not exist.`
...
+  ASSERT_FIELD_LENGTH: {
+    number: 16,
+    message: "Invalid Field Length Expected: %exp% Got: %got% - Parse Error, Line : %line% ",
+    title: "Invalid Field Length",
+    description: ``
+  },
+  INVALID_ADDR: {
+    number: 17,
+    message: "Input Parse Error - Invalid Address %addr%, Line: %line%",
+    title: "Invalid Address",
```
robert-zaremba, line 126:
```suggestion
    title: "Parse Error",
```

packages/algorand-js/src/errors/errors-list.ts
```diff
@@ -113,6 +113,24 @@ by an index that does not exist.`
     message: "Input is not uint8",
     title: "Input is outside the valid uint8 range of 0 to 255",
     description: `Input is outside the valid uint8 range of 0 to 255`
+  },
+  ASSERT_FIELD_LENGTH: {
+    number: 16,
+    message: "Invalid Field Length Expected: %exp% Got: %got% - Parse Error, Line : %line% ",
+    title: "Invalid Field Length",
```
robert-zaremba, line 120:
```suggestion
    message: "Invalid Field Length Expected: '%exp%' Got: '%got%', Line : %line% ",
    title: "Parse Error",
```

packages/algorand-js/src/interpreter/parser.ts
```diff
@@ -0,0 +1,155 @@
...
+ * @param counter: line number in TEAL file
+ */
+export function opcodeFromFields (fields: string[], counter: number): Operator {
```
robert-zaremba, line 116: let's rename here as well:
`opcodeFromFields` -> `opcodeFromSentence`
`fields` -> `words`

robert-zaremba (approved):
Approving

PR #158, opened 2020-12-10
Interpreter: Txn opcodes

ratik21:
- Added opcodes: `txn`, `gtxn`, `txna`, `gtxna`
- Added tests for each `txn field`

Transaction Reference: [https://developer.algorand.org/docs/reference/transactions/]()
Enum tables: [https://developer.algorand.org/docs/reference/teal/opcodes/#txn]()


packages/algorand-js/src/interpreter/txn.ts
```diff
@@ -0,0 +1,307 @@
...
+}
+
+/**
+ * Description: returns specific transaction field value from tx object
+ * @param txField: transaction field
+ * @param interpreter: interpreter
+ */
+export function txnSpecbyField (txField: TxnField, tx: Transaction, gtxs: Transaction[]): StackElem {
+  // common fields (for all transaction types)
+  switch (txField) {
```
amityadav0, line 77: I have a suggestion, as we are moving logic to constructors, therefore a list of args will be passed to `txn` constructor.

Instead of having `enums` and `swtich` conditions. I suggest to have a map. This way we can make our code easily maintainable and remove repetitive code.
Ex:
```
// Suppose this is transaction interface
interface tran {
   a: number,
   b: string
}
// tx object
const tx: tran = {
  a: 3,
  b: "string"
}
// we can have a single map like this for txfield and similary for other fields
let mp: {[key: string]: string } = {
  "Fee": "a",
   "B": "b"
}

// To access these values, you can simply pass arg[0]
const val = mp[arg[0]];
res = tx[val as keyof tran];
// result is your value.
```
This way we remove too many swtich cases

amityadav0: I have tested this code, this is working

packages/algorand-js/src/interpreter/txn.ts
```diff
@@ -0,0 +1,307 @@
...
+    return new Uint8Array(0);
+  }
+  if (addr.includes(txField)) {
+    return ZERO_ADDRESS;
+  }
+
+  return BigInt('0');
+}
+
+function assertBytesDefined (a?: Uint8Array): Uint8Array {
```
amityadav0, line 48: is it possible to have undefined field after getting transaction from `mkTransaction` ?

packages/algorand-js/src/interpreter/txn.ts
ratik21: This has some loopholes. Continuing discussion about the structure on discord

packages/algorand-js/src/interpreter/txn.ts
ratik21: yes

amityadav0: how is it handled in TEAL ?

packages/algorand-js/src/interpreter/txn.ts
amityadav0: yes there maybe some cases, but it can be done

packages/algorand-js/src/interpreter/txn.ts
```diff
@@ -0,0 +1,307 @@
...
+
+import { TealError } from "../errors/errors";
+import { ERRORS } from "../errors/errors-list";
+import { ZERO_ADDRESS } from "../lib/constants";
+import { toBytes } from "../lib/parse-data";
+import { StackElem, TxnField, TxnType } from "../types";
+import { Op } from "./opcode";
+
+// regular bytes
+const bytes = [
```
robert-zaremba, line 11: Let's rename to `byteTxFields`

packages/algorand-js/src/interpreter/txn.ts
```diff
@@ -0,0 +1,307 @@
...
+  TxnField.ApprovalProgram,
+  TxnField.ClearStateProgram,
+  TxnField.ConfigAssetUnitName,
+  TxnField.ConfigAssetName,
+  TxnField.ConfigAssetURL,
+  TxnField.ConfigAssetMetadataHash
+];
+
+// 32 byte addresses
+const addr = [
```
robert-zaremba, line 21: Let's rename to `addrTxFields`

packages/algorand-js/src/interpreter/txn.ts
```diff
@@ -0,0 +1,307 @@
...
+];
+
+// returns default value of txn field
+function defaultSpec (txField: TxnField): StackElem {
+  if (bytes.includes(txField)) {
+    return new Uint8Array(0);
+  }
+  if (addr.includes(txField)) {
+    return ZERO_ADDRESS;
+  }
```
robert-zaremba, line 43: checking in a map / set is much faster then iterating over an array. 
I suggest to use `Set` type above, instead of an Array.

packages/algorand-js/src/interpreter/txn.ts
robert-zaremba, line 48: `assert` suggest that we will return an error or throw an exception. How about renaming to `bytesOrZero`?

packages/algorand-js/src/interpreter/txn.ts
```diff
@@ -0,0 +1,307 @@
...
+}
+
+function assertBytesDefined (a?: Uint8Array): Uint8Array {
+  if (a === undefined) {
+    return new Uint8Array(0);
+  }
+  return a;
+}
+
+function assertUint64Defined (a?: number): bigint {
```
robert-zaremba, line 55: as above: `uint64OrZero` (below as well)

packages/algorand-js/src/interpreter/txn.ts
```diff
@@ -0,0 +1,307 @@
...
+          return assertBytesDefined(tx.appApprovalProgram);
+        }
+        case TxnField.ClearStateProgram: {
+          return assertBytesDefined(tx.appClearProgram);
+        }
+        default: {
+          return defaultSpec(txField);
+        }
+      }
+    }
```
robert-zaremba, line 275: I would prefer to use the `get_obj_for_encoding` as noted in the Discord chat.
This way we don't need to do that big switch, and if the `Transaction` object change (which is internal to the SDK, NOT the Algorand spec) our code will still work.
So the flow would be:

```
// in interpreter (constructor, or other method where we set the transactions):
this.tx = txSDK.get_obj_for_encoding()
```
TxnField can be defined as a map instead of enum in `types.ts`:
```
const TxnFields = {
   Amount: "amt"
   AssetAmount: "aamt"
   ...
}
// now we can declare a valid type for the TxField:
type TxField = keyof TxFields

const TxFieldDefaults: {[key: TxField]: any}  = {
  Amount: zeroUint64
  ...
}
```
Execute is much simpler and we don't need `txnSpecbyField`:
```
export class Txna extends Op {
   ...
   field: string;

  constructor(field: TxField, .... ) {
    ...
    this.field = types.TxnFields[field] ;
  }

    const r = this.interpreter.tx[this.field]  || types.TxFieldDefaults[this.field];
    stack.push(r);
  }
}

Comment (robert-zaremba) [reply packages/algorand-js/src/interpreter/txn.ts:275]: Let me know what do you think. We can keep your implementation - it also works ;)

File: packages/algorand-js/src/interpreter/txn.ts [hunk 8]
```diff
@@ -0,0 +1,307 @@
...
+}
+
+/**
+ * Description: returns specific transaction field value from array
+ * of accounts or application args
+ * @param tx: current transaction
+ * @param txField: transaction field
+ * @param idx: array index
+ */
+export function txnaSpecbyField (txField: TxnField, tx: Transaction, idx: number, op: Op): Uint8Array {
```
robert-zaremba, line 290: Rename `txnaSpecbyField` -> `txAppArg`

packages/algorand-js/src/interpreter/txn.ts
ratik21: 1. Agree about using map instead of enums, `get_obj_for_encoding()`.
2.  But i still think we would need `txnSpecbyField()` function as there are some TxFields which we'll need to handle (eg `TypeEnum`, `GroupIndex`, `TxID`, `NumAppArgs`..)

packages/algorand-js/src/interpreter/txn.ts
ratik21: ok

packages/algorand-js/src/interpreter/txn.ts
ratik21: ok

packages/algorand-js/src/interpreter/txn.ts
ratik21: Removed this function.

packages/algorand-js/src/interpreter/txn.ts
ratik21: Removed this function.

packages/algorand-js/src/interpreter/opcode-list.ts
```diff
@@ -10,10 +10,12 @@ import { ERRORS } from "../errors/errors-list";
...
 import { compareArray } from "../lib/helpers";
 import { convertToBuffer, convertToString } from "../lib/parse-data";
-import type { TEALStack } from "../types";
-import { EncodingType } from "../types";
+import type { EncodingType, TEALStack } from "../types";
 import { Interpreter } from "./interpreter";
 import { Op } from "./opcode";
+import { txAppArg, txnSpecbyField } from "./txn";
+
+// import { txnaSpecbyField, txnSpecbyField } from "./txn";
```
amityadav0, line 18: Please remove these lines

packages/algorand-js/src/interpreter/txn.ts
```diff
@@ -0,0 +1,116 @@
...
+    return parseToStackElem(result, txField);
+  }
+
+  // handle other cases
+  switch (txField) {
+    case 'FirstValidTime': { // Causes program to fail; reserved for future use
+      throw new TealError(ERRORS.TEAL.LOGIC_REJECTION);
+    }
+    case 'TypeEnum': {
+      result = TxnType[tx.type as keyof typeof TxnType]; // TxnType['pay']
```
amityadav0, line 63: Please remove comment

packages/algorand-js/src/types.ts
```diff
@@ -13,58 +13,145 @@ export type AppArgs = Array<string | number>;
...
-  ConfigAssetURL, // URL
-  ConfigAssetMetadataHash, // 32 byte commitment to some unspecified asset metadata
-  ConfigAssetManager, // 32 byte address
-  ConfigAssetReserve, // 32 byte address
-  ConfigAssetFreeze, // 32 byte address
-  ConfigAssetClawback, // 32 byte address
-  FreezeAsset, // Asset ID being frozen or un-frozen
-  FreezeAssetAccount, // 32 byte address of the account whose asset slot is being frozen or un-frozen.
-  FreezeAssetFrozen // The new frozen value, 0 or 1
+export interface TxnEncodedObj {
```
amityadav0, line 16: can we move this to types-algosdk ? or is it more relevant to put it here. what do you think ?

packages/algob/src/types.ts
```diff
@@ -463,8 +463,8 @@ export interface SSCStateSchema {
 export interface SSCOptionalFlags {
   appArgs?: Uint8Array[]
   accounts?: string[]
-  foreignApps?: string
-  foreignAssets?: string
+  foreignApps?: number[]
+  foreignAssets?: number[]
```
amityadav0, line 467: i don't think `foreignAssets` type is `number[]`. 
i think it's a array of object with fields.(https://developer.algorand.org/docs/reference/teal/opcodes/#asset_params_get)

packages/algorand-js/src/types.ts
ratik21: yes, i was thinking the same. Will need to confirm with Robert. in types/algosdk i think 'unknown' is being used.

packages/algob/src/types.ts
ratik21: As per sdk it's an array of 'indexes'. 
image

amityadav0: ok, Transaction reference says `address`(https://developer.algorand.org/docs/reference/transactions/#asset-transfer-transaction). but this seems right.

packages/algorand-js/src/types.ts
robert-zaremba: Yes, very good call. Let's put it to the `types-algosd`.

packages/algorand-js/src/interpreter/txn.ts
```diff
@@ -0,0 +1,118 @@
...
+  'ConfigAssetMetadataHash',
+  'ConfigAssetManager',
+  'ConfigAssetReserve',
+  'ConfigAssetFreeze',
+  'ConfigAssetClawback'
+]);
+
+// return default value of txField if undefined,
+// otherwise return parsed data to interpreter
+export function parseToStackElem (a: unknown, field: TxField): any {
```
robert-zaremba, line 27: Please change return type to `StackElem`

packages/algorand-js/src/types.ts
```diff
@@ -13,58 +13,64 @@ export type AppArgs = Array<string | number>;
...
-  ConfigAssetClawback, // 32 byte address
-  FreezeAsset, // Asset ID being frozen or un-frozen
-  FreezeAssetAccount, // 32 byte address of the account whose asset slot is being frozen or un-frozen.
-  FreezeAssetFrozen // The new frozen value, 0 or 1
+export interface Txn extends TxnEncodedObj {
+  txID: string
 }
 
+// https://developer.algorand.org/docs/reference/teal/opcodes/#txn
+export const TxnFields: {[key: string]: any} = {
```
robert-zaremba, line 21:
```suggestion
export const TxnFields: {[key: string]: string} = {
```

robert-zaremba: all values are strings

packages/types-algosdk/index.d.ts
```diff
@@ -109,20 +109,20 @@ export interface CompileOut {
...
-  lease: number;
+  genesisHash: Buffer;
+  lease: Uint8Array;
 
-  closeRemainderTo: ParsedAddress;
-  voteKey: string;
-  selectionKey: string;
+  closeRemainderTo: Address;
+  voteKey: Buffer;
+  selectionKey: Buffer;
```
robert-zaremba, line 125: I think we should use `Uint8Array`. As far as I know SDK is using `Uint8Array` everywhere for bytes / buffer.

packages/types-algosdk/index.d.ts
```diff
@@ -109,20 +109,20 @@ export interface CompileOut {
 export interface Transaction {
   // fields copied from
   // https://github.com/algorand/js-algorand-sdk/blob/develop/src/transaction.js#L117
-  from: ParsedAddress;
-  to: ParsedAddress;
+  from: Address;
+  to: Address;
```
robert-zaremba, line 113: Why we are changing this type? Why it was wrong before?

packages/types-algosdk/index.d.ts
```diff
@@ -131,35 +131,36 @@ export interface Transaction {
...
+  assetFreeze: Address;
+  assetClawback: Address;
   assetUnitName: string;
   assetName: string;
   assetURL: string;
-  assetMetadataHash: string;
+  assetMetadataHash: Buffer;
 
-  freezeAccount: string;
+  freezeAccount: Address;
```
robert-zaremba, line 144: ditto

packages/types-algosdk/index.d.ts
```diff
@@ -560,3 +561,96 @@ export interface AccountInfo {
...
+  fv: number
+  lv: number
+  note: Buffer
+  snd: Buffer
+  type: string
+  gen: string
+  gh: Buffer
+  rekey: Buffer
+  lx: Buffer
+  grp: Buffer
```
robert-zaremba, line 578: ditto, I'm pretty sure we should use `Uint8Array` here, and `Uint8Array[]` instead of `Buffer[]`

robert-zaremba (commented):
Left few more comments, mostly related to types.

packages/algorand-js/src/interpreter/txn.ts
ratik21: yes, sorry.

packages/types-algosdk/index.d.ts
ratik21: Seems to me this is being set as `buffer`.
image
[https://github.com/algorand/js-algorand-sdk/blob/develop/src/transaction.js#L123]()

packages/types-algosdk/index.d.ts
ratik21: `ParsedAddress` is represented as (in types)
```
export interface ParsedAddress {
  publicKey: string;
}
```
where as this should be a decoded address. (With `publickey` and `checksum`, both of type Uint8Array). Which was represented by type Address:
```
 export interface Address {
  publicKey: Uint8Array;
  checksum: Uint8Array;
}
```

Here is the screenshot from txn obj: 
image

packages/types-algosdk/index.d.ts
ratik21: Encoded types were using `Buffer` everywhere instead of `Uint8Array` (eg. `note` is represented as `Uint8Array` in sdk object, but after `get_obj_for_encoding()` it is represented as a `buffer`). Attaching info.

- In SDK (you can check `note` is represented as a buffer)
image
https://github.com/algorand/js-algorand-sdk/blob/develop/src/transaction.js#L162

- Also tried logging the encoded txn object myself:
image

packages/types-algosdk/index.d.ts
robert-zaremba: You are right. Sorry.

packages/types-algosdk/index.d.ts
robert-zaremba: thanks

packages/types-algosdk/index.d.ts
robert-zaremba: ok, thanks.

robert-zaremba (approved):
ACK

PR #159, opened 2020-12-11
Infrastructure update

robert-zaremba:
Updated our default algod setup.

### DIFF
--- a/examples/algob.config-template.js
+++ b/examples/algob.config-template.js
@@ -12,15 +12,14 @@ const { mkAccounts } = require("algob");
 let accounts = mkAccounts([
   {
     name: "master-account",
-    // This account is exported from algorand an node.
-    // It must already exist or ALGO tokens have to be transferred from elsewhere.
-    // It contains a lot of ALGO tokens so it can fund other accounts.
-    // You should check your accounts and replace this with your own account.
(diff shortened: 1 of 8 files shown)


examples/algob.config-template.js
```diff
@@ -12,15 +12,14 @@ const { mkAccounts } = require("algob");
...
-    // This account is exported from algorand an node.
-    // It must already exist or ALGO tokens have to be transferred from elsewhere.
-    // It contains a lot of ALGO tokens so it can fund other accounts.
-    // You should check your accounts and replace this with your own account.
-    // Command to view KMD account addresses (and their balances):
-    // goal account list -d ~/.algorand-local/Node/
+    // This account is created using `make setup-master-account` command from our
+    // `/infrastructure` directory.
+    // It is used in all our examples to setup and fund other accounts (so it must posses ALGOs).
+    // If you don't an account with this address (can check that by running
```
amityadav0, line 18:
```suggestion
    // If you don't want an account with this address (can check that by running
```

amityadav0 (approved):

examples/algob.config-template.js
robert-zaremba: thanks

PR #160, opened 2020-12-11
update error logging

ratik21:


packages/algob/src/internal/core/errors.ts
```diff
@@ -190,16 +190,16 @@ function _applyErrorMessageTemplate (
...
 
-  if (e?.statusCode !== undefined) {
-    if (e.statusCode >= 400 && e.statusCode < 500) {
+  if (e.response?.statusCode !== undefined) {
+    if (e.response?.statusCode >= 400 && e.response?.statusCode < 500) {
       return new BuilderError(ERRORS.ALGORAND.BAD_REQUEST, {
-        status: e.statusCode,
-        message: e.body?.message ?? e.text,
+        status: e.response?.statusCode,
+        message: e.response?.body?.message ?? e.response?.text,
```
robert-zaremba, line 197: please add a check for `e.error.text` (or `e.response.error.text` - not sure which one is correct) as a third option if the 2 previous are not defined.

robert-zaremba (approved):
left one comment.

packages/algob/src/internal/core/errors.ts
ratik21: Sorry, i don't follow. I think the `response` part was added after the latest sdk update. Line 197 is checking both possibilites (`body.message` and `text`). 
BDW, this block will execute only if the error is `Bad Request`.

robert-zaremba: I was thinking about:
```
e.response?.body?.message ?? e.response?.text ?? e.response?error
```
Although I don't remember if the last part is `e.response?error` or `e.error`

ratik21: ok. It's `e.response?error`

PR #162, opened 2020-12-12
Parser support for stateless opcodes

amityadav0:


robert-zaremba:
Looks good to me. I will live the final review to @ratik21

packages/algorand-js/src/lib/helpers.ts
```diff
@@ -31,6 +31,40 @@ export function assertOnlyDigits (val: string): void {
...
+    throw new TealError(ERRORS.TEAL.ASSERT_LENGTH, { exp: expected, got: val, line: line });
+  }
+}
+
+/**
+ * Description: assert index bound
+ * @param val Comparsion result
+ * @param expected expected result
+ * @param line Line number in TEAL file
+ */
```
ratik21, line 43: Params and definitions are incorrect.

packages/algorand-js/src/lib/helpers.ts
```diff
@@ -31,6 +31,40 @@ export function assertOnlyDigits (val: string): void {
...
+export function assertBound (len: number, idx: number): void {
+  if (idx >= len) {
+    throw new TealError(ERRORS.TEAL.INDEX_OUT_OF_BOUND);
+  }
```
ratik21, line 47: We have a function in `Op` ( i think `checkIndexBound`). Can you check if that can be used here?

packages/algorand-js/src/lib/helpers.ts
amityadav0: yes

packages/algorand-js/src/lib/helpers.ts
amityadav0: yes

packages/algorand-js/src/errors/errors-list.ts
```diff
@@ -116,23 +116,47 @@ by an index that does not exist.`
...
+    number: 21,
+    message: "Invalid Decode Data - value %val% is invalid, Line: %line%",
+    title: "Decode Error",
+    description: `value %exp%`
+  },
+  UNKOWN_DECODE_TYPE: {
+    number: 22,
+    message: "Invalid Decode Type - value %val% is unknown, Line: %line%",
+    title: "Unkown Decode Type",
+    description: `value %exp% is unknown`
```
ratik21, line 159: Since we have quite a few parser errors now as well, do you think we should separate them from execution errors? this list seems too long. We can have 
+ execution error: stack related, logic rejection etc.. 
+ parsing errors: invalid string, encode/decode etc..
Just a thought. I think they would be more organized this way. Let me know what you think

packages/algorand-js/src/interpreter/opcode-list.ts
```diff
@@ -199,8 +232,19 @@ export class Intcblock extends Op {
...
 
-  constructor (interpreter: Interpreter, intcblock: Array<bigint>) {
+  /**
+   * @param args words list extracted from line
+   * @param line line number in TEAL file
+   * @param interpreter interpreter object
+   */
+  constructor (args: string[], line: number, interpreter: Interpreter) {
     super();
+    const intcblock: Array<bigint> = [];
```
ratik21, line 242:
```suggestion
    const intcblock: BigInt[];
```

packages/algorand-js/src/interpreter/parser.ts
```diff
@@ -3,24 +3,83 @@ import readline from "readline";
...
+  // Pseudo-Ops
+  addr: Addr,
+  int: Int,
+  byte: Byte
 };
 
+// list of opcodes that require one extra parameter than others: `interpreter`.
+const interpreterReqList = new Set();
+interpreterReqList.add("arg").add("bytecblock").add("bytec").add("intcblock").add("intc");
+interpreterReqList.add("store").add("load");
```
ratik21, line 81: We can do `new Set(["a", "ab", ..])`. Could you please check?

packages/algorand-js/src/lib/helpers.ts
```diff
@@ -31,6 +31,28 @@ export function assertOnlyDigits (val: string): void {
  */
 export function assertLen (val: number, expected: number, line: number): void {
   if (val !== expected) {
-    throw new TealError(ERRORS.TEAL.ASSERT_FIELD_LENGTH, { exp: expected, got: val, line: line });
+    throw new TealError(ERRORS.TEAL.ASSERT_LENGTH, { exp: expected, got: val, line: line });
+  }
+}
+
+/**
+ * Descrition: Checks if string is base64
```
ratik21, line 39:
```suggestion
 * Description: Checks if string is base64
```

packages/algorand-js/src/lib/helpers.ts
```diff
@@ -31,6 +31,28 @@ export function assertOnlyDigits (val: string): void {
...
+ * @param line : line number in TEAL file
+ */
+export function assertBase64 (str: string, line: number): void {
+  if (!reBase64.test(str)) {
+    throw new TealError(ERRORS.TEAL.INVALID_BASE64, { val: str, line: line });
+  }
+}
+
+/**
+ * Descrition: Checks if string is base32
```
ratik21, line 50:
```suggestion
 * Description: Checks if string is base32
```

packages/algorand-js/src/lib/parse-data.ts
```diff
@@ -23,8 +31,70 @@ export function convertToBuffer (s: string, encoding?: EncodingType): Buffer {
     case EncodingType.HEX: {
       return Buffer.from(s, 'hex');
     }
+    case EncodingType.STRING: {
```
ratik21, line 34: `STRING` does not represent an encoding type. Did you mean `UTF8`? If yes, then please change to utf8. (`Buffer.from(s)` defaults to utf8 only, which is the default case of switch)

packages/algorand-js/src/types.ts
```diff
@@ -117,7 +118,8 @@ export enum AssetParam {
 export enum EncodingType {
   BASE64,
   BASE32,
-  HEX
+  HEX,
+  STRING
```
ratik21, line 122: same as above

packages/algorand-js/test/fixture-projects/teal-files/test-bitwise.teal
```diff
@@ -0,0 +1,8 @@
+// comment
+    |
+
+      &
+
+                ^
+     ~
+// comments
```
ratik21, line 8: :+1:

packages/algorand-js/src/errors/errors-list.ts
amityadav0: Good idea

amityadav0: but i don't think it will be more organized as currently we have `tealErrors`, most of them are overlapping, only 2-3 errors will be moved to Parsing error. some are used in constructor also

packages/algorand-js/src/interpreter/parser.ts
amityadav0: ok

packages/algorand-js/src/lib/parse-data.ts
amityadav0: yes `UTF8`

packages/algorand-js/src/interpreter/opcode-list.ts
amityadav0: this dosn't work. because this should initialized before use

packages/algorand-js/src/errors/errors-list.ts
ratik21: Okay, if only 2-3 errors will be moved then it makes sense to leave as it is.

ratik21 (approved):

PR #163, opened 2020-12-14
Account info TEAL opcodes Interpreter

amityadav0:


packages/algorand-js/src/errors/errors-list.ts
```diff
@@ -133,6 +133,18 @@ by an index that does not exist.`
...
+    number: 19,
+    message: "Account Error - Account doesn't exist",
+    title: "Account Error",
+    description: ``
+  },
+  INVALID_FIELD_TYPE: {
+    number: 20,
+    message: "Asset Field Error - Field unknown",
+    title: "Asset Field Error",
+    description: ``
```
ratik21, line 147: Please add a description to all tests

ratik21, line 147: Please add a description to all tests.

packages/algorand-js/src/interpreter/opcode-list.ts
```diff
@@ -832,3 +832,177 @@ export class Addr extends Op {
...
+   */
+  constructor (args: string[], line: number, interpreter: Interpreter) {
+    super();
+    this.interpreter = interpreter;
+
+    assertLen(args.length, 0, line);
+  };
+
+  execute (stack: TEALStack): void {
+    let last = this.assertBigInt(stack.pop());
```
ratik21, line 856: check stack len before popping.

amityadav0:
This PR will be updated once Interpreter stateful PR gets merged because it has some dependency.

packages/algorand-js/src/interpreter/opcode-list.ts
```diff
@@ -832,3 +832,177 @@ export class Addr extends Op {
...
+    this.interpreter = interpreter;
+
+    assertLen(args.length, 1, line);
+    assertOnlyDigits(args[0]);
+
+    this.field = args[0];
+  };
+
+  execute (stack: TEALStack): void {
+    let last = this.assertBigInt(stack.pop());
```
ratik21, line 902: check stack length

packages/algorand-js/src/interpreter/opcode-list.ts
```diff
@@ -832,3 +832,177 @@ export class Addr extends Op {
...
```
ratik21, line 954: same as above

packages/algorand-js/src/lib/helpers.ts
```diff
@@ -34,3 +34,11 @@ export function assertLen (val: number, expected: number, line: number): void {
     throw new TealError(ERRORS.TEAL.ASSERT_FIELD_LENGTH, { exp: expected, got: val, line: line });
   }
 }
+
+/**
+ * Description: Coverts Buffer into string
+ * @param buf : Buffer
+ */
+export function convertToString (buf: Buffer | Uint8Array): string {
```
ratik21, line 42: We already have a function for this.

packages/algorand-js/src/errors/errors-list.ts
amityadav0: to all errors ?

packages/algorand-js/src/interpreter/opcode-list.ts
amityadav0: ok

packages/algorand-js/src/interpreter/opcode-list.ts
amityadav0: ok

packages/algorand-js/src/interpreter/opcode-list.ts
amityadav0: ok

packages/algorand-js/src/lib/helpers.ts
amityadav0: yes, found it. It's better we have removed helpers.ts now.

packages/algorand-js/src/errors/errors-list.ts
ratik21: yes

packages/algorand-js/src/interpreter/interpreter.ts
```diff
@@ -19,7 +19,9 @@ export class Interpreter {
   tx: Txn;
   gtxs: Txn[];
   accounts: Map<string, AccountState>;
+  accountsAssets: Map<string, Map<number, AccountAssetInfo>>;
   globalApps: Map<number, SSCParams>;
+  globalAssets: Map<number, AssetParams>;
```
robert-zaremba, line 24: For the moment this is fine. But we will need to move it to a `ctx` -> I was writing about this in @ratik21  PR. Please coordinate for the future PR to update it (to avoid conflicts).

packages/algorand-js/src/interpreter/interpreter.ts
```diff
@@ -85,6 +90,18 @@ export class Interpreter {
...
+        this.globalAssets.set(asset.index, asset.params);
+      }
+
+      const assets = acc.assets;
+      const assetInfo = new Map<number, AccountAssetInfo>();
+      for (const asset of assets) {
+        assetInfo.set(asset["asset-id"], asset);
+      }
+
+      this.accountsAssets.set(acc.address, assetInfo);
```
robert-zaremba, line 104: as above, this logic shouldn't be in interpreter. Please add TODO to clean it up in other PR.

packages/algorand-js/src/interpreter/opcode-list.ts
```diff
@@ -1602,3 +1602,160 @@ export class Addr extends Op {
...
+
+// get balance for the requested account specified
+// by Txn.Accounts[A] in microalgos. A is specified as an account
+// index in the Accounts field of the ApplicationCall transaction,
+// zero index means the sender
+export class Balance extends Op {
+  readonly interpreter: Interpreter;
+
+  /**
+   * @param args words list extracted from line: {}
```
robert-zaremba, line 1614: please be more precise in the documentation and describe what are the expected arguments in `args`

packages/algorand-js/src/interpreter/opcode-list.ts
```diff
@@ -1602,3 +1602,160 @@ export class Addr extends Op {
...
+        case "AssetFrozen":
+          value = toBytes(res["is-frozen"]);
+          break;
+        default:
+          throw new TealError(ERRORS.TEAL.INVALID_FIELD_TYPE);
+      }
+
+      stack.push(value);
+      stack.push(BigInt("1"));
+    }
```
robert-zaremba, line 1703:
```suggestion
    if (res === undefined) {
      stack.push(BigInt("0"));
      return;
    } 
    let value: StackElem;
    switch (this.field) {
      case "AssetBalance":
        value = BigInt(res.amount);
        break;
      case "AssetFrozen":
        value = toBytes(res["is-frozen"]);
        break;
      default:
        throw new TealError(ERRORS.TEAL.INVALID_FIELD_TYPE);
    }

    stack.push(value);
    stack.push(BigInt("1"));
```

packages/algorand-js/src/interpreter/opcode-list.ts
```diff
@@ -1602,3 +1602,160 @@ export class Addr extends Op {
...
+
+    stack.push(BigInt(bal));
+  }
+}
+
+// get Asset Holding Info for given account
+// read from account specified by Txn.Accounts[A]
+// and asset B holding field X (imm arg) => {0 or 1 (top), value}
+// params: account index, asset id. Return: did_exist flag (1 if exist and 0 otherwise), value.
```
robert-zaremba, line 1656:
```suggestion
// For ASA A, Account B (txn.accounts[B]) pushes to the 
// stack {...stack, 0} if account has no A holding, 
// otherwise {...stack, holding, 1}
```

packages/algorand-js/src/interpreter/opcode-list.ts
```diff
@@ -1602,3 +1602,160 @@ export class Addr extends Op {
...
+export class AssetHoldingGet extends Op {
```
robert-zaremba, line 1657: Let's rename to `GetASAHolding`

packages/algorand-js/src/interpreter/opcode-list.ts
robert-zaremba: Please update all other constructors documentation in this PR as well.

packages/algorand-js/src/interpreter/opcode-list.ts
```diff
@@ -1602,3 +1602,160 @@ export class Addr extends Op {
...
+    this.interpreter = interpreter;
+    assertLen(args.length, 1, line);
+
+    this.field = args[0];
+  };
+
+  execute (stack: TEALStack): void {
+    this.assertMinStackLen(stack, 2);
+    const last = this.assertBigInt(stack.pop());
+    let prev = this.assertBigInt(stack.pop());
```
robert-zaremba, line 1677: what is in `last` and in `prev`? Please rename this variables.

packages/algorand-js/src/interpreter/opcode-list.ts
```diff
@@ -1602,3 +1602,160 @@ export class Addr extends Op {
...
+
+    this.checkIndexBound(Number(--prev), this.interpreter.tx.apat);
+
+    const accBuffer = this.interpreter.tx.apat[Number(prev)];
+    const val = this.interpreter.accountsAssets.get(convertToString(accBuffer));
+    const res = val?.get(Number(last));
```
robert-zaremba, line 1683: Is `val` an account and `res` is amount? Please rename accordingly.

packages/algorand-js/src/interpreter/opcode-list.ts
```diff
@@ -1602,3 +1602,160 @@ export class Addr extends Op {
...
+
+      stack.push(value);
+      stack.push(BigInt("1"));
+    }
+  }
+}
+
+// get Asset Params Info for given account
+// read from asset Txn.ForeignAssets[A] params field X (imm arg) => {0 or 1 (top), value}
+// params: txn.ForeignAssets offset. Return: did_exist flag (1 if exist and 0 otherwise), value.
```
robert-zaremba, line 1709: Please improve the comment. You can look at my suggestion of the operator above.

packages/algorand-js/src/interpreter/opcode-list.ts
```diff
@@ -1602,3 +1602,160 @@ export class Addr extends Op {
...
+  };
+
+  execute (stack: TEALStack): void {
+    this.assertMinStackLen(stack, 1);
+    let last = this.assertBigInt(stack.pop());
+    this.checkIndexBound(Number(--last), this.interpreter.tx.apas);
+
+    const assetId = this.interpreter.tx.apas[Number(last)];
+
+    const res = this.interpreter.globalAssets.get(assetId);
```
robert-zaremba, line 1737: as above, please rename `res`, `last` ...

robert-zaremba (commented):
Please add TODOs to the code and coordinate with @ratik21  to approve this PR and create a new PR to update for interpreter for a State model.

packages/algorand-js/src/interpreter/opcode-list.ts
amityadav0: I left it blank because `balance` expect 0 arguments.

packages/algorand-js/src/interpreter/opcode-list.ts
amityadav0: ok, currently these comments are taken from TEAL documentation, updating

packages/algorand-js/src/interpreter/opcode-list.ts
amityadav0: updated all the other opcodes in parser support for branch PR

packages/algorand-js/src/interpreter/opcode-list.ts
amityadav0: ok

packages/algorand-js/src/interpreter/opcode-list.ts
amityadav0: ok

packages/algorand-js/src/interpreter/opcode-list.ts
amityadav0: ok

packages/algorand-js/src/interpreter/opcode-list.ts
amityadav0: updating for all other opcodes too.

packages/algorand-js/src/interpreter/opcode-list.ts
amityadav0: ok

packages/algorand-js/src/interpreter/interpreter.ts
amityadav0: Ok

packages/types-algosdk/index.d.ts
```diff
@@ -555,11 +555,26 @@ export interface SSCParams {
...
+  defaultFrozen: string;
+  unitName: string;
+  name: string;
+  url: string;
+  metadataHash: string;
+  manager: string;
+  reserve: string;
+  freeze: string;
+  clawback: string;
+}
```
ratik21, line 571: I think in `algosdk`, defaultFrozen is represented as 'default-frozen'. Can you please confirm?
If this was intentional (using our custom types for accounts), then i don't think `AssetParams` should be in `types/algosdk`. Can we move to `types/algorand-js`?

amityadav0: yes, i was matching it from constructor. I think updating here is better. Thanks for catching

packages/algorand-js/src/interpreter/opcode-list.ts
```diff
@@ -1744,3 +1744,167 @@ export class Addr extends Op {
...
+      const res = this.interpreter.runtime.ctx.state.accounts.get(convertToString(buf));
+
+      if (res === undefined) {
+        throw new TealError(ERRORS.TEAL.ACCOUNT_DOES_NOT_EXIST);
+      }
+
+      bal = res.amount;
+    }
+
+    stack.push(BigInt(bal));
```
robert-zaremba, line 1794:
```suggestion
    let res;

    if (accountIndex === BigInt("0")) {
      const sender = convertToString(this.interpreter.runtime.ctx.tx.snd);
      res = this.interpreter.runtime.ctx.state.accounts.get(sender);
    } else {
      this.checkIndexBound(Number(--accountIndex), this.interpreter.runtime.ctx.tx.apat);
      const buf = this.interpreter.runtime.ctx.tx.apat[Number(accountIndex)];
      res = this.interpreter.runtime.ctx.state.accounts.get(convertToString(buf));
    }
    if (val === undefined)
        throw new TealError(ERRORS.TEAL.ACCOUNT_DOES_NOT_EXIST);

    stack.push(BigInt(res.amount));
```

packages/algorand-js/src/interpreter/opcode-list.ts
```diff
@@ -1744,3 +1744,167 @@ export class Addr extends Op {
...
+  }
+}
+
+// For Account A, Asset B (txn.accounts[A]) pushes to the
+// stack [...stack, 0] if account has no B holding,
```
robert-zaremba, line 1799:
```suggestion
// push to stack [...stack, 0] if account has no B holding,
```

robert-zaremba: Please update the other comments accordingly

packages/algorand-js/src/runtime/runtime.ts
```diff
@@ -17,24 +17,29 @@ import { assertValidSchema, getKeyValPair } from "../lib/stateful";
...
+      globalAssets: new Map<number, AssetParams>()
     };
+
+    // intialize accounts (should be done during runtime initialization)
+    this.initializeAccounts(accounts);
+
     // context for interpreter
     this.ctx = {
-      state: <State>{},
+      state: cloneDeep(this.store), // state is a deep copy of store
```
robert-zaremba, line 38: I t hink @ratik21  already did this in his PR.

packages/algorand-js/src/types.ts
```diff
@@ -80,7 +66,9 @@ export interface AccountsMap {
 
 export interface State {
   accounts: Map<string, StoreAccount>
+  accountAssets: Map<string, Map<number, AccountAssetInfo>>
   globalApps: Map<number, SSCParams>
+  globalAssets: Map<number, AssetParams>
```
robert-zaremba, line 71: Why don't we keep assets in `StoreAccount` object?
I would see it as this:
```
type StoreAccount {
  ...
  assetHoldings: AssetHolding // renamed from AccountAssetInfo
}
```
`globalAssets` should be renamed to `assetDefs`  and `AssetParams` to `AssetDef`.
We can keep it directly State, or move it to `StoreAccount` as well. What do you think?

packages/types-algosdk/index.d.ts
```diff
@@ -555,11 +555,26 @@ export interface SSCParams {
...
+
 export interface AccountAssetInfo {
   amount: number;
   'asset-id': number;
   creator: string;
-  'is-frozen': boolean;
+  'is-frozen': string;
```
robert-zaremba, line 577: why this is changed to string?

robert-zaremba (changes requested):

packages/algorand-js/src/interpreter/opcode-list.ts
amityadav0: ok

packages/algorand-js/src/runtime/runtime.ts
amityadav0: Previously it was done in `executeTx` function only, but not at the time of construction. But for testing it properly, we need to have access without calling `executeTx`

packages/algorand-js/src/types.ts
amityadav0: I don't think we can move globalAssets to StoreAccount because `storeAccount` contains information linked to each account.
But we need have to information about assets globally, so that we can return queries about AssetParams

amityadav0: assetHoldings is already in StoreAccount in `assets` object.
globalAssets is for Assets Parameters(Total, decimals, clawback etc..)

packages/algorand-js/src/runtime/runtime.ts
robert-zaremba: But how do you change `runtime.ctx` without calling `executeTx`? For which test do you need it?

packages/algorand-js/src/types.ts
robert-zaremba: AFAIK, in Algorand, all assets definitions are also stored in an Account. That's why I was thinking to put it there as well. But if it's not needed, or easier to do it other way around then I'm fine to keep `assetDefs` directly in `Store`.

robert-zaremba: > assetHoldings is already in StoreAccount in assets object.

Then why did you add `accountAssets`?

packages/types-algosdk/index.d.ts
amityadav0: I checked this in js-algorand repo.
https://github.com/algorand/js-algorand-sdk/blob/a5309ee57dddbf6f9db5f95dc8a82eb1ae03c326/tests/cucumber/steps/steps.js#L2092

packages/algorand-js/src/runtime/runtime.ts
amityadav0: Here, only `store` is changed.
while testing opcodes, for ex: assetHolding,
assetHolding opcode require `ctx` but if we don't initialize it here, ctx will be blank as per previous code. One way will be remove `private` from store variable and initialize ctx in test.
In tests:
`interpreter.runtime.ctx.state = runtime.store`

packages/algorand-js/src/types.ts
amityadav0: because I think this is faster. For ex: If we want to get user asset balance, One way is to traverse the whole assets array and stop when we find required `asset-id` for every query.
I think this will be slow, that's why `accountAssets` traverse accounts during initialization and creates a map<asset-id, asset-holdings>. when querying we get the values directly from map

> AFAIK, in Algorand, all assets definitions are also stored in an Account. That's why I was thinking to put it there as well. But if it's not needed, or easier to do it other way around then I'm fine to keep `assetDefs` directly in `Store`.

yes, they are stored in ["created-assets"] in every account, but for a query we  will have to traverse all the accounts to get a specific asset parameter. that's why i though global map for assets is better.

amityadav0: If you want i can move assetDefs to store, and move the logic in storeAccount, (find value by traversing)

packages/algorand-js/src/runtime/runtime.ts
robert-zaremba: OK, let's keep what you wrote with `clone`.  Please add a comment to the code.

packages/algorand-js/src/types.ts
robert-zaremba: OK, let's keep `assetDefs` outside of StoreAccount.

robert-zaremba: >  If we want to get user asset balance, One way is to traverse the whole assets array and stop when we find required asset-id for every query.

But we still need to go through a user, right? user_account_address => asset_id => user_asset_holding

amityadav0: accountAssets handles both, accountAssets contains <address, assets> and assets is <assetId, assetHoldings>

packages/algorand-js/src/runtime/runtime.ts
amityadav0: ok

robert-zaremba (approved):

PR #164, opened 2020-12-15
Interpreter stateful opcodes

ratik21:
- Added opcodes
```
app_opted_in
app_local_get
app_local_get_ex
app_global_get
app_global_get_ex
app_local_put
app_global_put
app_local_del
app_global_del
```
- Added tests for each opcode
- Updated types for stateful in `types/algosdk`


packages/algorand-js/src/types.ts
```diff
@@ -125,3 +125,7 @@ export enum EncodingType {
 export interface AccountsMap {
   [addr: string]: AccountInfo
 }
+
+export interface ApplicationMap {
+  [appId: number]: SSCParams
+}
```
robert-zaremba, line 131: Why not using a `Map` type?
eg:
```
Map<number, SSCParams>  // instead of ApplicationMap
Map<string, AccountInfo>  // instead of AccountsMap
```

packages/types-algosdk/index.d.ts
```diff
@@ -539,13 +546,38 @@ export interface Address {
...
+  params: SSCParams;
+}
+
+export interface AppLocalState {
+  id: number;
+  'key-value': SSCStateSchema[];
+  schema: SSCSchemaConfig;
+}
+
 export interface AccountInfo {
```
robert-zaremba, line 581: There is no such structure in `algosdk`.
How about renaming it to `AccountState`? It would make more sense in my opinion.

packages/types-algosdk/index.d.ts
```diff
@@ -200,15 +200,22 @@ export interface MultiSigAccount {
...
 }
 
+// Stateful Smart Contract Schema
 export interface SSCStateSchema {
-  key: string;
+  key: string | Uint8Array;
   value: {
     type: number;
-    bytes: string;
+    bytes: string | Uint8Array;
```
robert-zaremba, line 208: why do we have `string` type for bytes? Can we remove `string`? Otherwise, can we add a ne `value` filed (string: string)?

robert-zaremba (changes requested):

packages/types-algosdk/index.d.ts
ratik21: SDk returns base64 encoded data for key value pairs represent as bytes. here is the chat link: [https://discord.com/channels/754286650540032030/754286650540032033/788166652642328596]() 

Adding a new value to this type would lead in an incorrect structure.

packages/types-algosdk/index.d.ts
ratik21: This was present by before. But i think this structure is present in `algosdk`.
```
const accountInfo = await deployer.algodClient.accountInformation(account).do();
```
Changing name to `AccountState`.

packages/algorand-js/src/types.ts
ratik21: Actually we are fetching info from nested objects in global and local state. So using `Object` key - value pairs seemed easier.
eg.
```
this.interpreter.globalApps[appId]["global-state"]
```
Should i change to `Map`?

ratik21: Update: Changed to using `Map`

packages/types-algosdk/index.d.ts
robert-zaremba: good, thanks.

packages/algorand-js/src/interpreter/opcode-list.ts
```diff
@@ -1217,6 +1219,302 @@ export class Return extends Op {
...
+  execute (stack: TEALStack): void {
+    this.assertMinStackLen(stack, 3);
+    const key = this.assertBytes(stack.pop());
+    const appId = this.assertBigInt(stack.pop());
+    const accountIndex = this.assertBigInt(stack.pop());
+
+    const account = this.getAccount(accountIndex, this.interpreter);
+    const val = getLocalState(Number(appId), account, key);
+    if (val) {
+      stack.push(val);
```
amityadav0, line 1313: Please push did_exist flag in stack. (according to docstring)

packages/algorand-js/src/interpreter/opcode-list.ts
```diff
@@ -1217,6 +1219,302 @@ export class Return extends Op {
...
+    if (appIndex === BIGINT0) {
+      appId = this.interpreter.tx.apid; // zero index means current app
+    } else {
+      this.checkIndexBound(Number(--appIndex), foreignApps);
+      appId = foreignApps[Number(appIndex)];
+    }
+
+    const val = getGlobalState(appId, key, this.interpreter);
+    if (val) {
+      stack.push(val);
```
amityadav0, line 1382: Please push did_exist flag

packages/algorand-js/src/interpreter/opcode-list.ts
ratik21: yes, sorry

packages/algorand-js/src/interpreter/opcode-list.ts
ratik21: ok

packages/algob/src/types.ts
```diff
@@ -450,16 +450,6 @@ export interface FundASCFlags {
...
-export interface SSCStateSchema {
-  key: string
-  value: {
-    type: number
-    bytes: string
-    uint: number
-  }
-}
-
 export interface SSCOptionalFlags {
```
robert-zaremba, line 453: Please add a doc comment here.

packages/algorand-js/src/interpreter/opcode.ts
```diff
@@ -71,4 +76,33 @@ export class Op {
...
+      const senderAccount = convertToString(interpreter.tx.snd);
+      account = interpreter.accounts.get(senderAccount);
+    } else {
+      accountIndex--;
+      this.checkIndexBound(Number(accountIndex), interpreter.tx.apat);
+      const pkBuffer = interpreter.tx.apat[Number(accountIndex)];
+      account = interpreter.accounts.get(convertToString(pkBuffer));
+    }
+    return this.assertAccountDefined(account);
+  }
```
robert-zaremba, line 107: this should be a method of a runtime. You can do this later, once you will create runtime -> in that case please add TODO in the file, and add a subtask to the Runtime PoC pivotal task.

packages/algorand-js/src/interpreter/ssc.ts
```diff
@@ -0,0 +1,150 @@
+/* eslint sonarjs/no-duplicate-string: 0 */
```
robert-zaremba, line 1: plase rename this file to `stateful.ts`

packages/algorand-js/src/interpreter/ssc.ts
```diff
@@ -0,0 +1,150 @@
...
+  const data = localState.find(state => state.id === appId)?.["key-value"]; // can be undefined (eg. app opted in)
+  if (data) {
+    const keyValue = data.find(schema => compareArray(schema.key as Uint8Array, key));
+    const value = keyValue?.value;
+    if (value) {
+      return value?.bytes as Uint8Array || BigInt(value?.uint);
+    }
+  }
+  return undefined;
+}
```
robert-zaremba, line 59: this should be a method of the `AccountState` class.

packages/algorand-js/src/interpreter/ssc.ts
```diff
@@ -0,0 +1,150 @@
...
+  }
+  const globalState = appDelta["global-state"];
+
+  const keyValue = globalState.find(schema => compareArray(schema.key as Uint8Array, key));
+  const value = keyValue?.value;
+  if (value) {
+    return value?.bytes as Uint8Array || BigInt(value?.uint);
+  }
+  return undefined;
+}
```
robert-zaremba, line 84: this should be a method of a runtime. You can do this later, once you will create runtime -> in that case please add TODO in the file, and add a subtask to the `Runtime PoC` pivotal task.

packages/algorand-js/src/interpreter/ssc.ts
```diff
@@ -0,0 +1,150 @@
...
+        localDelta.push(data); // push new pair if key not found
+      } else {
+        localDelta[idx].value = data.value; // update value if key found
+      }
+      l["key-value"] = localDelta; // save updated state
+
+      assertValidSchema(l["key-value"], l.schema); // verify if updated schema is valid by config
+      return localState;
+    }
+  }
```
robert-zaremba, line 114: this should be a method of `AccountState`

packages/algorand-js/src/interpreter/ssc.ts
```diff
@@ -0,0 +1,150 @@
...
+  if (idx === -1) {
+    globalState.push(data); // push new pair if key not found
+  } else {
+    globalState[idx].value = data.value; // update value if key found
+  }
+  appDelta["global-state"] = globalState; // save updated state
+
+  assertValidSchema(appDelta["global-state"], appDelta["global-state-schema"]); // verify if updated schema is valid by config
+  return globalState;
+}
```
robert-zaremba, line 150: this should be a method of a runtime. You can do this later, once you will create runtime -> in that case please add TODO in the file, and add a subtask to the Runtime PoC pivotal task.

packages/algorand-js/src/interpreter/opcode.ts
```diff
@@ -71,4 +76,33 @@ export class Op {
...
+    return a;
+  }
+
+  assertAppDefined (appId: number, interpreter: Interpreter): SSCParams {
+    const globalDelta = interpreter.globalApps.get(appId);
+    if (globalDelta === undefined) {
+      throw new TealError(ERRORS.TEAL.APP_NOT_FOUND);
+    }
+    return globalDelta;
+  }
```
robert-zaremba, line 93: this should be a method of a runtime. You can do this later, once you will create runtime -> in that case please add TODO in the file, and add a subtask to the Runtime PoC pivotal task.

packages/types-algosdk/index.d.ts
robert-zaremba: so, if we store here `base64` string, then why do we need `Uint8Array`?

robert-zaremba (changes requested):

packages/types-algosdk/index.d.ts
ratik21: Because we are representing bytes as `Uint8Array`. Could you check: [https://discord.com/channels/754286650540032030/754286650540032033/788171546651918376]()

packages/algorand-js/src/interpreter/ssc.ts
ratik21: ok

packages/algorand-js/src/interpreter/ssc.ts
ratik21: Sorry, i don't follow. Did you mean the `AccountState` interface? If we add this method to the interface, then we will defer from algosdk's representation of `Account Information`. ( `getLocalState` is a custom function)

Or did you mean the `MockAccount` class which we will be adding in PoC? Should this method be moved there?
Or something else?

packages/algob/src/types.ts
ratik21: ok

packages/algorand-js/src/interpreter/ssc.ts
ratik21: ok, will add a TODO

packages/algorand-js/src/interpreter/opcode.ts
ratik21: ok

packages/algorand-js/src/interpreter/ssc.ts
ratik21: ok

packages/algorand-js/src/interpreter/ssc.ts
ratik21: I have added subtask for moving this to the `MockAccount` class in PoC

packages/types-algosdk/index.d.ts
robert-zaremba: Yes, that why in my first comment I was hoping to use only `Uint8Array`, not a string.  It's still not clear to me why we can't operate only on bytes and use base64 only for IO.

packages/algorand-js/src/interpreter/opcode-list.ts
```diff
@@ -1201,6 +1203,304 @@ export class Return extends Op {
...
+    } else {
+      stack.push(BIGINT0); // The value is zero if the key does not exist.
+    }
+  }
+}
+
+// read from account specified by Txn.Accounts[A] from local state of the
+// application B key C => {0 or 1 (top), value}
```
robert-zaremba, line 1273:
```suggestion
// read from application local state at Txn.Accounts[A] => app B => key C from local state.
// Pushes to the stack [...stack, val, 1] if the key exists, otherwise [...stack, 0]
```

robert-zaremba (approved):
Approving, however this still bothers me: https://github.com/scale-it/algorand-builder/pull/164#discussion_r543706630

ratik21:
> Approving, however this still bothers me: [#164 (comment)](https://github.com/scale-it/algorand-builder/pull/164#discussion_r543706630)

I have restored to using `Uint8Array` only

PR #165, opened 2020-12-15
remove helpers.ts

amityadav0:
- Move functions to parse-data.ts


packages/algorand-js/src/interpreter/opcode-list.ts
```diff
@@ -702,7 +701,7 @@ export class EqualTo extends Op {
           stack.push(BIGINT0);
         }
       } else {
-        if (compareArray(this.assertBytes(last), this.assertBytes(prev))) {
+        if (JSON.stringify(this.assertBytes(last)) === JSON.stringify(this.assertBytes(prev))) {
```
ratik21, line 704: I think using `JSON.stringify` has some edge cases. Can you check: https://www.30secondsofcode.org/blog/s/javascript-array-comparison

packages/algorand-js/src/lib/parse-data.ts
```diff
@@ -2,8 +2,53 @@ import * as base32 from "hi-base32";
...
+/**
+ * Description: Checks if string is base32
+ * @param str : string that needs to be checked
+ * @param line : line number in TEAL file
+ */
+export function assertBase32 (str: string, line: number): void {
+  if (!reBase32.test(str)) {
+    throw new TealError(ERRORS.TEAL.INVALID_BASE32, { val: str, line: line });
+  }
+}
```
ratik21, line 51: how about renaming this file to `parsing.ts`?

ratik21 (changes requested):
I think using or having `compareArray` was better. Could you please move that function to a new package (It's being used in many stateful opcodes as well)? We can have a generic function like
```
function compareArray (a: any[], b: any[]): Boolean {
 if (typeof a === typeof b) {
  // compare & return
 }
 return false;
} 
```

packages/algorand-js/src/interpreter/opcode-list.ts
amityadav0: Yes, I read it, it's not the case here as we have bytes array. Also if you see one line above we are using assert for that.

But if it is being used in stateful also then it is better to have compareArray like before.

Restoring

robert-zaremba: Also, could you restructure the code a bit? Having a waterfall of if's is hard to read. Example of improvement:
```
if (typeof last !== typeof prev)
   throw new TealError(ERRORS.TEAL.INVALID_TYPE);    // NOTE: are you sure we should throw an exception instead pushing 0 to the stack?
if (typeof last === "bigint")
   stack.push(last === prev? BIGINT1 : BIGINT0)
else {
  ...
}

**Review by robert-zaremba — CHANGES_REQUESTED** [review]

File: packages/algorand-js/src/interpreter/opcode-list.ts (hunk 1 above)
Comment (amityadav0) [reply packages/algorand-js/src/interpreter/opcode-list.ts:704]: yes

opEqq

File: packages/algorand-js/src/lib/parse-data.ts (hunk 2 above)
Comment (amityadav0) [reply packages/algorand-js/src/lib/parse-data.ts:51]: yes

File: packages/algorand-js/src/lib/compare.ts [hunk 3]
```diff
@@ -0,0 +1,9 @@
+/**
+ * Description: Compare Uint8Arrays
+ * @param a Uint8Array
+ * @param b Uint8Array
+ */
+export function compareArray (a: Uint8Array, b: Uint8Array): Boolean {
```
ratik21, line 6: Why only `Uint8Array`? This means i can't compare two string arrays. Can we make this more general? (using `any[]`)

amityadav0: because `(method) Op.assertBytes(b: unknown): Uint8Array
Argument of type 'Uint8Array' is not assignable to parameter of type 'any[]'`

amityadav0: also using any[] is not good. what if you pass one array of numbers and one Uint8array ? The code will return undesired results.
If you want to compare string we can add `Uint8array | string[] `.

robert-zaremba: how about just using `unknown[]`?

ratik21: OK. Then please change the function name to `compareUint8Array` / `compareBytes`. 
`compareArray` suggests we can compare any two arrays.

amityadav0: unkown also gives same results.
`Argument of type 'Uint8Array' is not assignable to parameter of type 'unknown[]'.`

ratik21:
```
function compareArray (a: unknown, b: unknown): Boolean {
  if (Array.isArray(a) && Array.isArray(b)) {
    return a.length === b.length &&
      a.every((v, i) => v === b[i]);
  }
  return false;
}
```
This works

amityadav0: ok, should we ignore this case ? `what if you pass one array of numbers and one Uint8array ?`

amityadav0: we won't be using this in our code, i think we can

ratik21: @amityadav0 i think just renaming `compareArray` would be better in this case. Somehow, `Array.isArray(a)` is not accepting `Uint8Array`. :/

amityadav0:
```
function compareArray (a: unknown, b: unknown): Boolean {
  if (Array.isArray(a) && Array.isArray(b)) {
    return a.length === b.length &&
      a.every((v, i) => v === b[i]);
  }
  return false;
}
```
this code doesn't work because, it is considering Uint8array as type, so it requires array  of Uint8array

amityadav0: updated

amityadav0: @ratik21 I have updated code to take Uin8array as well. let me know what do you think ?

ratik21: yes, i think we can use `instance of Uint8Array`. your call.

packages/algorand-js/src/lib/compare.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * Description: compare two array: returns true if they are equal, else false.
+ * @param a unkown (array)
```
ratik21, line 3:
```suggestion
 * @param a unknown (array)
```

packages/algorand-js/src/lib/compare.ts
```diff
@@ -0,0 +1,17 @@
...
+ * @param b unkown (array)
```
ratik21, line 4:
```suggestion
 * @param b unknown (array)
```

packages/algorand-js/src/lib/compare.ts
```diff
@@ -0,0 +1,17 @@
...
+ */
+export function compareArray (a: unknown, b: unknown): Boolean {
+  if ((Array.isArray(a) && Array.isArray(b))) {
+    return a.length === b.length &&
+      a.every((v, i) => v === b[i]);
+  }
+
+  if (a instanceof Uint8Array && b instanceof Uint8Array) {
```
ratik21, line 12: :+1:

ratik21 (approved):

packages/algorand-js/src/interpreter/opcode-list.ts
```diff
@@ -694,22 +694,20 @@ export class EqualTo extends Op {
...
-    if (typeof last === typeof prev) {
-      if (typeof last === "bigint") {
-        if (last === prev) {
-          stack.push(BIGINT1);
-        } else {
-          stack.push(BIGINT0);
-        }
+    if (typeof last !== typeof prev) {
+      throw new TealError(ERRORS.TEAL.INVALID_TYPE);
+    } else if (typeof last === "bigint") {
```
robert-zaremba, line 699: else not needed here, because we throw exception.
So we can do a normal block:
```
if (typeof last !== typeof prev)
        throw new TealError(ERRORS.TEAL.INVALID_TYPE);
if (typeof last === "bigint") {
   ...
```

packages/algorand-js/src/interpreter/opcode-list.ts
```diff
@@ -729,22 +727,20 @@ export class NotEqualTo extends Op {
...
-        }
+        stack.push(BIGINT1);
       }
     } else {
-      throw new TealError(ERRORS.TEAL.INVALID_TYPE);
+      if (compareArray(this.assertBytes(last), this.assertBytes(prev))) {
+        stack.push(BIGINT0);
+      } else {
+        stack.push(BIGINT1);
+      }
```
robert-zaremba, line 743: this thing is repeated many times. Please move it to function:
```
function pushBolleanCheck(stack, ok: boolean) {
  if (ok) 
    stack.push(BIGINT0)
  else ...
}

robert-zaremba: BTW, are you sure we should push `0` when things are equal?

robert-zaremba (changes requested):

packages/algorand-js/src/interpreter/opcode-list.ts
amityadav0: ok

packages/algorand-js/src/interpreter/opcode-list.ts
amityadav0: yes, because here opcode is `notEqualTo`.

robert-zaremba (approved):

PR #166, opened 2020-12-16
Add more tests(Edge cases)

amityadav0:
- Added more edge cases tests.
- I have tried adding all the cases, Please let me know if you think if a case needs to be added.


ratik21 (approved):
LGTM

PR #167, opened 2020-12-16
Parser: support branch, label and transaction opcodes

amityadav0:
- Add support for branch label and transaction opcodes
- Add tests


packages/algorand-js/src/interpreter/opcode-list.ts
```diff
@@ -1005,9 +1005,17 @@ export class Txn extends Op {
...
+  /**
+   * @param args words list extracted from line
+   * @param line line number in TEAL file
+   * @param interpreter interpreter object
+   */
+  constructor (args: string[], line: number, interpreter: Interpreter) {
     super();
-    this.field = txField;
+    assertLen(args.length, 1, line);
+    assertOnlyDigits(args[0]);
```
ratik21, line 1016: I don't think we'll be receiving digits here. (it should be a string eg. `txn Fee`). Could you please validate?

amityadav0: checked with Teal opcode documentation , it mentions we will receive numbers

wer

just checked, it doesn't work with compiler.
`txn Fee` works
`txn 0` doesn't

what is this documentation ?

ratik21: `txn Fee` is processed by `assember.go` in go-algorand. Here since we have a parser, we should be processing strings. 
Seems like they documented the opcodes based on assembler input. :laughing:

amityadav0: Is this documentation for their engineers or a person who wants to write code in TEAL ?

robert-zaremba: @ratik21  - in go-algorand, the teal is compiled to a bytecode, and the assembler operates on the bytecode, not on the teal source code.

robert-zaremba: @amityadav0  Please document the `args` - describe what is inside. (in all new constructors as well)

robert-zaremba (approved):
Pre-approving. Please update the doc strings.

packages/algorand-js/src/interpreter/opcode-list.ts
ratik21: > @ratik21 - in go-algorand, the teal is compiled to a bytecode, and the assembler operates on the bytecode, not on the teal source code.

ah, okay. Thanks for note.

amityadav0: yes, adding args docs

PR #169, opened 2020-12-17
PoC for integrating Mocha with TEAL interpreter

ratik21:
- Added simple asc logic to transfer algo's from `john` to `bob`
- Added mocha test
- Added `Runtime` class to manage state (intialization + update after successful transaction)

NOTES:
1. I wanted to add a better .teal code (something like `Txn Fee` check + `Txn TypeEnum`..etc), but the parser doesn't support transaction opcodes yet (this is the PR: #167 )
2. State will be furthur updater after #164 is merged. (Edit: this is done)


packages/algorand-js/src/interpreter/opcode-list.ts
```diff
@@ -1033,7 +1033,7 @@ export class Gtxn extends Op {
 
   execute (stack: TEALStack): void {
     this.assertUint8(BigInt(this.txIdx));
-    this.checkIndexBound(this.txIdx, this.interpreter.gtxs);
+    this.checkIndexBound(this.txIdx, this.interpreter.storageBranch.gtxs);
```
robert-zaremba, line 1036: `storage` should only contain the "blockhain" state, not current transactions. This should be in the runtime context or in interpreter context. So there are 2 options:
1. Interpreter has a reference to runtime, and uses runtime functions (`getTx`, `getAccount` etc...) - so it doesn't have a direct reference to a state and transaction context.
2. Similar to what you have now, we will create a `ctx` object in the interpreter (`interpreter.ctx`) which will be initialized by runtime when starting the interpreter (interpreter.execute). The ctx should be something like:

        ctx = {
          state,   // note: please rename storageBranch to state, interpreter shouldn't know if he operates on a branch or not. If we need are keeping a copy of the state in runtime, then in runtime we should use `stateBranch`.
          tx / gtx,
          currentAccount .... // not sure if this is needed.
        }
```

File: packages/algorand-js/src/interpreter/opcode.ts [hunk 2]
```diff
@@ -98,13 +98,13 @@ export class Op {
   getAccount (accountIndex: bigint, interpreter: Interpreter): AccountState {
```
robert-zaremba, line 98: this function should be in in runtime, or be a method of `state` object.

packages/algorand-js/src/types.ts
```diff
@@ -121,3 +121,14 @@ export enum EncodingType {
   HEX,
   UTF8
 }
+
+export interface AccountsMap {
+  [addr: string]: AccountState
+}
+
+export interface Storage {
```
robert-zaremba, line 129: Please rename to `State`.

packages/algorand-js/src/types.ts
```diff
@@ -121,3 +121,14 @@ export enum EncodingType {
...
+  accounts: Map<string, AccountState>
+  globalApps: Map<number, SSCParams>
+  tx: Txn // current txn
+  gtxs: Txn[] // all transactions
```
robert-zaremba, line 133: `tx` and `gtxs` is not a part of State. This is a part of the current transaction context / runtime context.

packages/algob/test/smart-contracts/escrow-account.ts
```diff
@@ -0,0 +1,78 @@
...
+import { ERRORS } from "algorand-js/src/errors/errors-list";
+import { expectTealErrorAsync } from "algorand-js/test/helpers/errors";
+import { assert } from "chai";
+
+import { useEnvironment } from "../helpers/environment";
+import { useFixtureProject } from "../helpers/project";
+import { Runtime } from "../helpers/runtime";
+import { MockAccount } from "../mocks/state";
+
+describe("Algorand Smart Contracts", function () {
```
robert-zaremba, line 11: I think this test should be an integration test in `algorand-js` (eg: algorand-js/tests/integration/escrow-account.ts). It doesn't use `algob`.

In `algob`, in the future we will add a functional test - a fixture algob project with teal code in `/assets`, tests in `/tests`.

robert-zaremba (changes requested):

packages/algob/src/internal/core/config/config-errors.ts
```diff
@@ -30,7 +30,7 @@ export default class CfgErrors {
       expectedType));
   }
 
-  public concatenate (errors: string[]): void{
+  public concatenate (errors: string[]): void {
```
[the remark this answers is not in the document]
ratik21: ok

packages/algorand-js/src/interpreter/interpreter.ts
```diff
@@ -105,32 +53,30 @@ export class Interpreter {
...
     while (this.instructionIndex < this.instructions.length) {
       const instruction = this.instructions[this.instructionIndex];
-      instruction.execute(this.stack);
+      if (!(instruction instanceof Pragma)) {
+        instruction.execute(this.stack);
+      }
       this.instructionIndex++;
     }
 
     if (this.stack.length() === 1) {
```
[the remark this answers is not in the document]
ratik21: I have tested this. If there is no `return` statement in the TEAL code, and if the stack length (after execution) is greator than 1  or if top of stack is `bytes`, it throws an error.
We return the last value on stack if there is a `return` statement in teal code.

packages/algorand-js/src/types.ts
ratik21: ok

packages/algob/test/mocks/state.ts
```diff
@@ -0,0 +1,41 @@
+import type { Account, AccountAssetInfo } from "algosdk";
+import { generateAccount } from "algosdk";
+
+export class MockAccount {
+  readonly account: Account;
```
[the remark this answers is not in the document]
ratik21: Renaming. 
Why is this not needed? Here is the reference from test :-
```
const john = new sdkAccount(100) // initializes new account with 100 microAlgos
```

packages/algob/test/smart-contracts/escrow-account.ts
```diff
@@ -0,0 +1,78 @@
...
+    const johnBal = john.balance();
+
+    const txnParams = {
+      type: 0, // payment
+      sign: 0,
+      fromAccount: escrow.account,
+      toAccountAddr: john.address,
+      amountMicroAlgos: 50,
+      payFlags: { totalFee: 1000 }
+    };
```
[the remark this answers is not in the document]
ratik21: Ok. thanks

packages/algob/test/mocks/state.ts
```diff
@@ -0,0 +1,41 @@
...
+  "amount-without-pending-rewards": number;
+  'pending-rewards': number;
+  'reward-base': number;
+  rewards: number;
+  round: number;
+  status: string;
+  'apps-local-state': any;
+  'apps-total-schema': any;
+  'created-apps': any;
+  'created-assets': any;
```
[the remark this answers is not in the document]
ratik21: Okay, i think that makes more sense.

packages/algob/test/smart-contracts/escrow-account.ts
ratik21: Alright.

packages/algorand-js/src/interpreter/interpreter.ts
```diff
@@ -105,32 +53,30 @@ export class Interpreter {
...
-   * @returns {boolean} : transaction accepted/rejected based on ASC logic
+   * Description: this function executes TEAL code after parsing
+   * @param {string} path: path to teal code
+   * @param {Uint8Array[]} args : external arguments
+   * @param {Storage} state : current state as input
    */
-  execute (txnParams: ExecParams | ExecParams[],
-    logic: Operator[], args: Uint8Array[],
-    accounts: AccountState[]): boolean {
+  async execute (path: string, args: Uint8Array[],
```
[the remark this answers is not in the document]
ratik21: yes, updated.

packages/algorand-js/src/types.ts
ratik21: Okay, sorry. I am adding this as a part of `runtime.ctx`

packages/algob/test/mocks/state.ts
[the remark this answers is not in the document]
robert-zaremba: This is an Account Storage object, right? Do we use somehow account keys in the interpreter? if yes, then keep it, if not then it's not needed. The only thing we need here is an address I think.

packages/algorand-js/src/interpreter/interpreter.ts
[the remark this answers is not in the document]
robert-zaremba: OK, thanks

packages/algob/test/mocks/state.ts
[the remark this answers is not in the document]
ratik21: Yes, this represents the account we use in the interpreter. So all the state manipulation (eg putting in local state or getting from local state), we get from this representation only.

packages/algorand-js/src/interpreter/opcode.ts
ratik21: okay, moved to runtime

packages/algorand-js/src/interpreter/opcode-list.ts
ratik21: ok

ratik21:
Update:
- [ ] Changed File Structure

- [ ] Updated parser to take teal code as string

- [ ] Moved global functions (`getGlobalState`, `putGlobalState`, `getAccount`, ..) to `Runtime`

- [ ] Moved local functions (`getLocalState`, `putLocalState`) to `SdkAccount`

packages/algorand-js/src/interpreter/interpreter.ts
```diff
@@ -105,20 +50,17 @@ export class Interpreter {
...
-   * Description: this function executes set of Operator[] passed after
-   * parsing teal code
-   * @param {ExecParams} txn : Transaction parameters
-   * @param {Logic[]} logic : smart contract instructions
-   * @param {AppArgs} args : external arguments
-   * @returns {boolean} : transaction accepted/rejected based on ASC logic
+   * Description: this function executes TEAL code after parsing
+   * @param {string} program: teal code
+   * @param {Uint8Array[]} args : external arguments
+   * @param {State} state : current state as input
```
amityadav0, line 56: Please update it to `runtime`

packages/algorand-js/src/runtime/runtime.ts
```diff
@@ -0,0 +1,227 @@
...
+    }
+    return this.assertAccountDefined(account);
+  }
+
+  /**
+   * Description: fetches global state value for key present app's global data
+   * returns undefined otherwise
+   * @param appId: current application id
+   * @param key: key to fetch value of from local state
+   * @param interpreter: interpreter object
```
amityadav0, line 69: please remove this

packages/algorand-js/src/runtime/runtime.ts
```diff
@@ -0,0 +1,227 @@
...
+    return undefined;
+  }
+
+  /**
+   * Description: add new key-value pair or updating pair with existing key in
+   * app's global data for application id: appId, throw error otherwise
+   * @param appId: current application id
+   * @param key: key to fetch value of from local state
+   * @param value: key to fetch value of from local state
+   * @param interpreter: interpreter object
```
amityadav0, line 94: please remove this

packages/algorand-js/test/src/interpreter/opcode-list.ts
```diff
@@ -1980,16 +1990,25 @@ describe("Teal Opcodes", function () {
   describe("StateFul Opcodes", function () {
     const stack = new Stack<StackElem>();
     const interpreter = new Interpreter();
-    interpreter.tx = TXN_OBJ;
+    const runtime = new Runtime(interpreter);
+    interpreter.runtime = runtime;
+
+    interpreter.ctx.tx = TXN_OBJ;
+
+    interpreter.ctx.tx.snd = Buffer.from("addr-1");
```
amityadav0, line 1998: please remove lines in between and add comments after each line

packages/algorand-js/src/lib/stateful.ts
```diff
@@ -0,0 +1,37 @@
+/* eslint sonarjs/no-duplicate-string: 0 */
```
amityadav0, line 1: I don't think we need this line here

packages/algorand-js/src/runtime/runtime.ts
```diff
@@ -0,0 +1,227 @@
...
+import { Interpreter } from "../index";
+import { BIGINT0 } from "../interpreter/opcode-list";
+import { checkIndexBound, compareArray } from "../lib/compare";
+import { convertToString } from "../lib/parsing";
+import { assertValidSchema, getKeyValPair } from "../lib/stateful";
+import type { Context, SdkAccount, StackElem, Txn } from "../types";
+
+export class Runtime {
+  interpreter: Interpreter;
+  // storage: State;
```
amityadav0, line 20: let's remove this

packages/algorand-js/src/runtime/runtime.ts
```diff
@@ -0,0 +1,227 @@
...
+  /**
+   * Description: fetches global state value for key present app's global data
+   * returns undefined otherwise
+   * @param appId: current application id
+   * @param key: key to fetch value of from local state
+   * @param interpreter: interpreter object
+   */
+  getGlobalState (appId: number, key: Uint8Array): StackElem | undefined {
+    const appDelta = this.interpreter.ctx.state.globalApps.get(appId);
+    if (!appDelta) {
```
amityadav0, line 73: can we use `assertAppDefined` here ?

packages/algorand-js/src/runtime/runtime.ts
```diff
@@ -0,0 +1,227 @@
...
+   * Description: add new key-value pair or updating pair with existing key in
+   * app's global data for application id: appId, throw error otherwise
+   * @param appId: current application id
+   * @param key: key to fetch value of from local state
+   * @param value: key to fetch value of from local state
+   * @param interpreter: interpreter object
+   */
+  updateGlobalState (appId: number, key: Uint8Array, value: StackElem): SSCStateSchema[] {
+    const appDelta = this.interpreter.ctx.state.globalApps.get(appId);
+    if (!appDelta) {
```
amityadav0, line 98: can we use `assertAppDefined` here ?

robert-zaremba:
> I wanted to add a better .teal code 

@ratik21  - we will do it in other task, which we discussed yesterday.

packages/algorand-js/src/runtime/runtime.ts
```diff
@@ -0,0 +1,215 @@
...
+      // if not array, then create a single transaction
+      const mockParams = mockSuggestedParams(txnParams.payFlags);
+      const tx = mkTransaction(txnParams, mockParams);
+
+      const encodedTxnObj = tx.get_obj_for_encoding() as Txn;
+      encodedTxnObj.txID = tx.txID();
+      this.store.tx = encodedTxnObj; // assign current txn
+      this.store.gtxs = [this.store.tx]; // assign single txn to grp
+    }
+  }
```
robert-zaremba, line 153: This has to be changed. When I was talking about context, I was specifically mention that we need to protect the state and separate it from transactions.
Context should have:
* copy of store (so when transaction will fail, we will be able to remove context without modifying the store). We need to protect the store. If transaction fails, it MUST NOT modify the store - hence we copy it to the context.
* transactions and all other required data - please don't put transactions in store.

robert-zaremba (changes requested):

packages/algorand-js/src/runtime/runtime.ts
ratik21: Ok, i am separating `[tx, gtxs]` from store.
But i don't understand why. Transaction data is READ ONLY (opcodes only get a txn property and put to stack. eg. `txn Fee`).

packages/algorand-js/src/interpreter/interpreter.ts
```diff
@@ -105,20 +50,17 @@ export class Interpreter {
...
-    logic: Operator[], args: Uint8Array[],
-    accounts: AccountState[]): boolean {
+  async execute (program: string, args: Uint8Array[],
+    runtime: Runtime): Promise<State> {
     assert(Array.isArray(args));
-    this.createTxnContext(txnParams);
-    this.createStatefulContext(accounts);
-    this.instructions = logic;
+    this.runtime = runtime;
+    this.ctx = runtime.store; // set local context of interpreter
```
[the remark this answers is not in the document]
ratik21: updating here.

packages/algorand-js/src/interpreter/interpreter.ts
```diff
@@ -1,91 +1,36 @@
...
   intcblock: BigInt[];
   scratch: StackElem[];
-  tx: Txn;
-  gtxs: Txn[];
-  accounts: Map<string, AccountState>;
-  globalApps: Map<number, SSCParams>;
   instructions: Operator[];
   instructionIndex: number;
   args: Uint8Array[];
+  ctx: Context; // interpreter's 'local' context
```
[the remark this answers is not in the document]
ratik21: yes, but it still needs to assigned as an interpreter's property, right? Otherwise how will the opcodes access the data in `ctx`?
(One option is to pass `ctx` to each `execute()` of opcode)

packages/algorand-js/src/runtime/account.ts
```diff
@@ -0,0 +1,94 @@
+import { SdkAccount, StackElem } from "algorand-js/src/types";
+import type { Account, AccountAssetInfo, AppLocalState, CreatedApps, SSCSchemaConfig } from "algosdk";
+import { generateAccount } from "algosdk";
+
+import { TealError } from "../errors/errors";
+import { ERRORS } from "../errors/errors-list";
+import { compareArray } from "../lib/compare";
+import { assertValidSchema, getKeyValPair } from "../lib/stateful";
+
+export class SdkAccountImpl implements SdkAccount {
```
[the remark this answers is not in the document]
ratik21: ok

packages/algorand-js/src/interpreter/interpreter.ts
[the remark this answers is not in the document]
amityadav0: If we are passing context to execute, i don't think we need to assign interpreter's property because `ctx` itself has all the data.

packages/algorand-js/src/runtime/runtime.ts
```diff
@@ -0,0 +1,215 @@
...
+   * Description: this function executes a transaction based on a smart
+   * contract logic and updates state afterwards
+   * @param txn : Transaction parameters
+   * @param fileName : smart contract file (.teal) name in assets/
+   * @param args : external arguments to smart contract
+   * @param accounts : accounts passed by the user
+   */
+  async executeTx (txnParams: ExecParams | ExecParams[], fileName: string,
+    args: Uint8Array[], accounts: SdkAccount[]): Promise<void> {
+    this.prepareInitialState(txnParams, accounts); // prepare initial state
```
amityadav0: If we have prepared state here, why are we passing whole runtime object to interpreter ?

packages/algorand-js/src/runtime/runtime.ts
```diff
@@ -0,0 +1,215 @@
...
+import { ERRORS } from "../errors/errors-list";
+import { Interpreter } from "../index";
+import { BIGINT0 } from "../interpreter/opcode-list";
+import { checkIndexBound, compareArray } from "../lib/compare";
+import { convertToString } from "../lib/parsing";
+import { assertValidSchema, getKeyValPair } from "../lib/stateful";
+import type { Context, SdkAccount, StackElem, Txn } from "../types";
+
+export class Runtime {
+  interpreter: Interpreter;
```
amityadav0, line 19: we have `interpreter` in `runtime` and 
`runtime` in `interpreter`. I don't think it is good to have cycle. Linear Flow is easy to understand for developers and users.
(just a thought)

packages/algorand-js/src/interpreter/interpreter.ts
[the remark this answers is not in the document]
amityadav0: also if we are passing `runtime` object to `interpreter`, we can access `ctx` using `runtime.ctx`

packages/algorand-js/src/interpreter/interpreter.ts
```diff
@@ -105,20 +50,18 @@ export class Interpreter {
...
+   * Description: this function executes TEAL code after parsing
+   * @param {string} program: teal code
+   * @param {Uint8Array[]} args : external arguments
+   * @param {Runtime} runtime : runtime object
+   * @param {Context} context: interpreter context (state and txns)
    */
-  execute (txnParams: ExecParams | ExecParams[],
-    logic: Operator[], args: Uint8Array[],
-    accounts: AccountState[]): boolean {
+  async execute (program: string, args: Uint8Array[],
```
amityadav0, line 59: I think `args` should be put in context also.

ratik21: `args` are provided by the user. We are keeping the state we prepared in the `ctx` (eg, `encoded tx, gtxs`, `global assets, apps..`). Is it okay to put `args` inside ctx?

packages/algorand-js/src/interpreter/interpreter.ts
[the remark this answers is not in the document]
ratik21: @amityadav0 the idea here is that interperter will use a 'copy' of store. Interpreter shouldn't use `runtime.store` directly. `ctx` is kept inside interpreter. This is because if the transaction fails, then we ignore the changes in the interpreter's local ctx, and the store remains unchanged.

packages/algorand-js/src/runtime/runtime.ts
ratik21: We need to access state functions as well, rather than just state data (eg. `runtime.getGlobalState`, `updateGlobal..`)

packages/algorand-js/src/runtime/runtime.ts
ratik21: I agree. Interpreter needs runtime to access runtime state functions, and Runtime should take interpreter (as we are executing .teal code). 
One idea is to represent `interpreter` as a child class of `Runtime` and interpreter can access runtime functions using `super`.
Did you have something in mind?

packages/algorand-js/src/interpreter/interpreter.ts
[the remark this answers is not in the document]
ratik21: > also if we are passing runtime object to interpreter, we can access ctx using runtime.ctx

packages/algorand-js/src/interpreter/interpreter.ts
amityadav0: I think so, because args are also runtime context, specific to each transaction.

packages/algorand-js/src/runtime/runtime.ts
amityadav0: One idea that  i have we can put these functions in `runtime.ts` but out of runtime class, because these functions are only using `this.interpreter.ctx`. we can export them and use them directly from interpreter. But I am not sure if this is a good way

amityadav0: another way is to remove `interpreter` from runtime, create a new object in executeTx function only.
Update globalState and other functions to work from `ctx` passed(as i can see they require this.interpreter.ctx).

ratik21: > One idea that i have we can put these functions in runtime.ts but out of runtime class, because these functions are only using this.interpreter.ctx. we can export them and use them directly from interpreter. But I am not sure if this is a good way

Agreed, not a good way.

ratik21: > another way is to remove `interpreter` from runtime, create a new object in executeTx function only.
Update globalState and other functions to work from `ctx` passed(as i can see they require this.interpreter.ctx only).

Yes, i think we can:
- remove interpreter from runtime (constructor), and declare it directly in `runtime.executeTx()`. Or
- Make `interpreter` class child of `Runtime`, and interpreter can access runtime functions using `this.super...`.

But in both approaches, it means we are using `interpreter` internally and user is only exposed to `Runtime`. @robert-zaremba  what do you think?

packages/algorand-js/src/interpreter/interpreter.ts
ratik21: Ok, moving `args` to ctx.

packages/algorand-js/src/runtime/runtime.ts
```diff
@@ -0,0 +1,215 @@
...
+    return globalDelta;
+  }
+
+  getAccount (accountIndex: bigint): SDKAccount {
+    let account: SDKAccount | undefined;
+    if (accountIndex === BIGINT0) {
+      const senderAccount = convertToString(this.interpreter.ctx.tx.snd);
+      account = this.interpreter.ctx.state.accounts.get(senderAccount);
+    } else {
+      accountIndex--;
```
robert-zaremba, line 51: let's create a new variable instead:
```
const accountIndex = accountIndex - 1;
```

This will prevent bad bugs when we use `accountIndex` in other places as well.

packages/algorand-js/src/runtime/runtime.ts
```diff
@@ -0,0 +1,215 @@
...
+  }
+
+  /**
+   * Description: fetches global state value for key present app's global data
+   * returns undefined otherwise
+   * @param appId: current application id
+   * @param key: key to fetch value of from local state
+   */
+  getGlobalState (appId: number, key: Uint8Array): StackElem | undefined {
+    const appDelta = this.assertAppDefined(appId);
```
robert-zaremba, line 66: why do we call it `appDelta`?

robert-zaremba: let's rename `appDelta` to `app`.

packages/algorand-js/src/runtime/runtime.ts
```diff
@@ -0,0 +1,215 @@
...
+
+  assertAccountDefined (a?: SDKAccount): SDKAccount {
+    if (a === undefined) {
+      throw new TealError(ERRORS.TEAL.ACCOUNT_DOES_NOT_EXIST);
+    }
+    return a;
+  }
+
+  assertAppDefined (appId: number): SSCParams {
+    const globalDelta = this.interpreter.ctx.state.globalApps.get(appId);
```
robert-zaremba, line 38: let's rename `globalDelta` to `appSatate`.

packages/algorand-js/src/runtime/runtime.ts
```diff
@@ -0,0 +1,215 @@
...
+  /**
+   * Description: add new key-value pair or updating pair with existing key in
+   * app's global data for application id: appId, throw error otherwise
+   * @param appId: current application id
+   * @param key: key to fetch value of from local state
+   * @param value: key to fetch value of from local state
+   */
+  updateGlobalState (appId: number, key: Uint8Array, value: StackElem): SSCStateSchema[] {
+    const appDelta = this.assertAppDefined(appId);
+    const globalState = appDelta["global-state"];
```
robert-zaremba, line 86: as above, let's rename to: app, and appGlobalState.

packages/algorand-js/src/runtime/runtime.ts
```diff
@@ -0,0 +1,215 @@
...
+
+    assertValidSchema(appDelta["global-state"], appDelta["global-state-schema"]); // verify if updated schema is valid by config
+    return globalState;
+  }
+
+  /**
+   * Description: set accounts for context as {address: SDKAccount}
+   * @param accounts: array of account info's
+   */
+  createStatefulContext (accounts: SDKAccount[]): void {
```
robert-zaremba, line 105: Please rename to `initializeAccounts`

packages/algorand-js/src/runtime/runtime.ts
```diff
@@ -0,0 +1,215 @@
...
+      globalState[idx].value = data.value; // update value if key found
+    }
+    appDelta["global-state"] = globalState; // save updated state
+
+    assertValidSchema(appDelta["global-state"], appDelta["global-state-schema"]); // verify if updated schema is valid by config
+    return globalState;
+  }
+
+  /**
+   * Description: set accounts for context as {address: SDKAccount}
```
robert-zaremba, line 102:
```suggestion
   * Setup initial accounts as {address: SDKAccount}. This should be called only when initializing Runtime.
```

packages/algorand-js/src/runtime/runtime.ts
robert-zaremba: This assumes that we will use this function only for test setup, not for transaction context preparation.

packages/algorand-js/src/runtime/runtime.ts
robert-zaremba: We don't put transactions is state. If we need to keep the transaction data for whatever reason (but not in store) then it's better to copy it. Otherwise, we don't need to copy and we can pass it to the context.

packages/algorand-js/src/runtime/runtime.ts
```diff
@@ -0,0 +1,215 @@
...
+      const mockParams = mockSuggestedParams(txnParams.payFlags);
+      const tx = mkTransaction(txnParams, mockParams);
+
+      const encodedTxnObj = tx.get_obj_for_encoding() as Txn;
+      encodedTxnObj.txID = tx.txID();
+      return [encodedTxnObj, [encodedTxnObj]];
+    }
+  }
+
+  prepareInitialState (txnParams: ExecParams| ExecParams[], accounts: SDKAccount[]): Context {
```
robert-zaremba, line 150: Now I see what you want to do. It's not Runtime setup, it's Tx context creation.
1. The way you create copy is wrong
2. Please rename this function, eg: `createTxContext`

packages/algorand-js/src/runtime/runtime.ts
```diff
@@ -0,0 +1,215 @@
...
+   * Description: this function executes a transaction based on a smart
+   * contract logic and updates state afterwards
+   * @param txn : Transaction parameters
+   * @param fileName : smart contract file (.teal) name in assets/
+   * @param args : external arguments to smart contract
+   * @param accounts : accounts passed by the user
+   */
+  async executeTx (txnParams: ExecParams | ExecParams[], fileName: string,
+    args: Uint8Array[], accounts: SDKAccount[]): Promise<void> {
+    const context = this.prepareInitialState(txnParams, accounts); // prepare initial state
```
robert-zaremba, line 208: let's refactore the functions above, and make more clear function names.

packages/algorand-js/src/runtime/runtime.ts
```diff
@@ -0,0 +1,215 @@
...
+
+    const program = getProgram(fileName);
+    const updatedState = await this.interpreter.execute(program, args, this, context);
+    this.store = updatedState; // update state after successful execution('local-state', 'global-state'..)
```
robert-zaremba, line 212: if we will keep the `ctx` only in runtime, then here we will need to do:
```
this.store = this.ctx.store
```
We don't need to do a full deep copy here.

packages/algorand-js/src/types.ts
```diff
@@ -70,3 +73,36 @@ export enum EncodingType {
...
+// describes interpreter's local context (state + txns)
+export interface Context {
+  state: State
+  tx: Txn // current txn
+  gtxs: Txn[] // all transactions
+}
+
+// represent account used in tests and by the context
+// NOTE: custom notations are used rather than SDK AccountState notations
+export interface SDKAccount {
```
robert-zaremba, line 95: How about renaming `SDKAccount` to `StoreAccount`? This is to avoid a name clash with the `algosdk.Account`

robert-zaremba (changes requested):

packages/algorand-js/src/runtime/runtime.ts
ratik21: Ok, thanks

packages/algorand-js/src/runtime/runtime.ts
ratik21: ok

packages/algorand-js/src/runtime/runtime.ts
ratik21: yes, i have removed this function.

packages/algorand-js/src/types.ts
ratik21: Good point. Thanks

packages/algorand-js/src/runtime/runtime.ts
ratik21: I have removed this function.

packages/algorand-js/src/runtime/runtime.ts
ratik21: ok

packages/algorand-js/src/runtime/runtime.ts
ratik21: done

packages/algorand-js/src/runtime/runtime.ts
ratik21: Renamed to `app` (inline with previous declarations)

packages/algorand-js/src/runtime/runtime.ts
ratik21: ok

packages/algorand-js/src/runtime/runtime.ts
ratik21: Ok, thanks

packages/algorand-js/src/runtime/runtime.ts
```diff
@@ -0,0 +1,226 @@
...
+      gtxs: gtxs,
+      args: args
+    };
+
+    const program = getProgram(fileName); // get TEAL code as string
+    const interpreter = new Interpreter();
+    await interpreter.execute(program, this);
+
+    this.store = this.ctx.state; // update state after successful execution('local-state', 'global-state'..)
+    this.prepareFinalState(txnParams, this.accounts); // update account balances
```
robert-zaremba, line 224: Why do we have accounts both in `this.accounts` and `this.store.accounts`? Please update if this is a mistake. Probably we should only have `this.store.accounts`

packages/algorand-js/src/types.ts
```diff
@@ -70,3 +73,37 @@ export enum EncodingType {
...
+  appsLocalState: AppLocalState[]
+  appsTotalSchema: SSCSchemaConfig
+  createdApps: CreatedApps[]
+  createdAssets: CreatedAssets[]
+  account?: Account
+
+  balance: () => number
+  getLocalState: (appId: number, key: Uint8Array) => StackElem | undefined
+  updateLocalState: (appId: number, key: Uint8Array, value: StackElem) => AppLocalState[]
+}
```
robert-zaremba, line 109: `algorand-js` shouldn't have a dependency to algob.
The dependency graph is that algob import algorand-js, not other way around.
So, for functional tests, where we create an algob project, we should use the algob package tests  (with a fixture project).
For simple runtime testing (this is what you are doing in the tests you created) we don't need a fixture project. You just set some accounts and that's it.

robert-zaremba (approved):
Pre approving. Please solve all comments before merging.

packages/algorand-js/src/runtime/runtime.ts
ratik21: `this.accounts` was kept to update the accounts passed by the user (in runtime), along with accountMap. So that it's easier for the end user to check the updates in original account (rather then getting from map).
Reverting.

packages/algorand-js/package.json
```diff
@@ -58,10 +58,12 @@
     "sinon": "^9.0.2"
   },
   "dependencies": {
+    "@types/lodash": "^4.14.165",
     "algosdk": "^1.8.0",
     "glob": "^7.1.6",
     "js-sha256": "^0.9.0",
     "js-sha512": "^0.8.0",
+    "loadash": "^1.0.0",
```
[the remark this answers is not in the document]
ratik21: ok

packages/algorand-js/src/types.ts
ratik21: removed dependency

packages/algorand-js/src/runtime/runtime.ts
amityadav0: was this resolved @ratik21 ?

robert-zaremba: Yes, Ratik moved interpreter creation to `runtime.executeTx`: https://github.com/scale-it/algorand-builder/pull/169/files#diff-614a0db207a77ce83f6fd3b00f9c0e43c40ec8f7d010aa2b15cbfe5d9439c438

PR #171, opened 2020-12-17
Update examples and Add docs for checkpoints

amityadav0:
- Store empty checkpoints for scripts:
    - ran in deploy mode
    - doesn't have properties that are stored in checkpoints. Ex: DeploySSC, delpoyASA, delegatedLsig
  
- Current structure only creates checkpoint if there exist one of  the property in script that require checkpoint storage


robert-zaremba:
The idea is that the deployer should store checkpoints whenever it's saving a smart-contract or ASA, not empty checkpoints. User should be able to append a checkpoint as well if needed.
This is the logic:
+ Whenever we deploy something through deployer (ASA, smart contract etc...) we store  a checkpoint
+ User should be able to store his own checkpoints as well. For example, if we have a deployment script which fund accounts (so only Algo transfer, nothing else), and we want to make sure that it won't be called twice, then a user must create his own checkpoint.

robert-zaremba:
Let me know if this is clear. Could you update docs as well?

amityadav0:
understood

robert-zaremba (approved):

PR #172, opened 2020-12-17
Rework tx builder

robert-zaremba:
Updated transaction creation and parameters packing.

### DIFF
--- a/packages/algob/src/lib/tx.ts
+++ b/packages/algob/src/lib/tx.ts
@@ -1,4 +1,4 @@
-import tx, { decodeSignedTransaction, SuggestedParams, Transaction } from "algosdk";
+import algosdk, { Algodv2, SuggestedParams, Transaction } from "algosdk";
 import { TextEncoder } from "util";
 
 import { BuilderError } from "../internal/core/errors";
@@ -7,7 +7,7 @@ import {
   AlgobDeployer,
   ASADef,
   ASADeploymentFlags,
-  execParams,
+  ExecParams,
   SignType,
   TransactionType,
   TxParams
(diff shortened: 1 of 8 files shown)


packages/algob/src/lib/tx.ts
```diff
@@ -189,49 +192,43 @@ function signTransaction (txn: Transaction, txnParam: execParams): Uint8Array {
...
+    mkTransaction(p,
+      await mkTxParams(deployer.algodClient, p.payFlags, Object.assign({}, suggestedParams)));
+
+  let signedTxn;
+  if (Array.isArray(execParams)) {
+    if (execParams.length > 16) { throw new Error("Maximum size of an atomic transfer group is 16"); }
+
+    let txns = [];
+    for (const t of execParams) { txns.push(await mkTx(t)); }
+    txns = algosdk.assignGroupID(txns);
```
robert-zaremba, line 218: here, it's safer to assign the result of `assignGroupID`.

packages/algob/src/lib/tx.ts
```diff
@@ -189,49 +192,43 @@ function signTransaction (txn: Transaction, txnParam: execParams): Uint8Array {
...
+    signedTxn = txns.map((txn: Transaction, index: number) => {
```
amityadav0, line 219: I am curious about this. does this creates a map and stores it in signedTxn ? How is it different than using `forEach` ?

amityadav0 (approved):
LGTM
tested atomic transaction example also.

packages/algob/src/lib/tx.ts
robert-zaremba: `map` returns a list. `forEach` only iterates
explanation of map:
```
let L = some_list;
L.map(f) == [f(L[0]), f(L[1]), .... f(L[n])]
```

PR #173, opened 2020-12-21
Update documentation of opcodes

amityadav0:
- Update documentation of opcodes

### DIFF
--- a/packages/algorand-js/src/interpreter/opcode-list.ts
+++ b/packages/algorand-js/src/interpreter/opcode-list.ts
@@ -1231,7 +1231,7 @@ export class BranchIfZero extends Op {
 
   /**
    * Description: Sets `label` according to arguments passed.
-   * @param args Expected arguments: [label(branch name)]
+   * @param args Expected arguments: [label of branch]
    * @param line line number in TEAL file
    * @param interpreter interpreter object
    */
@@ -1259,7 +1259,7 @@ export class BranchIfNotZero extends Op {
 
   /**
    * Description: Sets `label` according to arguments passed.
-   * @param args Expected arguments: [label(branch name)]
+   * @param args Expected arguments: [label of branch]
    * @param line line number in TEAL file
    * @param interpreter interpreter object
    */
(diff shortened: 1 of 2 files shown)


packages/algorand-js/src/interpreter/opcode-list.ts
```diff
@@ -153,10 +164,12 @@ export class Mul extends Op {
 
 // pushes argument[N] from argument array to stack
 export class Arg extends Op {
-  readonly _arg;
+  readonly _arg: Uint8Array;
 
   /**
-   * @param args words list extracted from line: {argument number}
+   * Description: Gets the argument value from interpreter.args array.
+   * store the vale in _arg variable
```
ratik21, line 171:
```suggestion
   * store the value in _arg variable
```

packages/algorand-js/src/interpreter/opcode-list.ts
```diff
@@ -183,7 +196,8 @@ export class Bytecblock extends Op {
   readonly interpreter: Interpreter;
 
   /**
-   * @param args words list extracted from line: {byte block}
+   * Description: Store blocks of bytes in bytecblock
+   * @param args Expected arguments: [byte block]
```
ratik21, line 200: What does `[byte block]` mean? Expected arguments is an array of bytes. (eg. `bytecblock "a" "b" "c"`)

packages/algorand-js/src/interpreter/opcode-list.ts
```diff
@@ -1144,7 +1205,8 @@ export class Branch extends Op {
   readonly interpreter: Interpreter;
 
   /**
-   * @param args words list extracted from line: {branch name}
+   * Description: Sets `label` according to arguments passed.
+   * @param args Expected arguments: [label(branch name)]
```
ratik21, line 1209:
```suggestion
   * @param args Expected arguments: [label of branch]
```

packages/algorand-js/src/interpreter/opcode-list.ts
amityadav0: hmm, just checked TEAL opcode documentation `intcblock` and `bytecblock` are wrongly implementated.

amityadav0: ** Discussion 
- On call discussed and checked both the opcodes, docs mentioned that we require a length for both, but when running teal codes, that's not the case.

ratik21 (approved):

PR #175, opened 2020-12-22
PoC for stateless TEAL processing

ratik21:

PR #177, opened 2020-12-23
PoC for stateful TEAL processing

ratik21:
Updated `Runtime` to support stateful txns (application creation + Opt-In)
Added a simple stateful smart contract which updates global and local "counter" by 1 on each application call.
Fixed some functions.


packages/algorand-js/test/integration/escrow-account.ts
```diff
@@ -0,0 +1,89 @@
...
+
+import { ERRORS } from "../../src/errors/errors-list";
+import { Runtime } from "../../src/index";
+import { StoreAccountImpl } from "../../src/runtime/account";
+import { getAcc } from "../helpers/account";
+import { expectTealErrorAsync } from "../helpers/errors";
+import { johnAccount } from "../mocks/account";
+
+describe("Algorand Stateless Smart Contracts", function () {
+  const escrow = new StoreAccountImpl(1000000000); // 1000 ALGO
```
robert-zaremba, line 12: Use variables for numbers:
```
const initialAlgoHolding = 100e6
```
and use it in tests below.  In checks you can use: `assert.equal(..., initialAlgoHolding - 100)`

packages/algorand-js/test/integration/escrow-account.ts
```diff
@@ -0,0 +1,89 @@
...
+    invalidParams.payFlags = { totalFee: 12000 };
+
+    // execute transaction (should fail as fee is 12000)
+    await expectTealErrorAsync(
+      async () => await runtime.executeTx(invalidParams, 'escrow.teal', []),
+      ERRORS.TEAL.INVALID_STACK_ELEM
+    );
+  });
+
+  it("should reject transaction type is not `pay`", async function () {
```
robert-zaremba, line 64:
```suggestion
  it("should reject transaction if type is not `pay`", async function () {
```

packages/algorand-js/test/integration/escrow-account.ts
```diff
@@ -0,0 +1,89 @@
...
+    const invalidParams = {
+      ...txnParams,
+      type: 1, // asset transfer
```
robert-zaremba, line 67: You don't need to use `1`. We should have defined (or it's in SDK) the constants for types.

packages/algorand-js/test/mocks/txn.ts
```diff
@@ -9,6 +9,9 @@ const GENESIS_HASH = 'SGO1GKSzyE7IEPItTxCByw9x8FmnrCDexi9/cOUJOiI=';
 const account = generateAccount();
 const addr = decodeAddress(account.addr);
 
+export const addr1 = 'WHVQXVVCQAD7WX3HHFKNVUL3MOANX3BYXXMEEJEJWOZNRXJNTN7LTNPSTY';
+export const addr2 = '2UBZKFR6RCZL7R24ZG327VKPTPJUPFM6WTG7PJG2ZJLU234F5RGXFLTAKA';
```
robert-zaremba, line 13: please use `joshnAccount.address` here. 

If we copy things in other places, then update them as well.

robert-zaremba (changes requested):

amityadav0:
is this PR extension of stateless ? should we merge that first or we can merge this before ?

ratik21:
> is this PR extension of stateless ? should we merge that first or we can merge this before ?

If you look at the code, yes. This PR had some dependency so i took a pull from the previous one. We should merge #175 before this

packages/algorand-js/src/runtime/account.ts
```diff
@@ -101,4 +102,45 @@ export class StoreAccountImpl implements StoreAccount {
...
+      if (this.appsLocalState.length === 10) {
+        throw new Error('Maximum Opt In applications per account is 10');
+      }
+
+      const localParams: AppLocalState = {
+        id: appId,
+        "key-value": [],
+        schema: appParams["local-state-schema"]
+      };
+      this.appsLocalState.push(localParams); // push
```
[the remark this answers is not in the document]
amityadav0: > should we store in a map?
> 
> ```
> appLocalState: Map<app_id, AppLocalState>
> ```

I have a suggestion, extending to this comment, i think we should change the structure of Store, instead of storing them as in `algosdk` we can have them according to our requirements because we won't be interacting with algorand.

- For ex: In assets, we can directly use Map structure directly, instead of having, assets: {asset-id, assetbalance, ....}
we can directly store assets: {asset-id -> data}, 
this way we won't have to copy the data,(updating will be easier and faster too)
- in some places we are using traversing whole objects to find the value, it will be faster to use map instead.
what do you think ?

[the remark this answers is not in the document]
ratik21: > in some places we are traversing whole objects to find the value, it will be faster to use map instead.
what do you think ?
In `algorand` we only have max 10 applications per account. So using a map in that case seemed excessive.

I am also thinking to use maps instead. This way, i think we don't need to limit the app in account to 10. It will be more easier for end user also (getting data directly)

packages/algorand-js/src/types.ts
```diff
@@ -89,9 +90,12 @@ export interface StoreAccount {
   appsTotalSchema: SSCSchemaConfig
   createdApps: CreatedApps[]
   createdAssets: CreatedAssets[]
-  account?: Account
+  account: Account
+  accountsAssets: Map<string, Map<number, AssetHolding>>
 
   balance: () => number
+  createApp: (params: SSCDeploymentFlags) => CreatedApps
```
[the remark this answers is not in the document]
ratik21: ok

packages/algorand-js/test/integration/escrow-account.ts
ratik21: yes

packages/algorand-js/src/runtime/account.ts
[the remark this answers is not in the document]
amityadav0: yes, but each account may have long list of `appLocal-state`. as we will be in test environment multiple queries for the same application may occur. in terms of implementation and understanding complexity , it is easy to implement and understand so i don't think it will be a overkill.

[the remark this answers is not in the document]
ratik21: > yes, but each account may have long list of appLocal-state.

No, each account in `algorand` has also maximum 10 applications. So if we are limiting to 10, then simple search should be sufficient enough. 
But i don't think we should limit to 10 here in tests. Same for assets.
image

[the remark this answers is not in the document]
amityadav0: i meant, what if local state has storage variable `counter`, `counter1`, ........
all these variables have some value.  
For ex: 
```
export LOCAL_BYTESLICES=0
export LOCAL_INTS=10
```
state storage.
this means LOCAL_INTS has 10 int state storage for this application.

we can search directly if you want.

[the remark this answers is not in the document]
ratik21: > this means LOCAL_INTS has 10 int state storage for this application.

Yes, in local schema we can have 16 max key-value pairs. These are also stored as array.
> we can search directly if you want.

I think using map on the top level will be better with (maybe) increasing the limit.

packages/algorand-js/test/helpers/account.ts
```diff
@@ -0,0 +1,12 @@
+import { Runtime } from "../../src";
+import { StoreAccount } from "../../src/types";
+
+// returns account from runtime store (updated state)
+export function getAcc (runtime: Runtime, acc: StoreAccount): StoreAccount {
+  const account = runtime.ctx.state.accounts.get(acc.address);
+  if (account === undefined) {
```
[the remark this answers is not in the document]
ratik21: Yes, added assert

packages/algorand-js/test/integration/escrow-account.ts
ratik21: ok, thanks

packages/algorand-js/test/mocks/txn.ts
ratik21: ok

packages/algorand-js/src/runtime/account.ts
```diff
@@ -101,4 +102,45 @@ export class StoreAccountImpl implements StoreAccount {
       appId: appId
     });
   }
+
+  createApp (params: SSCDeploymentFlags): CreatedApps {
+    if (this.createdApps.length === 10) {
+      throw new Error('Maximum created applications for an account is 10');
+    }
+
+    const appId = Math.floor(1000 + Math.random() * 9000);
```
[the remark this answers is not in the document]
ratik21: Okay

packages/algorand-js/test/integration/stateful-counter.ts
```diff
@@ -0,0 +1,67 @@
...
+import { Runtime } from "../../src/index";
+import { toBytes } from "../../src/lib/parsing";
+import { StoreAccountImpl } from "../../src/runtime/account";
+import { getAcc } from "../helpers/account";
+
+describe("Algorand Smart Contracts - Stateful Counter example", function () {
+  const john = new StoreAccountImpl(1000);
+
+  const txnParams = {
+    type: 2, // callNoOpSSC
```
amityadav0, line 12: please update here as well

packages/algorand-js/test/mocks/account.ts
```diff
@@ -0,0 +1,11 @@
+export const johnAccount = {
+  addr: '2UBZKFR6RCZL7R24ZG327VKPTPJUPFM6WTG7PJG2ZJLU234F5RGXFLTAKA',
```
amityadav0, line 2: is it possible to use addresses from `config` files ? maybe using mnenomic phrase ?

packages/algorand-js/test/mocks/txn.ts
```diff
@@ -53,7 +56,7 @@ export const TXN_OBJ = {
   apap: Buffer.from("approval"),
   apsu: Buffer.from("clear"),
   apaa: [Buffer.from("arg1"), Buffer.from("arg2")],
-  apat: [Buffer.from("addr-2"), Buffer.from("addr-2")],
+  apat: [Buffer.from(decodeAddress(addr2).publicKey), Buffer.from(decodeAddress(addr2).publicKey)],
```
amityadav0, line 59: can we update second address to something else ? it will be good for testing.

packages/algorand-js/src/runtime/account.ts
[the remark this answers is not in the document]
ratik21: Update: Adding to TODO, and will do this in a new tasks. Some other structures needs to be updated as well.

packages/algorand-js/test/mocks/account.ts
ratik21: I think this is fine. In algob also we use mock accounts as `{ addr: .., sk: ..}`, as it's type is `AccountSDK`. Did you mean something else?

packages/algorand-js/test/mocks/txn.ts
ratik21: Okay, i will add one more account.

packages/algorand-js/src/runtime/runtime.ts
```diff
@@ -27,6 +26,7 @@ export class Runtime {
    */
   private store: State;
   ctx: Context;
+  private counter: number;
```
robert-zaremba, line 29: please rename to `appCounter`

packages/algorand-js/src/runtime/runtime.ts
```diff
@@ -184,6 +185,22 @@ export class Runtime {
...
+    const sender = params.sender;
+    const senderAccountDef = this.assertAccountDefined(this.store.accounts.get(sender.addr));
+    const app = senderAccountDef.createApp(++this.counter, params);
+
+    this.store.globalApps.set(app.id, app.params); // update globalApps Map
+    return app.id;
+  }
+
+  optInToApp (appId: number, accountAddr: string): void {
+    const appParams = this.store.globalApps.get(appId);
```
robert-zaremba, line 199: you need to check if the app exists, and fail / throw an exception if it doesn't.

packages/algorand-js/src/runtime/runtime.ts
```diff
@@ -194,7 +211,7 @@ export class Runtime {
             break;
           }
           case txnParam.toAccountAddr: {
-            account.amount += txnParam.amountMicroAlgos; // add 'x' algo to receiver
+            account.amount += txnParam.amountMicroAlgos as number; // add 'x' algo to receiver
```
robert-zaremba, line 214: We should use big numbers here, because we need to support uint64.

packages/algorand-js/test/integration/stateful-counter.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    // execute transaction
+    await runtime.executeTx(txnParams, 'counter-approval.teal', []);
+
+    // after execution the counters should be updated by +1
+    const newGlobalCounter = runtime.getGlobalState(txnParams.appId, toBytes(key));
+    const newLocalCounter = getAcc(runtime, john).getLocalState(txnParams.appId, toBytes(key));
+
+    assert.equal(newGlobalCounter, globalCounter + BigInt('1'));
+    assert.equal(newLocalCounter, localCounter + BigInt('1'));
```
robert-zaremba, line 66: You can write `1n` instead of `BigInt(1)` or `BigInt('1')`.

packages/algorand-js/test/mocks/txn.ts
```diff
@@ -9,6 +11,9 @@ const GENESIS_HASH = 'SGO1GKSzyE7IEPItTxCByw9x8FmnrCDexi9/cOUJOiI=';
 const account = generateAccount();
 const addr = decodeAddress(account.addr);
 
+export const addr1 = elonMuskAccount.addr;
+export const addr2 = johnAccount.addr;
```
robert-zaremba, line 15: please rename to elonAddr and johnAddr

packages/algorand-js/src/runtime/runtime.ts
ratik21: ok

packages/algorand-js/src/runtime/runtime.ts
ratik21: added

packages/algorand-js/src/runtime/runtime.ts
ratik21: Sorry, i don't follow. This amount will be passed by the user via `ExecParams` and not TEAL code. So it's maximum value should be javascript max number. (`Amount must be a positive number and smaller than 2^53-1`)

packages/algorand-js/test/integration/stateful-counter.ts
ratik21: ok, thanks

ratik21: I am having this error: `BigInt literals are not available when targeting lower than ES2020`.
For now i have updated `BigInt('1')` to `BIGINT1`. Let me know if i need to update furthur.

packages/algorand-js/src/runtime/runtime.ts
robert-zaremba: So the argument type should be `number | BigInt`
```
amountMicroAlgos: number | BigInt
```

packages/algorand-js/test/integration/stateful-counter.ts
robert-zaremba: OK, I'm using `1n` in Node console and it works. But compiler targets es2017.

packages/algorand-js/src/runtime/runtime.ts
ratik21: Alright, i'll update the `ExecParams` type. But i still don't see user passing `BigInt(..)` as transfer amount :/

packages/algorand-js/src/runtime/account.ts
```diff
@@ -60,7 +61,7 @@ export class StoreAccountImpl implements StoreAccount {
       const keyValue = data.find(schema => compareArray(schema.key, key));
       const value = keyValue?.value;
       if (value) {
-        return value?.bytes || BigInt(value?.uint);
+        return value.type === 1 ? value.bytes : BigInt(value.uint);
```
robert-zaremba, line 64: please don't use `1`. Use a const named value instead.

packages/algorand-js/src/runtime/account.ts
```diff
@@ -98,4 +99,44 @@ export class StoreAccountImpl implements StoreAccount {
       appId: appId
     });
   }
+
+  createApp (appId: number, params: SSCDeploymentFlags): CreatedApp {
+    if (this.createdApps.length === 10) {
+      throw new Error('Maximum created applications for an account is 10');
+    }
+
+    const appParams: SSCParams = {
```
robert-zaremba, line 108: Why do we call it `SSCParams`? How about callling it `SSCAttributes`?

packages/algorand-js/src/runtime/account.ts
```diff
@@ -98,4 +99,44 @@ export class StoreAccountImpl implements StoreAccount {
...
+      'approval-program': '',
+      'clear-state-program': '',
+      creator: params.sender.addr,
+      'global-state': [],
+      'global-state-schema': { 'num-byte-slice': params.globalBytes, 'num-uint': params.globalInts },
+      'local-state-schema': { 'num-byte-slice': params.localBytes, 'num-uint': params.localInts }
+    };
+    // create new app in sender's account
+    const newApp: CreatedApp = { id: appId, params: appParams };
```
robert-zaremba, line 117: Let's rename this to `SSC` or `App` and make it a class. This function should be a class constructor, and here, instead of `account.createApp` we should have `account.addApp`

packages/algorand-js/src/runtime/account.ts
```diff
@@ -98,4 +99,44 @@ export class StoreAccountImpl implements StoreAccount {
...
+    this.createdApps.push(newApp); // push newly created app
+
+    console.log('Created new app with id:', appId);
+    return newApp;
+  }
+
+  // opt in to application
+  optInToApp (appId: number, appParams: SSCParams): void {
+    const localState = this.appsLocalState.find(cfg => cfg.id === appId);
```
robert-zaremba, line 126:
```suggestion
    const localState = this.appsLocalState.find(app => app.id === appId);
```

packages/algorand-js/src/runtime/runtime.ts
```diff
@@ -184,6 +185,26 @@ export class Runtime {
     }
   }
 
+  // creates new application and returns application id
+  createApp (params: SSCDeploymentFlags): number {
+    const sender = params.sender;
+    const senderAccountDef = this.assertAccountDefined(this.store.accounts.get(sender.addr));
+    const app = senderAccountDef.createApp(++this.appCounter, params);
```
robert-zaremba, line 192:
```suggestion
    const senderAcc = this.assertAccountDefined(this.store.accounts.get(sender.addr));
    const app = senderAcc.createApp(++this.appCounter, params);
```

robert-zaremba (approved):
Preapproving, but left few comments.

packages/algorand-js/src/runtime/account.ts
ratik21: ok

packages/algorand-js/src/runtime/account.ts
ratik21: renamed

packages/algorand-js/src/runtime/account.ts
ratik21: Done. Thanks for suggestion

PR #181, opened 2021-01-05
Update pseudo-op int

ratik21:
Updated `int` to accept txnOnComplete enum.
added test

NOTE:
+ Other than `int`, `txn OnCompletion` is also using oncomplete enum, but this is already handled during encoding the txn details. (by `tx.get_obj_for_encoding()`)

### DIFF
--- a/packages/algorand-js/src/interpreter/opcode-list.ts
+++ b/packages/algorand-js/src/interpreter/opcode-list.ts
@@ -1893,7 +1893,7 @@ export class Int extends Op {
 
     let uint64;
     const txOnComplete = TxnOnComplete[args[0] as keyof typeof TxnOnComplete]; // eg. TxnOnComplete['NoOp']
-    if (txOnComplete !== undefined && typeof txOnComplete === 'number') { // check if string is keyof TxnOnComplete Enum
+    if (txOnComplete !== undefined) { // check if string is keyof TxnOnComplete Enum
       uint64 = BigInt(txOnComplete);
     } else {
       assertOnlyDigits(args[0]);
--- a/packages/algorand-js/src/interpreter/opcode-list.ts
+++ b/packages/algorand-js/src/interpreter/opcode-list.ts
@@ -1892,9 +1892,9 @@ export class Int extends Op {
     assertLen(args.length, 1, line);
 
     let uint64;
-    if (Object.values(TxnOnComplete).includes(args[0])) { // check if string is keyof TxnOnComplete Enum
-      uint64 = TxnOnComplete[args[0] as keyof typeof TxnOnComplete]; // eg. TxnOnComplete['NoOp']
-      uint64 = BigInt(uint64);
+    const txOnComplete = TxnOnComplete[args[0] as keyof typeof TxnOnComplete]; // eg. TxnOnComplete['NoOp']
+    if (txOnComplete !== undefined && typeof txOnComplete === 'number') { // check if string is keyof TxnOnComplete Enum
+      uint64 = BigInt(txOnComplete);
     } else {
       assertOnlyDigits(args[0]);
       uint64 = BigInt(args[0]);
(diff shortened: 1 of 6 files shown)


packages/algorand-js/src/interpreter/opcode-list.ts
```diff
@@ -1945,3 +1875,83 @@ export class GetAssetDef extends Op {
...
+   * Description: Sets uint64 variable according to arguments passed.
+   * @param args Expected arguments: [number]
+   * @param line line number in TEAL file
+   */
+  constructor (args: string[], line: number) {
+    super();
+    assertLen(args.length, 1, line);
+
+    let uint64;
+    if (Object.values(TxnOnComplete).includes(args[0])) { // check if string is keyof TxnOnComplete Enum
```
robert-zaremba, line 1895: Why can't we use:
```
let x = TxnOnComplete[args[0] as keyof typeof TxnOnComplete];
if (x !== undefined) ...
```

In your version we are creating a new list each time we call a constructor, and iterating over the list.

packages/algorand-js/src/types.ts
```diff
@@ -95,3 +95,13 @@ export interface StoreAccount {
...
 }
+
+// https://developer.algorand.org/docs/reference/teal/specification/#oncomplete
+export enum TxnOnComplete {
+  NoOp,
+  OptIn,
+  CloseOut,
+  ClearState,
+  UpdateApplication,
+  DeleteApplication
```
amityadav0, line 106: this `enum` will result in reverse mapping, (1 -> NoOp and NoOp -> 1), to solve this we can use `string` enum.
Ex: NoOp = '0'. 

https://www.typescriptlang.org/docs/handbook/enums.html

packages/algorand-js/src/interpreter/opcode-list.ts
ratik21: Thanks for suggestion. I was doing this way because the above code will work for `0` as well. (which we don't want, we want `NoOp`). 
I have updated using types only:
```
let x = TxnOnComplete[args[0] as keyof typeof TxnOnComplete];
if (x !== undefined && typeof x === "number") ...
```

packages/algorand-js/src/types.ts
ratik21: Reverse mapping is already handled in the constructor. I like the idea to use string enum. But shouldn't we use a map instead then? What's the difference?

amityadav0: yes, it is handled but i don't think it is good to create reverse mapping and then not use it.
I couldn't find the exact difference in typescript. Please let me know if you find it.

amityadav0: after reading more about it, i found that `enum` are stored as array like structure, that's why to find key, we have to traverse all the values in enum,
But Map is implemented using hashtable. 
Therefore it is way much faster, but in our case we have  only 5 or 6 pairs.

ratik21: Reverse Mapping is a general property of any enum. When we decided on call we should use enums, it was understood that there would be reverse mapping.
I would prefer to use map instead of enums then. Please let me know if you follow, i'll update it.

amityadav0: I don't think we discussed for this enum, In this case what if we have TypeEnumConstants also, then there will be two values mapped to `0` -> NoOp, unknown. But for other enums let's say TransactionType, we can use reverse mapping.

I agree, your call

ratik21: Yes, but in this case reverse mapping acutally makes sense, because for two strings we can have the same `int` as well. eg.
```
int NoOp -> int 0 // using TxnOnComplete Enum
int pay -> int 0 // using TypeEnum
```
So both these statements makes sense: 
a) `NoOp` and `pay` represent int 0, 
b) int `0` represents both `NoOp` and `pay` 

I am retaining the enum for now with number mapping.

robert-zaremba: The thing I didin't like in the code was when we use `Object.values().indexOf(...)` (it's ugly and not intuitive), otherwise @amityadav0  has a good point - iterating over a list of few elements doesn't change much.

In general we are using this in few places, so it's good to have some shorter, better way to do it: map, auxiliary function etc...

amityadav0 (approved):

PR #183, opened 2021-01-06
Add support for typeEnum Constants

amityadav0:


packages/algorand-js/src/interpreter/opcode-list.ts
```diff
@@ -1893,8 +1893,11 @@ export class Int extends Op {
 
     let uint64;
     const txOnComplete = TxnOnComplete[args[0] as keyof typeof TxnOnComplete]; // eg. TxnOnComplete['NoOp']
+    const typeEnumConst = TxnType[args[0] as keyof typeof TxnType]; // ex: TxnType['pay']
```
ratik21, line 1896:
```suggestion
    const typeEnumConst = TxnType[args[0] as keyof typeof TxnType]; // eg: TxnType['pay']
```

packages/algorand-js/src/interpreter/opcode-list.ts
```diff
@@ -1893,8 +1893,11 @@ export class Int extends Op {
...
     if (txOnComplete !== undefined) { // check if string is keyof TxnOnComplete Enum
       uint64 = BigInt(txOnComplete);
+    } else if (typeEnumConst !== undefined) { // check if string is keyof TxnType Enum
+      uint64 = BigInt(typeEnumConst);
     } else {
```
ratik21, line 1901: Can we do:
```
const intConst = TxnOnComplete[args[0] as ..] || TxnType[args[0] as...];
if (intConst !== undefined) { uint64 = BigInt(intConst); } 
else {..}
 ```

amityadav0: yes

ratik21 (approved):

PR #185, opened 2021-01-08
Update test file structure

amityadav0:
- Move `getProgram` to `algob`. Update runtime to use program string
- update test file structure


packages/algorand-js/test/helpers/integration.ts
```diff
@@ -0,0 +1,39 @@
...
+    "..",
+    "integration-test-files",
+    filename
+  );
+  if (!fs.existsSync(filePath)) {
+    throw new Error(`Integration files ${filename} doesn't exist`);
+  }
+
+  return fs.realpathSync(filePath);
+}
```
ratik21, line 39: I was doing something similar before. Can you please check: [https://discord.com/channels/754286650540032030/754286650540032033/790732752101572608](https://discord.com/channels/754286650540032030/754286650540032033/790732752101572608)

packages/algorand-js/test/helpers/integration.ts
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * This helper adds mocha hooks to run the tests inside one of the projects
+ * from test/integration-test-files.
+ *
+ * @param filename The base name of the folder with the integration files to use.
+ */
+export function useIntegrationFile (filename: string): void {
+  let filePath: string;
+  let prevWorkingDir: string;
```
robert-zaremba, line 12:
```suggestion
 * This helper adds mocha hooks to run the tests inside one of the projects in test/fixtures.
 * @param fixtureName: The base name of the folder in test/fixtures
 */
export function useFixture (fixtureName: string): void {
  let dirPath: string;
  let prevWorkingDir: string;
```

robert-zaremba: Note: I change the argument from filename to dirname, you need to update the code below as well.

Let's rename the directory as well: 
test/integration-test-files -> test/fixtures

packages/algorand-js/test/helpers/integration.ts
amityadav0: yes mean fixture-project part or using algob part ?
for algob part, yes i think we can copy getProgram function in test/helpers(to remove dependency)

for integration-files, i don't think it is good idea to use a single folder for all testing.
because in future if we add more complex integration files, we can easily create a different folder and use it.

packages/algorand-js/test/helpers/integration.ts
```diff
@@ -0,0 +1,39 @@
...
+    prevWorkingDir = process.cwd();
+    process.chdir(filePath);
+  });
+
+  after(() => {
+    process.chdir(prevWorkingDir);
+  });
+}
+
+export function getIntegrationFilePath (
```
robert-zaremba, line 25:
```suggestion
export function getFixtureWorkingDir (
```

robert-zaremba (approved):
pre-approving

packages/algorand-js/test/helpers/integration.ts
amityadav0: but i noticed in integration tests we are using `import { ExecParams, SignType, TransactionType } from "@algorand-builder/algob/src/types";` this creates a dependency also

packages/algorand-js/test/helpers/integration.ts
amityadav0: ok

packages/algorand-js/test/helpers/integration.ts
ratik21: Okay, makes sense

robert-zaremba: I think it's ok to import types for the moment, we may export some common types in the future to another package.

PR #190, opened 2021-01-12
Add delete function and add tests

amityadav0:


packages/algorand-js/src/runtime/runtime.ts
```diff
@@ -318,6 +318,16 @@ export class Runtime {
     }
   }
 
+  // Delete application from account's state and global state
```
ratik21, line 321:
```suggestion
  // Delete application from account's local state and global state
```

packages/algorand-js/src/runtime/runtime.ts
```diff
@@ -347,12 +357,18 @@ export class Runtime {
...
+        }
       }
     } else {
       // for a single (stand alone) transaction
       this.store.accounts.forEach((account, addr) => {
         this.updateBalance(txnParams, account);
       });
+      if (txnParams.type === TransactionType.DeleteSSC) {
+        this.deleteApp(txnParams.fromAccount.addr, txnParams.appId);
+      }
```
ratik21, line 371: We have repetitive code in this function (which will probably increase when we add more txn types). I would suggest to:
+ either move the common code to a function and reuse the function. OR
+ treat the single txn as a group of 1 txn (`const tx = [txnParams];`), and just operate the updates on an array once.

packages/algorand-js/test/integration/deleteApp.ts
```diff
@@ -0,0 +1,54 @@
...
+  it("should delete application", async function () {
+    const appId = await runtime.addApp(flags, {}, program);
+    deleteParams.appId = appId;
+    await runtime.executeTx(deleteParams, program, []);
+
+    // verify app is deleted
+    const res = runtime.ctx.state.globalApps.has(appId);
+    assert.equal(res, false);
+  });
+});
```
ratik21, line 54: Please add one more test: 
+ create an app
+ execute transaction which should be rejected by logic (maybe we need to modify teal code)
+ assert that the app is not deleted as txn is rejected.

packages/algorand-js/src/runtime/runtime.ts
```diff
@@ -318,6 +318,16 @@ export class Runtime {
     }
   }
 
+  // Delete application from account's state and global state
+  deleteApp (accountAddr: string, appId: number): void {
+    if (!this.store.globalApps.has(appId)) {
+      throw new TealError(ERRORS.TEAL.APP_NOT_FOUND);
+    }
+    this.store.globalApps.delete(appId);
+    const account = this.assertAccountDefined(this.store.accounts.get(accountAddr));
```
ratik21, line 327: What happens if we delete the `app` and account is not defined?

packages/algorand-js/src/runtime/runtime.ts
amityadav0: we are deleting it from global too

packages/algorand-js/src/runtime/runtime.ts
amityadav0: second one seems good

packages/algorand-js/src/runtime/runtime.ts
amityadav0: Nice catch

packages/algorand-js/test/integration/deleteApp.ts
amityadav0: thanks for suggestion

ratik21:
@amityadav0 the current code is only deleting application from global state (creator's account). Can you please check one thing: Let's say we have 3 accounts A, B, C. A has created an app and A, B, C opt-in to that app. Now if we delete the app, what happens to the `localState` of A, B, C? Does the app get's deleted from there as well? If yes then we need to modify the code. Can you check?

packages/algorand-js/src/runtime/runtime.ts
```diff
@@ -342,17 +356,19 @@ export class Runtime {
    * @param accounts : accounts passed by user
```
robert-zaremba, line 356: there is no `accounts` parameter in this function. Could you update the function description? Please be more specific which account balances are we updating - it seams that we update all existing accounts.

packages/algorand-js/test/integration/crowdfunding.ts
```diff
@@ -64,7 +63,7 @@ describe("Crowdfunding basic tests", function () {
     const appId = await runtime.addApp({ ...validFlags, appArgs: appArgs }, {}, program);
     const getGlobal = (key: string):
       bigint | Uint8Array |undefined => runtime.getGlobalState(appId, stringToBytes(key));
-    const johnPk = decodeAddress(john.address).publicKey;
+    const johnPk = addressToBytes(john.address);
```
robert-zaremba, line 66: how about renaming `addressToBytes` -> `addressToPk`? BTW, I thought that address doesn't have a reverse function to Public Key.

robert-zaremba (approved):
pre-approving.

amityadav0:
@ratik21 i think you are talking about closeout and clear, we will add them when we add full support for application calls, i.e after payment checks and asset checks

ratik21:
> @ratik21 i think you are talking about closeout and clear, we will add them when we add full support for application calls, i.e after payment checks and asset checks

Okay, understood. So that means delete call only delete's the app from creator's account.

packages/algorand-js/src/runtime/runtime.ts
amityadav0: ok

amityadav0: yes, we are traversing a list of accounts it is not required, updating. thanks

packages/algorand-js/test/integration/crowdfunding.ts
amityadav0: for int and string, we are using toBytes, and according to decodeAddress, it returns pk as Uint8Array. so looking from user perspective i think `addressToBytes` is ok(i added doc string addressToBytes about Public key).
should we rename it to `addressToPk` ?

amityadav0: updated to `addressToPk`

PR #191, opened 2021-01-12
bug fix

amityadav0:
- Bug fix: we were using test functions in src, that's why build folder had two folders (src and test)
and we were not able to import directly

### DIFF
--- a/examples/crowdfunding/test/crowdfundingTest.js
+++ b/examples/crowdfunding/test/crowdfundingTest.js
@@ -6,7 +6,7 @@ import {
   stringToBytes,
   TransactionType
 } from '@algorand-builder/algob';
-import { Runtime, StoreAccountImpl } from '@algorand-builder/algorand-js/build/src/index';
+import { Runtime, StoreAccountImpl } from '@algorand-builder/algorand-js';
 import { assert } from 'chai';
 
 import { getAcc } from './common';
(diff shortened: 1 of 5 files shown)


packages/algorand-js/tsconfig.json
```diff
@@ -1,10 +0,0 @@
-{
-  "extends": "../../tsconfig.json",
-  "compilerOptions": {
-    "outDir": "./build"
-  },
-  "include": [
-    "src",
-    "../../@types"
-  ]
-}
```
ratik21, line 10: What was different before? It wasn't a symlink?

examples/crowdfunding/test/crowdfundingTest.js
```diff
@@ -6,7 +6,7 @@ import {
   stringToBytes,
   TransactionType
 } from '@algorand-builder/algob';
-import { Runtime, StoreAccountImpl } from '@algorand-builder/algorand-js/build/src/index';
+import { Runtime, StoreAccountImpl } from '@algorand-builder/algorand-js';
 import { assert } from 'chai';
 
 import { getAcc } from './common';
```
ratik21, line 12: I noticed we're using `import {..}` here in `.js`, whereas in other files we are using `const {..} = require('..');`. Is there any specific reason for doing that?

packages/algorand-js/tsconfig.json
amityadav0: yes

examples/crowdfunding/test/crowdfundingTest.js
amityadav0: no, not any specific reason

robert-zaremba (approved):

PR #193, opened 2021-01-13
Update Runtime and StoreAccount Structure

ratik21:
+ Remove duplicate storage
+ Remove `accountAssets` and use `account.assets`
+ Change structures from array to Map

### DIFF
--- a/packages/algorand-js/src/runtime/account.ts
+++ b/packages/algorand-js/src/runtime/account.ts
@@ -23,7 +23,7 @@ export class StoreAccount implements StoreAccountI {
   amount: number;
   appsLocalState: Map<number, AppLocalState>;
   appsTotalSchema: SSCSchemaConfig;
-  createdApps: CreatedApp[];
+  createdApps: Map<number, SSCAttributes>;
   createdAssets: CreatedAssets[];
 
   constructor (balance: number, account?: Account) {
@@ -41,7 +41,7 @@ export class StoreAccount implements StoreAccountI {
     this.amount = balance;
     this.appsLocalState = new Map<number, AppLocalState>();
     this.appsTotalSchema = <SSCSchemaConfig>{};
-    this.createdApps = [];
+    this.createdApps = new Map<number, SSCAttributes>();
     this.createdAssets = [];
   }
 
@@ -101,12 +101,12 @@ export class StoreAccount implements StoreAccountI {
   }
 
   addApp (appId: number, params: SSCDeploymentFlags): CreatedApp {
-    if (this.createdApps.length === 10) {
+    if (this.createdApps.size === 10) {
       throw new Error('Maximum created applications for an account is 10');
     }
 
     const app = new App(appId, params);
-    this.createdApps.push(app);
+    this.createdApps.set(app.id, app.params);
     return app;
   }
 
--- a/packages/algorand-js/src/runtime/account.ts
+++ b/packages/algorand-js/src/runtime/account.ts
@@ -82,6 +82,7 @@ export class StoreAccount implements StoreAccountI {
     });
   }
 
+  // add application in account's state
   addApp (appId: number, params: SSCDeploymentFlags): CreatedAppM {
     if (this.createdApps.size === 10) {
       throw new Error('Maximum created applications for an account is 10');
@@ -109,6 +110,14 @@ export class StoreAccount implements StoreAccountI {
       };
       this.appsLocalState.set(appId, localParams); // push
     }
+  }
+
+  // delete application from account's state
+  deleteApp (appId: number): void {
+    if (!this.createdApps.has(appId)) {
+      throw new TealError(ERRORS.TEAL.APP_NOT_FOUND, { appId: appId });
+    }
+    this.createdApps.delete(appId);
   }
 }
 
--- a/packages/algorand-js/src/runtime/account.ts
+++ b/packages/algorand-js/src/runtime/account.ts
@@ -13,7 +13,7 @@ import { ERRORS } from "../errors/errors-list";
 import { assertValidSchema } from "../lib/stateful";
 import { AppLocalStateM, CreatedAppM, SSCAttributesM, StackElem, StoreAccountI } from "../types";
 
-const keyValue = "key-value";
+const StateMap = "key-value";
 export class StoreAccount implements StoreAccountI {
   readonly account: Account;
   readonly address: string;
@@ -56,7 +56,7 @@ export class StoreAccount implements StoreAccountI {
    */
   getLocalState (appId: number, key: Uint8Array): StackElem | undefined {
     const localState = this.appsLocalState;
-    const data = localState.get(appId)?.[keyValue]; // can be undefined (eg. app opted in)
+    const data = localState.get(appId)?.[StateMap]; // can be undefined (eg. app opted in)
     return data?.get(key.toString());
   }
 
@@ -69,12 +69,12 @@ export class StoreAccount implements StoreAccountI {
    */
   setLocalState (appId: number, key: Uint8Array, value: StackElem): AppLocalStateM {
     const localState = this.appsLocalState.get(appId);
-    const localApp = localState?.[keyValue];
+    const localApp = localState?.[StateMap];
     if (localState && localApp) {
       localApp.set(key.toString(), value);
-      localState[keyValue] = localApp; // save updated state
+      localState[StateMap] = localApp; // save updated state
 
-      assertValidSchema(localState[keyValue], localState.schema); // verify if updated schema is valid by config
+      assertValidSchema(localState[StateMap], localState.schema); // verify if updated schema is valid by config
       return localState;
     }
 
@@ -96,7 +96,7 @@ export class StoreAccount implements StoreAccountI {
 
   // opt in to application
   optInToApp (appId: number, appParams: SSCAttributesM): void {
-    const localState = this.appsLocalState.get(appId);
+    const localState = this.appsLocalState.get(appId); // fetch local state from account
     if (localState) {
       console.warn(`${this.address} is already opted in to app ${appId}`);
     } else {
@@ -104,12 +104,13 @@ export class StoreAccount implements StoreAccountI {
         throw new Error('Maximum Opt In applications per account is 10');
       }
 
+      // create new local app attribute
       const localParams: AppLocalStateM = {
         id: appId,
         "key-value": new Map<string, StackElem>(),
         schema: appParams["local-state-schema"]
       };
-      this.appsLocalState.set(appId, localParams); // push
+      this.appsLocalState.set(appId, localParams); // opt-in
     }
   }
 
--- a/packages/algorand-js/src/runtime/account.ts
+++ b/packages/algorand-js/src/runtime/account.ts
@@ -13,9 +13,9 @@ import { ERRORS } from "../errors/errors-list";
 import { compareArray } from "../lib/compare";
 import { SSC_BYTES } from "../lib/constants";
 import { assertValidSchema, getKeyValPair } from "../lib/stateful";
-import { StackElem, StoreAccount } from "../types";
+import { StackElem, StoreAccountI } from "../types";
 
-export class StoreAccountImpl implements StoreAccount {
+export class StoreAccount implements StoreAccountI {
   readonly account: Account;
   readonly address: string;
   assets: AssetHolding[]; // TODO: to be removed
--- a/packages/algorand-js/src/runtime/account.ts
+++ b/packages/algorand-js/src/runtime/account.ts
@@ -20,7 +20,7 @@ export class StoreAccount implements StoreAccountI {
   readonly address: string;
   assets: AssetHolding[]; // TODO: to be removed
   amount: number;
-  appsLocalState: AppLocalState[]; // TODO: update to map
+  appsLocalState: Map<number, AppLocalState>; // TODO: update to map
   appsTotalSchema: SSCSchemaConfig;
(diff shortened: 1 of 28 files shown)


packages/algorand-js/src/runtime/account.ts
```diff
@@ -75,22 +66,14 @@ export class StoreAccount implements StoreAccountI {
...
+  updateLocalState (appId: number, key: Uint8Array, value: StackElem): AppLocalStateM {
     const localState = this.appsLocalState.get(appId);
-    const data = getKeyValPair(key, value); // key value pair to put
-
-    const localApp = localState?.["key-value"];
+    const localApp = localState?.[keyValue];
     if (localState && localApp) {
-      const idx = localApp.findIndex(schema => compareArray(schema.key, key));
+      localApp.set(key.toString(), value);
+      localState[keyValue] = localApp; // save updated state
```
amityadav0, line 74: can we rename keyValue to `localState` or `localStateMap` or something else ?

packages/algorand-js/src/runtime/account.ts
```diff
@@ -57,15 +55,8 @@ export class StoreAccount implements StoreAccountI {
...
-    if (data) {
-      const keyValue = data.find(schema => compareArray(schema.key, key));
-      const value = keyValue?.value;
-      if (value) {
-        return value.type === SSC_BYTES ? value.bytes : BigInt(value.uint);
-      }
-    }
-    return undefined;
+    const data = localState.get(appId)?.[keyValue]; // can be undefined (eg. app opted in)
+    return data?.get(key.toString());
```
amityadav0, line 59: what happens if localstate is undefined ? does TEAL throws an error or returns 0 ?

packages/algorand-js/src/runtime/account.ts
```diff
@@ -120,9 +103,9 @@ export class StoreAccount implements StoreAccountI {
         throw new Error('Maximum Opt In applications per account is 10');
       }
 
-      const localParams: AppLocalState = {
+      const localParams: AppLocalStateM = {
         id: appId,
-        "key-value": [],
+        "key-value": new Map<string, StackElem>(),
         schema: appParams["local-state-schema"]
       };
```
amityadav0, line 110: what if we want to update the application ? does that mean we need to reset the local-state of opt-in accounts ?

packages/algorand-js/src/runtime/runtime.ts
```diff
@@ -222,11 +194,14 @@ export class Runtime {
...
+
+    this.ctx.state.globalApps.delete(0); // remove zero app from context after execution
+    senderAcc.createdApps.delete(0); // remove zero app from sender's account
 
-    // update local state
-    senderAcc.createdApps.pop();
+    // set new application in sender's account
+    // after setting in globalApps
     senderAcc.addApp(this.appCounter, flags);
     this.store.accounts.set(sender.addr, senderAcc);
```
amityadav0, line 206: please add this in deleteApp and optInToapp too, as we are doing same thing there. 
also please add TO-DO in these functions.

packages/algorand-js/test/mocks/stateful.ts
```diff
@@ -1,76 +1,58 @@
-import { AppLocalState } from "algosdk";
+import { AssetDef, AssetHolding } from "algosdk";
 
 import { stringToBytes } from "../../src/lib/parsing";
+import { AppLocalStateM, SSCAttributesM, StackElem } from "../../src/types";
+
+const byteToStr = (str: string): string => { return stringToBytes(str).toString(); };
```
amityadav0, line 6: can we rename it to convertToKey or something else ?

packages/algorand-js/test/mocks/stateful.ts
```diff
@@ -1,76 +1,58 @@
...
+const appLocalState = new Map<number, AppLocalStateM>();
+const createdApps = new Map<number, SSCAttributesM>();
+const createdAssets = new Map<number, AssetDef>();
+
+const assets = new Map<number, AssetHolding>();
+assets.set(3, { 'asset-id': 3, amount: 2, creator: "string", 'is-frozen': "false" });
+assets.set(32, { 'asset-id': 32, amount: 2, creator: "AS", 'is-frozen': "false" });
+
+const globalKeyVal = new Map<string, StackElem>();
+globalKeyVal.set(byteToStr('Hello'), stringToBytes('World'));
```
amityadav0, line 17: can we rename key-val in global as well ?

packages/algorand-js/test/src/interpreter/opcode-list.ts
```diff
@@ -2346,14 +2348,12 @@ describe("Teal Opcodes", function () {
 
         const appId = interpreter.runtime.ctx.tx.apid;
         const acc = interpreter.runtime.ctx.state.accounts.get(elonAddr) as StoreAccountI;
-        let localStateCurr = acc.appsLocalState.get(appId)?.["key-value"];
+        let localStateCurr = acc.appsLocalState.get(appId)?.[keyValue];
         assert.isDefined(localStateCurr);
```
amityadav0, line 2352: can we move this part to appsLocalState function or as how TEAL behaves in this case ?

packages/algorand-js/src/runtime/account.ts
ratik21: It pushes 0 to stack.

packages/algorand-js/test/src/interpreter/opcode-list.ts
```diff
@@ -2346,14 +2348,12 @@ describe("Teal Opcodes", function () {
...
         if (localStateCurr) {
-          // TODO: will be updated to Map
-          idx = localStateCurr.findIndex(a => compareArray(a.key, stringToBytes('New-Key')));
-          assert.isDefined(idx);
-          assert.notEqual(idx, -1); // idx should not be -1
-          if (idx) assert.deepEqual(localStateCurr[idx].value.bytes, stringToBytes('New-Val'));
+          value = localStateCurr.get(byteToStr('New-Key'));
```
amityadav0, line 2354: i think teal returns 0 for non-existing keys. can you please confirm ?

packages/algorand-js/test/src/interpreter/opcode-list.ts
```diff
@@ -2497,9 +2497,9 @@ describe("Teal Opcodes", function () {
         const op = new AppGlobalDel([], 1, interpreter);
         op.execute(stack);
 
-        const globalStateCurr = (interpreter.runtime.ctx.state.globalApps.get(1828) as SSCAttributes)["global-state"];
-        const idx = globalStateCurr.findIndex(a => compareArray(a.key, stringToBytes('global-key')));
-        assert.equal(idx, -1); // idx should be -1
+        const globalStateCurr = (interpreter.runtime.ctx.state.globalApps.get(1828) as SSCAttributesM)["global-state"];
+        const value = globalStateCurr.get(byteToStr('global-key'));
+        assert.isUndefined(value); // value should be undefined
```
amityadav0, line 2502: same.

packages/algorand-js/test/src/interpreter/opcode-list.ts
```diff
@@ -2478,16 +2477,17 @@ describe("Teal Opcodes", function () {
         op.execute(stack);
 
         localStateCurr = (interpreter.runtime.ctx.state.accounts.get(johnAddr) as StoreAccountI)
-          .appsLocalState.get(appId)?.["key-value"];
-        if (localStateCurr) idx = localStateCurr.findIndex(a => compareArray(a.key, stringToBytes('Local-key')));
-        assert.equal(idx, -1); // idx should be -1
+          .appsLocalState.get(appId)?.[keyValue];
+        if (localStateCurr) value = localStateCurr.get(byteToStr('Local-key'));
+        assert.isUndefined(value); // value should be undefined
```
amityadav0, line 2482: same

packages/algorand-js/src/runtime/account.ts
amityadav0: okay, please update it to as TEAL behaves

ratik21: That part is handled in the opcode itself. If we receive `undefined` in `AppLocalGet` opcode, it pushes `0n` to stack.

packages/algorand-js/src/runtime/account.ts
ratik21: yes. updating to `localStateMap`.

packages/algorand-js/src/runtime/account.ts
amityadav0: when we are testing TEAL, i would prefer it to behave same as TEAL behaves, 
because for some TEAL codes, if user wants to check what is there in localstate or how teal interprets it ? then it is  better.
what do you think ? what are the benefits of not doing it ?

amityadav0: can we check for particular values from goal ?

ratik21: Agreed, no issues. It's just that we only interact with TEAL during a transaction execution, but we interact with runtime throughout the tests, that's why i was returning `undefined` here, and keeping teal relevant logic in opcodes. 
Will update it as per TEAL behaviour, and resolve similar comments.

amityadav0: please add doc string for the same.

ratik21: > can we check for key-values(map) from goal ?

Yes, we can view account's local and global storage using `goal`. I think using `goal app read --local --app-id ..`

packages/algorand-js/test/src/interpreter/opcode-list.ts
ratik21: This is just an extra assert, because map can also return undefined value.

packages/algorand-js/src/runtime/account.ts
amityadav0: Ok thanks, I couldn't find searching from key. For ex : finding value from map with a specific key. I think this must be the only way.

packages/algorand-js/src/runtime/runtime.ts
ratik21: Ok, will do. 
Also this function needs to be updated also as per our discussion, line `205` will always create a fresh app with new state, not the update one. Adding TODO's and task link

packages/algorand-js/src/runtime/account.ts
ratik21: > Ok thanks, I couldn't find searching from key. For ex : finding value from map with a specific key. I think this must be the only way.

Yes, i don't think we can do that. We can view complete global and local state only.

packages/algorand-js/src/runtime/runtime.ts
ratik21: > please add this in deleteApp and optInToapp too, as we are doing same thing there.

I checked delete and opt-in functions, delete is fine, as we are just deleting the app from account and global Map. 
In optIn we need to update `globaApps` map as well, but since we will be removing this, i have added a TODO there.

packages/algorand-js/test/mocks/stateful.ts
ratik21: renaming to `convertToKey`

packages/algorand-js/src/runtime/account.ts
ratik21: Upon checking furthur, i think this can cause discrepency with `GetEx` opcodes, as we push did exist also with them. @amityadav0 can you please check: [https://github.com/scale-it/algorand-builder/blob/master/packages/algorand-js/src/interpreter/opcode-list.ts#L1581](https://github.com/scale-it/algorand-builder/blob/master/packages/algorand-js/src/interpreter/opcode-list.ts#L1581).

If we return `0n` here, then we won't know whether the value exists or not. One possible solution could be to return flag as well, but i am not sure if we should do that.

packages/algorand-js/src/runtime/account.ts
ratik21: Good point. I haven't tested that, will test it to check if local state is reset on update or not.

packages/algorand-js/test/mocks/stateful.ts
ratik21: I don't follow. What should we rename here?

packages/algorand-js/src/runtime/account.ts
amityadav0: maybe we can keep accounts functions as it is and update runtime functions to return 0, but if it causes discrepancy I agree with you, let's keep it as it is.

amityadav0: Please add todo that we need to pass address to global state function in updated single storage.

packages/algorand-js/test/mocks/stateful.ts
amityadav0: maybe globalstatemap

ratik21: works

packages/algorand-js/src/runtime/account.ts
ratik21: ok.

packages/algorand-js/src/runtime/account.ts
ratik21: Upon confirming: no, the state is not resetted. And in this function we are already checking previously if app is opted in or not. If app is not opted in, then only we are creating a new local state and setting. Please let me know if that makes sense.

amityadav0 (approved):

PR #194, opened 2021-01-14
improve devUX

amityadav0:


packages/algob/src/lib/status.ts
```diff
@@ -28,7 +28,7 @@ export async function readGlobalStateSSC (
   appId: number): Promise<SSCStateSchema[] | undefined> {
   const accountInfoResponse = await deployer.algodClient.accountInformation(creator).do();
   for (const value of accountInfoResponse['created-apps']) {
```
robert-zaremba, line 30: please rename `value` to `app`

robert-zaremba (approved):
preapproving. left one change request.

packages/algorand-js/src/runtime/runtime.ts
```diff
@@ -98,14 +98,24 @@ export class Runtime {
     return undefined;
   }
 
+  /**
+   * Fetches local state for account address and application index
+   * @param appId application index
+   * @param accountAddr address for which local state needs to be retrieved
+   */
+  getLocalState (appId: number, accountAddr: string, key: Uint8Array): StackElem | undefined {
```
ratik21, line 106: Please add one more param `key` in description

packages/algob/src/lib/status.ts
amityadav0: ok

packages/algorand-js/src/runtime/runtime.ts
amityadav0: ok

PR #195, opened 2021-01-14
Update all get/set state functions to support string key argument

ratik21:
#193 to be merged first


packages/algorand-js/src/lib/parsing.ts
```diff
@@ -17,7 +17,7 @@ export function assertOnlyDigits (val: string): void {
 }
 
 /**
- * Description: assert words length
+ * assert words length
```
robert-zaremba, line 20:
```suggestion
 * assert that a line has given number of words
```

packages/algorand-js/src/lib/parsing.ts
```diff
@@ -86,7 +94,7 @@ export function convertToBuffer (s: string, encoding?: EncodingType): Buffer {
 }
 
 /**
- * Description: Returns base64 or base32 string
+ * Returns base64 or base32 string
```
robert-zaremba, line 97: Please update the the description about the `args` argument and describe when we convert to base64 and when to base32

packages/algorand-js/src/lib/parsing.ts
```diff
@@ -110,7 +118,7 @@ function base64OrBase32 (args: string[], line: number): [string, EncodingType] {
 }
 
 /**
- * Description: returns encodingtype and string from words list
+ * returns encodingtype and string from words list
```
robert-zaremba, line 121: Please update the description to make it more intuitive. what string are we returning and what is an encodingtype.

robert-zaremba (approved):
looks good. left few comments to update function docs.

packages/algorand-js/src/lib/parsing.ts
ratik21: ok

packages/algorand-js/src/lib/parsing.ts
ratik21: yes, ok

PR #197, opened 2021-01-15
Report line

amityadav0:
- line number in all assert* statements 
- line number when parsing TEAL.
- line number while executing in interpreter


packages/runtime/src/interpreter/opcode.ts
```diff
@@ -6,85 +6,87 @@ import type { TEALStack } from "../types";
...
   }
 
   // assert if known global field is passed
-  assertGlobalDefined (str: string): void {
+  assertGlobalDefined (str: string, line: number): void {
     if (GlobalFields[str] === undefined) {
-      throw new TealError(ERRORS.TEAL.UNKOWN_GLOBAL_FIELD, { field: str });
+      throw new TealError(ERRORS.TEAL.UNKOWN_GLOBAL_FIELD, { field: str, line: line });
     }
   }
```
ratik21, line 91: Please add a doc string (maybe at the top of file) explaining `line`

packages/runtime/src/runtime.ts
```diff
@@ -46,29 +46,38 @@ export class Runtime {
...
     }
     return app;
   }
 
-  getAccount (accountIndex: bigint): StoreAccountI {
+  /**
+   * Fetch account using accountIndex from `Accounts` list
+   * Accounts: List of accounts in addition to the sender
+   * that may be accessed from the application's approval-program and clear-state-program.
+   * @param accountIndex account list
```
ratik21, line 70:
```suggestion
   * @param accountIndex index of account to fetch from account list
```

packages/runtime/src/runtime.ts
```diff
@@ -46,29 +46,38 @@ export class Runtime {
...
+      throw new TealError(ERRORS.TEAL.APP_NOT_FOUND, { appId: appId, line: lineNumber });
     }
     return app;
   }
 
-  getAccount (accountIndex: bigint): StoreAccountI {
+  /**
+   * Fetch account using accountIndex from `Accounts` list
+   * Accounts: List of accounts in addition to the sender
+   * that may be accessed from the application's approval-program and clear-state-program.
```
ratik21, line 69: Please also add `NOTE: index 0 represents txn sender account`

packages/runtime/src/lib/parsing.ts
```diff
@@ -10,9 +10,13 @@ import { reBase32, reBase64, reDigit } from "./constants";
  * "123" // ok.  "12+2" // error.
  * @param val : string
  */
-export function assertOnlyDigits (val: string): void {
+export function assertOnlyDigits (val: string, line: number): void {
   if (!reDigit.test(val)) {
-    throw new TealError(ERRORS.TEAL.INVALID_TYPE, { expected: "unsigned integer", actual: val });
+    throw new TealError(ERRORS.TEAL.INVALID_TYPE, {
+      expected: "unsigned integer",
```
ratik21, line 16:
```suggestion
      expected: "unsigned integer (upto 64 bit)",
```

packages/runtime/test/src/interpreter/opcode.ts
```diff
@@ -12,26 +12,26 @@ describe("Teal Opcodes basic assertions", function () {
...
-      () => op.checkUnderflow(min),
+      () => op.checkUnderflow(min, 1),
       ERRORS.TEAL.UINT64_UNDERFLOW
     );
   });
 
   it("check minimum stack length", function () {
     const stack = new Stack<StackElem>();
-    op.assertMinStackLen(stack, 0);
+    op.assertMinStackLen(stack, 0, 1);
```
ratik21, line 30: I think it can be confusing. From an outside eye i am not sure whether i am asserting for stack length `0` or `1`. Maybe we could pass source line in an object (eg. `op.assertMinStackLen(stack, 0, { line: 1})` ) ?

amityadav0: i don't follow, because there are many functions similar to this, we cannot pass objects instead of arguments(in every function),
we are in test folder, i think it will be better if we can store it as `line` `stackLength`.
In src we have name for arguments

packages/runtime/src/runtime.ts
amityadav0: yes

packages/runtime/src/interpreter/opcode.ts
amityadav0: yes

packages/runtime/test/src/interpreter/opcode.ts
ratik21: > we are in test folder, i think it will be better if we can store it as line stackLength.

That works too! Could you update for similar tests as well? Otherwise `op.checkOverflow(2, 1)`, `op.assertMinStackLen(stack, 0, 1)`... seems confusing.

amityadav0: yes

packages/runtime/src/interpreter/opcode.ts
```diff
@@ -6,88 +6,152 @@ import type { TEALStack } from "../types";
 import { BIGINT0, BIGINT1 } from "./opcode-list";
 
 export class Op {
-  assertMinStackLen (stack: TEALStack, minLen: number): void {
+  /**
+   * assert stack length is less than minLen
```
ratik21, line 10:
```suggestion
   * assert stack length is atleast minLen
```

packages/runtime/src/interpreter/opcode.ts
```diff
@@ -6,88 +6,152 @@ import type { TEALStack } from "../types";
...
+  assertMinStackLen (stack: TEALStack, minLen: number, line: number): void {
     if (stack.length() < minLen) {
-      throw new TealError(ERRORS.TEAL.ASSERT_STACK_LENGTH);
+      throw new TealError(ERRORS.TEAL.ASSERT_STACK_LENGTH, { line: line });
     }
   }
 
-  checkOverflow (num: bigint): void {
+  /**
+   * asserts number is less than equal to MAX_UINT64
```
ratik21, line 22:
```suggestion
   * asserts number is less than or equal to MAX_UINT64
```

packages/runtime/src/interpreter/opcode.ts
```diff
@@ -6,88 +6,152 @@ import type { TEALStack } from "../types";
...
+  checkOverflow (num: bigint, line: number): void {
     if (num > MAX_UINT64) {
-      throw new TealError(ERRORS.TEAL.UINT64_OVERFLOW);
+      throw new TealError(ERRORS.TEAL.UINT64_OVERFLOW, { line: line });
     }
   }
 
-  checkUnderflow (num: bigint): void {
+  /**
+   * asserts number is greater than MIN_UINT64
```
ratik21, line 33:
```suggestion
   * asserts number is greater than MIN_UINT64 (0n)
```

packages/runtime/src/interpreter/opcode.ts
amityadav0: atleast would mean we want stack to be greater than minLen

packages/runtime/src/interpreter/opcode.ts
```diff
@@ -6,88 +6,152 @@ import type { TEALStack } from "../types";
...
+  checkIndexBound (idx: number, arr: any[], line: number): void {
     if (!(idx >= 0 && idx < arr.length)) {
-      throw new TealError(ERRORS.TEAL.INDEX_OUT_OF_BOUND);
+      throw new TealError(ERRORS.TEAL.INDEX_OUT_OF_BOUND, { line: line });
     }
   }
 
-  assertArrLength (arr: Uint8Array[] | BigInt[]): void {
+  /**
+   * asserts if array length is less than equal to MAX_UINT8 and not equal to 0
```
ratik21, line 56:
```suggestion
   * asserts if array length is less than equal to MAX_UINT8 (255) and not equal to 0
```

packages/runtime/src/interpreter/opcode.ts
```diff
@@ -6,88 +6,152 @@ import type { TEALStack } from "../types";
...
+        actual: typeof b,
+        line: line
       });
     }
     return b;
   }
 
-  assertUint8 (a: bigint): bigint {
+  /**
+   * asserts if given bigint is a Uint8 integer
```
ratik21, line 99:
```suggestion
   * asserts if given bigint is an 8 bit unsigned integer
```

packages/runtime/src/interpreter/opcode.ts
amityadav0: we cannot use `atmost` here too, because that will include minLen

packages/runtime/src/interpreter/opcode.ts
```diff
@@ -6,88 +6,152 @@ import type { TEALStack } from "../types";
...
+   * @param a bigint
```
ratik21, line 100:
```suggestion
   * @param a  value to assert (in bigint)
```

packages/runtime/src/interpreter/opcode.ts
```diff
@@ -6,88 +6,152 @@ import type { TEALStack } from "../types";
...
     }
     return a;
   }
 
-  subString (start: bigint, end: bigint, byteString: Uint8Array): Uint8Array {
+  /**
+   * Returns substring from given string (if it exists)
+   * @param start starting index
+   * @param end ending index
+   * @param byteString given string
```
ratik21, line 114:
```suggestion
   * @param byteString given string as bytes
```

packages/runtime/src/lib/compare.ts
```diff
@@ -20,8 +20,9 @@ export function compareArray (a: unknown, b: unknown): boolean {
 }
 
 // check if index is accessible in an array
-export function checkIndexBound (idx: number, arr: any[]): void {
+export function checkIndexBound (idx: number, arr: any[], line?: number): void {
+  const lineNumber = line ?? 'unknown';
```
ratik21, line 24: I am curious regarding the `unknown` line number. When can the line number be unknown?

amityadav0: when user is testing from runtime object, let's say user access account from runtime.getAcc() we don't have line here. but when accessing from interpreter a line is definitely passed(similarly for some other functions).

packages/runtime/src/interpreter/opcode.ts
ratik21: > atleast would mean we want stack to be greater than minLen

Isn't this what we want? . `assertMinStackLen(stack, 2)` means that there should be atleast 2 elements in stack (probably because the `opcode` is looking to pop two elements. Am i missing something?

amityadav0: hmm, i was thinking of something else here, you are right

packages/runtime/src/lib/compare.ts
ratik21: Ok, makes sense. Please add a doc string regarding line number `unknown` too. (if user is accessing this function directly through runtime, the line number is unknown)

ratik21 (approved):

PR #198, opened 2021-01-15
Add support for arguments, similar to goal

ratik21:
- Updated `intToBigEndian` implementation to support `uint64` (Changed function because `bigint` and `number` gives different results with left shift `<<` operators)
- added test
- updated eg


packages/algob/test/lib/ssc.ts
```diff
@@ -11,20 +13,113 @@ describe("Convert integer to big endian", () => {
...
-    res = intToBigEndian(233654);
+    res = uint64ToBigEndian(233654);
     expected = new Uint8Array([0, 0, 0, 0, 0, 3, 144, 182]);
     assert.deepEqual(res, expected);
 
-    res = intToBigEndian(9007199254740991);
+    res = uint64ToBigEndian(Number.MAX_SAFE_INTEGER);
     expected = new Uint8Array([0, 31, 255, 255, 255, 255, 255, 255]);
     assert.deepEqual(res, expected);
   });
```
amityadav0, line 35: let's merge these tests with the test below or remove it(as they are same).

packages/algob/test/lib/ssc.ts
```diff
@@ -11,20 +13,113 @@ describe("Convert integer to big endian", () => {
...
+});
+
+describe("Parse appArgs to SSC to bytes", () => {
+  it("should return undefined if app Args are not defined", () => {
+    const res = parseSSCAppArgs(undefined);
+    assert.isUndefined(res);
+  });
+
+  it("should return same bytes if all bytes are passed", () => {
+    const res = parseSSCAppArgs(['a', 'b', 'c'].map(stringToBytes));
```
amityadav0, line 75: can we change it to numbers(like in line below) ?

amityadav0: or not this seems non-intutive what the next line is doing.

examples/crowdfunding/test/crowdfundingTest.js
```diff
@@ -74,11 +74,11 @@ describe('Crowdfunding Tests', function () {
   fundCloseDate.setSeconds(fundCloseDate.getSeconds() + 120000);
 
   const creationArgs = [
-    intToBigEndian(beginDate.getTime()),
-    intToBigEndian(endDate.getTime()),
-    intToBigEndian(goal),
+    uint64ToBigEndian(beginDate.getTime()),
+    uint64ToBigEndian(endDate.getTime()),
+    uint64ToBigEndian(goal),
```
amityadav0, line 79: can we use "int:${goal}" this here ?

packages/algob/test/lib/ssc.ts
ratik21: Can you please elaborate? In `res` i am using passing string in arrays and using `stringToBytes` because that's probably how the user will pass `appArgs`.
`expected` is the expected value from the parse function, since this was expected i added actual `byte` array here instead of using the `stringToBytes` function.

examples/crowdfunding/test/crowdfundingTest.js
ratik21: yes

packages/algob/test/lib/ssc.ts
amityadav0: a, b, c are 3 different strings, in case of `int` test(the next unit test) we have three [] arrays, but here we have 3 values in one array. that seems un-intuitive

ratik21: Here we are creating the array inside map function. I thought that was understood. Will update this as per next unit test.

packages/algob/test/lib/ssc.ts
ratik21: In the below tests we are operating on bigint not number (to represent upto 64 bit). I don't want to remove them, will merge.

packages/algob/src/lib/ssc.ts
```diff
@@ -1,23 +1,84 @@
...
+export function uint64ToBigEndian (x: number | bigint): Uint8Array {
+  const bytes = new Uint8Array(8);
+  let i = 0;
+  x = BigInt(x); // use x as bigint internally to support upto uint64
+  assertUint64(x);
+  while (x) {
+    bytes[i++] = Number(x % 256n);
+    x /= 256n;
+  }
+  return bytes.reverse();
```
robert-zaremba, line 28: If you start with `i=7` and use `bytes[i--] = Number(x % 256n)` then you don't need to call `bytes.reverse()`.

packages/algob/src/lib/ssc.ts
```diff
@@ -1,23 +1,84 @@
...
+      case 'b64': {
+        appArgs[idx] = new Uint8Array(Buffer.from(value, 'base64'));
+        break;
+      }
+      default: {
+        throwErr(appArg);
+      }
+    }
+  });
+  return appArgs as Uint8Array[];
```
robert-zaremba, line 73: Please don't overwrite. Create a new array, and return it:
```
const args = <Uint7Array>[]
for ... {
   ...
   args.push(new Uint...)
}
return args
```

robert-zaremba: BTW - I don't think in the return type we need `undefined` - we can remove it.

robert-zaremba (approved):
left few comments.

packages/algob/src/lib/ssc.ts
ratik21: Thanks for suggestion!

packages/algob/src/lib/ssc.ts
ratik21: > BTW - I don't think in the return type we need `undefined` - we can remove it.

We are returning undefined here to handle the case when user doesn't passes appArgs to Stateful smart contract. 
eg.
```
return algosdk.makeApplicationNoOpTxn(
  execParams.fromAccount.addr,
  suggestedParams,
  execParams.appId,
  parseSSCAppArgs(execParams.appArgs), // remains undefined if appArgs are undefined
  execParams.accounts,
  execParams.foreignApps,
  execParams.foreignAssets,
  note,
  execParams.lease,
  execParams.rekeyTo);
```

robert-zaremba: Please update function docs and explain that.

ratik21: Ok

PR #199, opened 2021-01-15
fix: algosdk type

ratik21:
While renaming devUx we updated types in `algosdk` as well. We shouldn't update keys in `@types/algosdk` (they are written as per sdk).

### DIFF
--- a/packages/algob/src/lib/status.ts
+++ b/packages/algob/src/lib/status.ts
@@ -28,7 +28,7 @@ export async function readGlobalStateSSC (
   appId: number): Promise<SSCStateSchema[] | undefined> {
   const accountInfoResponse = await deployer.algodClient.accountInformation(creator).do();
   for (const app of accountInfoResponse['created-apps']) {
-    if (app.id === appId) { return app.attributes['global-state']; }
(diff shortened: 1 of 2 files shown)


packages/algob/src/lib/status.ts
```diff
@@ -28,7 +28,7 @@ export async function readGlobalStateSSC (
   appId: number): Promise<SSCStateSchema[] | undefined> {
   const accountInfoResponse = await deployer.algodClient.accountInformation(creator).do();
   for (const app of accountInfoResponse['created-apps']) {
-    if (app.id === appId) { return app.attributes['global-state']; }
+    if (app.id === appId) { return app.params['global-state']; }
```
robert-zaremba, line 31: Hmm,  I thought we decided to use attributes?

ratik21: Yes, in `Runtime` we are using `attributes`. 
But `algosdk` uses `params`( i.e `app.params` is defined in accountInfoResponse,  not `app.attributes`).

robert-zaremba (approved):

PR #200, opened 2021-01-18
Update storage

amityadav0:
- Verify we have only single data source
- Set clear boundaries for `ctx` and `store`
- Fix tests


packages/runtime/src/interpreter/interpreter.ts
```diff
@@ -26,6 +37,86 @@ export class Interpreter {
...
+    const accAddress = this.runtime.assertAddressDefined(
+      this.runtime.ctx.state.globalApps.get(appId));
+    const account = this.runtime.ctx.state.accounts.get(accAddress);
+    return account?.createdApps.get(appId);
+  }
+
+  /**
+   * Fetch account using accountIndex from `Accounts` list
+   * Accounts: List of accounts in addition to the sender
+   * that may be accessed from the application's approval-program and clear-state-program.
```
robert-zaremba, line 71:
```suggestion
   * Queries account by accountIndex or `ctx.tx.snd` (if `accountIndex==0`).
   * Throws exception if account is not found.
```

packages/runtime/src/interpreter/interpreter.ts
```diff
@@ -1,13 +1,24 @@
...
-import { BIGINT0, Label } from "./opcode-list";
+import type { Operator, SSCAttributesM, StackElem, StoreAccountI, TEALStack } from "../types";
+import { BIGINT0, BIGINT1, Label } from "./opcode-list";
+
+const globalState = "global-state";
 
 export class Interpreter {
+  /**
+   * Note: Interpreter operates on only ctx, current context of runtime
+   * All the functions fetches and sets the values only in current context, not in `runtime.store`.
```
robert-zaremba, line 20:
```suggestion
   * All the functions query or update only a state copy from the interpreter, not the `runtime.store`.
```

packages/runtime/src/interpreter/interpreter.ts
```diff
@@ -26,6 +37,86 @@ export class Interpreter {
...
+    const globalKey = keyToBytes(key);
+    return appGlobalState.get(globalKey.toString());
+  }
+
+  /**
+   * Add new key-value pair or updating pair with existing key in
+   * app's global data for application id: appId, throw error otherwise
+   * @param appId: current application id
+   * @param key: key to fetch value of from local state
+   * @param value: key to fetch value of from local state
```
robert-zaremba, line 108:
```suggestion
   * Upserts app global state.
   * Throws error if app is not found.
   * @param appId: application id
   * @param key: app global state key
   * @param value: value associated with a key
```

robert-zaremba (commented):
Good catch - interpreter state should be used rather than a runtime one.

packages/runtime/src/interpreter/interpreter.ts
```diff
@@ -1,13 +1,24 @@
...
-import type { Operator, StackElem, TEALStack } from "../types";
-import { BIGINT0, Label } from "./opcode-list";
+import type { Operator, SSCAttributesM, StackElem, StoreAccountI, TEALStack } from "../types";
+import { BIGINT0, BIGINT1, Label } from "./opcode-list";
+
+const globalState = "global-state";
 
 export class Interpreter {
+  /**
+   * Note: Interpreter operates on only ctx, current context of runtime
```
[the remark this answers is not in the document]
amityadav0: ok, changing

packages/runtime/src/interpreter/interpreter.ts
```diff
@@ -26,6 +37,86 @@ export class Interpreter {
     this.runtime = <Runtime>{};
   }
 
+  /**
+   * Fetches Asset Definition for given asset index from current context (ctx)
```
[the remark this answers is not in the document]
amityadav0: ok

packages/runtime/src/interpreter/interpreter.ts
```diff
@@ -26,6 +37,86 @@ export class Interpreter {
...
+    account = this.runtime.assertAccountDefined(account);
+
+    return account.createdAssets.get(assetId);
+  }
+
+  /**
+   * fetches app from `this.store`
+   * @param appId Application Index
+   */
+  getApp (appId: number, line: number): SSCAttributesM | undefined {
```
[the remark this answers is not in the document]
amityadav0: we need `undefined` here because for some opcodes if app is not defined they push 0 to stack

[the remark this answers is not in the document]
ratik21: [`SSCAttributesM`](https://github.com/scale-it/algorand-builder/blob/master/packages/runtime/src/types.ts#L86) represents [`SSCAttributes`](https://github.com/scale-it/algorand-builder/blob/master/packages/types-algosdk/index.d.ts#L549) from algosdk, with one stipulation: we are using `maps` instead of `arrays` ( in `'global-state'` here). 
Added the suffix `M` as we are using maps. We can rename to something more meaningful if not clear.

packages/runtime/src/runtime.ts
```diff
@@ -62,55 +63,67 @@ export class Runtime {
...
   /**
    * asserts if application exists in state
+   * @param app application
    * @param appId application index
    * @param line line number in TEAL file
    * Note: if user is accessing this function directly through runtime,
    * the line number is unknown
    */
-  assertAppDefined (appId: number, line?: number): SSCAttributesM {
+  assertAppDefined (app: SSCAttributesM | undefined, appId: number, line?: number): SSCAttributesM {
```
ratik21, line 88:
```suggestion
  assertAppDefined (app?: SSCAttributesM, appId: number, line?: number): SSCAttributesM {
```
We are using `| undefined` in 1st arg and `?` in 3rd

packages/runtime/src/runtime.ts
```diff
@@ -62,55 +63,67 @@ export class Runtime {
...
-      const pkBuffer = this.ctx.tx.apat[Number(accIndex)];
-      account = this.ctx.state.accounts.get(encodeAddress(pkBuffer));
+  getApp (appId: number): SSCAttributesM {
+    if (!this.store.globalApps.has(appId)) {
+      throw new TealError(ERRORS.TEAL.APP_NOT_FOUND, { appId: appId, line: 'unknown' });
     }
+    const accAddress = this.assertAddressDefined(this.store.globalApps.get(appId));
+    const account = this.store.accounts.get(accAddress);
+    const app = account?.createdApps.get(appId);
+    return this.assertAppDefined(app, appId);
```
ratik21, line 107: Please create a function `getApp` in `StoreAccount`. We should operate on an account attribute (`createdApps` in this case) in account only. So `runtime.getApp() => calls account.getApp()` which get's the app from account's global state.

packages/runtime/src/runtime.ts
```diff
@@ -62,55 +63,67 @@ export class Runtime {
...
    * @param key: key to fetch value of from local state
    */
   getGlobalState (appId: number, key: Uint8Array | string): StackElem | undefined {
-    // TODO: will be updated in https://www.pivotaltracker.com/story/show/176487715
-    // we will operate on accounts rather than globalApp map
-    const app = this.assertAppDefined(appId);
+    const app = this.getApp(appId);
     const appGlobalState = app["global-state"];
     const globalKey = keyToBytes(key);
     return appGlobalState.get(globalKey.toString());
```
ratik21, line 129: Same as above. `runtime.getGlobalState => calls account.getGlobalState`. We can check the getLocalState function below (I think we discussed these updates in the `improve DevUx` task, but at that time we were operating on `globalApps` only and not `store.accounts`.

packages/runtime/src/runtime.ts
```diff
@@ -21,6 +21,7 @@ export class Runtime {
    * because of faster and easy querying.
    * This way when querying, instead of traversing the whole object,
    * we can get the value directly from Map
+   * Note: Runtime operates on `store`, it doesn't operate on `ctx`.
```
ratik21, line 24: :+1:

packages/runtime/src/runtime.ts
```diff
@@ -127,26 +140,6 @@ export class Runtime {
...
   }
 
-  /**
-   * Add new key-value pair or updating pair with existing key in
-   * app's global data for application id: appId, throw error otherwise
-   * @param appId: current application id
-   * @param key: key to fetch value of from local state
-   * @param value: key to fetch value of from local state
-   */
-  setGlobalState (appId: number, key: Uint8Array | string, value: StackElem): Map<string, StackElem> {
```
ratik21, line 137: Should we remove this function from runtime? Is it possible that during testing user wants to set global state explicitly before calling the smart contract?

packages/runtime/src/runtime.ts
```diff
@@ -155,13 +148,13 @@ export class Runtime {
...
+      for (const appId of acc.createdApps.keys()) {
+        this.store.globalApps.set(appId, acc.address);
+      }
 
-      acc.createdAssets.forEach((params, assetIndex) => {
-        this.store.assetDefs.set(assetIndex, params);
-      });
+      for (const assetId of acc.createdAssets.keys()) {
+        this.store.assetDefs.set(assetId, acc.address);
+      }
```
ratik21, line 157: We have `initializeAccounts()` in runtime constructor. But currently we create an app/asset after runtime initialization. eg.
```
const runtime = new Runtime([john, bob]); // sets this.store.accounts
runtime.addApp({..}), runtime.updateApp({..})..etc afterwards. 
```
Since we don't have any created apps or assets during initialization of runtime, do lines 151-153, 155-157 ever get executed?

packages/runtime/src/runtime.ts
```diff
@@ -225,29 +218,35 @@ export class Runtime {
...
+   */
   async addApp (flags: SSCDeploymentFlags, payFlags: TxParams, program: string): Promise<number> {
     const sender = flags.sender;
     const senderAcc = this.assertAccountDefined(this.store.accounts.get(sender.addr));
 
     // create app with id = 0 in globalApps for teal execution
     const app = senderAcc.addApp(0, flags);
-    this.ctx.state.globalApps.set(app.id, app.attributes);
+    this.ctx.state.accounts.set(senderAcc.address, senderAcc);
+    this.ctx.state.globalApps.set(app.id, senderAcc.address);
```
ratik21, line 234: Please don't operate on `ctx` in Runtime. We are directly contradicting the note above the runtime class here. Let's be consistent with our definition: `interpreter operates on ctx (using deepCopy of store) and runtime operates on store`. 

So, we can do these updates on `runtime.store` (if needed), and then assign a deepCopy to `ctx` before `run`. eg.
```
this.store.accounts.set(senderAcc.address, senderAcc);
this.store.globalApps.set(app.id, senderAcc.address);
this.ctx.state = deepCopy(this.store) // assign a deep copy here instead of setting directly
await this.run(..)
..
```

packages/runtime/src/runtime.ts
```diff
@@ -225,29 +218,35 @@ export class Runtime {
...
     this.ctx.state.globalApps.delete(0); // remove zero app from context after execution
     senderAcc.createdApps.delete(0); // remove zero app from sender's account
 
-    // set new application in sender's account
-    // after setting in globalApps
-    // TODO: will be updated in https://www.pivotaltracker.com/story/show/176487715
-    senderAcc.addApp(this.appCounter, flags);
+    senderAcc.createdApps.set(this.appCounter, attributes);
     this.store.accounts.set(sender.addr, senderAcc);
+    this.ctx.state.accounts.set(senderAcc.address, senderAcc);
```
ratik21, line 248: Same as above. Also, do we need this line (keeping `ctx` after transaction is executed) ? I mean `ctx` will be overwritten on `runtime.executeTx` and in `Runtime` we are using `store` to getAccount.

packages/runtime/src/runtime.ts
```diff
@@ -346,18 +344,23 @@ export class Runtime {
...
-    const accountAddr = this.store.globalApps.get(appId)?.creator;
+    const accountAddr = this.store.globalApps.get(appId);
     if (accountAddr === undefined) {
       throw new TealError(ERRORS.TEAL.ACCOUNT_DOES_NOT_EXIST);
     }
     const account = this.assertAccountDefined(this.store.accounts.get(accountAddr));
+
     account.deleteApp(appId);
     this.store.globalApps.delete(appId);
+    this.store.accounts.set(account.address, account);
```
ratik21, line 363: is `this.assertAccountDefined(this.store.accounts.get(accountAddr))` returning a deepCopy ?

packages/runtime/test/integration/deleteApp.ts
```diff
@@ -63,8 +65,9 @@ describe("Algorand Smart Contracts - Delete Application", function () {
       async () => await runtime.executeTx(deleteParams, program, []),
       ERRORS.TEAL.REJECTED_BY_LOGIC
     );
-    // verify app is not deleted
-    const res = runtime.ctx.state.globalApps.has(appId);
-    assert.equal(res, true);
+
+    // verify app is not deleted - use app get function
```
ratik21, line 69:
```suggestion
    // verify app is not deleted - using getApp function
```

packages/runtime/test/src/interpreter/opcode-list.ts
```diff
@@ -2348,13 +2344,10 @@ describe("Teal Opcodes", function () {
...
         const acc = interpreter.runtime.ctx.state.accounts.get(elonAddr) as StoreAccountI;
-        let localStateCurr = acc.appsLocalState.get(appId)?.[keyValue];
-        assert.isDefined(localStateCurr);
-        if (localStateCurr) {
-          value = localStateCurr.get(convertToKey('New-Key'));
-          assert.isDefined(value);
-          if (value) assert.deepEqual(value, stringToBytes('New-Val'));
-        }
+
+        value = acc.getLocalState(appId, 'New-Key');
```
ratik21, line 2348: Should we use `runtime.getLocalState` (Assuming user will do the same)?

packages/runtime/src/runtime.ts
amityadav0: yes

packages/runtime/src/runtime.ts
amityadav0: I don't think that's ideal, User will be testing their TEAL files, not our code. If that makes sense

packages/runtime/src/runtime.ts
amityadav0: yes, they do, when we call the constructor new Runtime([john, bob]). and if john and bob, have createdAssets or createdApps, we are mocking these.

packages/runtime/src/runtime.ts
amityadav0: we cannot do that, for deployer functions(addApp, updateApp).
because what if program gets failed, you would have updated the store already.
but according to spec, we cannot do that, first ctx will be updated, then only we move to store

also this is similar to what we do in `run` first we set the ctx and then we pass it to interpreter, and if everything goes correct, we update store.(it is more clear and avoids bug)
I can add a detailed doc for more clarification.

packages/runtime/src/runtime.ts
amityadav0: yes, similar to addApp, did you figure what was going on with addAPp ?

packages/runtime/src/runtime.ts
amityadav0: same as above, yes, we don't need it.

packages/runtime/src/runtime.ts
amityadav0: One other way would be moving appApp, updateApp to interpreter.

packages/runtime/src/runtime.ts
amityadav0: yes

packages/runtime/src/runtime.ts
ratik21: hmm, then we need to figure something out. The current implemention seems incorrect by definition.

packages/runtime/src/runtime.ts
ratik21: When i tested `addApp` it wasn't returning a deepCopy. That's why i asked here

packages/runtime/src/runtime.ts
ratik21: how can john or bob create new asset or app before initializing runtime?

packages/runtime/src/runtime.ts
ratik21: > because what if program gets failed, you would have updated the store already.

Agreed. The state would be updated even if program fails. But i see in the below code, we are deleting this app from the account (since this is a zero indexed app, we only need it's attribute and set new state)

packages/runtime/src/runtime.ts
amityadav0: If you see we have mocks, which has hard coded createdAssets, this is used for that,
I haven't structured it like that, I think it was added by you. We can update in new PR

packages/runtime/src/runtime.ts
amityadav0: One way would be to pass ctx to intrpreter.addApp and let it handle this and return updated ctx
Which we can update in runtime.addApp

packages/runtime/src/runtime.ts
amityadav0: I think it's good for user, user can send updated states if he wishes.
We can discuss it's pros and cons.

ratik21: I wasn't saying to remove it, was rather curious when this does gets executed.
For mocks the structure seems fine, as we set in globalassets and App directly there. From users perspective, I only see this code getting executed when the user explicitly modifies createdAssets or Map by using account.createdApps.set(..)..

packages/runtime/src/runtime.ts
amityadav0: verified it, working after removing it, will have a look at each function for this too.

packages/runtime/src/runtime.ts
amityadav0: > > because what if program gets failed, you would have updated the store already.
> 
> Agreed. The state would be updated even if program fails. But i see in the below code, we are deleting this app from the account anyway (since this is a zero indexed app, we only need it's attribute and set new state with id appCounter + 1). Will check more

if program fails, error will be thrown therefore we won't reach that line. store will be updated.

amityadav0: also i don't think we can move it to interpreter,, that will create cycle. best would be to add doc. what do you think?

packages/runtime/src/account.ts
```diff
@@ -85,6 +85,28 @@ export class StoreAccount implements StoreAccountI {
     });
   }
 
+  /**
+   * Fetches global state value for key present in account
+   * returns undefined otherwise
```
robert-zaremba, line 90:
```suggestion
   * Queries app global state value. Returns `undefined` if the key is not present.
```

packages/runtime/src/account.ts
```diff
@@ -85,6 +85,28 @@ export class StoreAccount implements StoreAccountI {
...
+  getGlobalState (appId: number, key: Uint8Array | string): StackElem | undefined {
+    const app = this.getApp(appId);
+    if (!app) return undefined;
+    const appGlobalState = app["global-state"];
+    const globalKey = keyToBytes(key);
+    return appGlobalState.get(globalKey.toString());
+  }
+
+  /**
+   * Queries application by application index
```
robert-zaremba, line 103:
```suggestion
   * Queries application by application index. Returns undefined if app is not found.
```

packages/runtime/src/interpreter/interpreter.ts
```diff
@@ -1,13 +1,24 @@
...
+import type { Operator, SSCAttributesM, StackElem, StoreAccountI, TEALStack } from "../types";
+import { BIGINT0, BIGINT1, Label } from "./opcode-list";
+
+const globalState = "global-state";
 
 export class Interpreter {
+  /**
+   * Note: Interpreter operates on `ctx`, it doesn't operate on `store`.
+   * All the functions query or update only a state copy from the interpreter, not the `runtime.store`.
+   */
```
robert-zaremba, line 21:
```suggestion
   */
   
```

packages/runtime/src/interpreter/interpreter.ts
```diff
@@ -26,6 +37,85 @@ export class Interpreter {
     this.runtime = <Runtime>{};
   }
 
+  /**
+   * Queries ASA Definitions data by assetID
```
robert-zaremba, line 41:
```suggestion
   * Queries ASA Definitions data by assetID.  Returns undefined if ASA is not deployed.
```

packages/runtime/src/interpreter/interpreter.ts
```diff
@@ -26,6 +37,85 @@ export class Interpreter {
...
+    if (!accountAddr) return undefined;
+
+    let account = this.runtime.ctx.state.accounts.get(accountAddr);
+    account = this.runtime.assertAccountDefined(account);
+
+    return account.createdAssets.get(assetId);
+  }
+
+  /**
+   * Queries app (SSCAttributesM) from state. Throws TEAL.APP_NOT_FOUND if app is not found.
```
robert-zaremba, line 55: Update the second part: explain when the app returns undefined and when it throws exception.

packages/runtime/src/interpreter/interpreter.ts
```diff
@@ -26,6 +37,85 @@ export class Interpreter {
...
+      checkIndexBound(Number(accIndex), this.runtime.ctx.tx.apat, line);
+      const pkBuffer = this.runtime.ctx.tx.apat[Number(accIndex)];
+      account = this.runtime.ctx.state.accounts.get(encodeAddress(pkBuffer));
+    }
+    return this.runtime.assertAccountDefined(account, line);
+  }
+
+  /**
+   * Fetches global state value for key present in creator's global state
+   * for given appId, returns undefined otherwise
```
robert-zaremba, line 91:
```suggestion
   * Queries application by application index. Returns undefined if app is not found.
```

robert-zaremba (approved):
Pre-approving. Left comments and suggestions. Please also wait for @ratik21  approval to merge.

packages/runtime/src/runtime.ts
ratik21: It just seems to me we're taking control away from the user. Eg For a stateful counter application, the user can use `runtime.setGlobalState('counter', 5)` explicitly, and then call the smart contract to assert if counter has incremented by 1 or not. What's the benefit of removing it? 
Anyway, if you will remove it, please update the `/docs` as well. It is mentioned there user can use `setGlobalState`.

packages/runtime/src/account.ts
```diff
@@ -85,6 +85,28 @@ export class StoreAccount implements StoreAccountI {
     });
   }
 
+  /**
+   * Fetches global state value for key present in account
+   * returns undefined otherwise
+   * @param appId: current application id
+   * @param key: key to fetch value of from local state
+   */
+  getGlobalState (appId: number, key: Uint8Array | string): StackElem | undefined {
```
ratik21, line 94: Just to confirm, we haven't removed this function from `runtime` right? `runtime.getGlobalState` will call `account.getGlobalState`. Right?

amityadav0: yes

packages/runtime/src/runtime.ts
ratik21: > if program fails, error will be thrown therefore we won't reach that line. store will be updated.

We can wrap the `run` in try {} catch () and throw error after deleting ?
Anyways, adding doc seems to be OK here. Please explain why are operating on `ctx` here in a doc string.

packages/runtime/src/runtime.ts
amityadav0: fair enough, restoring setGlobalState

packages/runtime/src/runtime.ts
```diff
@@ -225,29 +215,34 @@ export class Runtime {
...
+    const attributes = this.assertAppDefined(0, senderAcc.createdApps.get(0));
     this.ctx.state.globalApps.delete(0); // remove zero app from context after execution
     senderAcc.createdApps.delete(0); // remove zero app from sender's account
 
-    // set new application in sender's account
-    // after setting in globalApps
-    // TODO: will be updated in https://www.pivotaltracker.com/story/show/176487715
-    senderAcc.addApp(this.appCounter, flags);
+    senderAcc.createdApps.set(this.appCounter, attributes);
     this.store.accounts.set(sender.addr, senderAcc);
```
ratik21, line 244: Also, since `this.assertAccountDefined(this.store.accounts.get(sender.addr))` is not returning a deepCopy, do we need this line?

amityadav0: ok, removing it, please make sure tests are good for this one.

ratik21 (approved):
Please check `/docs/testing-teal.md` as well before merging

packages/runtime/src/runtime.ts
ratik21: checking

packages/runtime/src/runtime.ts
amityadav0: ok

packages/runtime/src/runtime.ts
ratik21: > ok, removing it, please make sure tests are good for this one.

Please check at other places as well

ratik21: Seems ok

PR #201, opened 2021-01-19
Update interpreter to calculate cost of the program - for all opcodes

ratik21:
- Updated parser to calculate gas for `crypto` opcodes.
- Added tests


packages/runtime/src/interpreter/opcode-list.ts
```diff
@@ -535,11 +537,13 @@ export class Sha256 extends Op {
...
    * @param args Expected arguments: [] // none
    * @param line line number in TEAL file
+   * @param interpreter interpreter object
    */
-  constructor (args: string[], line: number) {
+  constructor (args: string[], line: number, interpreter: Interpreter) {
     super();
     this.line = line;
     assertLen(args.length, 0, line);
+    interpreter.gas += interpreter.tealVersion === 1 ? 7 : 35;
```
amityadav0, line 546: can we use a map for gas ? Ex: version->opcodes-> gas (maybe two maps one for version1 and other for 2).
- They are more clear
- Easily maintainable(changing few according to current code is tedious task)
- Also values are constant, so they are unrelated to opcode logic.
what are the benefits of embedding it in constructor ?

packages/runtime/test/src/parser/parser.ts
```diff
@@ -898,4 +900,61 @@ describe("Parser", function () {
...
+      // eslint-disable-next-line
+      op = new Ed25519verify([], 4, interpreter);
+      assert.equal(interpreter.gas, 1900);
+
+      interpreter.gas = 0;
+      await parser(getProgram(cryptoFile), interpreter);
+      assert.deepEqual(interpreter.gas, 1942); // 7 + 26 + 9 + 1900
+    });
+
+    it("Should return correct gas cost for 'Crypto opcodes' for tealversion 2", async () => {
```
amityadav0, line 936: should we test constant map values ? I don't think that are going to change or their is any logic in them.
(just a thought) for logic parts we need more tests.

packages/runtime/src/interpreter/opcode-list.ts
ratik21: Yes. Also, only crypto ops have different values for v1 and v2. Rest all the opcodes have a gas cost of 1. So with them in constructor, we could just do `interpreter.gas++`, after verifying min version. We can create a 2 maps for crypto ops (v1, v2).

packages/runtime/test/src/parser/parser.ts
ratik21: Probably not. I don't see any downside of not doing it. Should i remove?

packages/runtime/src/interpreter/opcode-list.ts
amityadav0: Is it possible to move calculation out of constructor ? This way updating will be easier, let's say version 3 comes.

We can have one function which will add up gas using map, and store it, without going to each opcode constructor.

packages/runtime/test/src/parser/parser.ts
amityadav0: No, it's neutral to me.

packages/runtime/src/interpreter/opcode-list.ts
ratik21: We can take the calculation out of the constructor by moving gas cost calculation in the parser itself. I am not sure about using maps though, as most of the opcodes have cost 1 in both version. Maybe use a Set?

packages/runtime/test/src/parser/parser.ts
ratik21: Ok, will keep then.

ratik21: Ok, will keep then.

packages/runtime/src/interpreter/opcode-list.ts
amityadav0: Yes we can use set. Or check for opcode in map(it will only have opcodes that has cost other than 1).  If it is not in map we can add 1. We can put this logic in parser where we already checked if opcodes is known.

ratik21: Ok

packages/runtime/src/parser/parser.ts
```diff
@@ -267,7 +286,7 @@ export async function parser (program: string, interpreter: Interpreter): Promis
   let counter = 0;
 
   const lines = program.split('\n');
-  for await (const line of lines) {
```
amityadav0, line 270: i think we can remove Promise as well. can you please confirm ?

packages/runtime/src/parser/parser.ts
```diff
@@ -280,5 +299,9 @@ export async function parser (program: string, interpreter: Interpreter): Promis
       opCodeList.push(opcodeFromSentence(words, counter, interpreter));
     }
   }
+
+  if (interpreter.gas > LogicSigMaxCost) {
```
amityadav0, line 303: let's add a documentation about when are we calculating gas, at suitable place.

packages/runtime/src/errors/errors-list.ts
```diff
@@ -207,9 +207,15 @@ by an index that does not exist.`
   },
   UNKOWN_OPCODE: {
     number: 31,
-    message: "Error encountered while parsing teal file unkown opcode %opcode% , Line : %line% ",
+    message: "Error encountered while parsing teal file: unkown opcode \"%opcode%\" for teal version #%version%, Line : %line% ",
     title: PARSE_ERROR,
     description: `Unkown Opcode encountered`
+  },
+  INVALID_LOGICSIG_MAX_COST: {
```
amityadav0, line 214: Is this only for  logic signature or stateful too ? or are they same in this case ?

amityadav0: https://developer.algorand.org/docs/reference/teal/specification/
just saw this section In `Execution Mode`, it is mentioned we have different COST for both MaxAppProgramCost and LogicSigMaxCost. can you please confirm if their values are same ?

amityadav0: also there is Max program length (consensus parameters LogicSigMaxSize, MaxApprovalProgramLen and MaxClearStateProgramLen) different for different programs

ratik21: I was checking the statement above the `Execution Mode` and thought it was generic.
image

I think you are right, starting from version 2, stateless and stateful teal have different max cost. But i had question regarding this, currently can we know in parser whether TEAL is stateless or stateful?

packages/runtime/src/parser/parser.ts
ratik21: Alright

packages/runtime/src/errors/errors-list.ts
amityadav0: I don't think that's a good idea, because clear program doesn't have ant stateful opcode but it is stateful.(possible for approval too).
One other way would be:
I think we can run the parser to store interpreter.gas(as we do now). the only way to know if current TEAL is stateless, approval or clear is via execParams, We can add a check in interpreter which will throw error according to TransactionType or SignType

amityadav0: after parsing is done, we will have interpreter.gas, which we can use in interpreter to check for max gas, and throw error accordingly

amityadav0: first we can check for `appID` if it is passed, then program is approval or  clear
if not we can check for SignType
If it is logicSig then we proceed with LogicSigMaxCost.

ratik21: I don't think checking by `appId` is a good approach (it's not intuitive)
Extending you previous comment, we can do:
if SignType is LogicSignature, then it's stateless TEAL - `LogicSigMaxCost`, `LogicSigMaxSize`
If not, we can check if TransactionType is `ClearSSC` (For clear program) - `MaxAppProgramCost`, `MaxApprovalProgramLen`
if TransactionType is `NoOp | CloseSSC | DeleteSSC` (For approval program) - `MaxAppProgramCost`, `MaxClearStateProgramLen`

ratik21: Also, i think we can support this for a single tx for now. For group transaction we currently support only passing a single TEAL file (for stateless + stateful), whereas we should be able to pass teal code with each stateless tx (as they only approve/reject a txn).

packages/runtime/src/parser/parser.ts
ratik21: yes

packages/runtime/src/interpreter/interpreter.ts
```diff
@@ -8,16 +8,18 @@ import { DEFAULT_STACK_ELEM } from "../lib/constants";
...
 export class Interpreter {
   /**
    * Note: Interpreter operates on `ctx`, it doesn't operate on `store`.
    * All the functions query or update only a state copy from the interpreter, not the `runtime.store`.
    */
-
   readonly stack: TEALStack;
+  tealVersion: number; // LogicSigVersion
+  gas: number; // total gas cost of TEAL code
+  length: number; // total length of TEAL code
```
robert-zaremba, line 22: How this length is expressed? In number of op codes? If yes, how about renaming to `opCodesNum`?

packages/runtime/src/lib/constants.ts
```diff
@@ -8,36 +8,51 @@ export const MAX_CONCAT_SIZE = 4096;
 export const SSC_BYTES = 1; // 1 represents type 'byte' in stateful key-value pair
 export const SSC_UINT = 2; // 2 represents type 'uint' in stateful key-value pair
 export const ALGORAND_MIN_TX_FEE = 1000;
+
+// values taken from https://github.com/algorand/go-algorand/find/master#L689
```
robert-zaremba, line 12: Could you provide a link to a source file?

packages/runtime/src/errors/errors-list.ts
```diff
@@ -207,9 +207,21 @@ by an index that does not exist.`
   },
   UNKOWN_OPCODE: {
     number: 31,
-    message: "Error encountered while parsing teal file unkown opcode %opcode% , Line : %line% ",
+    message: "Error encountered while parsing teal file: unkown opcode \"%opcode%\" for teal version #%version%, Line : %line% ",
     title: PARSE_ERROR,
     description: `Unkown Opcode encountered`
+  },
+  INVALID_MAX_COST: {
```
amityadav0, line 214: can we rename  to `MAX_COST_EXCEEDED` or something else ?

packages/runtime/src/errors/errors-list.ts
```diff
@@ -207,9 +207,21 @@ by an index that does not exist.`
...
+    number: 32,
+    message: "Cost of provided TEAL code = %cost% exceeds max cost of %maxcost%, Mode: %mode%",
+    title: 'MaxCost Error',
+    description: `MaxCost Error`
+  },
+  INVALID_MAX_LEN: {
```
amityadav0, line 220: same

packages/runtime/src/runtime.ts
```diff
@@ -396,11 +397,19 @@ export class Runtime {
...
+    }
+    return ExecutionMode.STATEFUL;
+  }
+
   /**
    * This function executes a transaction based on a smart
    * contract logic and updates state afterwards
    * @param txnParams : Transaction parameters
-   * @param fileName : smart contract file (.teal) name in assets/
+   * @param program : teal code as string
```
robert-zaremba, line 412:
```suggestion
   * @param program : teal code as a string
```

packages/runtime/src/lib/constants.ts
```diff
@@ -8,36 +8,51 @@ export const MAX_CONCAT_SIZE = 4096;
...
+export const LogicSigMaxCost = 20000;
+export const MaxAppProgramCost = 700;
+export const LogicSigMaxSize = 1000;
+export const MaxAppProgramLen = 1024;
```
amityadav0, line 16: just to confirm. we don't have this for `MaxClearProgramLen` right ? same for cost

packages/runtime/src/errors/errors-list.ts
ratik21: Ok, will change to `MAX_COST_EXCEEDED`

packages/runtime/src/errors/errors-list.ts
ratik21: ok

packages/runtime/src/lib/constants.ts
ratik21: Yes, i couldn't find specfic max lengths for approval and clear programs in go-algorand. Please let me know if you will find it.

packages/runtime/src/parser/parser.ts
```diff
@@ -276,9 +342,13 @@ export async function parser (program: string, interpreter: Interpreter): Promis
...
     // Trim whitespace from line and extract words from line
     const words = wordsFromLine(line);
+    interpreter.length += words.join('').length; // calculate length of TEAL code
     if (words.length !== 0) {
       opCodeList.push(opcodeFromSentence(words, counter, interpreter));
     }
   }
+
+  assertMaxCost(interpreter.gas, mode);
+  // assertMaxLen(interpreter.length, mode); // TBD
```
amityadav0, line 352: can we uncomment it ?

packages/runtime/src/parser/parser.ts
```diff
@@ -241,33 +256,84 @@ export function opcodeFromSentence (words: string[], counter: number, interprete
...
+        cost: gas,
+        maxcost: MaxAppProgramCost,
+        mode: 'Stateful'
+      });
+    }
+  }
+}
+
+// verify max length of TEAL code is within consensus parameters
+function assertMaxLen (len: number, mode: ExecutionMode): void {
```
amityadav0, line 303: does this `len` includes blank lines or just the opcodes ?

packages/runtime/src/errors/errors-list.ts
```diff
@@ -189,13 +189,13 @@ by an index that does not exist.`
   },
   UNKOWN_TRANSACTION_FIELD: {
```
amityadav0, line 190:
```suggestion
  UNKNOWN_TRANSACTION_FIELD: {
```

packages/runtime/src/errors/errors-list.ts
```diff
@@ -189,13 +189,13 @@ by an index that does not exist.`
...
     number: 28,
-    message: "Transaction Field Error - Unknown transaction field %field% at %line%",
+    message: "Transaction Field Error - Unknown transaction field \"%field%\" for teal version #%version% at line %line%",
     title: "Transaction Field Error at %line%",
     description: `Transaction Field unknown`
   },
   UNKOWN_GLOBAL_FIELD: {
```
amityadav0, line 196:
```suggestion
  UNKNOWN_GLOBAL_FIELD: {
```

packages/runtime/src/errors/errors-list.ts
```diff
@@ -207,9 +207,21 @@ by an index that does not exist.`
   },
   UNKOWN_OPCODE: {
```
amityadav0, line 208:
```suggestion
  UNKNOWN_OPCODE: {
```

packages/runtime/src/errors/errors-list.ts
```diff
@@ -207,9 +207,21 @@ by an index that does not exist.`
...
     number: 31,
-    message: "Error encountered while parsing teal file unkown opcode %opcode% , Line : %line% ",
+    message: "Error encountered while parsing teal file: unkown opcode \"%opcode%\" for teal version #%version%, Line : %line% ",
```
amityadav0, line 210:
```suggestion
    message: "Error encountered while parsing teal file: unknown opcode \"%opcode%\" for teal version #%version%, Line : %line% ",
```

packages/runtime/src/errors/errors-list.ts
```diff
@@ -207,9 +207,21 @@ by an index that does not exist.`
...
     title: PARSE_ERROR,
     description: `Unkown Opcode encountered`
```
amityadav0, line 212:
```suggestion
    description: `Unknown Opcode encountered`
```

packages/runtime/src/lib/constants.ts
amityadav0: i think it's same for both.

amityadav0: i think it's same for both.

packages/runtime/src/interpreter/interpreter.ts
```diff
@@ -134,13 +139,14 @@ export class Interpreter {
   }
 
   /**
-   * Description: this function executes TEAL code after parsing
-   * @param {string} program: teal code
-   * @param {Runtime} runtime : runtime object
+   * This function executes TEAL code after parsing
+   * @param program: teal code
+   * @param mode : execution mode of TEAL code (Stateless or Application)
```
amityadav0, line 144:
```suggestion
   * @param mode : execution mode of TEAL code (Stateless or Stateful)
```

packages/runtime/src/lib/constants.ts
```diff
@@ -8,36 +8,51 @@ export const MAX_CONCAT_SIZE = 4096;
...
+
 const zeroAddress = new Uint8Array(32);
 const zeroUint64 = BigInt('0');
 const zeroByte = new Uint8Array(0);
 
 // https://developer.algorand.org/docs/reference/teal/opcodes/#txn
```
amityadav0, line 22:
```suggestion
// https://developer.algorand.org/docs/reference/teal/opcodes/
```

packages/runtime/src/parser/parser.ts
```diff
@@ -241,33 +256,84 @@ export function opcodeFromSentence (words: string[], counter: number, interprete
...
   }
 
-  if (opCodeMap[opCode] === undefined) {
-    throw new TealError(ERRORS.TEAL.UNKOWN_OPCODE, { opcode: opCode, line: counter });
+  if (opCodeMap[tealVersion][opCode] === undefined) {
+    throw new TealError(ERRORS.TEAL.UNKOWN_OPCODE, { opcode: opCode, version: tealVersion, line: counter });
   }
 
+  // increment gas of TEAL code
+  // select from OpGasCost map if cost is different than 1
```
amityadav0, line 270:
```suggestion
  // Add 1 if opCode is not present in OpGasCost map
```

packages/runtime/test/src/parser/parser.ts
```diff
@@ -766,30 +777,27 @@ describe("Parser", function () {
...
       const expected = [new Store(["0"], 1, interpreter), new Load(["0"], 2, interpreter)];
 
       assert.deepEqual(res, expected);
     });
 
     it("Should return correct opcode list for 'Crypto opcodes'", async () => {
-      const file = "test-crypto.teal";
-
-      const res = await parser(getProgram(file), interpreter);
+      const res = await parser(getProgram(cryptoFile), ExecutionMode.STATELESS, interpreter);
```
amityadav0, line 787:
```suggestion
      const res = parser(getProgram(cryptoFile), ExecutionMode.STATELESS, interpreter);
```

packages/runtime/src/lib/constants.ts
ratik21: Yes, sure

packages/runtime/src/interpreter/interpreter.ts
ratik21: It's actually the length of compiled teal code + arguments (for stateless) and 1kb for stateful execution mode. Not sure about how to calculate this, as we're not using the TEAL compiler.

packages/runtime/src/parser/parser.ts
ratik21: It's actually the length of compiled teal code + arguments (for stateless) and 1kb for stateful execution mode. Not sure about how to calculate this, as we're not using the TEAL compiler.

packages/runtime/src/interpreter/interpreter.ts
ratik21: Added a new task to think about it: [https://www.pivotaltracker.com/story/show/176623588](https://www.pivotaltracker.com/story/show/176623588)

packages/runtime/src/parser/parser.ts
ratik21: Added a new task to think about it: [https://www.pivotaltracker.com/story/show/176623588](https://www.pivotaltracker.com/story/show/176623588)

amityadav0:
LGTM

amityadav0 (approved):

robert-zaremba (approved):

PR #202, opened 2021-01-19
rename @types/algosdk to @algorand-builder/types-algosdk

amityadav0:


robert-zaremba:
@amityadav0  - why do we need to update the `typeRoots` in `tsconfig`?  Ideally, for users, it would be if they can just add this packages as a dependency and use it without modifying `tsconfig`.

Could you check [this](https://www.npmjs.com/package/@gooddata/typings) package and see if this works without any tsconfig modification?
 
1. Create a TS project
2. add https://www.npmjs.com/package/@gooddata/typings and https://www.npmjs.com/package/@gooddata/gooddata-js as a dependency
3. add some fake function which uses something from @gooddata/gooddata-js / @gooddata/typings and set it's type .The function can be empty, it should only take as an argument something defined in @gooddata/typings
4. Check if the project compiles without any tsconfig modification.

robert-zaremba:
This is related to: our TypeScript release task and https://github.com/algorand/js-algorand-sdk/issues/265

amityadav0:
checked by installing `@gooddata/gooddata-js`

import2

I found https://github.com/Microsoft/TypeScript/issues/9184 useful link for declarartions files.
From this link i found:
- Either declarations file should be in `package-name`
- Or in `@types` 
these two are default search paths for typescript while resolving typings
- Third way would be to add search path in `typeRoots`

Updating docs

packages/types-algosdk/README.md
```diff
@@ -1,3 +1,18 @@
 # Summary
 
 This package contains type definitions for [js-algorand-sdk](https://github.com/algorand/js-algorand-sdk)
+
+The following is a `tsconfig.json` requirement:
+
+By default all visible ”@types” packages are included in your compilation. Packages in `node_modules/@types` of any enclosing folder are considered visible.
+
+To include this typing `typeRoots` needs to be specified, only packages under typeRoots will be included.
```
robert-zaremba, line 9:
```suggestion
## Usage

By default, the TypeScript compiler searches for type definitions in 
* all packages discoverable (node [modules](https://docs.npmjs.com/cli/v6/configuring-npm/folders#node-modules): `./node_modles`) under  ”@types” organization (eg `node_modules/@types` of any enclosing folder are considered visible).
* local `*d.ts` files
* all subpackages defined `tsconfig.compilerOptions.typeRoots`.

To use this package in a Typescript you have to update your `tsconfig.json` file. The `typeRoots` attribute must be  updated:

```

robert-zaremba (approved):
Left a suggestion for README file

PR #203, opened 2021-01-20
Update doc

ratik21:


packages/algob/src/lib/algo-operator.ts
```diff
@@ -278,7 +278,7 @@ export class AlgoOperatorImpl implements AlgoOperator {
       flags.localBytes,
       flags.globalInts,
       flags.globalBytes,
-      parseSSCAppArgs(flags.appArgs),
+      parseSSCAppArgs(flags.appArgs), // parseSSCAppArgs returns undefined if flags.appArgs is undefined
```
robert-zaremba, line 281: we don't need this comment here. It was important to add it to the function doc string.

robert-zaremba (approved):
We can remove the comment from the following lines:
```
parseSSCAppArgs(flags.appArgs), // parseSSCAppArgs returns undefined if flags.appArgs is undefined
```

packages/algob/src/lib/algo-operator.ts
ratik21: Ok

PR #205, opened 2021-01-21
Logic signature validation

amityadav0:
- validate LSig (delegated signature or escrow account) when processing a transaction.
- added multisignature verification.
- add tests for single and multisignature
- Full runtime support will be done in another task

### DIFF
--- a/packages/runtime/src/logicsig.ts
+++ b/packages/runtime/src/logicsig.ts
@@ -79,7 +79,7 @@ export class LogicSig {
    * @param accAddr Sender's account address
    */
   verify (accAddr: string): boolean {
-    if (this.sig === new Uint8Array(0) && this.msig === undefined) {
+    if (compareArray(this.sig, new Uint8Array(0)) && this.msig === undefined) {
       if (accAddr === this.SignatureAddress) return true;
       return false;
     }
--- a/packages/runtime/src/logicsig.ts
+++ b/packages/runtime/src/logicsig.ts
@@ -1,0 +1,135 @@
+import { Account, decodeAddress, generateAccount, MultiSig, MultiSigAccount, signBytes, verifyBytes } from "algosdk";
+
+import { compareArray } from "../src/lib/compare";
+import { convertToString, stringToBytes } from "../src/lib/parsing";
+
+export class LogicSig {
+  logic: Uint8Array;
+  args: Uint8Array[];
+  sig: Uint8Array;
+  msig: MultiSig | undefined;
+  SignatureAddress: string;
+
+  constructor (program: string, args: Uint8Array[]) {
+    this.logic = stringToBytes(program);
+    this.args = args;
+    this.sig = new Uint8Array(0);
+    this.msig = undefined;
+    this.SignatureAddress = generateAccount().addr;
+  }
+
+  /**
+   * Creates signature (if no msig provided) or multi signature otherwise
+   * @param secretKey sender's secret key
+   */
+  sign (account: Account, msig?: MultiSigAccount): void {
+    if (msig === undefined) {
+      this.sig = signBytes(this.logic, account.sk);
+    } else {
+      const subsigs = msig.addrs.map(addr => {
+        return { pk: decodeAddress(addr).publicKey, s: new Uint8Array(0) };
+      });
+      this.msig = {
+        v: msig.version,
+        thr: msig.threshold,
+        subsig: subsigs
+      };
+
+      const [sig, index] = this.singleSignMultisig(account, this.msig);
+      this.msig.subsig[index].s = sig;
+    }
+  }
+
+  /**
+   * Sign Multisignature
+   * @param account Account
+   * @param msig Multisignature
+   */
+  singleSignMultisig (account: Account, msig: MultiSig): [Uint8Array, number] {
+    let index = -1;
+    const passedPk = decodeAddress(account.addr).publicKey;
+    for (let i = 0; i < msig.subsig.length; i++) {
+      const pk = msig.subsig[i].pk;
+      if (compareArray(pk, passedPk)) {
+        index = i;
+        break;
+      }
+    }
+    if (index === -1) {
+      throw new Error("invalid secret key");
+    }
+    const sig = signBytes(this.logic, account.sk);
+    return [sig, index];
+  }
+
+  /**
+   * Appends a signature to multi signature
+   * @param {Uint8Array} secretKey Secret key to sign with
+   */
+  appendToMultisig (account: Account): void {
+    if (this.msig === undefined) {
+      throw new Error("no multisig present");
+    }
+    const [sig, index] = this.singleSignMultisig(account, this.msig);
+    this.msig.subsig[index].s = sig;
+  }
+
+  /**
+   * Performs signature verification
+   * @param accAddr Sender's account address
+   */
+  verify (account: Account): boolean {
+    return verifyBytes(this.logic, this.sig, account.addr);
+  }
+
+  /**
+   * Verify mult-signature
+   * @param msig Msig
+   * @param publicKey Public key of sender
+   */
+  verifyMultisig (msig: MultiSig, publicKey: Uint8Array): boolean {
+    const threshold = msig.thr;
+    const subsigs = msig.subsig;
+
+    if (msig.subsig.length < threshold) {
+      return false;
+    }
+
+    let counter = 0;
+    for (const subsig of subsigs) {
+      if (subsig.s !== new Uint8Array(0)) {
+        counter += 1;
+      }
+    }
+    if (counter < threshold) {
+      return false;
+    }
+
+    let verifiedCounter = 0;
+    for (const subsig of subsigs) {
+      if (subsig.s !== new Uint8Array(0) && verifyBytes(this.logic, this.sig, publicKey)) {
+        verifiedCounter += 1;
+      }
+    }
+
+    if (verifiedCounter < threshold) {
+      return false;
+    }
+
+    return true;
+  }
+
+  /**
+   * Returns logic signature address
+   */
+  address (): string {
+    return this.SignatureAddress;
+  }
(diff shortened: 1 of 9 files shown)


packages/runtime/src/logicsig.ts
```diff
@@ -0,0 +1,173 @@
+import {
+  Account, decodeAddress, encodeAddress,
+  generateAccount, MultiSig, MultiSigAccount,
+  multisigAddress, signBytes, verifyBytes
+} from "algosdk";
+
+import { compareArray } from "../src/lib/compare";
+import { convertToString, stringToBytes } from "../src/lib/parsing";
+
+export class LogicSig {
```
robert-zaremba, line 10: Is this a copy paste from SDK? Why do we need it here? Maybe we can use inheritance and derive from SDK class if needed?

packages/types-algosdk/types/index.d.ts
```diff
@@ -435,7 +435,14 @@ declare module 'algosdk' {
 
   export function signBid(bid: any, sk: any): any;
 
-  export function signBytes(bytes: any, sk: any): any;
+  /**
+   * signBytes takes arbitrary bytes and a secret key, prepends the bytes with "MX" for domain separation, signs the bytes 
+   * with the private key, and returns the signature.
+   * @param bytes Uint8array
```
robert-zaremba, line 441: We don't need to repeat the type in the doc string. It's already provided in the function definition. So instead of writing  `Uint8array`, you should write what the bytes are representing.

packages/types-algosdk/types/index.d.ts
```diff
@@ -460,7 +467,15 @@ declare module 'algosdk' {
 
   export function signTransaction(txn: Transaction, sk: any): any;
 
-  export function verifyBytes(bytes: any, signature: any, addr: any): any;
+  /**
+   * verifyBytes takes array of bytes, an address, and a signature and verifies if the signature is correct for the public
+   * key and the bytes (the bytes should have been signed with "MX" prepended for domain separation).
+   * @param bytes Uint8Array
```
robert-zaremba, line 473: ditto

packages/runtime/test/src/runtime.ts
```diff
@@ -0,0 +1,88 @@
...
+const multiSigProg = "sample-asc.teal";
+
+describe("Logic Signature Test", () => {
+  useFixture("escrow-account");
+  const john = new StoreAccount(10);
+  const bob = new StoreAccount(100);
+  const runtime = new Runtime([john, bob]);
+
+  it("should sign the lsig by john(delegated signature)", () => {
+    const lsig = runtime.getLogicSig(getProgram(programName), []);
```
robert-zaremba, line 19: why don't we use an SDK for that? SDK contains all functionality for creating LogicSig.

robert-zaremba: same for multisig.

robert-zaremba (commented):
I'm not sure if this is the right way - creating logicsig can be don through SDK. We also have few helper functions in `algob`. If we need to use them, then the best way will be to move that functions to other package, or to this package (in both cases algob will import a package with that functions). 

Left few questions.

packages/runtime/src/logicsig.ts
amityadav0: some parts are same as SDK, but others are according to interpreter.
- in SDK i was not sure if program is in compiled bytes or that hash can only be calculated using the same function 
- or we can get address (escrow address) without network configuration or there are other dependencies on network config.
- i will check whole sdk function independently and confirm if there is dependency

packages/types-algosdk/types/index.d.ts
amityadav0: ok

packages/runtime/test/src/runtime.ts
amityadav0: the functionality that is offered by SDK takes compiled bytes of a program,
I wasn't sure if that will create any inconsistency, that's why i used `signBytes` and `verifyBytes` that signs bytes and verifies it. SDK functions need compiled bytes.
I will check again and see if we can use them without compiling program.

amityadav0:
> I'm not sure if this is the right way - creating logicsig can be don through SDK. We also have few helper functions in `algob`. If we need to use them, then the best way will be to move that functions to other package, or to this package (in both cases algob will import a package with that functions).
> The dependency flow is `runtime -> algob` (so algob depends on runtime, and runtime shouldn't import anything from algob).
> 
> Left few questions.

`algob` functions use compiled program. compiled program needs network connection.
so i think we can use `algosdk` directly if SDK functions can work without compiled bytes.
confirming by independent testing.

packages/runtime/src/logicsig.ts
amityadav0: error

confirmed, cannot use SDK function because of `compiled bytes`.

amityadav0:
error

confirmed, cannot use SDK function because of compiled bytes.

packages/runtime/src/logicsig.ts
robert-zaremba: OK, please add a comment to this class explaining why we can't use SDK type. 
PS: maybe we can still inherit and overwrite some functionality which is not working.

robert-zaremba (approved):
let's resolve comments before merging.

packages/runtime/src/logicsig.ts
```diff
@@ -0,0 +1,173 @@
...
+    this.sig = new Uint8Array(0);
+    this.msig = undefined;
+    this.SignatureAddress = generateAccount().addr;
+  }
+
+  /**
+   * Creates signature (if no msig provided) or multi signature otherwise
+   * @param secretKey sender's secret key
+   */
+  sign (account: Account, msig?: MultiSigAccount): void {
```
ratik21, line 29: Can you please explain as why we're passing account here? When using SDK we usually pass a sk (eg. `lsig.sign(delegator.sk)` )

packages/runtime/src/logicsig.ts
ratik21: Can we also add a doc string pointing out that the we are using the raw TEAL code as program, which is different from the SDK (as they use compiled bytes)?

packages/runtime/src/logicsig.ts
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  /**
+   * Sign Multisignature
+   * @param account Account
+   * @param msig Multisignature
+   */
+  singleSignMultisig (account: Account, msig: MultiSig): [Uint8Array, number] {
+    let index = -1;
+    const passedPk = decodeAddress(account.addr).publicKey;
```
ratik21, line 54:
```suggestion
    const accountPk = decodeAddress(account.addr).publicKey;
```

packages/runtime/src/logicsig.ts
```diff
@@ -0,0 +1,173 @@
...
+    for (let i = 0; i < msig.subsig.length; i++) {
+      const pk = msig.subsig[i].pk;
+      if (compareArray(pk, passedPk)) {
+        index = i;
+        break;
+      }
+    }
+    if (index === -1) {
+      throw new Error("invalid secret key");
```
ratik21, line 63: Can we update to invalid secret key for account: `${account.addr}`?

packages/runtime/src/logicsig.ts
```diff
@@ -0,0 +1,173 @@
...
+    }
+    const sig = signBytes(this.logic, account.sk);
+    return [sig, index];
+  }
+
+  /**
+   * Appends a signature to multi signature
+   * @param {Uint8Array} secretKey Secret key to sign with
+   */
+  appendToMultisig (account: Account): void {
```
ratik21, line 73: Please update the description

packages/runtime/src/logicsig.ts
amityadav0: ok, 
we can inherit but the functionality which is not working, is when class decodes the compiled bytes and check it.
this is happening in constructor, so while inheriting i couldn't figure a way - without calling super constructor.

packages/runtime/src/logicsig.ts
amityadav0: here we need a function to retrieve public key from secret key, i though that will create extra dependency, but i think sk is good, updating

amityadav0: @ratik21 do you know library which has 64 length secret key and public key or how to change it ? i could find tweetnacl , it has 32 length.

amityadav0: got it, nvm

packages/runtime/src/logicsig.ts
amityadav0: ok

packages/runtime/src/logicsig.ts
```diff
@@ -0,0 +1,186 @@
...
+ * because algosdk.makeLogicSig() takes compiled bytes as a argument
+ * and currently we don't calculate compiled bytes from TEAL program.
+ * We are using raw TEAL code as program.(by converting string into bytes)
+ */
+export class LogicSig {
+  logic: Uint8Array;
+  args: Uint8Array[];
+  sig: Uint8Array;
+  msig: MultiSig | undefined;
+  SignatureAddress: string;
```
ratik21, line 24: Should we start class property with caps?  I think using `address` here should be ok. eg. `new LogicSig(..).address` (LogicSig already represents 'Logic Signature', so i don't think we should add Signature prefix. right?

ratik21 (approved):
Left one comment, LGTM otherwise

packages/runtime/src/logicsig.ts
amityadav0: agreed

amityadav0: but can't update to address because we have address(), updating to lsigAddress

PR #206, opened 2021-01-25
Finalize the API and add full Runtime support for LSig

amityadav0:
- runtime, when it sees a LSig transaction, it should validate it and execute a TEAL code corresponding to lsig.
- Update Crowdfunding App tests.
- Add more tests / update old code where needed.

### DIFF
--- a/examples/crowdfunding/test/crowdfundingTest.js
+++ b/examples/crowdfunding/test/crowdfundingTest.js
@@ -78,7 +78,7 @@ describe('Crowdfunding Tests', function () {
     uint64ToBigEndian(fundCloseDate.getTime())
   ];
 
-  it('crowdfunding application', async () => {
+  it('crowdfunding application', () => {
     /**
      * This test demonstrates how to create a Crowdfunding Stateful Smart Contract Application
      * and interact with it. there are following operations that are performed:
@@ -93,7 +93,7 @@ describe('Crowdfunding Tests', function () {
     const creationFlags = Object.assign({}, flags);
 
     // create application
-    applicationId = await runtime.addApp({ ...creationFlags, appArgs: creationArgs }, {}, program);
+    applicationId = runtime.addApp({ ...creationFlags, appArgs: creationArgs }, {}, program);
     const creatorPk = addressToPk(creator.address);
 
     // setup escrow account
@@ -113,7 +113,7 @@ describe('Crowdfunding Tests', function () {
     // update application with correct escrow account address
     let appArgs = [addressToPk(lsig.address())]; // converts algorand address to Uint8Array
 
-    await runtime.updateApp(
+    runtime.updateApp(
       creator.address,
       applicationId,
       program,
@@ -125,8 +125,8 @@ describe('Crowdfunding Tests', function () {
     assert.deepEqual(getGlobal('Escrow'), escrowPk);
 
     // opt-in to app
-    await runtime.optInToApp(creator.address, applicationId, {}, {}, program);
-    await runtime.optInToApp(donor.address, applicationId, {}, {}, program);
+    runtime.optInToApp(creator.address, applicationId, {}, {}, program);
+    runtime.optInToApp(donor.address, applicationId, {}, {}, program);
 
     syncAccounts(lsig.address());
     assert.isDefined(creator.appsLocalState.get(applicationId));
@@ -157,7 +157,7 @@ describe('Crowdfunding Tests', function () {
       }
     ];
     // execute transaction
-    await runtime.executeTx(txGroup, program, []);
+    runtime.executeTx(txGroup, program, []);
 
     syncAccounts(lsig.address());
     assert.equal(escrow.balance(), donationAmount);
@@ -189,7 +189,7 @@ describe('Crowdfunding Tests', function () {
 
     syncAccounts(lsig.address());
     const donorBalance = donor.balance();
-    await runtime.executeTx(txGroup, program, []);
+    runtime.executeTx(txGroup, program, []);
 
     syncAccounts(lsig.address());
     assert.equal(escrow.balance(), 300000);
@@ -218,7 +218,7 @@ describe('Crowdfunding Tests', function () {
       }
     ];
     // execute transaction
-    await runtime.executeTx(txGroup, program, []);
+    runtime.executeTx(txGroup, program, []);
 
     appArgs = [stringToBytes('claim')];
     txGroup = [
@@ -240,14 +240,14 @@ describe('Crowdfunding Tests', function () {
         payFlags: { closeRemainderTo: creator.address }
       }
     ];
-    await runtime.executeTx(txGroup, program, []);
+    runtime.executeTx(txGroup, program, []);
     // TODO- close account and tranfer funds to closeRemainderTo in runtime
   });
 
-  it('should be rejected by logic when claiming funds if goal is not met', async () => {
+  it('should be rejected by logic when claiming funds if goal is not met', () => {
     // create application
     const creationFlags = Object.assign({}, flags);
-    const applicationId = await runtime.addApp({ ...creationFlags, appArgs: creationArgs }, {}, program);
+    const applicationId = runtime.addApp({ ...creationFlags, appArgs: creationArgs }, {}, program);
 
     // setup escrow account
     const escrowProg = getProgram('crowdFundEscrow.py', { APP_ID: applicationId });
@@ -257,7 +257,7 @@ describe('Crowdfunding Tests', function () {
 
     // update application with correct escrow account address
     let appArgs = [addressToPk(lsig.address())]; // converts algorand address to Uint8Array
-    await runtime.updateApp(
+    runtime.updateApp(
       creator.address,
       applicationId,
       program,
@@ -286,7 +286,7 @@ describe('Crowdfunding Tests', function () {
     ];
     // execute transaction: Expected to be rejected by logic because goal is not reached
     try {
-      await runtime.executeTx(txGroup, program, []);
+      runtime.executeTx(txGroup, program, []);
     } catch (e) {
       console.warn(e);
     }
--- a/examples/crowdfunding/test/crowdfundingTest.js
+++ b/examples/crowdfunding/test/crowdfundingTest.js
@@ -9,7 +9,7 @@ import {
 import { Runtime, StoreAccount } from '@algorand-builder/runtime';
 import { assert } from 'chai';
 
-const initialDonorBalance = 6e7;
+const initialDonorBalance = 60e6;
 const initialCreatorBalance = 1e4;
 const goal = 7e6;
 
--- a/examples/crowdfunding/test/crowdfundingTest.js
+++ b/examples/crowdfunding/test/crowdfundingTest.js
@@ -9,9 +9,9 @@ import {
 import { Runtime, StoreAccount } from '@algorand-builder/runtime';
 import { assert } from 'chai';
 
-const initialDonorBalance = 60000000;
-const initialCreatorBalance = 10000;
-const goal = 7000000;
+const initialDonorBalance = 6e7;
+const initialCreatorBalance = 1e4;
+const goal = 7e6;
 
 describe('Crowdfunding Tests', function () {
   let creator = new StoreAccount(initialCreatorBalance);
@@ -52,10 +52,10 @@ describe('Crowdfunding Tests', function () {
   const getGlobal = (key) => runtime.getGlobalState(applicationId, key);
 
   // fetch latest account state
-  function syncAccounts (escrowAddress) {
+  function syncAccounts () {
     creator = runtime.getAccount(creator.address);
-    escrow = runtime.getAccount(escrowAddress);
     donor = runtime.getAccount(donor.address);
+    escrow = runtime.getAccount(escrow.address);
   }
 
   // Get begin date to pass in
@@ -99,7 +99,11 @@ describe('Crowdfunding Tests', function () {
     // setup escrow account
     const escrowProg = getProgram('crowdFundEscrow.py', { APP_ID: applicationId });
     const lsig = runtime.getLogicSig(escrowProg, []);
-    console.log('Escrow Address: ', lsig.address());
+    const escrowAddress = lsig.address();
+
+    // sync escrow account
+    escrow = runtime.getAccount(escrowAddress);
+    console.log('Escrow Address: ', escrowAddress);
 
     // verify global state
     assert.isDefined(applicationId);
@@ -111,14 +115,14 @@ describe('Crowdfunding Tests', function () {
     assert.deepEqual(getGlobal('Total'), 0n);
 
     // update application with correct escrow account address
-    let appArgs = [addressToPk(lsig.address())]; // converts algorand address to Uint8Array
+    let appArgs = [addressToPk(escrowAddress)]; // converts algorand address to Uint8Array
 
     runtime.updateApp(
       creator.address,
       applicationId,
       program,
       {}, { appArgs: appArgs });
-    const escrowPk = addressToPk(lsig.address());
+    const escrowPk = addressToPk(escrowAddress);
 
     // verify escrow storage
     assert.isDefined(applicationId);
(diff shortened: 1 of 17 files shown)


examples/crowdfunding/test/crowdfundingTest.js
```diff
@@ -107,24 +111,24 @@ describe('Crowdfunding Tests', function () {
     assert.deepEqual(getGlobal('Total'), 0n);
 
     // update application with correct escrow account address
-    let appArgs = [addressToPk(escrow.address)]; // converts algorand address to Uint8Array
+    let appArgs = [addressToPk(lsig.address())]; // converts algorand address to Uint8Array
```
ratik21, line 114: Instead of using `lsig.address()`, we can declare a constant:
```
const escrowAddr = lsig.address();
```
It's more intuitive

examples/crowdfunding/test/crowdfundingTest.js
```diff
@@ -53,9 +52,9 @@ describe('Crowdfunding Tests', function () {
   const getGlobal = (key) => runtime.getGlobalState(applicationId, key);
 
   // fetch latest account state
-  function syncAccounts () {
+  function syncAccounts (escrowAddress) {
```
ratik21, line 55: I am not sure if this is the right way. What if we have multiple logic signatures?

examples/crowdfunding/test/crowdfundingTest.js
```diff
@@ -94,9 +93,14 @@ describe('Crowdfunding Tests', function () {
     const creationFlags = Object.assign({}, flags);
 
     // create application
-    applicationId = await runtime.addApp({ ...creationFlags, appArgs: creationArgs }, {}, program);
+    applicationId = runtime.addApp({ ...creationFlags, appArgs: creationArgs }, {}, program);
     const creatorPk = addressToPk(creator.address);
 
+    // setup escrow account
+    const escrowProg = getProgram('crowdFundEscrow.py', { APP_ID: applicationId });
+    const lsig = runtime.getLogicSig(escrowProg, []);
```
ratik21, line 101: Can we move these 2 lines during setUp of tests? (Maybe declare globally, or in `beforeEach(..)` hook if logic signature will be different in each test). I think this way we don't need to pass address in `syncAccounts()` also

examples/crowdfunding/test/crowdfundingTest.js
amityadav0: Then the user can modify the code, this is just for syncing accounts, this varies from user to user.
Also user can directly call in test instead of having this fucntion.

Here we are getting the escrow address after deploying escrow, therefore we get the address after deploying it, the other way would be having each line there or changing the constant.

examples/crowdfunding/test/crowdfundingTest.js
amityadav0: I am not sure that's the right way, test should be independent of each other, what if there are two different logic signatures ? We xan modify test to use them before each is more general, can only take same program name.

examples/crowdfunding/test/crowdfundingTest.js
amityadav0: Updating to that way

examples/crowdfunding/test/crowdfundingTest.js
amityadav0: ok

examples/crowdfunding/test/crowdfundingTest.js
amityadav0: updated @ratik21 please check the updated one.

examples/crowdfunding/test/crowdfundingTest.js
```diff
@@ -9,9 +9,9 @@ import {
 import { Runtime, StoreAccount } from '@algorand-builder/runtime';
 import { assert } from 'chai';
 
-const initialDonorBalance = 60000000;
-const initialCreatorBalance = 10000;
-const goal = 7000000;
+const initialDonorBalance = 6e7;
```
ratik21, line 12:
```suggestion
const initialDonorBalance = 60e6;
```

packages/algob/src/lib/load-program.ts
```diff
@@ -0,0 +1,29 @@
...
+import path from "path";
+import YAML from "yaml";
+
+import { ASSETS_DIR } from "../internal/core/project-structure";
+import { CompileOp, PyCompileOp, pyExt, tealExt } from "./compile";
+
+// takes file name as input and returns program as string
+export function getProgram (fileName: string, scInitParam?: unknown): string {
+  const filePath = path.join(process.cwd(), ASSETS_DIR, fileName);
+  const file = fs.readFileSync(filePath, 'utf8');
```
robert-zaremba, line 12: please rename `file` to `program`

packages/runtime/src/logicsig.ts
```diff
@@ -170,6 +171,14 @@ export class LogicSig {
     return true;
   }
 
+  /**
+   * Returns signed logic
+   * @param secretKey account's secret key
```
robert-zaremba, line 176:
```suggestion
   * @param secretKey: account's secret key
```

packages/runtime/test/src/logicsig.ts
```diff
@@ -0,0 +1,95 @@
...
+const multiSigProg = "sample-asc.teal";
+
+describe("Logic Signature Test", () => {
+  useFixture("escrow-account");
+  const john = new StoreAccount(10);
+  const bob = new StoreAccount(100);
+  const runtime = new Runtime([john, bob]);
+  const johnPk = decodeAddress(john.address).publicKey;
+
+  it("should sign the lsig by john(delegated signature)", () => {
```
robert-zaremba, line 19:
```suggestion
  it("john should be able to create a delegated signature", () => {
```

packages/runtime/test/src/logicsig.ts
```diff
@@ -0,0 +1,95 @@
...
+
+import { StoreAccount } from "../../src/account";
+import { Runtime } from "../../src/runtime";
+import { getProgram } from "../helpers/files";
+import { useFixture } from "../helpers/integration";
+
+const programName = "escrow.teal";
+const multiSigProg = "sample-asc.teal";
+
+describe("Logic Signature Test", () => {
```
robert-zaremba, line 12:
```suggestion
describe("Logic Signature", () => {
```

packages/runtime/test/src/logicsig.ts
```diff
@@ -0,0 +1,95 @@
...
+  const runtime = new Runtime([john, bob]);
+  const johnPk = decodeAddress(john.address).publicKey;
+
+  it("should sign the lsig by john(delegated signature)", () => {
+    const lsig = runtime.getLogicSig(getProgram(programName), []);
+
+    lsig.sign(john.account.sk);
+    const result = lsig.verify(johnPk);
+
+    assert.equal(result, true);
```
robert-zaremba, line 25:
```suggestion
    assert.isTrue(lsig.verify(johnPk));
```

robert-zaremba: We can do similar updates below (isTrue, isFalse...)

https://www.chaijs.com/api/assert/#method_istrue

packages/runtime/test/src/logicsig.ts
```diff
@@ -0,0 +1,95 @@
...
+  });
+
+  it("should return false if lsig is not signed by john(delegated signature)", () => {
```
robert-zaremba, line 28:
```suggestion
  it("should fail to verify delegated signature signed by someone else", () => {
```

packages/runtime/test/src/logicsig.ts
```diff
@@ -0,0 +1,95 @@
...
+    const lsig = runtime.getLogicSig(getProgram(programName), []);
+
+    lsig.sign(bob.account.sk);
+    const result = lsig.verify(johnPk);
+
+    assert.equal(result, false);
+  });
+
+  it("should verify only for lsig address(escrow account)", () => {
```
robert-zaremba, line 37:
```suggestion
  it("should handle contract lsig (escrow account) verification correctly", () => {
```

packages/runtime/test/src/runtime.ts
```diff
@@ -1,93 +1,69 @@
...
+    lsig = runtime.getLogicSig(getProgram(programName), []);
+    txnParam = {
+      type: TransactionType.TransferAlgo,
+      sign: SignType.LogicSignature,
+      fromAccount: john.account,
+      toAccountAddr: bob.account.addr,
+      amountMicroAlgos: 1000,
+      lsig: lsig,
+      payFlags: { totalFee: 1000 }
+    };
```
robert-zaremba, line 35: it's good to use `beforeAll` :+1:

packages/runtime/test/src/runtime.ts
```diff
@@ -1,93 +1,69 @@
...
-
-    const result = lsig.verify(multsigaddr);
-    assert.equal(result, false);
+  it("should verify signature but reject logic", async () => {
+    const logicSig = runtime.getLogicSig(getProgram("reject.teal"), []);
+    txnParam.lsig = logicSig;
+    txnParam.fromAccount = john.account;
+
+    logicSig.sign(john.account.sk);
+    // execute transaction (rejected by logic after signature validation)
```
robert-zaremba, line 63: Please add more details why it's rejected (eg, because fromAccount check failed in the program)

robert-zaremba (commented):
pre approving - left few comments.

packages/algob/src/lib/load-program.ts
amityadav0: ok

packages/runtime/src/runtime.ts
```diff
@@ -400,6 +401,23 @@ export class Runtime {
       txnParameters = txnParams;
     }
     for (const txnParam of txnParameters) {
+      // check if transaction is signed by logic signature,
+      // if yes verify signature and run logic
+      if (txnParam.sign === SignType.LogicSignature) {
+        if (txnParam.lsig === undefined) {
+          throw new TealError(ERRORS.TEAL.LOGIC_SIGNATURE_NOT_FOUND);
+        }
```
ratik21, line 409: What happens when user passes sign type as secrect key (instead of logic signature), and passes a logic signature with the txParams? Is it covered or should we add that check?

examples/crowdfunding/test/crowdfundingTest.js
```diff
@@ -252,8 +257,12 @@ describe('Crowdfunding Tests', function () {
...
-    syncAccounts(lsig.address());
+    const escrowAddress = lsig.address();
+
+    // sync escrow account
+    escrow = runtime.getAccount(escrowAddress);
+    console.log('Escrow Address: ', escrowAddress);
+    syncAccounts();
 
     // update application with correct escrow account address
     let appArgs = [addressToPk(lsig.address())]; // converts algorand address to Uint8Array
```
ratik21, line 268: Please update `lsig.address()` here as well

packages/runtime/src/runtime.ts
amityadav0: i don't think we can cover that. because this is clearly user's mistake, there is 50% chance that it is correct(and 50% it is wrong).

examples/crowdfunding/test/crowdfundingTest.js
amityadav0: ok

packages/runtime/test/src/runtime.ts
amityadav0: ok

packages/algob/src/lib/load-program.ts
```diff
@@ -0,0 +1,29 @@
...
+  }
+
+  if (fileName.endsWith(pyExt)) {
+    // convert initial parameters
+    let param: string | undefined = YAML.stringify(scInitParam);
+    if (scInitParam === undefined) { param = undefined; }
+
+    // mock algod credentials to initialize PyCompileOp
+    const mockAlgod = new Algodv2("dummyToken", "dummyNetwork", 8080);
+    const py = new PyCompileOp(new CompileOp(mockAlgod));
```
ratik21, line 25: Can we move the mock functionality to `/mock/algod.ts`? We may use the mock client in future as well?

packages/runtime/src/runtime.ts
ratik21: I wanted to know how the transaction is executed if the user makes that mistake (passing secretkey as sign type along with a logic signature). Because in algob i think `signTx` covers that case

amityadav0: what if he is right ? and mistakenly sent lsig

amityadav0: sending secret key was right, but lsig was mistake or otherwise. this is 50-50

amityadav0: also, if secret key doesn't matches with lsig address, logic signature validation will throw error.

ratik21: > what if he is right ? and mistakenly sent lsig

How can the user be right? Is there a transaction where user can send a secret key (knowingly), and passes lsig?
If the user is passing secret key knowingly, then the transaction must be other than stateless asc execution (algoTransfer, assetTransfer, stateful call, ..). Am i missing something?

amityadav0: yes, what if he makes a mistake in passing `lsig` , it should not be passed but he passed it anyway.
how will you decide what was the mistake ? passing extra `lsig` or passing wrong `sign`.

amityadav0: think about the first case, user passed `secret key` with `lsig` -> this case is clearly confusing for interpreter.
but the question is how will you decide the mistake of user ?

correct case may be `logicSignature` with `lsig` 
or
`SecretKey` with no `lsig`

50-50
does that makes sense ?

amityadav0: we can add a warning though, that these should not be passed simultaneously

ratik21: It's still not clear to me how the tx is executed if we pass sk with lsig. Please add a warning/note atleast. I will test it on my end later.

amityadav0: ok

packages/runtime/src/runtime.ts
```diff
@@ -388,6 +388,28 @@ export class Runtime {
     toAccount.amount += txnParam.amountMicroAlgos; // add 'x' algo to receiver
   }
 
+  /**
+   * validate logic signature and teal logic
+   * @param txnParam Transaction Parameters
+   */
+  validateLsig (txnParam: ExecParams): void {
```
ratik21, line 395: Can we change the function to `validateLsigAndRun`? We are executing TEAL code along with validating lsig in this function. Or we can `run` the code after validation

packages/runtime/src/runtime.ts
```diff
@@ -461,9 +466,15 @@ export class Runtime {
     let mode = ExecutionMode.STATEFUL;
     if (!Array.isArray(txnParams)) {
       mode = this.getExecutionMode(txnParams);
+      if (txnParams.sign === SignType.LogicSignature) {
+        this.validateLsig(txnParams);
```
ratik21, line 470: We have `run` in line 483 as well. This way we would run stateless teal code twice. Right?

amityadav0: No, because program passed is different.

packages/runtime/src/runtime.ts
amityadav0: validate lsig means we are checking logic too, but yes i can update

packages/runtime/src/runtime.ts
ratik21: Ok, so that means `program` in `executeTx` can be optional (if the tx or txGroup contains only stateless txns). We may need to update this is well. We can do it in the group txns task

ratik21 (approved):

packages/runtime/src/runtime.ts
amityadav0: exactly, also we don't need this for payment only too.
we need to restructure it.

PR #208, opened 2021-01-27
Add support to block height / round checks.

amityadav0:


packages/runtime/src/errors/errors-list.ts
```diff
@@ -253,6 +253,12 @@ by an index that does not exist.`
     message: "Secret key and Logic Signature should not be passed together",
     title: "Transaction params error",
     description: `Transaction params error`
+  },
+  INVALID_ROUND: {
+    number: 38,
+    message: "Transaction rounds are not valid, current round: %round%.",
```
ratik21, line 259:
```suggestion
    message: "Transaction rounds (firstValid, lastValid) are not valid, current round: %round%.",
```

packages/runtime/src/runtime.ts
```diff
@@ -174,6 +205,22 @@ export class Runtime {
...
+    const s = mockSuggestedParams(flags);
+
+    // https://developer.algorand.org/docs/features/transactions/#setting-first-and-last-valid
+    s.firstRound = flags.firstValid ?? (this.round - 1);
+    s.lastRound = flags.firstValid === undefined || flags.validRounds === undefined
+      ? s.firstRound + 1000
+      : Number(flags.firstValid) + Number(flags.validRounds);
+
+    return s;
+  }
```
ratik21, line 222: Can't we pass `this.round` to mockSuggestedParams, and calculate `firstRound` and `lastRound` at a single place?

packages/runtime/test/src/runtime.ts
```diff
@@ -4,7 +4,7 @@ import { assert } from "chai";
 import { StoreAccount } from "../../src/account";
 import { ERRORS } from "../../src/errors/errors-list";
 import { Runtime } from "../../src/runtime";
-import { ExecParams, SignType, TransactionType } from "../../src/types";
+import { AlgoTransferParam, ExecParams, SignType, TransactionType } from "../../src/types";
```
ratik21, line 7:
```suggestion
import type { AlgoTransferParam, ExecParams, SignType, TransactionType } from "../../src/types";
```

packages/runtime/src/runtime.ts
```diff
@@ -500,6 +547,8 @@ export class Runtime {
   executeTx (txnParams: ExecParams | ExecParams[], program: string,
     args: Uint8Array[]): void {
     const [tx, gtxs] = this.createTxnContext(txnParams); // get current txn and txn group (as encoded obj)
+    // validate first and last rounds
+    this.validateRounds(gtxs);
```
ratik21, line 551:
```suggestion
    this.validateTxRound(gtxs);
```

ratik21: I have a question. Here we are validating the round for the 'current' transaction only, and i don't think we're incrementing the round after the transaction is executed. Shouldn't we do that? By the current method, if i am executing two or more transactions then the current round (`this.round`) is same for them, which means that those different transactions were executed in the same round.

packages/runtime/src/runtime.ts
amityadav0: yes

packages/runtime/src/runtime.ts
amityadav0: yes, https://algoexplorer.io/
in each round you can have around 18 - 89 transactions committed (observed from the above link).
and it takes 4.5s to commit a block.
therefore it is not clear when to increment rounds(increment after 18 txns or 80 txns) , only user can decide that by setting rounds.

ratik21: Okay, so what does the current round represent in our case? Is it a particular timestamp? I thought we could manage time using rounds and update crowdfunding tests accordingly. Or are these two different?

amityadav0: yes, there is a slight difference, here is a good explanation https://developer.algorand.org/docs/features/transactions/#current-round

ratik21: Alright, will check.

amityadav0: also, in our case current round is not timestamp, it is value that can be set by user. let's say we want want to check crowdfunding, we will have to set the rounds and then check for validation

amityadav0: I have an better idea, let me know if you follow:
+ we will have start timestamp stored - (when calling constructor)
+ at each time when we call a check for transaction, we will have difference ( current time - construction time (in seconds)).
-- round will be `diff / 4.5`  this way we can have good testing for crowdfunding, this is a different idea than what is mentioned in task, what do you think ?

ratik21: yes, got it. I was referring to the commented code we have in crowdfunding tests (related to timestamps, start date, end date..), but logic of rounds is different from that. 
[https://www.pivotaltracker.com/story/show/176631770](https://www.pivotaltracker.com/story/show/176631770)

amityadav0: but in this case user cannot set the rounds because round calculation will be changed.

ratik21: I like the idea. 

> round will be diff / 4.5 this way we can have good testing for crowdfunding

I think 4.5s is the average block time for the algorand blockchain. Since we're not on a real blockchain, can't we keep the average block time to be 1s? This way `currentRound` = `lastest timestamp` - `construction timestamp`. 

I will read more about rounds. Because by looking at the difference, let's say transaction is being executed after 2s from the construction time, then `currentRound = 2`. But what does that mean (what does the '2' represent)? 
Also, should we take start timestamp as the a) timestamp when runtime is initialized or b) timestamp when previous tx was executed

amityadav0: - current round 2 means 9 seconds are passed i,e two blocks are committed
- if transaction is executed after 2 sec that will be round 1.
- we will take timestamp at time of validation of transaction- let's say you executed transaction after 20 sec after initializing the runtime, so calculation will be done current - init = 20 / 4.5 this will return round.

But i think the task idea is better because we still have to get the rounds from network.
this seems overengineered.

amityadav0: there is one more benefit of using task idea, let's say we want to test if application rejects logic after 1 hour, we can set the rounds and check.

packages/runtime/test/src/runtime.ts
```diff
@@ -69,3 +70,76 @@ describe("Logic Signature Transaction in Runtime", function () {
...
+  });
+
+  function syncAccounts (): void {
+    john = runtime.getAccount(john.address);
+    bob = runtime.getAccount(bob.address);
+  }
+
+  it("should pass if first and last valid are in current round range", () => {
+    txnParams.payFlags = { totalFee: 1000, firstValid: 5, validRounds: 200 };
+    // set round
```
robert-zaremba, line 111: this comment is not needed.

packages/runtime/test/src/runtime.ts
```diff
@@ -69,3 +70,76 @@ describe("Logic Signature Transaction in Runtime", function () {
...
+    runtime.setRound(20);
+
+    // execute transaction
```
robert-zaremba, line 114: same here (also in other unit tests as well).

packages/runtime/test/src/runtime.ts
```diff
@@ -69,3 +70,76 @@ describe("Logic Signature Transaction in Runtime", function () {
...
+    runtime.executeTx(txnParams, program, []);
+
+    // get final state (updated accounts)
+    syncAccounts();
+    assert.equal(john.balance(), minBalance - 1100);
+    assert.equal(bob.balance(), minBalance + 100);
+  });
+
+  it("should fail if first and last valid are not in current round range", () => {
```
robert-zaremba, line 123:
```suggestion
  it("should fail if current round is not between first and last valid", () => {
```

packages/runtime/test/src/runtime.ts
```diff
@@ -69,3 +70,76 @@ describe("Logic Signature Transaction in Runtime", function () {
...
+    runtime.setRound(3);
+
+    // execute transaction
+    expectTealError(
+      () => runtime.executeTx(txnParams, program, []),
+      ERRORS.TEAL.INVALID_ROUND
+    );
+  });
+
+  it("should pass if no rounds are passed", () => {
```
robert-zaremba, line 134:
```suggestion
  it("should succeeded by default (no round requirement is passed)", () => {
```

packages/runtime/test/src/runtime.ts
```diff
@@ -69,3 +70,76 @@ describe("Logic Signature Transaction in Runtime", function () {
...
+    txnParams.fromAccount = john.account;
+    txnParams.toAccountAddr = bob.address;
+  });
+
+  function syncAccounts (): void {
+    john = runtime.getAccount(john.address);
+    bob = runtime.getAccount(bob.address);
+  }
+
+  it("should pass if first and last valid are in current round range", () => {
```
robert-zaremba, line 109:
```suggestion
  it("should succeed if current round is between first and last valid", () => {
```

robert-zaremba (approved):
pre-approving, left few comments.

packages/runtime/src/runtime.ts
```diff
@@ -107,6 +110,34 @@ export class Runtime {
...
 
+  /**
+   * Validate first and last rounds of transaction using current round
+   * @param gtxns transactions
+   */
+  validateTxRound (gtxns: Txn[]): void {
+    // https://developer.algorand.org/docs/features/transactions/#current-round
+    for (const txn of gtxns) {
+      if (txn.fv >= this.round || txn.lv <= this.round) {
+        throw new TealError(ERRORS.TEAL.INVALID_ROUND, { round: this.round });
```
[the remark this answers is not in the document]
amityadav0: ok

packages/runtime/test/src/runtime.ts
amityadav0: ok

packages/runtime/test/src/runtime.ts
amityadav0: ok

PR #209, opened 2021-01-27
Remove dependency

amityadav0:
- Remove `algob` from `runtime`.
- moved required types and functions from algob to runtime

### DIFF
--- a/packages/runtime/src/errors/errors-list.ts
+++ b/packages/runtime/src/errors/errors-list.ts
@@ -242,10 +242,16 @@ by an index that does not exist.`
     description: `lsig validation error`
   },
   INVALID_TRANSACTION_PARAMS: {
-    number: 27,
+    number: 37,
     message: "Secret key and Logic Signature should not be passed together",
     title: "Transaction params error",
     description: `Transaction params error`
+  },
+  UNKNOWN_TRANSACTION_TYPE: {
+    number: 38,
+    message: "Unknown Transaction type: %transaction%",
+    title: "Transaction error",
+    description: `Transaction error`
   }
 };
 
--- a/packages/runtime/src/errors/errors-list.ts
+++ b/packages/runtime/src/errors/errors-list.ts
@@ -13,7 +13,8 @@ export function getErrorCode (error: ErrorDescriptor): string {
 }
 
 export const ERROR_RANGES = {
-  TEAL: { min: 0, max: 99, title: "TEAL opcode errors" }
+  TEAL: { min: 0, max: 99, title: "TEAL opcode errors" },
+  TRANSACTION: { min: 0, max: 99, title: "Transaction error" }
 };
 
 const PARSE_ERROR = "Parse Error";
@@ -252,9 +253,12 @@ by an index that does not exist.`
     message: "Secret key and Logic Signature should not be passed together",
     title: "Transaction params error",
     description: `Transaction params error`
-  },
+  }
+};
+
+const transactionErrors = {
   UNKNOWN_TRANSACTION_TYPE: {
-    number: 38,
+    number: 1,
     message: "Unknown Transaction type: %transaction%",
     title: "Transaction error",
     description: `Transaction error`
@@ -266,5 +270,6 @@ export const ERRORS: {
     [errorName: string]: ErrorDescriptor
   };
 } = {
-  TEAL: tealErrors
+  TEAL: tealErrors,
+  TRANSACTION: transactionErrors
 };
--- a/packages/runtime/src/index.ts
+++ b/packages/runtime/src/index.ts
@@ -1,9 +1,33 @@
+import { applyErrorMessageTemplate } from "../src/errors/errors";
 import { Interpreter } from "../src/interpreter/interpreter";
+import { addressToPk, parseSSCAppArgs, stringToBytes, uint64ToBigEndian } from "../src/lib/parsing";
+import { encodeNote, mkTransaction } from "../src/lib/txn";
+import {
+  AlgoTransferParam, AssetTransferParam, ExecParams,
+  SignType, SSCCallsParam, SSCDeploymentFlags,
+  SSCOptionalFlags, TransactionType, TxParams
+} from "../src/types";
 import { StoreAccount } from "./account";
 import { Runtime } from "./runtime";
 
 export {
   Interpreter,
   Runtime,
-  StoreAccount
+  StoreAccount,
+  mkTransaction,
+  applyErrorMessageTemplate,
+  parseSSCAppArgs,
+  addressToPk,
+  uint64ToBigEndian,
+  stringToBytes,
+  SSCCallsParam,
+  SSCDeploymentFlags,
+  ExecParams,
+  TransactionType,
+  SignType,
+  AlgoTransferParam,
+  AssetTransferParam,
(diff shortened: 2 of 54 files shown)


ratik21:
Is copy pasting is the right way? I think this could be difficult to maintain, let's say we need to update `mkTransaction` (either by us or through an `algosdk` update) then we will need to update both these functions. Same for others. 

Is it possible to move the 'absolutely' common functions to a new package and use them in both `algob` and `runtime`? This way we would need to maintain those functions at a single place only. Also, some functions are pretty common like `uint64ToBigEndian`, `addressToPk`.. and these can be used accross multiple projects as well (not just algob and runtime). 

We can also move more common functions (like `convertToBuffer`, `convertToString`, `stringToBytes`..) so that they can be used accross examples and packages. What do you think? Can we do that?

amityadav0:
That's why I asked specifically about this, this should have been brought up at time of discussion.

packages/runtime/src/lib/parsing.ts
```diff
@@ -97,6 +98,81 @@ export function convertToBuffer (s: string, encoding?: EncodingType): Buffer {
...
 
+/**
+ * Converts 64 bit unsigned integer to bytes in big endian.
+ */
+export function uint64ToBigEndian (x: number | bigint): Uint8Array {
+  x = BigInt(x); // use x as bigint internally to support upto uint64
+  const buff = Buffer.alloc(8);
+  buff.writeBigUInt64BE(x);
+  return Uint8Array.from(buff);
+}
```
robert-zaremba, line 109: this is useful!

packages/runtime/src/lib/parsing.ts
```diff
@@ -97,6 +98,81 @@ export function convertToBuffer (s: string, encoding?: EncodingType): Buffer {
...
+
+/**
+ * Takes an Algorand address in string form and decodes it into a Uint8Array (as public key)
+ * @param addr : algorand address
+ */
+export function addressToPk (addr: string): Uint8Array {
+  return decodeAddress(addr).publicKey;
+}
+
+const throwErr = (appArg: string): void => {
```
robert-zaremba, line 119: maybe we can rename to `throwFmtError`?

packages/runtime/src/lib/txn.ts
```diff
@@ -119,3 +119,106 @@ export function txAppArg (txField: TxField, tx: Txn, idx: number, op: Op,
...
+        note,
+        execParams.lease,
+        execParams.rekeyTo
+      );
+    }
+    default: {
+      throw new TealError(ERRORS.TEAL.ACCOUNT_DOES_NOT_EXIST, { transaction: transactionType });
+    }
+  }
+}
```
robert-zaremba, line 224: this we can leave here.

ratik21:
> That's why I asked specifically about this, this should have been brought up at time of discussion. Not after.
> 
> Let's discuss on discord

Sorry about that, will be more careful from now. I didn't know you were working on it or it is blocking another task. I think we should create a task/add in icebox first, then discuss about it, and then create a PR for it.

robert-zaremba (changes requested):
This is a good PR, but instead of copying types we should move them. We can discuss on Discord where to move them.

amityadav0:
> > That's why I asked specifically about this, this should have been brought up at time of discussion. Not after.
> > Let's discuss on discord
> 
> Sorry about that, will be more careful from now. I didn't know you were working on it or it is blocking another task. I think we should create a task/add in icebox first, then discuss about it, and then create a PR for it (for better communication)

ok, https://discord.com/channels/754286650540032030/754286650540032033/803302226688344094

packages/runtime/src/lib/parsing.ts
amityadav0: ok

ratik21:
> ok, https://discord.com/channels/754286650540032030/754286650540032033/803302226688344094

Yes, it's mentioned in the chat about moving code to runtime, not copying (I understood it that way). That's where i think the miscommunication happened. Will discuss more in future.

packages/runtime/src/lib/txn.ts
```diff
@@ -119,3 +119,106 @@ export function txAppArg (txField: TxField, tx: Txn, idx: number, op: Op,
...
+        execParams.accounts,
+        execParams.foreignApps,
+        execParams.foreignAssets,
+        note,
+        execParams.lease,
+        execParams.rekeyTo
+      );
+    }
+    default: {
+      throw new TealError(ERRORS.TEAL.UNKNOWN_TRANSACTION_TYPE, { transaction: transactionType });
```
ratik21, line 221: Isn't `transactionType` a number here? (`execParams.type` is an enum)

amityadav0: i will check, i just moved it here, is everything OK otherthan this @ratik21  ?

packages/runtime/src/errors/errors-list.ts
```diff
@@ -242,10 +242,16 @@ by an index that does not exist.`
     description: `lsig validation error`
   },
   INVALID_TRANSACTION_PARAMS: {
-    number: 27,
+    number: 37,
     message: "Secret key and Logic Signature should not be passed together",
     title: "Transaction params error",
     description: `Transaction params error`
+  },
+  UNKNOWN_TRANSACTION_TYPE: {
```
ratik21, line 250: should this be a TEAL error (as now we're using it within algob as well) ? We can move to a different class of errors later if you agree.

ratik21 (approved):
Left one comment, LGTM otherwise

packages/runtime/src/lib/txn.ts
ratik21: @amityadav0 looks good. Thanks for the PR.

amityadav0: i think this is not a problem just confirmed by running,
this shows string if we pass unknown because we are storing this before switch.

packages/runtime/src/errors/errors-list.ts
amityadav0: yes, good idea. creating `transactionErrors` class

packages/runtime/src/index.ts
```diff
@@ -1,9 +1,33 @@
+import { applyErrorMessageTemplate } from "../src/errors/errors";
 import { Interpreter } from "../src/interpreter/interpreter";
+import { addressToPk, parseSSCAppArgs, stringToBytes, uint64ToBigEndian } from "../src/lib/parsing";
+import { encodeNote, mkTransaction } from "../src/lib/txn";
+import {
+  AlgoTransferParam, AssetTransferParam, ExecParams,
+  SignType, SSCCallsParam, SSCDeploymentFlags,
+  SSCOptionalFlags, TransactionType, TxParams
+} from "../src/types";
```
robert-zaremba, line 9: One suggestion: let's don't export types in index. So, on importing, someone will need to use 
```
import types ... from "@algrorand-builder/runtime/types"
```

robert-zaremba (approved):
Pre-approving. I left one suggestion to reorganize the imports and exports.

packages/runtime/src/index.ts
amityadav0: ok

PR #211, opened 2021-01-28
Support group transactions

amityadav0:


packages/runtime/src/runtime.ts
```diff
@@ -156,6 +158,18 @@ export class Runtime {
     return account.getLocalState(appId, key);
   }
 
+  /**
+   * Returns programs from programMap
```
ratik21, line 162:
```suggestion
   * Returns approval and clear program from programMap
```

packages/runtime/src/runtime.ts
```diff
@@ -264,6 +281,8 @@ export class Runtime {
     this.ctx.state.globalApps.delete(0); // remove zero app from context after execution
     senderAcc.createdApps.delete(0); // remove zero app from sender's account
     senderAcc.createdApps.set(this.appCounter, attributes);
+    // save programs in programMap
+    this.programMap.set(this.appCounter, [approvalProgram, clearProgram]);
```
ratik21, line 285: Can we use an object instead of array to represent approval and clear programs?
Eg `appId => {approval: approvalProgram, clear: clearProgram}` instead of `[approval, clear]`

packages/runtime/src/runtime.ts
```diff
@@ -301,12 +320,13 @@ export class Runtime {
...
   optInToApp (accountAddr: string, appId: number,
-    flags: SSCOptionalFlags, payFlags: TxParams, program: string): void {
+    flags: SSCOptionalFlags, payFlags: TxParams): void {
     const appParams = this.getApp(appId);
     const account = this.assertAccountDefined(accountAddr, this.store.accounts.get(accountAddr));
     if (appParams) {
       this.createOptInTx(accountAddr, appId, payFlags, flags);
-      this.run(program, ExecutionMode.STATEFUL); // execute TEAL code
+      // Execute approval program for Opt-In
+      this.run(this.getProgram(appId)[0], ExecutionMode.STATEFUL);
```
ratik21, line 329: Same as above. I think `this.getProgram(appId).approval` will be more intuitive

packages/runtime/src/runtime.ts
```diff
@@ -49,6 +50,7 @@ export class Runtime {
       args: []
     };
     this.appCounter = 0;
+    this.programMap = new Map<number, string[]>();
```
ratik21, line 53: Can we rename to `appPrograms` ? Or something similar (to represent application mapped to approval & clear programs)

packages/runtime/src/runtime.ts
```diff
@@ -471,80 +500,60 @@ export class Runtime {
...
+      switch (txnParam.type) {
+        case TransactionType.CallNoOpSSC:
+          this.run(this.getProgram(txnParam.appId)[0], ExecutionMode.STATEFUL); // approval program
+          break;
+        case TransactionType.CloseSSC:
+          this.run(this.getProgram(txnParam.appId)[0], ExecutionMode.STATEFUL); // approval program
+          break;
+        case TransactionType.DeleteSSC:
+          this.run(this.getProgram(txnParam.appId)[0], ExecutionMode.STATEFUL); // approval program
+          break;
```
ratik21, line 544: Instead of repeating 3 times can we do:
```
 case TransactionType.CallNoOpSSC || CloseSSC || DeleteSSC:
   this.run(this.getProgram(txnParam.appId)[0], ExecutionMode.STATEFUL); // approval program
   break;
```

packages/runtime/src/runtime.ts
```diff
@@ -471,80 +500,60 @@ export class Runtime {
...
+  executeTx (txnParams: ExecParams | ExecParams[], args?: Uint8Array[]): void {
     const [tx, gtxs] = this.createTxnContext(txnParams); // get current txn and txn group (as encoded obj)
 
     // initialize context before each execution
     this.ctx = {
       state: cloneDeep(this.store), // state is a deep copy of store
       tx: tx,
       gtxs: gtxs,
-      args: args
+      args: args ?? []
```
ratik21, line 523: Why do we need args now? I don't think external arguments are passed to stateful smart contracts. And args for stateless should be inside logic signature. Right?

packages/runtime/src/runtime.ts
```diff
@@ -471,80 +500,60 @@ export class Runtime {
...
+          break;
+        case TransactionType.ClearSSC:
+          // if transaction type is Clear Call, remove the app first before throwing error (rejecting tx)
+          // https://developer.algorand.org/docs/features/asc1/stateful/#the-lifecycle-of-a-stateful-smart-contract
+          fromAccount = this.getAccount(txnParam.fromAccount.addr);
           fromAccount.closeApp(txnParam.appId); // remove app from local state
-        }
+
+          this.run(this.getProgram(txnParam.appId)[1], ExecutionMode.STATEFUL); // clear program
+          break;
```
ratik21, line 552: Please include this case in brackets `{ }`.

packages/runtime/src/runtime.ts
```diff
@@ -471,80 +500,60 @@ export class Runtime {
...
+          this.run(this.getProgram(txnParam.appId)[0], ExecutionMode.STATEFUL); // approval program
+          break;
+        case TransactionType.ClearSSC:
+          // if transaction type is Clear Call, remove the app first before throwing error (rejecting tx)
+          // https://developer.algorand.org/docs/features/asc1/stateful/#the-lifecycle-of-a-stateful-smart-contract
+          fromAccount = this.getAccount(txnParam.fromAccount.addr);
           fromAccount.closeApp(txnParam.appId); // remove app from local state
-        }
+
+          this.run(this.getProgram(txnParam.appId)[1], ExecutionMode.STATEFUL); // clear program
```
ratik21, line 551: I see that we're deleting the app from local state before running the clear call TEAL code. Should we do that? Is there a possible case where the clear call teal code is getting/setting something in local state of app (maybe while using debugger)? Because if so, then the tx will fail (which it shouldn't, as we are explicitley removing local state). That's why i didn't clear the state before execution, only after (for both pass and fail cases). 
But we can keep this if it doesn't make sense to operate on local state in clear call.

packages/runtime/test/integration/basic-teal.ts
```diff
@@ -42,9 +42,13 @@ describe("Algorand Smart Contracts", function () {
     // check initial balance
     assert.equal(john.balance(), initialJohnHolding);
     assert.equal(bob.balance(), initialBobHolding);
+    // get logic signature
+    const lsig = runtime.getLogicSig(getProgram('basic.teal'), []);
+    lsig.sign(john.account.sk);
```
ratik21, line 47: Please update the title to reflect stateless ASC `delegated signature mode`.

packages/runtime/src/runtime.ts
```diff
@@ -380,6 +406,7 @@ export class Runtime {
 
     account.deleteApp(appId);
     this.store.globalApps.delete(appId);
+    this.programMap.delete(appId);
```
ratik21, line 409: hmmm, i see that we're storing application data in multiple maps. Can we use a single map? Maybe rename `globalApps` to `appData`/ `appConfig` and store:
```
  appData: Map[ appId => {
    address: addr // StoreAccount address
    approval: approvalProgram // application's approval program
    clear: clearProgram // application's clear program
    ... // more data
  }]
  ```

File: packages/runtime/src/runtime.ts [hunk 11]
```diff
@@ -471,80 +500,60 @@ export class Runtime {
...
     const txnParameters = Array.isArray(txnParams) ? txnParams : [txnParams];
-    try {
-      this.run(program, mode);
-    } catch (error) {
-      // if transaction type is Clear Call, remove the app first before throwing error (rejecting tx)
+    // Run TEAL program associated with each transaction without interacting with store.
+    for (const txnParam of txnParameters) {
+      this.assertAmbiguousTxnParams(txnParam);
+      if (txnParam.sign === SignType.LogicSignature) {
+        this.validateLsigAndRun(txnParam);
```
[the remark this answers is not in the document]
amityadav0: no we cannot, there may be a chance call application with delegated signature

packages/runtime/src/runtime.ts
amityadav0: yes

packages/runtime/src/runtime.ts
amityadav0: what are the benefits ?

packages/runtime/src/runtime.ts
amityadav0: i am not sure, previoulsy we were deleting from local state two times, that would have thrown error, so i deleted one, if you think this is right way i can move it.(this is part of clear SSC PR)

packages/runtime/test/integration/basic-teal.ts
amityadav0: ok

packages/runtime/src/runtime.ts
amityadav0: i thought it has some logic, are you sure about this ? i will remove

packages/runtime/src/runtime.ts
amityadav0: thanks for suggestion

packages/runtime/src/runtime.ts
amityadav0: just checked we are using these args in args opcode

amityadav0: that means we have to keep this in logicSig part.

packages/runtime/src/runtime.ts
ratik21: It was difficult to read during the review (had to open the file). That's why i suggested.

packages/runtime/src/runtime.ts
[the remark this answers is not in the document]
ratik21: I don't follow. Is there a case where sign type is `LogicSignature` (representing stateless tx), but there is an application call involved?

packages/runtime/src/runtime.ts
ratik21: Exactly. For stateless external arguments (using `Arg` op) are passed in logicSig, and for stateful we have application args (in `txParams.payflags.appArgs`)

packages/runtime/src/runtime.ts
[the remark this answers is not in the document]
amityadav0: yes, it is possible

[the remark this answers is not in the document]
ratik21: ok, will check

ratik21 (approved):

PR #213, opened 2021-02-03
Fix local state initialization after opting-in

sebastiangula:
If I don't change the order of method execution in the `optInToApp` in the `runtime.ts` then I get the following errors when running the test:

```
1) ASC - CloseOut from Application and Clear State
       should successfully closeOut from app and update state according to asc:
     Error: TEAL_ERR24: Application Index 1 not found or is invalid at line unknown
      at StoreAccount.closeApp (src/account.ts:208:13)
      at /Users/sebastiangula/algorand-builder/packages/runtime/src/runtime.ts:502:25
      at Array.forEach (<anonymous>)
      at Runtime.updateFinalState (src/runtime.ts:483:15)
      at Runtime.executeTx (src/runtime.ts:581:10)
      at Context.<anonymous> (test/integration/close-clear-ssc.ts:58:13)
      at processImmediate (node:internal/timers:463:21)

  2) ASC - CloseOut from Application and Clear State
       should not delete application on CloseOut call if logic is rejected:
     AssertionError: expected undefined to not equal undefined
      at Context.<anonymous> (test/integration/close-clear-ssc.ts:106:12)
      at processImmediate (node:internal/timers:463:21)

  3) ASC - CloseOut from Application and Clear State
       should delete application on clearState call even if logic is rejected:
     AssertionError: expected undefined to not equal undefined
      at Context.<anonymous> (test/integration/close-clear-ssc.ts:126:12)
      at processImmediate (node:internal/timers:463:21)

  4) Algorand Smart Contracts - Stateful Counter example
       "before all" hook for "should initialize local counter to 0 after opt-in":
     Error: TEAL_ERR24: Application Index 1 not found or is invalid at line 57
      at StoreAccount.setLocalState (src/account.ts:92:11)
      at AppLocalPut.execute (src/interpreter/opcode-list.ts:1743:32)
      at Interpreter.execute (src/interpreter/interpreter.ts:158:19)
      at Runtime.run (src/runtime.ts:592:17)
      at Runtime.optInToApp (src/runtime.ts:341:12)
      at Context.<anonymous> (test/integration/stateful-counter.ts:39:13)
      at processImmediate (node:internal/timers:463:21)
```
I'm not sure if this is the correct way to fix this issue as I'm still learning about the inner workings of Algorand.

### DIFF
--- a/packages/runtime/src/runtime.ts
+++ b/packages/runtime/src/runtime.ts
@@ -369,13 +369,15 @@ export class Runtime {
   optInToApp (accountAddr: string, appId: number,
     flags: SSCOptionalFlags, payFlags: TxParams): void {
     const appParams = this.getApp(appId);
-    const account = this.assertAccountDefined(accountAddr, this.store.accounts.get(accountAddr));
     if (appParams) {
       this.createOptInTx(accountAddr, appId, payFlags, flags);
+      this.ctx.state = cloneDeep(this.store);
+      const account = this.assertAccountDefined(accountAddr, this.ctx.state.accounts.get(accountAddr));
+      account.optInToApp(appId, appParams);
+
       // Execute approval program for Opt-In
       this.run(this.getProgram(appId).approvalProgram, ExecutionMode.STATEFUL);
-
-      account.optInToApp(appId, appParams);
+      this.store = this.ctx.state;
     } else {
       throw new TealError(ERRORS.TEAL.APP_NOT_FOUND, { appId: appId, line: 'unknown' });
     }
--- a/packages/runtime/test/integration/stateful-counter.ts
+++ b/packages/runtime/test/integration/stateful-counter.ts
@@ -1,7 +1,7 @@
 import { assert } from "chai";
 
 import { Runtime, StoreAccount } from "../../src/index";
-import { BIGINT1 } from "../../src/interpreter/opcode-list";
+import { BIGINT0, BIGINT1 } from "../../src/interpreter/opcode-list";
 import { ALGORAND_ACCOUNT_MIN_BALANCE } from "../../src/lib/constants";
 import { ExecParams, SignType, TransactionType } from "../../src/types";
 import { getProgram } from "../helpers/files";
@@ -41,6 +41,13 @@ describe("Algorand Smart Contracts - Stateful Counter example", function () {
   });
 
   const key = "counter";
+
+  it("should initialize local counter to 0 after opt-in", function () {
+    const localCounter = runtime.getAccount(john.address).getLocalState(txnParams.appId, key); // get local value from john account
+    assert.isDefined(localCounter); // there should be a value present in local state with key "counter"
+    assert.equal(localCounter, BIGINT0);
+  });
+
   it("should initialize global and local counter to 1 on first call", function () {
     runtime.executeTx(txnParams);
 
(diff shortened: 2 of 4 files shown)


robert-zaremba:
Thanks for the pull request and for adding a tests. We will have a look.

packages/runtime/src/runtime.ts
```diff
@@ -338,9 +338,9 @@ export class Runtime {
     const account = this.assertAccountDefined(accountAddr, this.store.accounts.get(accountAddr));
     if (appParams) {
       this.createOptInTx(accountAddr, appId, payFlags, flags);
-      this.run(program, ExecutionMode.STATEFUL); // execute TEAL code
-
       account.optInToApp(appId, appParams);
+
+      this.run(program, ExecutionMode.STATEFUL); // execute TEAL code
```
ratik21, line 343: Thanks for catching. Yes we should have a local state present before executing teal code. However, if the transaction fails, then the local state should not exist. 
To achieve this behaviour we can do a [`deepCopy`](https://medium.com/@manjuladube/understanding-deep-and-shallow-copy-in-javascript-13438bad941c) first (of the current state), then execute the transaction, and then finally update the 'real/actual' state with the copied one (if the tx execution passes successfully). Please let me know if that makes sense ;)
So, I would suggest to update the code of this function to:
```
  optInToApp (accountAddr: string, appId: number,
    flags: SSCOptionalFlags, payFlags: TxParams, program: string): void {
    const appParams = this.getApp(appId);
    if (appParams) {
      this.createOptInTx(accountAddr, appId, payFlags, flags);
      this.ctx.state = cloneDeep(this.store);
      const account = this.assertAccountDefined(
        accountAddr, this.ctx.state.accounts.get(accountAddr));
      account.optInToApp(appId, appParams);

      this.run(program, ExecutionMode.STATEFUL); // execute TEAL code
      this.store = this.ctx.state;
    } else {
      throw new TealError(ERRORS.TEAL.APP_NOT_FOUND, { appId: appId, line: 'unknown' });
    }
  }
```

ratik21 (changes requested):
Thanks for taking the time to create this PR @sebastiangula. I left a comment.

packages/runtime/test/integration/stateful-counter.ts
```diff
@@ -40,6 +40,13 @@ describe("Algorand Smart Contracts - Stateful Counter example", function () {
   });
 
   const key = "counter";
+
+  it("should initialize local counter to 0 after opt-in", function () {
+    const localCounter = runtime.getAccount(john.address).getLocalState(txnParams.appId, key); // get local value from john account
+    assert.isDefined(localCounter); // there should be a value present in local state with key "counter"
+    assert.equal(localCounter, BIGINT0);
+  });
```
ratik21, line 48: Thanks for this test!

packages/runtime/src/runtime.ts
sebastiangula: @ratik21 Thanks for the help, it makes sense. I've updated and rebased the PR.

ratik21 (approved):

packages/runtime/src/runtime.ts
robert-zaremba: other idea would be to create a `context` object and use exactly same pattern we have when processing a transaction. @ratik21 - what do you think?

ratik21: Continued on discord.

PR #214, opened 2021-02-03
Support Asset creation transactions in Runtime

amityadav0:
- Support Asset Creation Transaction
- Moved BuilderError class to runtime
- Moved validation of ASADef to runtime
- Moved required types also.
- we need to restructure the errors-list for both `runtime` and `algob`. I have created a task for it https://www.pivotaltracker.com/story/show/176797697


packages/algob/src/internal/deployer.ts
```diff
@@ -1,4 +1,9 @@
-import type { SSCDeploymentFlags, SSCOptionalFlags, TxParams } from "@algorand-builder/runtime/build/types";
+import { BuilderError } from "@algorand-builder/runtime";
+import type {
+  Account, AccountMap, ASADefs, ASADeploymentFlags,
+  SSCDeploymentFlags, SSCOptionalFlags,
+  TxParams
+} from "@algorand-builder/runtime/build/types";
```
ratik21, line 6: Can we import from `src` ?

packages/algob/src/lib/tx.ts
```diff
@@ -1,12 +1,10 @@
 import { encodeNote, mkTransaction } from "@algorand-builder/runtime";
-import type { ExecParams, TxParams } from "@algorand-builder/runtime/build/types";
+import type { ASADef, ASADeploymentFlags, ExecParams, TxParams } from "@algorand-builder/runtime/build/types";
```
ratik21, line 2: same

packages/runtime/src/account.ts
```diff
@@ -144,6 +150,30 @@ export class StoreAccount implements StoreAccountI {
     return this.appsLocalState.get(appId);
   }
 
+  /**
+   * Returns asset definitions
```
ratik21, line 154:
```suggestion
   * Queries asset definition by assetId
```

packages/runtime/src/account.ts
```diff
@@ -144,6 +150,30 @@ export class StoreAccount implements StoreAccountI {
...
+  getAssetDef (assetId: number): AssetDef | undefined {
+    return this.createdAssets.get(assetId);
+  }
+
+  /**
+   * Creates Asset in account's state
+   * @param name Asset Name
+   * @param asaDef Asset Definitions
+   */
+  createAsset (assetId: number, name: string, asaDef: ASADef, creator: string): AssetDef {
```
ratik21, line 166: We don't need to pass the `creator` here, we are already inside an account. You can use `this.address` instead of passing creator explicitly

packages/runtime/src/account.ts
```diff
@@ -144,6 +150,30 @@ export class StoreAccount implements StoreAccountI {
...
+    if (this.createdAssets.size === 1000) {
+      throw new Error('Maximum created assets for an account is 1000');
```
ratik21, line 168:
```suggestion
      throw new Error('Error while creating asset \"${asaDef.unitName}\" for ${this.address}. Maximum created assets for an account is 1000');
```

packages/runtime/src/account.ts
```diff
@@ -144,6 +150,30 @@ export class StoreAccount implements StoreAccountI {
...
+    return this.createdAssets.get(assetId);
+  }
+
+  /**
+   * Creates Asset in account's state
+   * @param name Asset Name
+   * @param asaDef Asset Definitions
+   */
+  createAsset (assetId: number, name: string, asaDef: ASADef, creator: string): AssetDef {
+    if (this.createdAssets.size === 1000) {
```
ratik21, line 167: Let's create a constant for these as well. 
```
MAX_ALGORAND_ACCOUNT_ASSETS = 1000
MAX_ALGORAND_ACCOUNT_APPS = 10
```

packages/runtime/src/errors/errors-list.ts
```diff
@@ -264,18 +265,52 @@ by an index that does not exist.`
...
+    message: `Invalid ASA definition: '%filename%'.
+Reason:
+
+%reason%`,
+    title: "Invalid ASA definition",
+    description: `Invalid ASA definition: '%filename%'.
+
+Reason:
+%reason%
+
```
ratik21, line 287: Is the indentation here correct?

packages/runtime/src/errors/errors.ts
```diff
@@ -39,6 +41,44 @@ export class TealError extends Error {
   }
 }
 
+export class BuilderError extends Error {
```
ratik21, line 44: should we keep this in a separate file? Maybe `builder-errors.ts`. And rename this to `teal-errors.ts`?

packages/runtime/src/interpreter/opcode-list.ts
```diff
@@ -1982,6 +1982,9 @@ export class GetAssetDef extends Op {
         case "AssetDecimals":
           value = BigInt(AssetDefinition.decimals);
           break;
+        case "AssetDefaultFrozen":
+          value = AssetDefinition["default-frozen"] ? 1n : 0n;
+          break;
```
ratik21, line 1987: Was it missing before? We probably need to update the test as well for `GetAssetDef` Op, right?

packages/runtime/src/lib/asa.ts
```diff
@@ -1,36 +1,39 @@
...
+
+function validateSingle (accounts: AccountMap | RuntimeAccountMap, filename: string, asaDef: ASADef): void {
   if (!asaDef.optInAccNames || asaDef.optInAccNames.length === 0) {
     return;
   }
   for (const accName of asaDef.optInAccNames) {
     if (!accounts.get(accName)) {
       throw new BuilderError(
-        ERRORS.SCRIPT.ASA_PARAM_ERROR_NO_NAMED_OPT_IN_ACCOUNT, {
+        ERRORS.ASA.ASA_PARAM_ERROR_NO_NAMED_OPT_IN_ACCOUNT, {
```
ratik21, line 20: If using `ASA` instead of `SCRIPT`, then we don't need the `ASA_..` prefix. We can rename to `ERRORS.ASA.PARAM_ERROR_NO_NAMED_OPT_IN_ACCOUNT`. Or maybe in the new task.

packages/runtime/src/lib/asa.ts
```diff
@@ -1,36 +1,39 @@
...
           filename: filename,
           optInAccName: accName
         });
     }
   }
 }
 
-function validateParsedASADefs (accounts: AccountMap, asaDefs: ASADefs, filename: string): void {
+function validateParsedASADefs (
+  accounts: AccountMap | RuntimeAccountMap, asaDefs: ASADefs, filename: string): void {
```
ratik21, line 29: Please add a doc string explaining the two account types. `AccountMap` is the SDK account type, used in builder. `RuntimeAccountMap` is for `StoreAccount` used in runtime (where we use maps instead of arrays in sdk structures).

packages/runtime/src/lib/asa.ts
```diff
@@ -43,7 +46,7 @@ export function validateASADefs (obj: Object, accounts: AccountMap, filename: st
   } catch (e) {
     if (e instanceof z.ZodError) {
       throw new BuilderError(
-        ERRORS.SCRIPT.ASA_PARAM_PARSE_ERROR, {
+        ERRORS.ASA.ASA_PARAM_PARSE_ERROR, {
```
ratik21, line 49: Same as above. We can do it here, or in the new task you created

packages/runtime/src/lib/constants.ts
```diff
@@ -10,6 +10,7 @@ export const ALGORAND_ACCOUNT_MIN_BALANCE = 1e6; // 1 ALGO
 // values taken from: https://developer.algorand.org/docs/features/asc1/stateful/#minimum-balance-requirement-for-a-smart-contract
 // minimum balance costs (in microalgos) for ssc schema
 export const APPLICATION_BASE_FEE = 0.1e6; // base fee for creating or opt-in to application
+export const ASSET_CREATION_FEE = 0.1e6; // creation fee for asset
```
ratik21, line 13: Can we merge these constants with a single name? Maybe `BASE_FEE`, as it's value is also used in `app-opt-in` & `asset-opt-in`

packages/runtime/src/runtime.ts
```diff
@@ -259,6 +281,36 @@ export class Runtime {
     }
   }
 
+  /**
+   * Creates Asset in Runtime
+   * @param name ASA name
+   * @param flags ASA Deployment Flags
+   */
+  createAsset (name: string, flags: ASADeploymentFlags): number {
```
ratik21, line 289: Let's keep the naming similar to `app`. We are using `addApp`, so please rename this to `addAsset`, OR rename `addApp` to `createApp` as well. I would prefer (i think robert suggested this earlier): 
```
runtime.createAsset(..) => account.addAsset(..)
runtime.createApp(..) => account.addApp(..)
```

packages/runtime/src/runtime.ts
```diff
@@ -259,6 +281,36 @@ export class Runtime {
...
+    const sender = flags.creator;
+    const senderAcc = this.assertAccountDefined(sender.addr, this.store.accounts.get(sender.addr));
+
+    // create asset
+    senderAcc.createAsset(++this.assetCounter, name, this.loadedAssetsDefs[name], sender.addr);
+    this.store.assetDefs.set(this.assetCounter, sender.addr);
+    return this.assetCounter;
+  }
```
ratik21, line 297: I see that we're not creating the assetCreation tx from sdk (`makeAssetCreateTxn(..)` from `algosdk`). I realize that we don't need to set ctx here, as we're not executing TEAL code. But the SDK constructor validates the `txParams (flags, payflags, asaDefs, addresses..)`, so we should either cover these validations or use the `sdk tx constructor`. Ignore if validations are already covered.

packages/runtime/src/runtime.ts
```diff
@@ -259,6 +281,36 @@ export class Runtime {
...
+
+  /**
+   * Returns Asset Definitions
+   * @param assetId Asset Index
+   */
+  getAssetDef (assetId: number): AssetDef {
+    const addr = this.store.assetDefs.get(assetId);
+    if (addr === undefined) {
+      throw new TealError(ERRORS.ASA.ASSET_NOT_FOUND, { assetId: assetId });
+    }
```
ratik21, line 307: We should use `this.assertAddressDefined` here (which throws the correct error). Address is undefined here, not the asset.

packages/runtime/src/runtime.ts
```diff
@@ -259,6 +281,36 @@ export class Runtime {
...
+
+    const creatorAcc = this.assertAccountDefined(addr, this.store.accounts.get(addr));
+    const assetDef = creatorAcc.getAssetDef(assetId);
+    return this.assertAssetDefined(assetId, assetDef);
```
ratik21, line 311: Here we can throw `ASSET_NOT_FOUND`.

packages/runtime/test/src/runtime.ts
```diff
@@ -136,3 +136,20 @@ describe("Rounds Test", function () {
...
+  this.beforeAll(() => {
+    runtime = new Runtime([john]);
+  });
+
+  it("should create asset using asa.yaml file", () => {
+    const assetId = runtime.createAsset('gold',
+      { creator: { name: "john", addr: john.address, sk: john.account.sk } });
+
+    const res = runtime.getAssetDef(assetId);
+    assert.equal(res.decimals, 0);
```
ratik21, line 153: Please update to check few more params (manager addr, unitName, ..)

packages/algob/src/internal/deployer.ts
amityadav0: tried but cannot import types

packages/algob/src/lib/tx.ts
amityadav0: same

packages/runtime/src/errors/errors.ts
amityadav0: i am thinking of creating `runtime-error.ts` and restructrue erros properly in different task,, will update this one

packages/runtime/src/interpreter/opcode-list.ts
amityadav0: no, previoulsy we were considering default-frozen as string but it is boolean

packages/runtime/src/lib/asa.ts
amityadav0: ?

packages/runtime/src/lib/asa.ts
amityadav0: yes

packages/runtime/src/lib/constants.ts
amityadav0: not sure, this gives more clarity

packages/runtime/src/runtime.ts
amityadav0: ok

packages/runtime/src/runtime.ts
amityadav0: if payFlags are validated then we can add it, only asaDefs are covered.

packages/runtime/src/runtime.ts
amityadav0: if assetId doesn't exist in map, what does it mean ?

packages/runtime/src/runtime.ts
amityadav0: we are doing that already.

packages/runtime/test/src/runtime.ts
amityadav0: ok, they are constant though

packages/runtime/src/runtime.ts
ratik21: hmm, right. Possible that the account & address exists but asset in account doesn't. Then we should update `getApp` as well, as it's using `assertAddressDefined` with globalApps.

packages/runtime/src/runtime.ts
ratik21: i know

packages/runtime/src/lib/constants.ts
ratik21: ok

packages/runtime/src/lib/asa.ts
ratik21: `ERRORS.ASA.ASA_PARAM_ERROR_NO_NAMED_OPT_IN_ACCOUNT` => `ERRORS.ASA.PARAM_ERROR_NO_NAMED_OPT_IN_ACCOUNT`

packages/runtime/src/interpreter/opcode-list.ts
ratik21: ok

packages/runtime/src/errors/errors.ts
ratik21: ok

packages/runtime/src/lib/asa.ts
amityadav0: done in this one

packages/runtime/src/lib/asa.ts
amityadav0: hmm

packages/algob/src/internal/deployer.ts
ratik21: I confirm. I think tests are working because we use `ts-node` there. But in examples we use the `tsc` compiler, and the build is only exposing the main file (`index.js`).

packages/runtime/src/runtime.ts
```diff
@@ -259,6 +281,59 @@ export class Runtime {
     }
   }
 
+  // creates new asset creation transaction object and update context
+  addCtxAssetCreate (
```
ratik21, line 285:
```suggestion
  mkAssetCreateTx (
```

packages/runtime/src/account.ts
```diff
@@ -144,15 +151,41 @@ export class StoreAccount implements StoreAccountI {
...
+
+  /**
+   * Creates Asset in account's state
+   * @param name Asset Name
+   * @param asaDef Asset Definitions
+   */
+  addAsset (assetId: number, name: string, asaDef: ASADef): AssetDef {
+    if (this.createdAssets.size === MAX_ALGORAND_ACCOUNT_ASSETS) {
+      throw new TealError(ERRORS.TEAL.MAX_LIMIT_ASSETS,
+        { name: name, address: this.address, max: MAX_ALGORAND_ACCOUNT_APPS });
```
ratik21, line 170:
```suggestion
        { name: name, address: this.address, max: MAX_ALGORAND_ACCOUNT_ASSETS });
```

packages/runtime/src/errors/errors-list.ts
```diff
@@ -259,23 +260,63 @@ by an index that does not exist.`
     message: "Transaction rounds (firstValid: %first%, lastValid: %last%) are not valid, current round: %round%.",
     title: 'Round Error',
     description: `Round Error`
+  },
+  MAX_LIMIT_ASSETS: {
+    number: 39,
+    message: "Error while creating asset %name% for %address%. Maximum created assets for an account is %max%}'",
```
ratik21, line 266:
```suggestion
    message: "Error while creating asset %name% for %address%. Maximum created assets for an account is %max%'",
```

packages/runtime/src/errors/errors-list.ts
```diff
@@ -259,23 +260,63 @@ by an index that does not exist.`
...
+    title: 'Asset Creation Error',
+    description: `Asset Creation Error`
+  },
+  MAX_LIMIT_APPS: {
+    number: 40,
+    message: "Error while creating app for %address%. Maximum created apps for an account is %max%}'",
```
ratik21, line 272:
```suggestion
    message: "Error while creating app for %address%. Maximum created apps for an account is %max%",
```

packages/algob/src/internal/deployer.ts
ratik21: Could you check if `@algorand-builder/runtime/types` works ? (Since `/src` folder is not built, i think we can do this)

amityadav0: already checked, doesn't work, also in htlc-example we are using `build` that's why it is working

ratik21 (approved):

PR #218, opened 2021-02-04
Remove symlinks for tsconfig.json and .mocharc.json

### LINKED ISSUE #216: Tsc error during first build
During the installation, the yarn build gives an error because the tsconfig of algob and runtime packages seem to be invalid 

image

ratik21:
fixes #216

### DIFF
--- a/packages/runtime/LICENSE
+++ b/packages/runtime/LICENSE
@@ -1,28 +1,201 @@
-The ALGOB is licensed under the Apache-2.0 license as defined in the
-algorand-builder repository (https://github.com/scale-it/algorand-builder).
-Code originating between commits
-eb85a4b299841431c40cf27bafc57ad2e567b1ee and 1dca01a038fd4d68b238f2310921c8654182be65
-It is licensed under MIT license as defined below.
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
 
-The MIT License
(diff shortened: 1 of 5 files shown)


packages/runtime/tsconfig.json
```diff
@@ -1 +1,10 @@
...
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "outDir": "./build"
+  },
+  "include": [
+    "src",
+    "../../@types"
+  ]
+}
```
amityadav0, line 10: Please update the license file as well.

ratik21: thanks

packages/runtime/LICENSE
```diff
@@ -1 +1,28 @@
-../../LICENSE
\ No newline at end of file
+The ALGOB is licensed under the Apache-2.0 license as defined in the
```
robert-zaremba, line 1: Please copy the license file from the root directory.

robert-zaremba (approved):
Preapproving - please change the license file.

packages/runtime/LICENSE
robert-zaremba: so we should just use the Apache-2.0 license.

ratik21: ok

PR #220, opened 2021-02-04
Support Asset transfer and opt-in in Runtime

ratik21:
- Added asset opt-in, tests
- Added asset transfer, tests


packages/runtime/src/account.ts
```diff
@@ -227,6 +235,22 @@ export class StoreAccount implements StoreAccountI {
     }
   }
 
+  // opt-in to asset
+  optInToASA (assetIndex: number, assetHolding: AssetHolding): void {
+    const localAsset = this.assets.get(assetIndex); // fetch asset holding of account
```
amityadav0, line 240: how about renaming to `localAsset` to `assetHolding` ?

packages/runtime/src/errors/errors-list.ts
```diff
@@ -308,6 +314,12 @@ const asaErrors = {
     message: `Asset with Index %assetId% not found`,
     title: "Asset Not Found",
     description: "Asset Not Found"
+  },
+  ASSET_HOLDING_NOT_FOUND: {
+    number: 903,
+    message: `Account %address% does not hold asset index %assetId%`,
```
amityadav0, line 320:
```suggestion
    message: `Account %address% doesn't hold asset index %assetId%`,
```

packages/runtime/src/interpreter/opcode-list.ts
```diff
@@ -1921,7 +1921,7 @@ export class GetAssetHolding extends Op {
         value = BigInt(assetInfo.amount);
         break;
       case "AssetFrozen":
-        value = stringToBytes(assetInfo["is-frozen"]);
+        value = BigInt(assetInfo["is-frozen"]);
```
amityadav0, line 1924: assetInfo["is-frozen"] ?  1n : 0n;
how about using the above ?
i think this is more intuitive.

packages/runtime/src/runtime.ts
```diff
@@ -314,11 +314,35 @@ export class Runtime {
...
+    this.mkAssetCreateTx(name, flags, asset);
     this.store.assetDefs.set(this.assetCounter, sender.addr);
     return this.assetCounter;
   }
 
+  /**
+   * Asset Opt-In for account in Runtime
+   * @param assetIndex Asset Index
+   * @param address Account address to opt-into asset
+   * @param flags txParams
```
amityadav0, line 326:
```suggestion
   * @param flags Transaction Parameters
```

packages/runtime/src/runtime.ts
```diff
@@ -580,6 +621,29 @@ export class Runtime {
     }
   }
 
+  // transfer ASSET as per transaction parameters
+  transferAsset (txnParam: AssetTransferParam): void {
+    const fromAssetHolding = this.getAssetHolding(txnParam.assetID, txnParam.fromAccount.addr);
+    const toAssetHolding = this.getAssetHolding(txnParam.assetID, txnParam.toAccountAddr);
+
```
amityadav0, line 628: let's add a check here for asset frozen.

packages/runtime/src/account.ts
ratik21: renamed

packages/runtime/src/interpreter/opcode-list.ts
ratik21: I don't have a particular preference. But will update if you think this is more intuitive

packages/runtime/src/runtime.ts
ratik21: hmm, thanks. So we should also check if asset is frozen or not during opt-in?

ratik21: Added test as well.

amityadav0: Thanks

packages/runtime/src/runtime.ts
```diff
@@ -580,6 +632,30 @@ export class Runtime {
...
+    this.assertAssetNotFrozen(txnParam.assetID, txnParam.fromAccount.addr);
```
amityadav0, line 640: does it mean `to` can receive assets even if it is frozen ? could you please confirm ?

ratik21: It's probably both. But i can't test it to be 100% sure, as algob doesn't support asset freeze right now. Will validate after we add this support.
```
An Asset Freeze Transaction is issued by the Freeze Address and results in the asset receiver address losing or being granted the ability to send or receive the frozen asset.
```
Adding assert for `to` as well.

packages/runtime/src/interpreter/opcode-list.ts
robert-zaremba: I like @amityadav0  suggestion.

amityadav0 (approved):

PR #222, opened 2021-02-05
Export types

amityadav0:


examples/htlc-pyteal-ts/scripts/withdraw/htlc-withdraw.ts
```diff
@@ -7,21 +7,22 @@
...
   const wrongSecret = 'hero wisdom red split loop element vote belt';
 
   let lsig = await deployer.loadLogic('htlc.py', [stringToBytes(wrongSecret)], scTmplParams);
   let sender = lsig.address();
 
-  const txnParams: AlgoTransferParam = {
-    type: TransactionType.TransferAlgo,
-    sign: SignType.LogicSignature,
+  const txnParams: runtime.types.AlgoTransferParam = {
+    type: runtime.types.TransactionType.TransferAlgo,
```
ratik21, line 24: Is there a reason in `ts` files we are using `runtime.types` and in other js examples we are using only types? BDW, i think using `runtime.types` is more intuitive. However, to avoid long names, can we do:
```
import * as runtime from "@algorand-builder/runtime";
const { TransactionType, SignType } = runtime.types;
...
use types below..

```

examples/asa/scripts/transfer/gold-contract-sc.js
```diff
@@ -3,7 +3,7 @@
  * This file demonstrates the example to transfer MicroAlgos
  * from contract account (contract approval mode) to another according to smart contract (ASC) logic
 */
-const { TransactionType, SignType } = require('@algorand-builder/runtime/build/types');
+const { types } = require('@algorand-builder/runtime');
```
ratik21, line 6: can we do: 
```
const { TransactionType, SignType  } = require('@algorand-builder/runtime').types;
..
```
OR
```
const runtime = require('@algorand-builder/runtime');
const { TransactionType, SignType  } = runtime.types;
..
```
Or we can keep this as well.

examples/htlc-pyteal-ts/scripts/withdraw/htlc-withdraw.ts
amityadav0: we can use runtime.types in js files also, but there are only two imports that only from runtime, so I think types. is better

I am not sure about the second point, in other files we have algob and runtime, so runtime.types is more intuitive

examples/asa/scripts/transfer/gold-contract-sc.js
robert-zaremba: I don't have a preference.

packages/algob/src/types.ts
```diff
@@ -1,9 +1,4 @@
-import type {
-  Account, AccountMap,
-  ASADeploymentFlags,
-  SSCDeploymentFlags, SSCOptionalFlags,
-  TxParams
-} from "@algorand-builder/runtime/build/types";
+import * as runtime from "@algorand-builder/runtime";
```
robert-zaremba, line 1: we can add a variable to make things shorter:
```
const rtypes = runtime.types
```

robert-zaremba (approved):
utACK

packages/algob/src/types.ts
amityadav0: ok

examples/asa/scripts/transfer/gold-contract-sc.js
amityadav0: ok, keeping same with shorter variable.

PR #223, opened 2021-02-07
Add support for all other Assets transactions in Runtime

amityadav0:
- Asset modification/configuration
   - Asset freeze
   - Asset Revoke
   - Asset Destroy


packages/runtime/src/account.ts
```diff
@@ -184,6 +184,58 @@ export class StoreAccount implements StoreAccountI {
...
+   */
+  modifyAsset (assetId: number, fields: AssetModFields): void {
+    const asset = this.getAssetDef(assetId);
+    if (asset === undefined) {
+      throw new TealError(ERRORS.ASA.ASSET_NOT_FOUND, { assetId: assetId });
+    }
+    // check for blank fields
+    if ((fields.reserve && asset.reserve === "") || (fields.freeze && asset.freeze === "") ||
+      (fields.clawback && asset.clawback === "")) {
+      throw new Error("Cannot reset a blank address");
```
robert-zaremba, line 200: why? 
I think this should be correct. Example: What if `asset.reserve = ""` and we want to update `asset.freeze`? Then we will have `{reserve: "", freeze: "something else"}`

packages/runtime/src/account.ts
```diff
@@ -184,6 +184,58 @@ export class StoreAccount implements StoreAccountI {
...
+    asset.freeze = fields.freeze;
+    asset.clawback = fields.clawback;
+  }
+
+  /**
+   * Freeze asset
+   * @param assetId Asset Index
+   */
+  freezeAsset (assetId: number, state: boolean): void {
+    const holding = this.assets.get(assetId);
```
robert-zaremba, line 214:
```suggestion
    const account = this.assets.get(assetId);
```

robert-zaremba: I don't think holding is the right word to be used here. We are getting an account, host account. So other good name would be `host`.

packages/runtime/src/account.ts
```diff
@@ -184,6 +184,58 @@ export class StoreAccount implements StoreAccountI {
...
+    if (holding === undefined) {
+      throw new TealError(ERRORS.ASA.ASSET_HOLDING_NOT_FOUND,
```
robert-zaremba, line 216: As above, please rename HOLDING in the errors.

packages/runtime/src/runtime.ts
```diff
@@ -210,6 +211,28 @@ export class Runtime {
     return account.getLocalState(appId, key);
   }
 
+  /**
+   * Returns asset creator account or throws error is it doesn't exist
+   * @param Asset Index
+   */
+  getAssetCreatorAccount (assetId: number): StoreAccountI {
```
robert-zaremba, line 218:
```suggestion
  getAssetAccount (assetId: number): StoreAccountI {
```

packages/runtime/src/runtime.ts
```diff
@@ -325,18 +363,143 @@ export class Runtime {
...
+   */
+  revokeAsset (
+    sender: string, recipient: string,
+    assetID: number, revocationTarget: string,
+    amount: number, payFlags: TxParams
+  ): void {
+    const asset = this.getAssetDef(assetID);
+    if (asset.clawback !== sender) {
+      throw new Error("Only Clawback account can revoke assets");
+    }
```
robert-zaremba, line 456: We must do this checks when we process transactions.

robert-zaremba (changes requested):
I left few remarks.

packages/runtime/src/account.ts
amityadav0: Here we are getting assetHolding from local state of a account using
This.assets.get(), assets is a map which stores AssetHolding Type as value.

Should I change it to something else ?

packages/runtime/src/account.ts
amityadav0: Yes, in this idea I thought user will set reserve as undefined if he doesn't want to update.

I will update to use "" blank address

packages/runtime/src/runtime.ts
amityadav0: Ok moving to mkTx function

packages/runtime/src/runtime.ts
```diff
@@ -324,19 +347,160 @@ export class Runtime {
...
+    this.mkAssetConfigTx(sender, assetId, fields, payFlags);
+    const creatorAcc = this.getAssetAccount(assetId);
+    creatorAcc.modifyAsset(assetId, fields);
+  }
+
+  // creates asset freeze transaction object.
+  mkAssetFreezeTx (
+    sender: string, assetId: number, target: string, state: boolean, flags: TxParams): void {
+    // this funtion is called only for validation of parameters passed
+    algosdk.makeAssetFreezeTxnWithSuggestedParams(
```
ratik21, line 388: Is it possible to have these types in `ExecParams` itself (Since we already have `AssetTransfer`) ? Then i think we could make these transactions at a single place (in mkTransaction) instead of in different functions. Also, it would be common for `algob` & `runtime` as well. Can we do that?

packages/runtime/src/runtime.ts
```diff
@@ -324,19 +347,160 @@ export class Runtime {
...
+      sender,
+      encodeNote(flags.note, flags.noteb64),
+      assetId,
+      target,
+      state,
+      mockSuggestedParams(flags, this.round)
+    );
+    const asset = this.getAssetDef(assetId);
+    if (asset.freeze !== sender) {
+      throw new Error("Only Freeze account can freeze asset");
```
ratik21, line 398:
```suggestion
      throw new Error(`Only Freeze account(${asset.freeze}) can freeze asset`);
```

packages/runtime/src/account.ts
```diff
@@ -184,6 +184,58 @@ export class StoreAccount implements StoreAccountI {
...
+    asset.reserve = fields.reserve;
+    asset.freeze = fields.freeze;
+    asset.clawback = fields.clawback;
+  }
+
+  /**
+   * Freeze asset
+   * @param assetId Asset Index
+   */
+  freezeAsset (assetId: number, state: boolean): void {
```
ratik21, line 213: Please update the description

packages/runtime/src/runtime.ts
```diff
@@ -324,19 +347,160 @@ export class Runtime {
...
+      flags.closeRemainderTo,
+      revocationTarget,
+      amount,
+      encodeNote(flags.note, flags.noteb64),
+      assetId,
+      mockSuggestedParams(flags, this.round)
+    );
+    const asset = this.getAssetDef(assetID);
+    if (asset.clawback !== sender) {
+      throw new Error("Only Clawback account can revoke assets");
```
ratik21, line 438:
```suggestion
      throw new Error(`Only Clawback account(${asset.clawback}) can revoke assets`);
```

ratik21: Also, should we move this check before `makeAssetTransferTxnWithSuggestedParams`?

packages/runtime/src/runtime.ts
```diff
@@ -324,19 +347,160 @@ export class Runtime {
...
+    this.assertAssetNotFrozen(assetID, recipient);
+
+    if (fromAssetHolding.amount - amount < 0) {
+      throw new TealError(ERRORS.TRANSACTION.INSUFFICIENT_ACCOUNT_ASSETS, {
+        amount: amount,
+        address: revocationTarget
+      });
+    }
+    fromAssetHolding.amount -= amount;
+    toAssetHolding.amount += amount;
```
ratik21, line 472: Can we use the `runtime.transferAsset` function here?

packages/runtime/src/runtime.ts
```diff
@@ -324,19 +347,160 @@ export class Runtime {
...
+  // creates asset destroy transaction object.
+  mkDestroyAssetTx (sender: string, assetId: number, flags: TxParams): void {
+    // this funtion is called only for validation of parameters passed
+    algosdk.makeAssetDestroyTxnWithSuggestedParams(
+      sender, encodeNote(flags.note, flags.noteb64),
+      assetId, mockSuggestedParams(flags, this.round)
+    );
+    const asset = this.getAssetDef(assetId);
+    if (asset.manager !== sender) {
+      throw new Error("Only Manager account can destroy assets");
```
ratik21, line 484: Please update this error as well. BDW - shouldn't this be a `TRANSACTION` / `ASA` error ?

packages/runtime/src/runtime.ts
```diff
@@ -324,19 +347,160 @@ export class Runtime {
...
+   * @param sender sender's address
+   * @param assetId asset index
+   * @param payFlags transaction parameters
+   */
+  destroyAsset (sender: string, assetId: number, payFlags: TxParams): void {
+    this.mkDestroyAssetTx(sender, assetId, payFlags);
+    const creatorAcc = this.getAssetAccount(assetId);
+    // destroy asset from creator's account
+    creatorAcc.destroyAsset(assetId);
+    // delete assets from all accounts
```
ratik21, line 500:
```suggestion
    // delete asset holdings from all accounts
```

packages/runtime/test/src/runtime.ts
```diff
@@ -286,4 +293,164 @@ describe("Algorand Standard Assets", function () {
...
+  it("should throw error if asset is not found while modifying", () => {
+    expectTealError(
+      () => runtime.modifyAsset(john.address, 120, modFields, {}),
+      ERRORS.ASA.ASSET_NOT_FOUND
+    );
+  });
+
+  it("should modify asset", () => {
+    const assetId = runtime.createAsset('gold',
+      { creator: { name: "john", addr: john.address, sk: john.account.sk } });
```
ratik21, line 306: Can we use: `creator: { ...john.account, name: "john" }`

packages/runtime/src/runtime.ts
amityadav0: no, because that's not the flow.
first we create the tx and then send it to network

packages/runtime/src/runtime.ts
amityadav0: no,If you look at code we need secret key for the revocationTarget and we don't have that.

packages/runtime/src/runtime.ts
amityadav0: update to what ? yes, we can but we will only throw this error here, i don't think it will be repetitive.
If you think it will i can add all error in error list

packages/runtime/src/account.ts
```diff
@@ -184,6 +184,58 @@ export class StoreAccount implements StoreAccountI {
...
+    const holding = this.assets.get(assetId);
+    if (holding === undefined) {
+      throw new TealError(ERRORS.ASA.ASSET_HOLDING_NOT_FOUND,
+        { address: this.address, assetId: assetId });
+    }
+    holding["is-frozen"] = state;
```
ratik21, line 219: What happens if asset is already frozen?  Is an error thrown or does it has no effect?
Also, could we rename this function to something else, `freezeAsset` suggests `is-frozen` is always true (i.e we are 'freezing' the asset). Maybe `setFreezeState()` ?

packages/runtime/src/account.ts
```diff
@@ -184,6 +184,58 @@ export class StoreAccount implements StoreAccountI {
...
+  }
+
+  /**
+   * Destroys asset
+   * @param assetId Asset Index
+   */
+  destroyAsset (assetId: number): void {
+    const holding = this.assets.get(assetId);
+    const asset = this.getAssetDef(assetId);
+    if (holding === undefined || asset === undefined) {
```
ratik21, line 229: Isn't it possible that `holding` can be `undefined` ? What if account A created the asset, and accounts B, C opt-in. Then i do `destroyAsset`, in this case assetHolding of creator is undefined but i should be able to delete the asset, right?

packages/runtime/src/account.ts
amityadav0: no effect, yeah setFreezeState is better, algorand uses makeFreezeAsset though

packages/runtime/test/src/runtime.ts
```diff
@@ -286,4 +293,164 @@ describe("Algorand Standard Assets", function () {
...
+    const modFields: AssetModFields = {
+      manager: bob.address,
+      reserve: bob.address,
+      clawback: john.address,
+      freeze: alice.address
+    };
+
+    expect(() => {
+      runtime.modifyAsset(elon.address, assetId, modFields, {});
+    }).to.throw("Cannot reset a blank address");
```
ratik21, line 330: Can we use `assert.throws` ? Since in the other tests we are using `assert`

amityadav0: this is more intuitive and clear

packages/runtime/src/account.ts
amityadav0: will check more, couldn't find about this

packages/types-algosdk/types/index.d.ts
```diff
@@ -375,7 +375,7 @@ declare module 'algosdk' {
 
   export function makeAssetConfigTxnWithSuggestedParams(from: any, note: any, assetIndex: any,
     manager: any, reserve: any, freeze: any, clawback: any, suggestedParams: any,
-    strictEmptyAddressChecking: any): any;
+    strictEmptyAddressChecking?: any): any;
```
ratik21, line 378: Please update this in the above function type as well (`makeAssetConfigTxn`) as well

amityadav0: ?

amityadav0: ah, okay

packages/runtime/test/src/runtime.ts
ratik21: Ok. Then can we update other tests (in ASA) as well? Currently there is mix up of `asset.throws` and `expect.to.throw` here

packages/runtime/src/runtime.ts
ratik21: Update to `Only Manager account(<manager_address>) can destroy assets`, to make the error more intuitive. 
Don't have a particular preference regarding the error list. Your call.

packages/runtime/src/runtime.ts
ratik21: Yes, but this entire function constitues mkTx right? I was thinking to avoid calling SDK constructor, if we can throw this error beforeHand.

amityadav0: can you please elaborate ?

packages/runtime/test/src/runtime.ts
amityadav0: changed these errors to ASA errors, using expectTealError is better

packages/runtime/src/account.ts
amityadav0: yes, because algorand states that all the assets must be present in creator's account and that is only possible when creator opts-in. we can't destroy asset until it is fully present in asset creator's account

packages/runtime/src/runtime.ts
amityadav0: seeing from user perspective i think using functions for ASA is easier because ExecParams already have application types + payment + lsig, not sure though if we should change current structure

packages/runtime/test/src/runtime.ts
ratik21: Yes, but it should be `expectRuntimeError` (TEAL is not associated here). We should update this in restructuring

amityadav0: Yes that's why added that task. Not possible to fix all things in one big task

ratik21: I know, was confirming so that we don't miss it.

packages/runtime/src/runtime.ts
ratik21: This function is `mkTx` function, i am assuming that line `437` is our validation during createTx (after SDK). So we could do this validation before calling sdk (so if error is thrown we don't need to call an external module(`algosdk`) constructor). Ignore if that's not case.

amityadav0: we cannot do that because information for asset.clawback is stored in network, therfore first we create tx and then do what network does

packages/runtime/src/runtime.ts
ratik21: I think from the user perspective, keeping all execution transactions in `TransactionType` makes more sense, user can use `TransactionType.ModifyAsset`, `TransactionType.TransferAsset`,.. instead of looking for each specific function.

Also think about group transactions, by this structure we can't support asset `freeze`, `clawback`,.. in a group of transactions.

packages/runtime/src/runtime.ts
```diff
@@ -324,19 +347,160 @@ export class Runtime {
...
+    sender: string, recipient: string,
+    assetID: number, revocationTarget: string,
+    amount: number, payFlags: TxParams
+  ): void {
+    this.mkAssetRevokeTx(sender, recipient, assetID, revocationTarget, amount, payFlags, assetID);
+    // Transfer assets
+    const fromAssetHolding = this.getAssetHolding(assetID, revocationTarget);
+    const toAssetHolding = this.getAssetHolding(assetID, recipient);
+    this.assertAssetNotFrozen(assetID, revocationTarget);
+    this.assertAssetNotFrozen(assetID, recipient);
```
ratik21, line 463: I don't think this is correct. I was reading the docs and found that asset `clawback` can revoke an asset **even if the asset is frozen**. Can you please confirm?

packages/runtime/src/runtime.ts
amityadav0: shouldn't we do this for opt-in for asa or app as well ? also for update, app ?
we should be consistent then.

what if group transaction include opt-in + payment ?

packages/runtime/src/runtime.ts
amityadav0: please share the link

amityadav0: couldn't find anything in this main doc https://developer.algorand.org/docs/features/asa/#revoking-an-asset

packages/runtime/src/runtime.ts
ratik21: For update app: yes, i agree. 

For createApp, opt-in i think these are used for setting up (deploying) the application/asset. In examples we use them during the deployment mode (not while RUN, where we actually use `executeTransaction`). Also couldn't find any examples in reference where creation/opt-in transactions are used in a group, but that doesn't mean it's not possible.

Anyway, for asset freeze, clawback,.. transactions, we should keep them in `ExecParams` because they are used in the RUN mode, they are not related to setting up of an asset. (User will use them in `/scripts/transfer` in examples probably).

I was planning to add this example: [https://developer.algorand.org/solutions/assets-and-custom-transfer-logic/](https://developer.algorand.org/solutions/assets-and-custom-transfer-logic/), but we require asset clawback tx in a group here. What do you think?

amityadav0: I think create should in transaction type as well, because group transaction create + pay.

packages/runtime/src/runtime.ts
ratik21: It's mentioned here: [https://developer.algorand.org/tutorials/asa-javascript/#part-6-revoke](https://developer.algorand.org/tutorials/asa-javascript/#part-6-revoke) and here: [https://developer.algorand.org/solutions/assets-and-custom-transfer-logic/](https://developer.algorand.org/solutions/assets-and-custom-transfer-logic/) as well.

amityadav0: ok

packages/runtime/src/runtime.ts
ratik21: Yes, possible. But please update the asset transactions here, this way we would be moving in the right direction, not the opposite.

amityadav0: agree for asset freeze, clawback etc... but for create and opt-in we should keep them in ExecParams too because that's the right way,
also if a user is making an application using algob, what if there is a transaction with create and pay or opt-in and pay
or pay and opt-in, for ex: you can only use our application if you pay some amount.
do you think we should avoid that ? @ratik21

ratik21: Currently we are avoiding it, but we shouldn't avoid that. Continued on discord.

packages/runtime/test/src/runtime.ts
```diff
@@ -286,4 +302,281 @@ describe("Algorand Standard Assets", function () {
...
+    };
+    expectTealError(
+      () => runtime.executeTx(modifyParam),
+      ERRORS.ASA.ASSET_NOT_FOUND
+    );
+  });
+
+  it("should modify asset", () => {
+    const assetId = runtime.createAsset('gold',
+      { creator: { ...john.account, name: "john" } });
```
ratik21, line 323: I think we are using this in each test. How about putting this in `beforeEach()` hook ?

packages/runtime/src/types.ts
```diff
@@ -211,6 +219,40 @@ export interface Sign {
...
+  freezeState: boolean
+  payFlags: TxParams
+}
+
+export interface RevokeAssetParam extends Sign {
+  type: TransactionType.RevokeAsset
+  fromAccount: AccountSDK
+  recipient: AccountAddress
+  assetID: number
+  revocationTarget: AccountAddress
```
ratik21, line 244: Please add a comment in `fromAccount`, `revocationTarget`. `fromAccount` is the clawback address, and the revocation target is the account from which the clawback revokes/retrieves the asset.

packages/runtime/src/account.ts
robert-zaremba: no, it's fine, I read it too fast.

packages/runtime/src/account.ts
robert-zaremba: So we are in the account context, maybe we should rename to `ERRORS.ACCOUNT.ASA_NOT_OPTIN` ?

robert-zaremba (approved):
approving - but check the new comment please.

packages/runtime/src/account.ts
amityadav0: yes

packages/runtime/test/src/runtime.ts
amityadav0: yes

packages/runtime/src/types.ts
amityadav0: ok

PR #226, opened 2021-02-08
Update algob test 

ratik21:
Issue Description/Summary:
- The test is creating a file (using `fsExtra.createFile(..))`, and setting it's permission to `0` (no permission for read, write & execute). Then later we're trying to read the file (using `fs.accessSync(..)`), which should expectedly throw an error (PERMISSION_DENIED), but doesn't on his system.

Updates to test:
- Used absolute file path to create and read from file (as previously we found his symlinks are broken)
- Updated test code from `async` to `sync`. (Using `fs.createFileSync` instead of `await fs.createFile`..)
- Updated no-permission code to octal (`0` -> `0o000`)

NOTE: not `100%` sure if this would fix it, as both ways of testing are working on my system. I am also using Ubuntu 20.04 LTS


amityadav0:
err

I was able to recreate this by using `sudo yarn run test`. tested this branch also.
the issue persists. this is probably because of using `sudo` it overrides the permission for the file and read it any way.

ratik21:
> I was able to recreate this by using `sudo yarn run test`. tested this branch also.
> the issue persists. this is probably because of using `sudo` it overrides the permission for the file and read it any way.

When i use `sudo` 9 errors fail on my system (Probably on yours too). On his system only 1 error is failing. Maybe ask him to run without using `sudo` first?

packages/algob/test/internal/core/params/argument-types.ts
```diff
@@ -232,19 +232,21 @@ describe("argumentTypes", () => {
...
+    it("Should throw if the file isn't readable", function () {
+      var isRoot = process.getuid && process.getuid() === 0;
+      if (os.type() === "Windows_NT" || isRoot) {
         this.skip();
       }
-      await fsExtra.createFile("A");
-      await fsExtra.chmod("A", 0);
+
+      fsExtra.createFileSync("A");
+      fsExtra.chmodSync("A", 0o000); // assign no permission [read-execute-write: 000]
```
amityadav0, line 242: what is the benefit of using 0o000 instead of 0 ?

amityadav0: also why are we removing await. is it possible that may result in error in some systems ?

amityadav0:
can we add a check in those 8 tests `if pyteal doesn't exist in sudo env` error explaining `pyteal doesn't exist in sudo env` will be thrown ?

packages/algob/test/internal/core/params/argument-types.ts
robert-zaremba: `0o000` is more readable because you show all flags. - each digit is rights for user, group and anyone else receptively.

packages/algob/test/internal/core/params/argument-types.ts
```diff
@@ -232,19 +232,21 @@ describe("argumentTypes", () => {
       );
     });
 
-    it("Should throw if the file isn't readable", async function () {
-      if (os.type() === "Windows_NT") {
+    it("Should throw if the file isn't readable", function () {
+      var isRoot = process.getuid && process.getuid() === 0;
+      if (os.type() === "Windows_NT" || isRoot) {
         this.skip();
```
robert-zaremba, line 238: Please add a warning log here.

robert-zaremba (approved):
Pre-approving. Please add a warning log.

packages/algob/test/internal/core/params/argument-types.ts
robert-zaremba: `console.warn(...)`

ratik21: Adding

ratik21:
> can we add a check in those 8 tests `if pyteal doesn't exist in sudo env` error explaining `pyteal doesn't exist in sudo env` will be thrown ?

Yes. Added a warning in `Compile` tests. If user is running as root, then pyteal should exist in sudo env. Otherwise tests will fail.

packages/algob/test/builtin-tasks/compile.ts
```diff
@@ -58,6 +59,15 @@ describe("Compile task", () => {
   const f4 = "gold-asa.teal";
   const fhash = 2374470440; // murmur3 hash for f1 file
 
+  before(function () {
+    var isRoot = process.getuid && process.getuid() === 0;
+    if (isRoot) {
+      const warnMsg = chalk.yellowBright(`*** If running tests as root, make sure to have pyteal exist in sudo env (or install pyteal as root).
```
amityadav0, line 65: how about renaming pyteal exist to pyteal installed ?

ratik21: ok

PR #227, opened 2021-02-09
Update branch ops

ratik21:


packages/runtime/src/interpreter/interpreter.ts
```diff
@@ -134,6 +134,15 @@ export class Interpreter {
...
       const instruction = this.instructions[this.instructionIndex];
       if (instruction instanceof Label && instruction.label === label) {
+        // if next immediate op is also label, then keep continuing, otherwise return
+        while (true) {
+          const nextInstruction = this.instructions[this.instructionIndex + 1];
+          if (nextInstruction instanceof Label) {
+            this.instructionIndex++;
+            continue;
+          }
+          break;
```
robert-zaremba, line 144:
```suggestion
        for (; this.instructionIndex < this.instructions.length-1; ++this.instructionIndex) {
          const nextInstruction = this.instructions[this.instructionIndex + 1];
          if (!(nextInstruction instanceof Label))
            break;
```

robert-zaremba (approved):
pre approving. In the inner loop we also need to check the boundaries of the index.

packages/runtime/src/interpreter/interpreter.ts
ratik21: Thanks for suggestion. Previously for `undefined` case (index out of bound for nextInstruction), `(nextInstruction instanceof Label)` was handling that. 
But this implementation seems more intuitive. Updating

PR #228, opened 2021-02-09
Bug fix

amityadav0:
Bug: Cannot set manager, freeze, reserve or clawback fields to blank in `algob` package.

Following the below discussion we can set the blank address by passing them as `undefined`. 

https://discord.com/channels/491256308461207573/631209194967531559/808681496498929724

Fix Bug in algob package and update tesla example


packages/algob/src/lib/tx.ts
```diff
@@ -53,10 +53,10 @@ export function makeAssetCreateTxn (
...
     asaDef.decimals,
     asaDef.defaultFrozen,
-    asaDef.manager,
-    asaDef.reserve,
-    asaDef.freeze,
-    asaDef.clawback,
+    asaDef.manager !== "" ? asaDef.manager : undefined,
+    asaDef.reserve !== "" ? asaDef.reserve : undefined,
+    asaDef.freeze !== "" ? asaDef.freeze : undefined,
+    asaDef.clawback !== "" ? asaDef.clawback : undefined,
```
ratik21, line 59: Can we do this during parsing of `asaDefs` (i.e when we read `" "` from YAML, we set it to undefined). Otherwise we would need to do repeat this each time for an asset related tx (eg. `AssetConfig`)

amityadav0: yes

ratik21: thanks

ratik21 (approved):

PR #230, opened 2021-02-10
Restructure errors

amityadav0:
- restructure error-list
- remove duplicate errors
- classifiy errors in both(runtime, algob, teal)


packages/runtime/src/errors/errors-list.ts
```diff
@@ -8,291 +9,297 @@ export interface ErrorDescriptor {
...
-export const ERROR_RANGES = {
-  TEAL: { min: 0, max: 99, title: "TEAL opcode errors" },
-  TRANSACTION: { min: 100, max: 199, title: "Transaction error" },
-  ASA: { min: 900, max: 999, title: "ASA File Parse Error" }
+export function getRuntimeErrorCode (error: ErrorDescriptor): string {
+  return `${RUNTIME_ERROR_PREFIX}${error.number}`;
+}
+
+export const TEAL_ERROR_RANGES = {
+  TEAL: { min: 1000, max: 1099, title: "TEAL opcode errors" }
```
amityadav0, line 21: What do you think about renaming this error to `EXE` title: "TEAL execution error" ?

ratik21: Agree with title change. Did you mean to rename the name as well (`TEAL` to `EXE`)?

packages/runtime/src/errors/errors-list.ts
```diff
@@ -8,297 +9,309 @@ export interface ErrorDescriptor {
...
+};
+
+export const RUNTIME_ERROR_RANGES = {
+  GENERAL: { min: 1300, max: 1399, title: "Runtime General Error" },
+  TRANSACTION: { min: 1400, max: 1499, title: "Transaction error" },
+  ASA: { min: 1500, max: 1599, title: "ASA Error" }
```
ratik21, line 27: Shouldn't `TEAL` error be also part of `RUNTIME_ERRORS` ? `Runtime` is a superset which does TEAL execution while processing transactions

packages/runtime/src/errors/errors-list.ts
```diff
@@ -8,297 +9,309 @@ export interface ErrorDescriptor {
...
+    number: 1004,
     message: "Result of current operation caused integer overflow at %line%",
     title: "Uint64 Overflow at %line%",
     description: `You are tying to perform operation where the result has exceeded
 maximun uint64 value of 18446744073709551615`
   },
   UINT64_UNDERFLOW: {
-    number: 6,
+    number: 1005,
     message: "Result of current operation caused integer underflow at %line%",
```
ratik21, line 69:
```suggestion
    message: "Result of current operation caused integer underflow at line %line%",
```

ratik21: Please update other messages as well

packages/runtime/src/errors/errors-list.ts
```diff
@@ -316,55 +329,63 @@ const asaErrors = {
...
+export const RUNTIME_ERRORS: {
+  [category in keyof typeof RUNTIME_ERROR_RANGES]: {
     [errorName: string]: ErrorDescriptor
   };
 } = {
-  TEAL: tealErrors,
+  GENERAL: runtimeGeneralErrors,
   TRANSACTION: transactionErrors,
-  ASA: asaErrors
+  ASA: runtimeAsaErrors
```
ratik21, line 390: Same as above. Is there a reason to keep TEAL errors separate?

packages/runtime/src/interpreter/interpreter.ts
```diff
@@ -137,7 +138,7 @@ export class Interpreter {
         return;
       }
     }
-    throw new TealError(ERRORS.TEAL.LABEL_NOT_FOUND, {
+    throw new TealError(TEAL_ERRORS.TEAL.LABEL_NOT_FOUND, {
```
ratik21, line 141: `TEAL_ERRORS.TEAL.LABEL_NOT_FOUND` has `TEAL` two times. 
Is it possible to have only Runtime Errors (then we won't need `TEAL_ERRORS`, `RUNTIME_ERRORS`), and we could just use `ERRORS.TEAL.LABEL_NOT_FOUND`, `ERRORS_TRANSACTION.TYPE_Err..`).
Basically i am asking if we could use runtime error class only and move teal errors into runtime errors

packages/runtime/src/errors/errors-list.ts
amityadav0: Yes good idea, so as you suggest removing teal errors, and making it as runtime errors with teal error as subset ?
What do you think ?

packages/runtime/src/errors/errors-list.ts
amityadav0: Just that in opcode list we are using teal opcode errors, and in runtime and account we throw runtime errors

packages/runtime/src/errors/errors-list.ts
amityadav0: But that will result in all errors in one class, adding more errors will be confusing, and if we want to import them in algob that will be not good also

packages/runtime/src/errors/errors-list.ts
amityadav0: Ok

packages/runtime/src/interpreter/interpreter.ts
amityadav0: Yes I was also thinking about EXEC, 'teal execution error' 
Renaming to TEAL_ERRORS.EXEC.

packages/runtime/src/errors/errors-list.ts
amityadav0: good idea, changing TEAL to  EXE

packages/runtime/src/errors/errors-list.ts
amityadav0: adding it as subset so there will be different classes for both

robert-zaremba (approved):

PR #231, opened 2021-02-10
TS example doesn't work properly

amityadav0:
https://github.com/scale-it/algorand-builder/issues/229

### DIFF
--- a/examples/htlc-pyteal-ts/.gitignore
+++ b/examples/htlc-pyteal-ts/.gitignore
@@ -1,2 +1,11 @@
 node_modules/
 build/
+scripts/deploy.d.ts
+scripts/deploy.js
+scripts/deploy.js.map
+scripts/withdraw/common.d.ts
+scripts/withdraw/common.js
+scripts/withdraw/common.js.map
+scripts/withdraw/htlc-withdraw.d.ts
+scripts/withdraw/htlc-withdraw.js
+scripts/withdraw/htlc-withdraw.js.map
--- a/examples/htlc-pyteal-ts/.gitignore
+++ b/examples/htlc-pyteal-ts/.gitignore
@@ -1,11 +1,2 @@
 node_modules/
 build/
-scripts/deploy.d.ts
-scripts/deploy.js
-scripts/deploy.js.map
-scripts/withdraw/common.d.ts
-scripts/withdraw/common.js
-scripts/withdraw/common.js.map
-scripts/withdraw/htlc-withdraw.d.ts
-scripts/withdraw/htlc-withdraw.js
-scripts/withdraw/htlc-withdraw.js.map
--- a/examples/htlc-pyteal-ts/tsconfig.json
+++ b/examples/htlc-pyteal-ts/tsconfig.json
@@ -1,7 +1,7 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
-    "outDir": "./scripts"
+    "outDir": "./build/scripts"
   },
   "include": [
     "scripts",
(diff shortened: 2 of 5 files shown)


examples/htlc-pyteal-ts/.gitignore
```diff
@@ -1,2 +1,11 @@
...
+build/
+scripts/deploy.d.ts
+scripts/deploy.js
+scripts/deploy.js.map
+scripts/withdraw/common.d.ts
+scripts/withdraw/common.js
+scripts/withdraw/common.js.map
+scripts/withdraw/htlc-withdraw.d.ts
+scripts/withdraw/htlc-withdraw.js
+scripts/withdraw/htlc-withdraw.js.map
```
ratik21, line 11: Can we do 
```
scripts/*.js // should cover .js.map as well
scripts/*.d.ts
```

examples/htlc-pyteal-ts/tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
-    "outDir": "./build/scripts"
+    "outDir": "./scripts"
```
ratik21, line 4: This would be a workaround to get the example working. But i am not sure if we should mix `/build` files in root. 

As i think it will complicate the current structure (have multiple files in root). Also, if the user decides to add one more file (say `scripts/1-deploy.ts`) then, for each file the `build` will generate 3 extra files (`.js, .d.ts, .js.map`). And if the above suggestion doesn't work, user will also need to manually edit `.gitignore` as well. So i don't think this would be a good user & developer experience.

Did you check why the deployer is not loading the file from `/build/scripts`? Particularly this line: [https://github.com/scale-it/algorand-builder/blob/master/packages/algob/src/builtin-tasks/deploy.ts#L35](https://github.com/scale-it/algorand-builder/blob/master/packages/algob/src/builtin-tasks/deploy.ts#L35). Interestingly, when i was trying to run the example, i tried to add a `console.log()` after `executeDeployTask` was called (to check what's happing). But it didn't show the log when i ran `yarn run algob deploy` in `/examples`. I don't know why, but i am unable to debug because of this. Can you check on your system as well?

ratik21:
Update: It's working now on my system and was able to view console logs in as well (master and not this one).
image

I don't think logic/code is incorrect here (to load compiled ts files from `/build`), issue is something else. Previously in the day (after fresh install) it wasn't working with me. Some major things i did:
+ i reinstalled `yarn`: `yarn global bin` was not correctly set on my system, and `$PATH` wasn't added in `~/.bashrc` as well.
+ installed `algob` as global binary (using `yarn global add @algorand-builder/algob`)

Also, this could be helpful: when i ran the examples using the latest published binary ( `algob v0.4.0`), it doesn't work. But on doing `yarn run algob deploy` it works. (Attaching screenshot as well)
image
If we check the later message its getting the path correctly (in `/build`): `Checkpoint exists for script build/scripts/deploy.js`

examples/htlc-pyteal-ts/tsconfig.json
amityadav0: working example would be a good user experience, what do you suggest to resolve this issue  ?

examples/htlc-pyteal-ts/.gitignore
amityadav0: yes

examples/htlc-pyteal-ts/tsconfig.json
amityadav0: nvm

robert-zaremba (approved):

Issue #224, opened 2021-02-08
yarn workspaces run test fails at @algorand-builder/algob


Lumene98:
Only one test fails:

```
1) argumentTypes
       Input file type
         Should throw if the file isn't readable:
     AssertionError: BuilderError number 302 expected, but no Error was thrown
      at Object.expectBuilderError (test/helpers/errors.ts:65:9)
      at Context.<anonymous> (test/internal/core/params/argument-types.ts:242:7)
```

ratik21:
@Lumene98 could you please share your OS?

Lumene98:
> 
> 
> @Lumene98 could you please share your OS?

Ubuntu 20.04 LTS

ratik21:
@Lumene98 could you try running the tests now (after taking `git pull`) ?

Lumene98:
> 
> 
> @Lumene98 could you try running the tests now (after taking `git pull`) ?

Same result after last change

PR #234, opened 2021-02-12
update test and exports

ratik21:
#224

### DIFF
--- a/packages/algob/test/internal/core/params/argument-types.ts
+++ b/packages/algob/test/internal/core/params/argument-types.ts
@@ -243,9 +243,14 @@ describe("argumentTypes", () => {
       fsExtra.createFileSync("A");
       fsExtra.chmodSync("A", 0o000); // assign no permission [read-execute-write: 000]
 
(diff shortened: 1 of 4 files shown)


packages/algob/test/internal/core/params/argument-types.ts
```diff
@@ -243,6 +243,11 @@ describe("argumentTypes", () => {
       fsExtra.createFileSync("A");
       fsExtra.chmodSync("A", 0o000); // assign no permission [read-execute-write: 000]
 
+      // check if permission set or not
```
robert-zaremba, line 246: Please update the comment to check why we are skipping the test. Also add  a warning that some tests are skipped with a reason why.

robert-zaremba (approved):
Please update the comment and add a warning.

packages/algob/test/internal/core/params/argument-types.ts
ratik21: ok

PR #236, opened 2021-02-15
update AppLocalGetEx and AppGlobalGetEx opcodes

### LINKED ISSUE #235: AppLocalGetEx and AppGlobalGetEx return values
In [packages/runtime/src/interpreter/opcode-list.ts](https://github.com/scale-it/algorand-builder/blob/master/packages/runtime/src/interpreter/opcode-list.ts), 
`AppLocalGetEx` and `AppGlobalGetEx` ops execution results in pushing `BIGINT0` to the stack in case the key doesn't have a corresponding value in app's state.

The actual behaviour should be pushing `BIGINT0` twice: one for the did_exist flag and one for the value.
This could result in an unwanted behaviour since the stack length is not as expected.

I've tested the above scenario with a simple TEAL script before posting this issue just to make sure but please validate as well.

ratik21:
closes #235

### DIFF
--- a/packages/runtime/src/interpreter/opcode-list.ts
+++ b/packages/runtime/src/interpreter/opcode-list.ts
@@ -1594,8 +1594,7 @@ export class AppLocalGet extends Op {
 }
 
 // read from application local state at Txn.Accounts[A] => app B => key C from local state.
-// Pushes to the stack [...stack, val, 1] if the key exists,
-// otherwise [...stack, 0]
+// push to stack [...stack, value, 1] (Note: value is 0 if key does not exist)
 export class AppLocalGetEx extends Op {
   readonly interpreter: Interpreter;
   readonly line: number;
@@ -1666,8 +1665,7 @@ export class AppGlobalGet extends Op {
 }
 
 // read from application Txn.ForeignApps[A] global state key B pushes to the stack
-// push to stack [...stack, 0] if key doesn't exist
-// otherwise push to stack [...stack, value, 1]
+// push to stack [...stack, value, 1] (Note: value is 0 if key does not exist)
 // A is specified as an account index in the ForeignApps field of the ApplicationCall transaction,
 // zero index means this app
 export class AppGlobalGetEx extends Op {
--- a/packages/runtime/src/interpreter/opcode-list.ts
+++ b/packages/runtime/src/interpreter/opcode-list.ts
@@ -1626,6 +1626,7 @@ export class AppLocalGetEx extends Op {
       stack.push(BIGINT1);
     } else {
       stack.push(BIGINT0); // The value is zero if the key does not exist.
+      stack.push(BIGINT0); // did_exist_flag
     }
   }
 }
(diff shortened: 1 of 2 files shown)


packages/runtime/src/interpreter/opcode-list.ts
```diff
@@ -1626,6 +1626,7 @@ export class AppLocalGetEx extends Op {
       stack.push(BIGINT1);
     } else {
       stack.push(BIGINT0); // The value is zero if the key does not exist.
+      stack.push(BIGINT0); // did_exist_flag
```
robert-zaremba, line 1629: Why adding another zero? According to spec, we should push only one `0`.

packages/runtime/test/src/interpreter/opcode-list.ts
```diff
@@ -2306,8 +2306,10 @@ describe("Teal Opcodes", function () {
         let op = new AppLocalGetEx([], 1, interpreter);
         op.execute(stack);
 
-        let flag = stack.pop();
-        assert.equal(BigInt('0'), flag);
+        let didExistFlag = stack.pop();
+        let val = stack.pop();
+        assert.equal(BigInt('0'), didExistFlag);
+        assert.equal(BigInt('0'), val);
```
robert-zaremba, line 2312: you can write `0n` instead of `BigInt('0')`. Also, for small numbers, we don't need to pass string into `BigInt`

packages/runtime/src/interpreter/opcode-list.ts
ratik21: Yes, i thought the same, that's why we were previously pushing only one `0`. But i confirmed this using the `teal-debugger`, it pushes two `0's` to stack (if value is not found)

packages/runtime/test/src/interpreter/opcode-list.ts
ratik21: Yes, actually we are updating all `BigInt('X')` => `Xn` in [this](https://github.com/scale-it/algorand-builder/pull/184) PR.
Will update this one for now.

packages/runtime/src/interpreter/opcode-list.ts
robert-zaremba: Maybe there is a bug in the debugger. Let's validate on Discord with Algorand Team. Clearly there is something wrong: either in debugger or in docs.

robert-zaremba: I was reading the https://developer.algorand.org/docs/reference/teal/opcodes/#app_local_get_ex
Yes, it will put two values.

We also need to update the doc string here.

robert-zaremba (approved):
pre-approving. Please update the class/function doc-strings as well

packages/runtime/src/interpreter/opcode-list.ts
ratik21: Alright, updating doc string.

PR #237, opened 2021-02-16
Add tests in /examples/restricted-assets

ratik21:


examples/restricted-assets/assets/poi-approval.teal
```diff
@@ -226,8 +228,8 @@ bz failed
 // verify second transaction
 // tx 1 - clawback transactions that moves the frozen asset from sender to receiver - signed by clawback-escrow
 // verify the account sent in the accounts array is
-// Note: the receiver of the asset is asset xfer
-gtxn 0 Accounts 1
+// actually the receiver of the asset in asset xfer
```
amityadav0, line 231:
```suggestion
// Note: the receiver of the asset is asset xfer
```

examples/restricted-assets/test/asset-txfer-test.js
```diff
@@ -0,0 +1,357 @@
...
+    creationFlags = {
+      sender: alice.account,
+      localInts: 1,
+      localBytes: 0,
+      globalInts: 2,
+      globalBytes: 1
+    };
+  });
+
+  this.afterEach(async function () {
```
amityadav0, line 38: how about renaming this to `beforeEach` ? and remove `beforeAll`.

examples/restricted-assets/test/asset-txfer-test.js
```diff
@@ -0,0 +1,357 @@
...
+    let assetDef = runtime.getAssetDef(assetId);
+
+    assert.equal(assetDef.creator, alice.address);
+    assert.equal(assetDef['default-frozen'], true);
+    assert.equal(assetDef.total, 1000000);
+    assert.equal(assetDef['unit-name'], 'GLD');
+    assert.equal(assetDef.manager, alice.address);
+    assert.equal(assetDef.clawback, alice.address);
+
+    runtime.optIntoASA(assetId, bob.address, {});
```
amityadav0, line 87: shouldn't we add opt-in for alice as well ?

packages/runtime/src/runtime.ts
```diff
@@ -355,7 +357,7 @@ export class Runtime {
       amount: address === creatorAddr ? assetDef.total : 0, // for creator opt-in amount is total assets
       'asset-id': assetIndex,
       creator: creatorAddr,
-      'is-frozen': assetDef["default-frozen"]
+      'is-frozen': address === creatorAddr ? false : assetDef["default-frozen"]
```
amityadav0, line 360: can you please elaborate why are we doing this ?

packages/runtime/test/src/interpreter/opcode-list.ts
```diff
@@ -1979,7 +1979,7 @@ describe("Teal Opcodes", function () {
         op.execute(stack);
 
         assert.equal(1, stack.length());
-        assert.deepEqual(TXN_OBJ.apat[1], stack.pop());
+        assert.deepEqual(TXN_OBJ.apat[0], stack.pop());
```
amityadav0, line 1982: why are doing this ?
i think `const op = new Txna(["Accounts", "1"], 1, interpreter);` means second account.

packages/runtime/src/runtime.ts
```diff
@@ -335,6 +335,8 @@ export class Runtime {
     const asset = senderAcc.addAsset(++this.assetCounter, name, this.loadedAssetsDefs[name]);
     this.mkAssetCreateTx(name, flags, asset);
     this.store.assetDefs.set(this.assetCounter, sender.addr);
+
+    this.optIntoASA(this.assetCounter, sender.addr, {}); // opt-in for creator
```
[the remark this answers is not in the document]
ratik21: yes, confirmed.

packages/runtime/src/runtime.ts
ratik21: Because for creator even if the asset is default frozen, for his holding the frozen is `false`. Confirmed by testing.
```
Note: A frozen account can always close out to the asset creator
```
Reference: [https://developer.algorand.org/tutorials/asa-javascript/](https://developer.algorand.org/tutorials/asa-javascript/)

examples/restricted-assets/test/asset-txfer-test.js
```diff
@@ -0,0 +1,357 @@
...
+    /* Setup Escrow Account */
+    const escrowProg = getEscrowProg(assetId, applicationId);
+    const escrowLsig = runtime.getLogicSig(escrowProg, []);
+    const escrowAddress = escrowLsig.address();
+
+    // sync escrow account
+    escrow = runtime.getAccount(escrowAddress);
+    console.log('Escrow Address: ', escrowAddress);
+
+    // fund escrow with some minimum balance first
```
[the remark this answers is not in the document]
ratik21: To pay tx fees. Initially it has 0 balance.

examples/restricted-assets/test/asset-txfer-test.js
```diff
@@ -0,0 +1,357 @@
...
+
+    applicationId = runtime.addApp(
+      { ...creationFlags, appArgs: creationArgs }, {}, approvalProgram, clearProgram);
+    const app = alice.getApp(applicationId);
+    assert.isDefined(app);
+
+    // opt in to app + verify optin
+    runtime.optInToApp(alice.address, applicationId, {}, {});
+    runtime.optInToApp(bob.address, applicationId, {}, {});
+
```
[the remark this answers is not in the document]
ratik21: ok

packages/runtime/test/src/interpreter/opcode-list.ts
ratik21: I though so too. But `Txna Account 0` means sender's address, and the above means first account. 
Tested (on blockchain) using: `gtxn 0 Accounts 1` (this pushed the first account from the account array). Something seems wrong. Will confirm on discord as well.

ratik21: Update: Confimed `const op = new Txna(["Accounts", "1"], 1, interpreter);` means first account.

Will do this in a new PR first, with more tests. We also need to update `Txn`, `Gtxn` as well.

amityadav0: i think we already have tests but they are written according to code.
or you can update on this PR as well, In interpreter file we already have `getAccount` function.
update lines with this.getAccount(1,line).your call

ratik21: Yes, but we need to add "more" tests with the correct behaviour, since we previously assumed wrong one, according to which the tests were written (your initial comment). 
Also found out that `Txn`, `Gtxn` needs to be updated as well because they also support array arguments. (eg. `txn Accounts 0`, `gtxn 0 Accounts 0` also works), so it's better to handle these fixes along with this fix in a new PR.

amityadav0: like i said, your call

examples/restricted-assets/assets/poi-approval.teal
```diff
@@ -227,7 +229,7 @@ bz failed
 // tx 1 - clawback transactions that moves the frozen asset from sender to receiver - signed by clawback-escrow
 // verify the account sent in the accounts array is
 // Note: the receiver of the asset is asset xfer
-gtxn 0 Accounts 1
+gtxna 0 Accounts 1
```
amityadav0, line 232: we have support for array in gtxna. right ?

amityadav0: or we will update it after fix gets merged ?

ratik21: Yes, we already had support for array in `gtxna`. After fix we will have support for array with `txn`, `gtxn` as well. 
Updated TEAL code here because i think `gtxna 0 Accounts 1` is more intuitive than `gtxn 0 Accounts 1`. Both will work after the fix. Let me know what you think we should keep here

examples/restricted-assets/test/asset-txfer-test.js
```diff
@@ -0,0 +1,332 @@
...
+    assert.deepEqual(getGlobal('Creator'), alicePk);
+    assert.deepEqual(getGlobal('AssetID'), BigInt(assetId));
+    assert.deepEqual(getGlobal('AssetLevel'), 2n); // we are setting level 2 in examples
+  });
+
+  const getGlobal = (key) => runtime.getGlobalState(applicationId, key);
+  const getEscrowProg = (assetId, appId) =>
+    getProgram('clawback-escrow.py', { ASSET_ID: assetId, APP_ID: appId });
+
+  // fetch latest account state
```
amityadav0, line 80:
```suggestion
  // Update account state
```

examples/restricted-assets/test/asset-txfer-test.js
```diff
@@ -0,0 +1,332 @@
...
+    const bobAssetHolding = runtime.getAssetHolding(assetId, bobAddr);
+    assert.isDefined(aliceAssetHolding);
+    assert.isDefined(bobAssetHolding);
+
+    /* Create application + optIn to app */
+    const creationArgs = [
+      `int:${assetId}`,
+      'int:2' // set min user level(2) for asset transfer ("Accred-level")
+    ];
+
```
amityadav0, line 62:
```suggestion
```

examples/restricted-assets/test/asset-txfer-test.js
```diff
@@ -0,0 +1,332 @@
...
+    };
+    runtime.executeTx(assetConfigParams);
+
+    // verify clawback is updated & manager, freeze address is set to ""
+    assetDef = runtime.getAssetDef(assetId);
+    assert.equal(assetDef.clawback, escrowAddress);
+    assert.equal(assetDef.manager, '');
+    assert.equal(assetDef.freeze, '');
+
+    /* Updating clawback again should throw error now (since we locked the asset) */
```
amityadav0, line 153:
```suggestion
    /* Updating clawback again should throw error now (since we set the asset manager to ``) */
```

examples/restricted-assets/test/asset-txfer-test.js
```diff
@@ -0,0 +1,332 @@
...
+
+    runtime.executeTx(setLevelParams); // set level for alice
+    runtime.executeTx({ ...setLevelParams, accounts: [bob.address] }); // set level for bob
+    syncAccounts();
+
+    // verify level is set in local-state
+    assert.equal(alice.getLocalState(applicationId, ACCRED_LEVEL), 2n);
+    assert.equal(bob.getLocalState(applicationId, ACCRED_LEVEL), 2n);
+
+    /* Transfer 1000 assets from Alice to Bob */
```
amityadav0, line 189: Please add 1000 assets revoked from clawback

examples/restricted-assets/test/asset-txfer-test.js
```diff
@@ -0,0 +1,332 @@
...
+    /**
+     * This test demonstrates how to transfer assets from account A to B using custom logic
+     * based on a smart contract. Asset is actually transferred by the clawback address (an escrow
+     * account in this case). Following operations are performed
+     * - Create the asset + optIn
+     * - Create the application + optIn
+     * - Setup Escrow Account
+     * - Update Asset clawback + lock manager and freeze address
+     * - Set Accred-Level
+     * - Transfer Asset from Alice -> Bob
```
amityadav0, line 97: Please modify this line to revoke assets

packages/runtime/src/runtime.ts
```diff
@@ -814,9 +816,10 @@ export class Runtime {
 
     const txnParameters = Array.isArray(txnParams) ? txnParams : [txnParams];
     // Run TEAL program associated with each transaction without interacting with store.
-    for (const txnParam of txnParameters) {
+    for (const [index, txnParam] of txnParameters.entries()) {
       this.assertAmbiguousTxnParams(txnParam);
       if (txnParam.sign === SignType.LogicSignature) {
+        this.ctx.tx = this.ctx.gtxs[index]; // update current tx
```
amityadav0, line 822: can you please explain this ?

amityadav0: should we set this.ctx.tx back to previous after we are done with lsig ?

examples/restricted-assets/test/asset-txfer-test.js
ratik21: Adding

examples/restricted-assets/test/asset-txfer-test.js
ratik21: Will add revoking instead of modifying. From the top level it's just a transfer of asset. Revoking via clawback escrow is internal logic

amityadav0: ok

packages/runtime/src/runtime.ts
ratik21: > can you please explain this ?

We need to set the current transaction (for correct `GroupIndex`) before executing stateless TEAL

> should we set this.ctx.tx back to previous after we are done with lsig ?

Not sure if setting previous config the right way. What if the first tx is the stateless one?

amityadav0: for the ssc `tx` is first transaction in `gtxs`, right ?

ratik21: > should we set this.ctx.tx back to previous after we are done with lsig ?

Doing this.

amityadav0 (approved):
Please verify that tests are working after updating builds.

PR #239, opened 2021-02-16
Optimize Package

amityadav0:


packages/algob/package.json
```diff
@@ -51,6 +51,14 @@
     "publish": "npm publish --access public",
     "algob:dev": "sh ./setup_dev_project.sh"
   },
+  "files": [
+    "src/",
```
robert-zaremba, line 55: do we need `src`? `build` already has `.d.ts` files.

amityadav0: i don't think so. 
let me know if you want to exclude it from package ?

amityadav0: removing src as well

robert-zaremba (approved):

PR #240, opened 2021-02-16
Update rounds to use timestamps with SSC

amityadav0:
https://discord.com/channels/754286650540032030/754286650540032033/811328347522203738

### DIFF
--- a/packages/runtime/test/src/interpreter/opcode-list.ts
+++ b/packages/runtime/test/src/interpreter/opcode-list.ts
@@ -2126,7 +2126,7 @@ describe("Teal Opcodes", function () {
     });
 
     it("should push Round to stack", function () {
-      interpreter.runtime.setBlockGlobalData(500, 1);
+      interpreter.runtime.setRoundAndTimestamp(500, 1);
       const op = new Global(['Round'], 1, interpreter);
       op.execute(stack);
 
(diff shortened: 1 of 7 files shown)


packages/runtime/test/src/interpreter/opcode-list.ts
```diff
@@ -2134,14 +2135,12 @@ describe("Teal Opcodes", function () {
...
+      interpreter.runtime.setBlockGlobalData(500, 100);
       const op = new Global(['LatestTimestamp'], 1, interpreter);
       op.execute(stack);
 
       const top = stack.pop();
-      const ts = Math.round((new Date()).getTime() / 1000);
-
-      const diff = Math.abs(ts - Number(top)); // for accuracy
-      assert.isBelow(diff, 10);
+      assert.equal(BigInt('100'), top);
```
robert-zaremba, line 2143: we can use `100n` instead of `BigInt('100')`

robert-zaremba (approved):
preapproving - let's also consider `setRoundAndTimestamp` instead of `setBlockGlobalData`

packages/runtime/test/src/interpreter/opcode-list.ts
amityadav0: ok

amityadav0:
Updated to `setRoundAndTimestamp`

PR #241, opened 2021-02-17
More tests for crowdfunding - Happy Paths

ratik21:
Happy Paths: 
- Creating Application
- Setup Escrow + update app
- opt in
- donate funds to escrow
- claim if goal is met
- reclaim if goal is not met
- delete application (using single tx)
- delete application (using group tx)


amityadav0:
how is it different than test we have in crowdfunding.js ?

ratik21:
In `crowdfunding.js` we only have a one big unit test. That's why we needed more tests. The issue with one big unit test is it only tests the entire flow of the application, whereas we should be able to test each aspect/behaviour of the smart contract. 
eg. In the single big unit test, if there is an issue with app creation part of the smart contract (say `Creator` was misspelled to `Creatorr`), then the test will fail at that point only, and the user does not know regarding the correctness of rest of the TEAL code (he doesn't know if `donate` branch works or not, as the test was rejected before this!). So initially he'll correct the app creation part, and then run that test again, which can still fail (say on `donate` this time). And so on..

I think we needed to address this behaviour and make each step of the crowdfunding test independent from each other. This is what i feel is different here. We are only testing each "unit"/"step" of the application one by one, so now even if there is an issue with app creation in approvalProgram, we will still be able to test `donate/reclaim` as we are preparing the state now (without being dependent on the app creation part of smart contract).
 
Each step with the happy flow (points mentioned in description) is tested here. I didn't add mix cases because we already have the big unit test to test dependent flow. Please let me know if that makes sense.

examples/crowdfunding/test/happypaths.js
```diff
@@ -0,0 +1,396 @@
+import {
```
amityadav0, line 1: Please update this test to use timestamp, you can refer to failing-paths PR.

amityadav0:
Ok, got it, they are same but divided into more unit tests.

examples/crowdfunding/test/happypaths.js
ratik21: I am already using `runtime.setRoundAndTimeStamp` in these tests. Since user has the control, i am using simple numbers to represent `timestamp`. Or did you mean something else?
eg. `runtime.setRoundAndTimestamp(2, 5);` (setting current `ts` to 5, and startTs, endTs to `1, 10` respectively).
Will update if you think we should use actual timestamps instead.

amityadav0: you have uncomment the lines in teal files, so setting timestamp, rounds has no effect unless tested.

ratik21: ah, updating .teal, yes.

robert-zaremba:
What is blocking this PR to merge it?

amityadav0:
I think this is ready, there is a task for it, https://www.pivotaltracker.com/story/show/177054963

examples/crowdfunding/assets/crowdFundApproval.teal
```diff
@@ -83,11 +83,11 @@ txn Sender
 ==
 
 // check that we are past fund close date
-// global LatestTimestamp
-// byte "FundCloseDate"
-// app_global_get
-// >=
-// &&
+global LatestTimestamp
```
amityadav0, line 86: Please merge master(this will be updated)

ratik21: This was discussed on call, as a task @man-of-code will be reviewing this PR, so i haven't merged master yet (intentionally).

ratik21:
@robert-zaremba there is a [task](https://www.pivotaltracker.com/story/show/177054963) in pivotal tracker for @man-of-code to review this PR (for understanding the code).

man-of-code (approved):
Understood what I could. @amityadav0 is right,  the original test is broken into many unit tests which will help in recognizing any potential bugs more efficiently.

PR #242, opened 2021-02-18
Add more tests in crowdfunding example [failing-paths]

amityadav0:
Failing Scenarios:
- Donate before or after startDate, endDate respectively
- Donate with insufficient fund
- Claim, reclaim before expiry date
- Wrong lsig passed, lsig missing
- Transaction missing from group transaction

### DIFF
--- a/examples/crowdfunding/test/failing-tests.js
+++ b/examples/crowdfunding/test/failing-tests.js
@@ -12,7 +12,7 @@ const initialDonorBalance = minBalance + 60e6;
 const initialCreatorBalance = minBalance + 0.01e6;
 const goal = 7e6;
 
-describe('Failing Scenario', function () {
+describe('Crowdfunding Test - Failing Scenarios', function () {
   const master = new StoreAccount(1000e6);
   let creator = new StoreAccount(initialCreatorBalance);
   let escrow = new StoreAccount(minBalance);
--- a/examples/crowdfunding/test/failing-tests.js
+++ b/examples/crowdfunding/test/failing-tests.js
@@ -315,7 +315,7 @@ describe('Crowdfunding Test - Failing Scenarios', function () {
     assert.throws(() => runtime.executeTx(txGroup), 'RUNTIME_ERR1301: logic signature validation failed.');
   });
 
-  it('should fail if transaction is lsig is missing', () => {
+  it('should fail transaction if logic signature is not passed', () => {
     updateAndOptIn();
     runtime.executeTx(donateTxGroup);
     runtime.executeTx(donateTxGroup);
@@ -359,7 +359,7 @@ describe('Crowdfunding Test - Failing Scenarios', function () {
     assert.throws(() => runtime.executeTx(deleteTx), 'RUNTIME_ERR1008: Index out of bound');
   });
 
-  it('should fail if trying to update application where sender is not creator', () => {
+  it('should fail on trying to update application where sender is not creator', () => {
     appArgs = [addressToPk(escrowAddress)]; // converts algorand address to Uint8Array
 
     assert.throws(() =>
--- a/examples/crowdfunding/test/failing-tests.js
+++ b/examples/crowdfunding/test/failing-tests.js
@@ -402,4 +402,36 @@ describe('Crowdfunding Test - Failing Scenarios', function () {
 
     assert.throws(() => runtime.executeTx(txGroup), rejectMsg);
   });
+
+  // uncomment when
+  // https://github.com/algorand/js-algorand-sdk/commit/b18e3beab8004d7e53a5370334b8e9f5c7699146#diff-75520b02c557ab3f0b89e5f03029db31af2f0dc79e5215d3e221ed9ea59fe441
+  // commit is released
+  /* it('should fail if ReKeyTo is not zeroAddress in transaction', () => {
+    updateAndOptIn();
+    runtime.executeTx(donateTxGroup);
+    runtime.executeTx(donateTxGroup);
+    runtime.setRoundAndTimestamp(5, endDate.getTime() + 122);
+    appArgs = [stringToBytes('claim')];
+    const txGroup = [
+      {
+        type: types.TransactionType.CallNoOpSSC,
+        sign: types.SignType.SecretKey,
+        fromAccount: creator.account,
+        appId: applicationId,
+        payFlags: { totalFee: 1000, ReKeyTo: donor.address },
+        appArgs: appArgs
+      },
+      {
+        type: types.TransactionType.TransferAlgo,
+        sign: types.SignType.LogicSignature,
+        fromAccount: escrow.account,
+        toAccountAddr: creator.address,
+        amountMicroAlgos: 0,
+        lsig: lsig,
+        payFlags: { totalFee: 1000, closeRemainderTo: creator.address, ReKeyTo: donor.address }
+      }
+    ];
+
+    assert.throws(() => runtime.executeTx(txGroup), rejectMsg);
+  }); */
 });
--- a/examples/crowdfunding/test/failing-tests.js
+++ b/examples/crowdfunding/test/failing-tests.js
@@ -207,7 +207,7 @@ describe('Crowdfunding Test - Failing Scenarios', function () {
     assert.throws(() => runtime.executeTx(txGroup), rejectMsg);
   });
 
-  it('should fail if goal is not met, donor tries to reclaim funds before fund close date', () => {
+  it('should fail if goal is not met, but donor tries to reclaim funds before fund close date', () => {
     updateAndOptIn();
     runtime.executeTx(donateTxGroup);
 
@@ -266,7 +266,7 @@ describe('Crowdfunding Test - Failing Scenarios', function () {
     assert.throws(() => runtime.executeTx(txGroup), rejectMsg);
   });
 
-  it('should fail if a transaction is missing in group transaction', () => {
+  it('should fail if a transaction is missing in group transaction while donating', () => {
     updateAndOptIn();
     appArgs = [stringToBytes('donate')];
     // Atomic Transaction (Stateful Smart Contract call + Payment Transaction)
--- a/examples/crowdfunding/test/failing-tests.js
+++ b/examples/crowdfunding/test/failing-tests.js
@@ -1,0 +1,348 @@
+import {
+  addressToPk,
+  getProgram,
+  stringToBytes,
+  uint64ToBigEndian
+} from '@algorand-builder/algob';
+import { Runtime, StoreAccount, types } from '@algorand-builder/runtime';
+import { assert } from 'chai';
+
+const minBalance = 10e6; // 10 ALGO's
+const initialDonorBalance = minBalance + 60e6;
+const initialCreatorBalance = minBalance + 0.01e6;
+const goal = 7e6;
+
+describe('Failing Scenario', function () {
+  const master = new StoreAccount(1000e6);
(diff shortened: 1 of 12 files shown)


ratik21:
Please also add (ignore if already added):
+ trying to update application where sender is not creator (this is imp and should fail)
+ closing the funds in escrow but `closeRemainderTo !== fundReceiver`
+ any one tx where common property is incorrect (eg. `ReyTo !== ZeroAddress`)

amityadav0:
thanks adding these.

examples/crowdfunding/test/failing-tests.js
```diff
@@ -0,0 +1,348 @@
...
+} from '@algorand-builder/algob';
+import { Runtime, StoreAccount, types } from '@algorand-builder/runtime';
+import { assert } from 'chai';
+
+const minBalance = 10e6; // 10 ALGO's
+const initialDonorBalance = minBalance + 60e6;
+const initialCreatorBalance = minBalance + 0.01e6;
+const goal = 7e6;
+
+describe('Failing Scenario', function () {
```
ratik21, line 15:
```suggestion
describe('Crowdfunding Test - Failing Scenarios', function () {
```

examples/crowdfunding/test/failing-tests.js
```diff
@@ -0,0 +1,348 @@
...
+  const master = new StoreAccount(1000e6);
+  let creator = new StoreAccount(initialCreatorBalance);
+  let escrow = new StoreAccount(minBalance);
```
ratik21, line 18: Please don't do this, it's confusing. We are initializing it later (using `runtime.getLogicSig`), so it doesn't make sense to create a store account here

amityadav0: ok

examples/crowdfunding/test/failing-tests.js
```diff
@@ -0,0 +1,437 @@
...
+    // set timestamp to out of endDate
+    runtime.setRoundAndTimestamp(5, beginDate.getSeconds() - 100);
+
+    assert.throws(() => runtime.executeTx(donateTxGroup), rejectMsg);
+  });
+
+  it('should fail if goal is met, and donor tries to reclaim funds', () => {
+    updateAndOptIn();
+    runtime.executeTx(donateTxGroup);
+    runtime.executeTx(donateTxGroup);
```
ratik21, line 180: Why are we doing this twice here? Can you please add a comment?

examples/crowdfunding/test/failing-tests.js
```diff
@@ -0,0 +1,437 @@
...
+    runtime.setRoundAndTimestamp(5, endDate.getTime() + 100); // end date is passed
+
+    // donor should be able to reclaim if goal is NOT met
```
ratik21, line 183: Please update this comment, this is a failing scenario

examples/crowdfunding/test/failing-tests.js
ratik21: Also, i wanted to understand, why can't we just fund the escrow on our own in this case while preparing the state (with a simple `algotransfer` txn, and setting "Total" field in smart contract). Is it necessary to go through the `donate` tx? Currently in this test there is dependency on `donate:` branch, is it possible to avoid that?

examples/crowdfunding/test/failing-tests.js
```diff
@@ -0,0 +1,437 @@
...
+        amountMicroAlgos: 300000,
+        lsig: lsig,
+        payFlags: { totalFee: 1000 }
+      }
+    ];
+
+    assert.throws(() => runtime.executeTx(txGroup), rejectMsg);
+  });
+
+  it('should fail if goal is not met, donor tries to reclaim funds before fund close date', () => {
```
ratik21, line 210:
```suggestion
  it('should fail if goal is not met, but donor tries to reclaim funds before fund close date', () => {
```

examples/crowdfunding/test/failing-tests.js
```diff
@@ -0,0 +1,437 @@
...
+    updateAndOptIn();
+    runtime.executeTx(donateTxGroup);
+
+    // donor should be able to reclaim if goal is NOT met
```
ratik21, line 214: Please update this comment

examples/crowdfunding/test/failing-tests.js
amityadav0: we are doing this to meet the goal of the donation.

we cannot do that because contract has variables which needs to be updated. we are using this donate: branch because we cannot test claim if we haven't donated to escrow.
changing this will be like updating app without even creating it. dependency will be there to properly test claim part.

examples/crowdfunding/test/failing-tests.js
```diff
@@ -0,0 +1,437 @@
...
+      }
+    ];
+
+    assert.throws(() => runtime.executeTx(txGroup), rejectMsg);
+  });
+
+  it('should fail if creator tries to claim funds before fund end date', () => {
+    updateAndOptIn();
+    runtime.executeTx(donateTxGroup);
+    runtime.executeTx(donateTxGroup);
```
ratik21, line 244: Same as above. If we won't be removing donate then please add a comment. Also, if we are doing this twice in multiple tests, not sure but it would probably affect the speed of tests as well.

examples/crowdfunding/test/failing-tests.js
amityadav0: yes

examples/crowdfunding/test/failing-tests.js
```diff
@@ -0,0 +1,437 @@
...
+      }
+    ];
+
+    assert.throws(() => runtime.executeTx(txGroup), rejectMsg);
+  });
+
+  it('should fail if a transaction is missing in group transaction', () => {
+    updateAndOptIn();
+    appArgs = [stringToBytes('donate')];
+    // Atomic Transaction (Stateful Smart Contract call + Payment Transaction)
```
ratik21, line 272: Please remove this comment

examples/crowdfunding/test/failing-tests.js
```diff
@@ -0,0 +1,437 @@
...
+        amountMicroAlgos: 0,
+        lsig: lsig,
+        payFlags: { totalFee: 1000, closeRemainderTo: creator.address }
+      }
+    ];
+
+    assert.throws(() => runtime.executeTx(txGroup), rejectMsg);
+  });
+
+  it('should fail if a transaction is missing in group transaction', () => {
```
ratik21, line 269:
```suggestion
  it('should fail if a transaction is missing in group transaction while donating', () => {
```

examples/crowdfunding/test/failing-tests.js
amityadav0: we can do one more thing increase donation amount to greater than goal.

examples/crowdfunding/test/failing-tests.js
amityadav0: why ?

examples/crowdfunding/test/failing-tests.js
```diff
@@ -0,0 +1,437 @@
...
+    ];
+
+    assert.throws(() => runtime.executeTx(txGroup), 'RUNTIME_ERR1008: Index out of bound');
+  });
+
+  it('should fail if transaction is signed by wrong lsig', () => {
+    updateAndOptIn();
+    runtime.executeTx(donateTxGroup);
+    runtime.executeTx(donateTxGroup);
+    const escrowProg = getProgram('wrongEscrow.teal', { APP_ID: applicationId });
```
ratik21, line 291: What happens if i don't update the crowdfunding application with the escrow address? Does that constitute the same case as this one, or is it different?

examples/crowdfunding/test/failing-tests.js
```diff
@@ -0,0 +1,437 @@
...
+        amountMicroAlgos: 0,
+        lsig: wrongLsig,
+        payFlags: { totalFee: 1000, closeRemainderTo: creator.address }
+      }
+    ];
+
+    assert.throws(() => runtime.executeTx(txGroup), 'RUNTIME_ERR1301: logic signature validation failed.');
+  });
+
+  it('should fail if transaction is lsig is missing', () => {
```
ratik21, line 318:
```suggestion
  it('should fail transaction if logic signature is not passed', () => {
```

examples/crowdfunding/test/failing-tests.js
```diff
@@ -0,0 +1,437 @@
...
+      type: types.TransactionType.DeleteSSC,
+      sign: types.SignType.SecretKey,
+      fromAccount: creator.account,
+      appId: applicationId,
+      payFlags: { totalFee: 1000 },
+      appArgs: [],
+      accounts: [escrow.address] //  AppAccounts
+    };
+
+    assert.throws(() => runtime.executeTx(deleteTx), 'RUNTIME_ERR1008: Index out of bound');
```
ratik21, line 359: Why `Index out of bound`? Shouldn't the transaction be rejected? Can you please confirm?

examples/crowdfunding/test/failing-tests.js
amityadav0: that will be different i think, this case checks logic signature, in this case sender is escrow but logic signature is different.
adding that case as well.

examples/crowdfunding/test/failing-tests.js
```diff
@@ -0,0 +1,437 @@
...
+  });
+
+  it('should fail if trying to update application where sender is not creator', () => {
```
ratik21, line 362:
```suggestion
  it('should fail on trying to update application where sender is not creator', () => {
```

examples/crowdfunding/test/failing-tests.js
amityadav0: the way logic is written it will throw index out of bound, the other way would be to change crowdfunding.teal
but i think we shouldn't because it is same in tutorial

examples/crowdfunding/test/failing-tests.js
ratik21: It does not reflect the txGroup below

examples/crowdfunding/test/failing-tests.js
ratik21: Ok, agreed. Just wanted to make sure it's not a bug.

examples/crowdfunding/test/failing-tests.js
amityadav0: ok

examples/crowdfunding/test/failing-tests.js
ratik21: I was thinking that we can prepare the state without calling the smart contract. eg.
```
runtime.transferAlgo(from, to, ..) // amount = goal, to = escrow
runtime.setGlobalState(appId, "Total", `${goal}`)
creator.setLocalState(appId, "MyAmountGiven",`${goal}`)
```
This way we prepare the correct state before doing claim/reclaim. I just think that `runtime.executeTx(donateTxGroup);` can fail due to number of reasons (eg - mistake in `donate:` in written smart contract, sender not creator, txGroup incorrect, common property incorrect ..etc). So if we can avoid that, and just prepare the correct state, it would be good (as we will be able to test claim/reclaim now with less failures due to dependency). 
Anyways, just a thought. This works too.

examples/crowdfunding/test/failing-tests.js
ratik21: thanks

ratik21 (approved):

PR #246, opened 2021-02-19
ScTmplParam Type update

amityadav0:
https://github.com/scale-it/algorand-builder/issues/243

### DIFF
--- a/packages/algob/src/types.ts
+++ b/packages/algob/src/types.ts
@@ -426,7 +426,7 @@ export interface AlgobDeployer {
     flags: FundASCFlags,
     payFlags: rtypes.TxParams,
     scParams: LogicSigArgs,
-    scTmplParams?: ScParamsMap
+    scTmplParams?: SCParams
   ) => void
 
   /**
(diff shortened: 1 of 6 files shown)


packages/algob/src/types.ts
```diff
@@ -545,6 +545,10 @@ export interface StrMap {
   [key: string]: string
 }
 
+export interface ScParamsMap {
+  [key: string]: string | bigint
```
ratik21, line 549: Shouldn't we include numbers as well? We recommend `bigint`, but i think we should have `number` as well. Right?

amityadav0: https://github.com/scale-it/algorand-builder/issues/243#issuecomment-781930008

ratik21: I am not sure if it means to remove the support for `number` and use only `bigint`. Ignore if i misunderstood.

packages/algob/src/types.ts
```diff
@@ -545,6 +545,10 @@ export interface StrMap {
   [key: string]: string
 }
 
+export interface ScParamsMap {
```
robert-zaremba, line 548: How about renaming to `SCParams`?

robert-zaremba (approved):
Left one suggestion

packages/algob/src/types.ts
amityadav0: ok

PR #248, opened 2021-02-19
[Fix] Txn related ops

### LINKED ISSUE #245: Restricted-access example has multiple errors
- The bob account should be slightly fund to opt-in for the asset
- In the teal, pragma version must be on the first line

ratik21:
- Support for array arguments for `txn`, `gtxn` (eg. `txn Accounts 0`, `gtxn 0 Accounts 0`)
- Fix indexing with `txna`, `gtxna` (0 index meaning the sender's address)
- Fix tests in Parser + add more tests

Closes #245 as well.

### DIFF
--- a/packages/runtime/src/interpreter/opcode-list.ts
+++ b/packages/runtime/src/interpreter/opcode-list.ts
@@ -1178,6 +1178,7 @@ export class Substring3 extends Op {
 // push to stack [...stack, transaction field]
 export class Txn extends Op {
   readonly field: string;
+  readonly idx: number | undefined;
   readonly interpreter: Interpreter;
   readonly line: number;
 
@@ -1192,7 +1193,14 @@ export class Txn extends Op {
   constructor (args: string[], line: number, interpreter: Interpreter) {
     super();
     this.line = line;
-    assertLen(args.length, 1, line);
+    this.idx = undefined;
+    if (args[0] === "ApplicationArgs" || args[0] === "Accounts") { // eg. txn Accounts 1
+      assertLen(args.length, 2, line);
+      assertOnlyDigits(args[1], line);
+      this.idx = Number(args[1]);
+    } else {
+      assertLen(args.length, 1, line);
+    }
     this.assertTxFieldDefined(args[0], interpreter.tealVersion, line);
 
     this.field = args[0]; // field
@@ -1200,11 +1208,17 @@ export class Txn extends Op {
   }
 
   execute (stack: TEALStack): void {
-    const result = txnSpecbyField(
-      this.field,
-      this.interpreter.runtime.ctx.tx,
-      this.interpreter.runtime.ctx.gtxs,
-      this.interpreter.tealVersion);
+    let result;
+    if (this.idx !== undefined) { // if field is an array use txAppArg (with "Accounts"/"ApplicationArgs")
+      result = txAppArg(this.field, this.interpreter.runtime.ctx.tx, this.idx, this,
+        this.interpreter.tealVersion, this.line);
+    } else {
+      result = txnSpecbyField(
+        this.field,
+        this.interpreter.runtime.ctx.tx,
+        this.interpreter.runtime.ctx.gtxs,
+        this.interpreter.tealVersion);
+    }
     stack.push(result);
   }
 }
@@ -1215,12 +1229,13 @@ export class Txn extends Op {
 export class Gtxn extends Op {
   readonly field: string;
   readonly txIdx: number;
+  readonly txFieldIdx: number | undefined;
   readonly interpreter: Interpreter;
   readonly line: number;
 
   /**
    * Sets `field`, `txIdx` values according to arguments passed.
-   * @param args Expected argumensts: [transaction group index, transaction field]
+   * @param args Expected arguments: [transaction group index, transaction field]
    * // Note: Transaction field is expected as string instead of number.
    * For ex: `Fee` is expected and `0` is not expected.
    * @param line line number in TEAL file
@@ -1229,7 +1244,14 @@ export class Gtxn extends Op {
   constructor (args: string[], line: number, interpreter: Interpreter) {
     super();
     this.line = line;
-    assertLen(args.length, 2, line);
+    this.txFieldIdx = undefined;
+    if (args[1] === "ApplicationArgs" || args[1] === "Accounts") {
+      assertLen(args.length, 3, line); // eg. gtxn 0 Accounts 1
+      assertOnlyDigits(args[2], line);
+      this.txFieldIdx = Number(args[2]);
+    } else {
+      assertLen(args.length, 2, line);
+    }
     assertOnlyDigits(args[0], line);
     this.assertTxFieldDefined(args[1], interpreter.tealVersion, line);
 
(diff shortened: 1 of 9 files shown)


packages/runtime/test/src/parser/parser.ts
```diff
@@ -394,9 +394,12 @@ describe("Parser", function () {
...
 
     it("should return correct objects for `txn`", () => {
-      const res = opcodeFromSentence(["txn", "Fee"], 1, interpreter);
-      const expected = new Txn(["Fee"], 1, interpreter);
+      let res = opcodeFromSentence(["txn", "Fee"], 1, interpreter);
+      let expected = new Txn(["Fee"], 1, interpreter);
+      assert.deepEqual(res, expected);
 
+      res = opcodeFromSentence(["txn", "Accounts", "1"], 1, interpreter);
+      expected = new Txn(["Accounts", "1"], 1, interpreter);
```
amityadav0, line 402: why do we need to update these tests ? aren't we testing the array part in opcode-list.ts ?
this tests was only to check if correct object us created.

packages/runtime/test/src/parser/parser.ts
```diff
@@ -428,11 +434,19 @@ describe("Parser", function () {
...
+      let expected = new Txna(["Accounts", "0"], 1, interpreter);
+      assert.deepEqual(res, expected);
 
+      res = opcodeFromSentence(["txna", "ApplicationArgs", "2"], 1, interpreter);
+      expected = new Txna(["ApplicationArgs", "2"], 1, interpreter);
       assert.deepEqual(res, expected);
 
+      expectRuntimeError(
+        () => opcodeFromSentence(["txna", "Fee", "2"], 1, interpreter),
+        RUNTIME_ERRORS.TEAL.INVALID_OP_ARG
```
amityadav0, line 447: i think this test should be in opcode-list, because this tests the field. parser only parses the sentence.

packages/runtime/test/src/parser/parser.ts
```diff
@@ -445,11 +459,19 @@ describe("Parser", function () {
...
+      let expected = new Gtxna(["1", "Accounts", "1"], 1, interpreter);
+      assert.deepEqual(res, expected);
 
+      res = opcodeFromSentence(["gtxna", "1", "ApplicationArgs", "4"], 1, interpreter);
+      expected = new Gtxna(["1", "ApplicationArgs", "4"], 1, interpreter);
       assert.deepEqual(res, expected);
 
+      expectRuntimeError(
+        () => opcodeFromSentence(["gtxna", "1", "Fee", "4"], 1, interpreter),
+        RUNTIME_ERRORS.TEAL.INVALID_OP_ARG
```
amityadav0, line 472: same

packages/runtime/test/src/parser/parser.ts
ratik21: No, this should be in parser because this validation is being done in the opcode constructor. Parser returns a new opcode, so validations in constructor should be handled in the parser tests. In `opcode-list.ts`, we are testing `op.execute()`, and not `new Op(..)`.

packages/runtime/test/src/parser/parser.ts
ratik21: This is related to `assertLen` in constructor. Since we are passing a txField of array type we should have 2 args now, otherwise 1. Hence added two tests - a) where num of args = 1 with `Fee`, and b) where num of args = 2 with `Accounts/ApplicationArgs`. So this is related to creation of object only, not execution.
In opcode-list we test the execution (where we check if correct value was pushed to stack or not)

amityadav0: ok

packages/runtime/test/src/parser/parser.ts
amityadav0: ok

packages/runtime/src/lib/txn.ts
```diff
@@ -104,20 +104,18 @@ export function txnSpecbyField (txField: string, tx: Txn, gtxns: Txn[], tealVers
...
 
-  throw new RuntimeError(RUNTIME_ERRORS.TEAL.INVALID_OP_ARG, {
-    opcode: "txna or gtxna"
-  });
+  if (txField === 'Accounts') {
+    if (idx === 0) { return parseToStackElem(tx.snd, txField) as Uint8Array; }
+    idx--; // if not sender, then reduce index by 1
+  }
+  op.checkIndexBound(idx, result, line);
+  return parseToStackElem(result[idx], txField) as Uint8Array;
```
amityadav0, line 118: are we handling `ApplicationArgs` here ?

packages/runtime/test/src/interpreter/opcode-list.ts
```diff
@@ -1951,6 +1951,35 @@ describe("Teal Opcodes", function () {
         assert.equal(1, stack.length());
         assert.deepEqual(TXN_OBJ.apsu, stack.pop());
       });
+
+      it("should push value from accounts or args array by index", function () {
```
amityadav0, line 1955: please add test for `ApplicationArgs`.

packages/runtime/test/src/parser/parser.ts
amityadav0: please add test for `ApplicationArgs`.

packages/runtime/src/interpreter/opcode-list.ts
```diff
@@ -1288,6 +1318,8 @@ export class Txna extends Op {
 
 // push value of a field to the stack from a transaction in the current transaction group
 // push to stack [...stack, value of field]
+// NOTE: for arg="Accounts" index 0 means sender's address, and index 1 means first address from accounts
```
amityadav0, line 1321: add comments for application args also

packages/runtime/test/src/interpreter/opcode-list.ts
ratik21: [https://github.com/scale-it/algorand-builder/pull/248/files/71ca98c1ebf4be194883bca6637fff28fc203b9e#diff-0bd11169d40e3be81fef88fe8a8e4cf0795255b19d1f4d44d615ea872a1e0e90R1977](https://github.com/scale-it/algorand-builder/pull/248/files/71ca98c1ebf4be194883bca6637fff28fc203b9e#diff-0bd11169d40e3be81fef88fe8a8e4cf0795255b19d1f4d44d615ea872a1e0e90R1977)

packages/runtime/src/lib/txn.ts
ratik21: yes. For `ApplicationArgs` it's normal indexing, that's why if we have `Accounts`, we are reducing the index by 1, otherwise normal.

packages/runtime/src/interpreter/opcode-list.ts
ratik21: For `ApplicationArgs` index 0 means 1st application arg from array. This is the normal case. Should i add a comment for this?

packages/runtime/test/src/parser/parser.ts
ratik21: ok, it will be similar to Accounts though

packages/runtime/src/interpreter/opcode-list.ts
amityadav0: yes, i think that will be good, because we are taking index both in application and accounts. but application args index is different

ratik21: ok, fair point.

amityadav0 (approved):

PR #249, opened 2021-02-22
[fix] group tx, ctx, store flow

amityadav0:
- Fix group transaction flow
- Fix Clear SSC
- Fix tests


packages/runtime/src/runtime.ts
```diff
@@ -736,141 +595,139 @@ export class Runtime {
...
+          this.ctxOp.destroyAsset(txnParam.assetID);
           break;
         }
       }
+    });
+    if (clearErrorBool) {
+      this.store = this.ctx.state;
+      throw new RuntimeError(RUNTIME_ERRORS.TEAL.REJECTED_BY_LOGIC);
     }
+  }
```
ratik21, line 702: Can we move `this.processTransactions()` to `Ctx` class (if we are operating entirely on `ctx` in this function) ?

packages/runtime/src/runtime.ts
```diff
@@ -736,141 +595,139 @@ export class Runtime {
...
           const appParams = this.assertAppDefined(txnParam.appId, this.getApp(txnParam.appId));
-          const fromAccount = this.getAccount(txnParam.fromAccount.addr);
           try {
             this.run(appParams["clear-state-program"], ExecutionMode.STATEFUL);
           } catch (error) {
             // if transaction type is Clear Call, remove the app first before throwing error (rejecting tx)
             // https://developer.algorand.org/docs/features/asc1/stateful/#the-lifecycle-of-a-stateful-smart-contract
-            fromAccount.closeApp(txnParam.appId); // remove app from local state
-            throw error;
+            if (error instanceof RuntimeError && error.number === 1007) {
```
ratik21, line 642: Please add a comment regarding error no. `1007`.

packages/runtime/src/ctx.ts
```diff
@@ -0,0 +1,247 @@
...
+      });
+    }
+  }
+
+  /**
+   * Fetches account from `runtime.ctx`
+   * @param address account address
+   */
+  getAccount (address: string): StoreAccountI {
+    const account = this.runtime.ctx.state.accounts.get(address);
```
ratik21, line 42: Can you please explain why we're using `runtime.ctx` here? We are in `Ctx` class but we are operating on `runtime.ctx.state` here. Shouldn't this be `this.state` (which would be a deepCopy of `runtime.store`)?

examples/crowdfunding/test/crowdfundingTest.js
```diff
@@ -108,14 +108,15 @@ describe('Crowdfunding Tests', function () {
     console.log('Escrow Address: ', escrowAddress);
 
     // fund escrow with some minimum balance first
-    runtime.transferAlgo({
+    const param = {
```
ratik21, line 111:
```suggestion
    const fundEscrowParam = {
```

examples/crowdfunding/test/crowdfundingTest.js
```diff
@@ -108,14 +108,15 @@ describe('Crowdfunding Tests', function () {
...
       type: types.TransactionType.TransferAlgo,
       sign: types.SignType.SecretKey,
       fromAccount: master.account,
       toAccountAddr: escrowAddress,
       amountMicroAlgos: minBalance,
       payFlags: {}
-    });
+    };
+    runtime.executeTx(param);
```
ratik21, line 119:
```suggestion
    runtime.executeTx(fundEscrowParam);
```

packages/runtime/src/ctx.ts
```diff
@@ -0,0 +1,247 @@
...
+    if (txnParam.payFlags.closeRemainderTo) {
+      const closeRemToAcc = this.getAccount(txnParam.payFlags.closeRemainderTo);
+
+      closeRemToAcc.amount += fromAccount.amount; // transfer funds of sender to closeRemTo account
+      fromAccount.amount = 0; // close sender's account
+    }
+  }
+
+  /**
+   * Deduct fee from sender.
```
ratik21, line 108:
```suggestion
   * Deduct transaction fee from sender account.
```

packages/runtime/src/ctx.ts
```diff
@@ -0,0 +1,247 @@
...
+   * @param sender Sender address
+   * @param index Index of transaction in a group
```
ratik21, line 110:
```suggestion
   * @param index Index of transaction in tx group
```

packages/runtime/src/runtime.ts
amityadav0: ok, this is because we only closeout if error is rejected by logic, other than that we should throw error,
we were missing this earlier. will add in code also

packages/runtime/src/ctx.ts
amityadav0: Because it is not required to make a deepcopy of a deepcopy.
think about the flow, first we make a deepcopy and send it to interpreter(interpreter works on runtime.ctx) if the execution fails at some point we don't update the runtime.store.

same is the case with processTransaction, we process transactions on deepcopy of store, if it fails we don;t update store with ctx, else we update store.

amityadav0: let me know it that makes sense.

packages/runtime/src/runtime.ts
amityadav0: yes

packages/runtime/src/ctx.ts
ratik21: No i understand about the deepCopy, i am talking about the intuitiveness here (using  `runtime.ctx.state`) inside a `Ctx` class.
When i was talking about using `Ctx` as a class, i was referring to `runtime.ctx` only (not `ctxOp`). Basically having `runtime.ctx = new Ctx(deepCopy(this.store), tx, gtxs, args..)`. Class `Ctx` would be of type `Context` in `/runtime/types` only. Interpreter would have used this class functions (instead of having local interpreter functions). This way the separation b/w `ctx & store` would be better also.

For here, maybe during initialization we can do `this.state = runtime.ctx.state` in the constructor? Otherwise it seems a little confusing to me. We can keep as it is if you think otherwise.

packages/runtime/src/runtime.ts
ratik21: ok

packages/runtime/src/ctx.ts
amityadav0: ok, do you mean ``` export interface Context {
  state: State
  tx: Txn // current txn
  gtxs: Txn[] // all transactions
  args: Uint8Array[]
  transferAlgo()
  transferAsset()
}```

we will have these in types and we will have this class as Context. right ?
yes, this seems better.

ctx = deepCopy(this.store)
yes in constructor we can also new Ctx(ctx, tx, gtxs, args..)
use this.state = ctx
this.tx = tx
like this

ratik21: > we will have these in types and we will have this class as Context. right ?

Exactly. Instead of having `runtime.ctx` as an obj, we can convert that to a class (which we can define in a new file, as done here). So this:
```
this.ctx = {
  state: cloneDeep(this.store), // state is a deep copy of store
  tx: tx,
  gtxs: gtxs,
  args: []
};
 ```   
 would update to:
 ```
 this.ctx = new Ctx(cloneDeep(this.store), tx, gtxs, [])
 ```
 And all ctx related functions would be in `Ctx` class.
 
 If you like it then we can update. Your call. 
 Otherwise we can just update the constructor here like you said above.

Comment (amityadav0) [reply packages/runtime/src/ctx.ts]: yes, i like it, updating to suggested one

File: packages/runtime/src/runtime.ts (hunk 2 above)
Comment (ratik21) [reply packages/runtime/src/runtime.ts]: just to confirm something, what about the errors thrown during execution other than `rejected_by_logic`? Let's say `txn Accounts` throws `index_out_of_bound` error (which happens during execution of `clear.teal`). We clear the state in these cases as well, Right?

Comment (amityadav0) [reply packages/runtime/src/runtime.ts]: No we don't.

Comment (ratik21) [reply packages/runtime/src/runtime.ts]: I am not sure, that's what i wanted to confirm here. Basically when we execute TEAL on blockchain (with index out of bound error, or something similar), does that constitute a tx `rejected_by_logic` there or not

Comment (amityadav0) [reply packages/runtime/src/runtime.ts]: https://developer.algorand.org/docs/features/asc1/teal/#getting-transaction-properties

Comment (amityadav0) [reply packages/runtime/src/runtime.ts]: updated as per network behaviour

File: packages/runtime/src/ctx.ts [hunk 8]
```diff
@@ -0,0 +1,370 @@
...
+
+      closeRemToAcc.amount += fromAccount.amount; // transfer funds of sender to closeRemTo account
+      fromAccount.amount = 0; // close sender's account
+    }
+  }
+
+  /**
+   * Deduct transaction fee from sender account.
+   * @param sender Sender address
+   * @param index Index of transaction in tx group
```
ratik21, line 135:
```suggestion
   * @param index Index of current tx being processed in tx group
```

packages/runtime/src/ctx.ts
```diff
@@ -0,0 +1,370 @@
...
+      });
+    }
+    return assetHolding;
+  }
+
+  /**
+   * Fetches app from `ctx state`
+   * @param appId Application Index
+   */
+  getApp (appId: number, line?: number): SSCAttributesM {
```
ratik21, line 103: Please update the params in description

packages/runtime/src/ctx.ts
```diff
@@ -0,0 +1,370 @@
...
+  }
+
+  /**
+   * https://developer.algorand.org/docs/features/asa/#destroying-an-asset
+   * Destroy asset
+   * @param sender sender's address
+   * @param assetId asset index
+   * @param payFlags transaction parameters
+   */
+  destroyAsset (assetId: number): void {
```
ratik21, line 231: Please update the description

packages/runtime/src/ctx.ts
```diff
@@ -0,0 +1,370 @@
...
+    const fromAccount = this.getAccount(sender);
+    // https://developer.algorand.org/docs/reference/cli/goal/app/closeout/#search-overlay
+    this.runtime.assertAppDefined(appId, this.getApp(appId));
+    fromAccount.closeApp(appId); // remove app from local state
+  }
+
+  /**
+   * Process transactions in ctx
+   * - Runs TEAL code if associated with transaction
+   * - Executes the transaction on ctx
```
ratik21, line 276: Please also add the explanation (if not added somewhere). That we're doing  this because if any one tx in group fails, then it does not affect `runtime.store`, otherwise we just update `store` with `ctx` (if all transactions are executed successfully).

packages/runtime/src/runtime.ts
```diff
@@ -736,144 +587,29 @@ export class Runtime {
...
-      }
-    }
+    // Run TEAL program associated with each transaction and
+    // then execute the transaction without interacting with store.
+    this.ctx.processTransactions(txnParameters);
 
     // update store only if all the transactions are passed
-    this.updateFinalState(txnParameters);
+    this.store = this.ctx.state;
   }
```
ratik21, line 613: just to confirm, i don't see the update  `this.ctx = {..}` to `this.ctx = new Ctx(..)` for app/asset creation & opt-in. Is that fine the previous way?

ratik21 (approved):
Left few comments. Please also check if we need to update `/docs/testing-teal.md` before merging.

packages/runtime/src/ctx.ts
amityadav0: ok

packages/runtime/src/runtime.ts
amityadav0: i think it is fine, otherwise test should fail.
-  asset creation, opt-in doesn't require ctx update.
- app creation, update and optin, we are updating ctx in functions

PR #251, opened 2021-02-26
Support logic sig for opt-in ASA (Contract Account)

ratik21:


packages/algob/src/lib/algo-operator.ts
```diff
@@ -109,20 +110,33 @@ export class AlgoOperatorImpl implements AlgoOperator {
   }
 
   async _optInToASA (
-    asaName: string, assetIndex: number, account: rtypes.Account, params: algosdk.SuggestedParams
+    asaName: string, assetIndex: number, account: rtypes.Account | undefined,
+    params: algosdk.SuggestedParams, signature: rtypes.Sign
```
robert-zaremba, line 114: Instead of using `(account, signature)` as parameters, how about using only one argument:
```
signer: account | Lsig
```
The problem with current design is that `account` is only used when `signature.sign === SignType.SecretKey`. Moreover in that case the `signature.lsig` should not be defined. So we have some dependency between parameters and that my not  be the best thing.
Other solution would be to have two functions: `accountASAOptIn` and `lsigASAOptIn`.

robert-zaremba (changes requested):
Let's change the arguments of `optInToASA` or break it in 2 functions. Moreover detailed docs must be provided for interface functions - because user usually sees the interface.

robert-zaremba:
Please update CHANGELOG as well

packages/algob/src/lib/algo-operator.ts
ratik21: Thanks for suggestion. I think using 2 functions would be more intuitive.

packages/algob/src/types.ts
```diff
@@ -476,7 +476,7 @@ export interface AlgobDeployer {
   /**
    * Creates an opt-in transaction for given ASA name, which must be defined in
    * `/assets/asa.yaml` file. */
-  optInToASA: (name: string, accountName: string, flags: rtypes.TxParams) => Promise<void>
+  optInToASA: (name: string, accountName: string, flags: rtypes.TxParams, sign: rtypes.Sign) => Promise<void>
```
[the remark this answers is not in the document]
ratik21: Ok, will update doc string

ratik21:
Breaked `optInToASA` into 2 functions. Updated ChangeLog as well.

robert-zaremba (approved):
Pre approving. Things to update:

* rename  `accountOptInToASA` -> `optInAcountToASA`  (same with lsig)
* Add / update docs related to optIn (in /docs)

PR #252, opened 2021-03-01
Add Approval and Clear checks and tests

amityadav0:
empClear

- Add checks because we cannot pass empty programs


ratik21:
We need to handle this for update application as well. And also for stateless TEAL - creating/executing logicSig with empty program (if not handled already)

robert-zaremba (approved):
Looks good, thanks for fixing.

PR #256, opened 2021-03-08
Add algob tests

man-of-code:


examples/crowdfunding/package.json
```diff
@@ -4,7 +4,8 @@
   "main": "index.js",
   "license": "Apache-2.0",
   "dependencies": {
-    "algosdk": "^1.8.0"
+    "algosdk": "^1.8.0",
+    "glob": "^7.1.6"
```
ratik21, line 8: Why do we need this dependency in crowdfunding?

packages/algob/src/builtin-tasks/test.ts
```diff
@@ -1,33 +1,25 @@
...
-    throw new Error("unexpected non-project execution");
-  }
-  return glob.sync(path.join(config.paths.tests, "**/*.js")).sort(cmpStr);
+async function runTests (config: AlgobConfig): Promise<void> {
+  const testFiles = loadFilenames(testsDirectory);
+  console.log("Test files:", testFiles);
+  const mocha = new Mocha(config.mocha);
+  console.log("Adding test files to mocha");
+  testFiles.forEach((file) => mocha.addFile(file));
+  console.log("Executing test files ");
```
ratik21, line 15: Please remove these `console.logs`. If you think it's necessary to keep them, then we can use `console.info`/ `debug logs`. But i don't think we need them (You can just add comments to code, if not clear).

packages/algob/src/builtin-tasks/test.ts
```diff
@@ -1,33 +1,25 @@
...
+    mocha.run(resolve);
+  });
+  console.log(testFailures);
 }
 
-async function runTests ({ fileNames }: TaskArgs, { config }: AlgobRuntimeEnv): Promise<void> {
-  const testfiles = loadFiles;
-  console.log("Test files:", testfiles);
+export default function (): void {
+  task(TASK_TEST, "Runs tests using mocha in project root")
```
ratik21, line 23:
```suggestion
  task(TASK_TEST, "Run tests using mocha in project root")
```

packages/algob/src/builtin-tasks/test.ts
man-of-code: I thought the command would be more informative by these statements, informing the user about every stage of the process

examples/crowdfunding/package.json
man-of-code: You're right, we don't need this.  I have removed it.

packages/algob/src/lib/script-checkpoints.ts
```diff
@@ -18,6 +18,7 @@ import {
 } from "../types";
 
 export const scriptsDirectory = "scripts";
+export const testsDirectory = "test";
```
amityadav0, line 21: Please rename this to `tests`

packages/algob/src/builtin-tasks/test.ts
```diff
@@ -1,33 +1,25 @@
...
-  }
-  return glob.sync(path.join(config.paths.tests, "**/*.js")).sort(cmpStr);
+async function runTests (config: AlgobConfig): Promise<void> {
+  const testFiles = loadFilenames(testsDirectory);
+  console.log("Test files:", testFiles);
+  const mocha = new Mocha(config.mocha);
+  console.log("Adding test files to mocha");
+  testFiles.forEach((file) => mocha.addFile(file));
+  console.log("Executing test files ");
+  const testFailures = await new Promise<number>((resolve, reject) => {
```
amityadav0, line 16: do we need `reject` here ?
or can we use this for error handling ?

packages/algob/src/lib/script-checkpoints.ts
man-of-code: I was confused with this, but then I found the tests folder named as `test` in crowdfunding example. 
I'll change them both appropriately.

packages/algob/src/builtin-tasks/test.ts
man-of-code: we don't need `reject` here.

packages/algob/src/builtin-tasks/test.ts
ratik21: `algob help` should be more than intuitive enough. Anyways this should be changed to `console.info` then (not `console.log`)

packages/algob/src/lib/script-checkpoints.ts
amityadav0: we can have multiple files in `test`. therefore `tests` is more "intuitive"

amityadav0: but does it work ?

ratik21: > we can have multiple files in `test`. therefore `tests` is more "intuitive"

Okay, make sense. Let's add somewhere in `docs` (or maybe in algob help) that test directory should be named `tests` then. I don't think tests will run if user adds/renames it to test

ratik21: But usually in all packages i find the test folder to be named as "test" (instead of "tests", even with algob & runtime as well).

man-of-code: @robert-zaremba  What do you think?

robert-zaremba: I think both works, it's a small  detail. Whatever you decide will work, let's just be consistent.

man-of-code: I think I agree with Amit here, `tests`  is the one we should use. 
I'll update the name of the directories, then.

man-of-code: I am changing this to `test` as most test packages like mocha or jest, look for the test files in that.

packages/algob/src/builtin-tasks/test.ts
```diff
@@ -1,33 +1,24 @@
...
+async function runTests (config: AlgobConfig): Promise<void> {
+  const testFiles = loadFilenames(testsDirectory);
+  console.log("Test files:", testFiles);
+  const mocha = new Mocha(config.mocha);
+  // Adding test files to mocha object
+  testFiles.forEach((file) => mocha.addFile(file));
+  const testFailures = await new Promise<number>((resolve) => {
+    mocha.run(resolve);
+  });
+  console.log(testFailures);
```
ratik21, line 18: I don't think we need this log. `mocha.run` shows test failures very intuitively (as showed on call).

packages/algob/src/lib/script-checkpoints.ts
ratik21, line 21: Please add documentation regarding this (that test directory should be named "test" only, otherwise `algob test` will fail.

ratik21: Also rename `/sample-project/tests` -> `/sample-project/test`

ratik21 (changes requested):
Let's update the error message here (i ran `algob test` in `examples/asa`, where test directory doesn't exist)
image

docs/user-script-execution.md
```diff
@@ -36,4 +36,11 @@ Example:
 
     algob deploy // will run all the scripts present in scripts folder
     algob deploy scripts/script1.js // will run script1.js.
-    algob deploy scripts/script2.js scripts/script1.js // will run script2.js and script1.js in given order
\ No newline at end of file
+    algob deploy scripts/script2.js scripts/script1.js // will run script2.js and script1.js in given order
+
+
```
amityadav0, line 41:
```suggestion
```

amityadav0 (approved):
LGTM

docs/user-script-execution.md
```diff
@@ -36,4 +36,11 @@ Example:
...
+## Tests
+
+In algob, test files are JS files located in `test` directory and their purpose is to test the smart contract logic present in `assets` directory.
```
ratik21, line 44:
```suggestion
In algob, test files are JS files located in `test` directory and their purpose is to test algorand smart contracts present in `/assets`. Check `@algorand-builder/runtime` package for more info on tests.
```

ratik21: Why only JS files? What if user writes tests in `.ts` (eg. in `/htlc-ts`)

packages/algob/src/builtin-tasks/test.ts
```diff
@@ -1,33 +1,32 @@
...
-    )
-    .setAction(runTests);
-}
-
-function loadFiles (fileNames: string[], config: ResolvedAlgobConfig): string[] {
-  if (fileNames.length !== 0) {
-    return fileNames;
+async function runTests (config: AlgobConfig): Promise<void> {
+  if (!fs.existsSync(testsDirectory)) {
+    console.log("Test directory doesn't exists or exists with a different name. Please ensure that it's name is 'test'.");
```
ratik21, line 12:
```suggestion
    console.log("Test directory does not exist or exists with a different name. Please ensure that it's name is 'test'.");
```

ratik21 (approved):
Pre-approving, left few comments

packages/algob/src/builtin-tasks/test.ts
man-of-code: I have completely removed this line. I have added a builtin error now which will be thrown by `loadfiles` func.

PR #258, opened 2021-03-12
Add new command: algob unbox-template

ratik21:
`algob unbox-template <template-name> <destination> --force (flag)`
 - if `destination` is not passed then `cwd` will be used
 - if `template-name` is not passed, then default template "/bare" is unboxed
 - if `--force` is passed, then files are overwritten
 - if `templateName` passed is not present in `scale-it/algorand-builder-templates`, user is made to choose from one of the existing templates or exit
 - template is copied from a temporary directory to destination
 - If `--force` is not used and there are conflicting files, then user chooses to overwrite those files or not.
 - After copying user we install the dependencies in `/destination` (after confirming from user)


packages/algob/src/builtin-tasks/unbox-template.ts
```diff
@@ -0,0 +1,24 @@
+import { unbox } from "../internal/cli/unbox-template";
+import { task } from "../internal/core/config/config-env";
+import { TASK_UNBOX_TEMPLATE } from "./task-names";
+
+export default function (): void {
+  task(TASK_UNBOX_TEMPLATE, "Unboxes a new dapp template using algorand-builder")
+    .addFlag(
+      "force",
+      "Unbox project in the current directory regardless of its " +
+      "state. Be careful, this\n                will potentially overwrite files " +
```
amityadav0, line 10:
```suggestion
      "state. Be careful, this\n                will overwrite files " +
```

packages/algob/src/internal/cli/unbox-template.ts
```diff
@@ -0,0 +1,191 @@
...
+      createConfirmationPrompt(
+        "shouldInstall",
+        `Do you want to install dapp template ${name}'s dependencies with ${packageManager} ?`
+      )
+    ]);
+  } catch (e) {
+    if (e === "") {
+      return false;
+    }
+
```
amityadav0, line 41:
```suggestion
```

packages/algob/src/internal/cli/unbox-template.ts
```diff
@@ -0,0 +1,191 @@
...
+    throw e;
+  }
+
```
amityadav0, line 44:
```suggestion
```

packages/algob/src/internal/cli/unbox-template.ts
```diff
@@ -0,0 +1,191 @@
...
+          createConfirmationPrompt(
+            "shouldProceedWithNonEmptyDir",
+            `Do you want to proceed with the unboxing?`
+          )
+        ]);
+      } catch (e) {
+        if (e === "") {
+          return;
+        }
+
```
amityadav0, line 74:
```suggestion
```

amityadav0 (commented):
Please update Readme.

packages/algob/src/builtin-tasks/unbox-template.ts
```diff
@@ -0,0 +1,24 @@
+import { unbox } from "../internal/cli/unbox-template";
+import { task } from "../internal/core/config/config-env";
+import { TASK_UNBOX_TEMPLATE } from "./task-names";
+
+export default function (): void {
+  task(TASK_UNBOX_TEMPLATE, "Unboxes a new dapp template using algorand-builder")
+    .addFlag(
+      "force",
+      "Unbox project in the current directory regardless of its " +
+      "state. Be careful, this\n                will overwrite files " +
```
amityadav0, line 10: do we need the spaces here ?

packages/algob/src/builtin-tasks/unbox-template.ts
```diff
@@ -0,0 +1,24 @@
...
+      "that exist in the directory."
+    )
+    .addOptionalPositionalParam<string>(
+    "destination",
+    "Path to the directory in which you would like to unbox the project files. " +
+      "If destination is\n                not provided, this defaults to the current directory.\n"
```
amityadav0, line 16: spaces here as well ?

packages/algob/src/internal/cli/unbox-template.ts
```diff
@@ -0,0 +1,188 @@
+import chalk from "chalk";
+import enquirer from "enquirer";
+import fse from "fs-extra";
+import path from "path";
+
+import { copyTemplatetoDestination, fetchRepository, setUpTempDirectory } from "../util/unpack";
+import { createConfirmationPrompt, installDependencies, printSuggestedAlgobCommands, printWelcomeMessage } from "./project-creation";
+
+const ALGOB_DAPP_TEMPLATES_GIT_REMOTE = 'scale-it/algorand-builder-templates';
```
amityadav0, line 9: please rename this to `algo`

packages/algob/src/internal/cli/unbox-template.ts
```diff
@@ -0,0 +1,188 @@
...
+};
+
+// commands to start and build the react app
+function printSuggestedDAppCommands (packageManager: string): void {
+  console.log(`To use the react app, try running:`);
+  console.log(`  ${packageManager} start`);
+  console.log(`  ${packageManager} build`);
+}
+
+export async function unbox ({ force, destination, templateName }:
```
amityadav0, line 138: Please add doc string.

amityadav0 (commented):
is it desired ?
i think we should be able to unbox it in any directory.

packages/algob/src/builtin-tasks/unbox-template.ts
ratik21: Yes, for displaying help. 
image

packages/algob/src/internal/cli/unbox-template.ts
ratik21: added

packages/algob/src/internal/cli/unbox-template.ts
ratik21: `algob`? I renamed `'scale-it/algorand-builder-templates'` -> `'scale-it/algo-builder-templates'`

ratik21:
> i think we should be able to unbox it in any directory.

Thanks, i agree. Updated code to unbox in any directory.

amityadav0:
df

how about creating a folder in`cwd` instead of directly copying all the things ? i think that is more clear.

packages/algob/src/builtin-tasks/unbox-template.ts
amityadav0: ok

amityadav0:
also, i am not able to run the command twice in same directory

ratik21:
> how about creating a folder in`cwd` instead of directly copying all the things ? i think that is more clear.

We are not directly copying everything. Please read the PR description first (or `algob unbox-template --help)`. Creating a folder should be determined by the user, if you don't pass <destination>, then it copies everything in the `cwd`. 
So if you do `algob unbox-template bare ./some-folder`, then it copies everything from `/bare` to `/some-folder` (user has the control). Or do you mean directly copying the entire folder (`/bare` in this case) and not it's contents? If yes, then i am not sure that's a good approach.

 > also, i am not able to run the command twice in same directory
 
 `Cannot find module @algorand-builder/algob `: Are you sure this error is related to the unbox command & not algob config ? I use algob with `yarn run ..` which is working fine on my system (i am able to use `algob unbox-template` multiple times in the same directory).

amityadav0:
> > how about creating a folder in`cwd` instead of directly copying all the things ? i think that is more clear.
> 
> We are not directly copying everything. You can check the PR description (or `algob unbox-template --help)`. Creating a folder should be determined by the user, if you don't pass , then it copies everything in the `cwd`.
> So if you do `algob unbox-template bare ./some-folder`, then it copies everything from `/bare` to `/some-folder` (user has the control). Or do you mean directly copying the entire folder (`/bare` in this case) and not it's contents? If yes, then i am not sure that's a good approach.
> 
> > also, i am not able to run the command twice in same directory
> 
> `Cannot find module @algorand-builder/algob `: Are you sure this error is related to the unbox command & not algob config ? I use algob with `yarn run ..` which is working fine on my system (i am able to use `algob unbox-template` multiple times in the same directory). If you're unable to resolve this, could you please provide steps so that i can replicate it on my system?

- This one is OK too.

- cannot find module, i am getting this error if i use the command in a folder in already copied template.

man-of-code:
image

Why is this happening? Am I doing something wrong..?
I'm in the `/home` directory and `temp` is a folder in it.

man-of-code:
image


This is the error @amityadav0 pointed out.

man-of-code (commented):
@ratik21 Please look into the errors I pointed out. Please let me know in case I did something wrong.

ratik21:
> Why is this happening? Am I doing something wrong..?
> I'm in the `/home` directory and `temp` is a folder in it.

Thanks for checking. Yes your syntax is incorrect. You haven't passed the template name with the command (which is the first param). So in your case `temp/` is being used as the template name, and since there are no templates with that name in algo-builder-templates, user is asked a prompt to select from an existing template OR exit. 

Correct way in your case would be: `algob unbox-template bare ./temp`

amityadav0:
> image
> 
> Why is this happening? Am I doing something wrong..?
> I'm in the `/home` directory and `temp` is a folder in it.
> 
> Apart from this, I really like how the command interacts with the user. 

I think template name will go first and then directory.
unbox-template template-name folder-name

ratik21:
> image
> 
> This is the error @amityadav0 pointed out.

Yeah, as discussed on discord, the only working solution currently would be to do `yarn link/add @algo-builder/algob ` to use algob with `algob.config.js`. @man-of-code I have already updated the sample-project readme regarding this (for `algob init`), and you will need to add/update the algob-unbox-template docs readme you're working on. Discussion: [https://discord.com/channels/754286650540032030/754286650540032033/822494123478483014](https://discord.com/channels/754286650540032030/754286650540032033/822494123478483014)

packages/algob/src/builtin-tasks/unbox-template.ts
```diff
@@ -0,0 +1,24 @@
+import { unbox } from "../internal/cli/unbox-template";
+import { task } from "../internal/core/config/config-env";
+import { TASK_UNBOX_TEMPLATE } from "./task-names";
+
+export default function (): void {
+  task(TASK_UNBOX_TEMPLATE, "Unboxes a new dapp template using algorand-builder")
```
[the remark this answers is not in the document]
ratik21: Weird, It's already updated.

man-of-code:
> Correct way in your case would be: `algob unbox-template bare ./temp`

You are right, but you mentioned that whenever template name is not provided then the command provides a list from which user can select the template. I was checking that case.

ratik21:
> You are right, but you mentioned that whenever template name is not provided then the command provides a list from which user can select the template. I was checking that case.

No. Whenever template name is not provided, then default template (/bare) is used. It's also mentioned in the PR description. You can check it

man-of-code (approved):
The command looks good to me.
It is descriptive and intuitive enough for a new user.

PR #259, opened 2021-03-12
In algob deploy/run show line number when handling script errors

ratik21:
previous:
image
updated to:
image

Also updated `Bad Request` error.
eg. 
image
updated to
image


amityadav0 (commented):
Please update message to be more clear.
Instead of `:9:46` it can be `Line: 9..` or something else.

ratik21:
> Instead of :9:46 it can be Line: 9.. or something else

`:line:charAtline` is how error stack trace (in general) shows line number in an error, which should be clear enough. But will update if you think `error at Line..` is more clear.

man-of-code (commented):
I think we should remove `Algorand` name from variables and functions too. We can have this function as `parseBuilderError`. What do you think? @robert-zaremba

robert-zaremba:
> I think we should remove `Algorand` name from variables and functions too. We can have this function as `parseBuilderError`. What do you think? @robert-zaremba

What kind of variables? We can use it in a code. Algorand is a name of product we use. We can't create other product with that name.

packages/algob/src/internal/util/scripts-runner.ts
```diff
@@ -18,17 +18,32 @@ async function loadScript (relativeScriptPath: string): Promise<any> {
   }
 }
 
+// returns error line number with path. eg: scripts/2-gold-asc.js => scripts/2-gold-asc.js:11:3
```
amityadav0, line 21: Please add doc string for passing variables

amityadav0 (approved):

packages/algob/src/internal/util/scripts-runner.ts
ratik21: ok

PR #261, opened 2021-03-15
Fix checkpoints data in ts-project

ratik21:
Updated implementation to load `.ts` files in a typescript-project.


packages/algob/src/builtin-tasks/deploy.ts
```diff
@@ -35,10 +35,10 @@ export function loadFilenames (directory: string, taskType?: string): string[] {
...
   }
 
-  const jsFiles = glob.sync(path.join(directory, "*.js")).sort(cmpStr);
-  if (jsFiles.length) { return jsFiles; }
-  // if not js file, return the compiled ts files from ./build
-  return glob.sync(path.join("build", directory, "*.js")).sort(cmpStr);
+  // return .js, .ts files from /scripts directory
+  return glob.sync(path.join(directory, "*.js"))
+    .concat(glob.sync(path.join(directory, "*.ts")))
+    .sort(cmpStr);
```
amityadav0, line 41: what is the difference between previous and current implementation ?
i think we added the code to support `htlc-ts-project`.

packages/algob/src/internal/util/scripts-runner.ts
```diff
@@ -40,6 +40,11 @@ export async function runScript (
   runtimeEnv: AlgobRuntimeEnv,
   deployer: AlgobDeployer
 ): Promise<void> {
+  // if .ts file is encountered, load from /build/scripts/file.ts
+  if (relativeScriptPath.endsWith('.ts')) {
+    relativeScriptPath = path.join('build', relativeScriptPath.split('.ts')[0] + '.js');
+  }
```
amityadav0, line 46: we are doing loading of js files here ?
why were we getting `undefined` error due to this ?

packages/algob/src/builtin-tasks/deploy.ts
ratik21: By the previous implementation we were 'changing' the relative paths. Eg. `/scripts/file.ts` -> `/build/scripts/file.js`, which was affecting the `checkpoints`, `txWriter` as well. Since the path was changed everything was relative to `/build` (the checkpoints were being saved in `/artifacts/build/scripts/..`, which we don't want. (As we're handling loading the file from build internally). Also since the cp path was changed, `getSSC`, `.asa.get()` were returning `undefined`, which was the bug. So i decided to change the implementation. 

By the current one, we're not changing the path, instead we're loading directly ts, js files (which now returns `[/scripts/file.ts, ..`] instead of [`/build/scripts/file.js, ..`]. This ensures that the paths are correct (with cp, txwriter as well), checkpoints are also stored in `/artifacts/scripts/file.ts.cp.yaml`. We only load the script from build when we need to run it. If build is not present, it throws the error with stacktrace that `/build/scripts/file..` is not found (as we suggest in the readme). 

Please let me know if that makes sense.

packages/algob/src/internal/util/scripts-runner.ts
ratik21: Yes, as explained above we are loading the js files here (during run only), without changing the relative path in the beginning (in `loadFileNames`). 
No, we were not getting `undefined` because of this, this is kept here because of changed implementation. The other way (keeping the previous implementation) would be to handle loading from `/build` at all other places (as we're changing the script path in the deployer). Getting `undefined` bug was here - [https://github.com/scale-it/algo-builder/blob/7354b021f5a6424593e13cd4dfe591100b8e224c/packages/algob/src/builtin-tasks/run.ts#L94](https://github.com/scale-it/algo-builder/blob/7354b021f5a6424593e13cd4dfe591100b8e224c/packages/algob/src/builtin-tasks/run.ts#L94). We can fix the bug by updated the line above, but i think changing and handling path everywhere is not good, ad could maybe lead to more bugs. What do you think?

packages/algob/src/builtin-tasks/deploy.ts
amityadav0: makes sense

packages/algob/src/internal/util/scripts-runner.ts
```diff
@@ -40,6 +40,11 @@ export async function runScript (
   runtimeEnv: AlgobRuntimeEnv,
   deployer: AlgobDeployer
 ): Promise<void> {
+  // if .ts file is encountered, load from /build/scripts/file.ts
```
amityadav0, line 43:
```suggestion
  // if .ts file is encountered, load from /build/scripts/file.js
```

packages/algob/src/internal/util/scripts-runner.ts
amityadav0: yes, i think keeping it as single is less prone to bugs.

packages/algob/src/builtin-tasks/deploy.ts
```diff
@@ -35,10 +35,10 @@ export function loadFilenames (directory: string, taskType?: string): string[] {
     }
   }
 
-  const jsFiles = glob.sync(path.join(directory, "*.js")).sort(cmpStr);
-  if (jsFiles.length) { return jsFiles; }
-  // if not js file, return the compiled ts files from ./build
-  return glob.sync(path.join("build", directory, "*.js")).sort(cmpStr);
+  // return .js, .ts files from /scripts directory
```
amityadav0, line 38: Please add this comment as doc string of function

amityadav0 (approved):
Thanks for the PR
Please update the last line of htlc-ts-project readme, from `js` to `ts`.

packages/algob/src/builtin-tasks/deploy.ts
ratik21: ok

PR #265, opened 2021-03-19
Algob README update and typedoc entrypoint

robert-zaremba:


packages/algob/README.md
```diff
@@ -1,10 +1,15 @@
 # Builder
 
-This package provides the main entry point into the application.
+Algob is the CLI (Command Line Interface) for Algo Builder. Think about it as a Ethereum Truffle but for Algorand projects.
```
ratik21, line 3:
```suggestion
Algob is the CLI (Command Line Interface) for Algo Builder. Think about it as an Ethereum Truffle but for Algorand projects.
```

ratik21 (approved):
Thanks for the update.

PR #268, opened 2021-03-24
Fix bug, Add typedoc pages plugin, custom README for docs

man-of-code:


package.json
```diff
@@ -64,6 +64,7 @@
     "nyc": "^15.1.0",
     "ts-node": "^9.0.0",
     "typedoc": "^0.20.32",
+    "typedoc-plugin-pages": "^1.1.0",
```
ratik21, line 67: Why are we just adding a dependency here? Please remove it. You can create another PR or update this one where we're using `typedoc-plugin-pages`.

man-of-code: There is some error of configurations, I have the pagesconfig ready,  I just need to properly configure typedoc... I'll update the PR once it's done.

PR #273, opened 2021-03-26
Updata /asa example: show how to create ASA by a smart contract (stateless)

amityadav0:


packages/algob/src/internal/deployer.ts
```diff
@@ -222,6 +222,55 @@ export class DeployerDeployMode extends DeployerBasicMode implements AlgobDeploy
     return asaInfo;
   }
 
+  /**
+   * Deploys ASA owned by smart contract
+   * @param name ASA name
+   * @param lsig logic signature
+   * @param payFlags Transaction Parameters
```
robert-zaremba, line 229:
```suggestion
   * Deploys ASA owned by a smart contract
   * @param name: ASA name
   * @param lsig: logic signature
   * @param payFlags: Transaction Parameters
```

packages/algob/src/lib/algo-operator.ts
```diff
@@ -214,6 +218,41 @@ export class AlgoOperatorImpl implements AlgoOperator {
     };
   }
 
+  /**
+   * Deploys ASA owned by smart contract
+   * @param name ASA name
+   * @param asaDef ASA definition read from asa.yaml
+   * @param flags deployment flags
+   * @param txWriter transaction writer object
+   * @param lsig logic signature of smart contract
```
robert-zaremba, line 227:
```suggestion
   * Deploys ASA owned by a smart contract
   * @param name: ASA name
   * @param asaDef: ASA definition read from asa.yaml
   * @param flags: deployment flags
   * @param txWriter: transaction writer object
   * @param lsig: logic signature of smart contract
```

examples/asa/scripts/transfer/contract-to-alice.js
```diff
@@ -0,0 +1,38 @@
+/**
+ * Description:
+ * This file demonstrates the example to transfer contract owned ASA's
+ * from contract account (lsig) to a owner account.
```
robert-zaremba, line 4:
```suggestion
 * This file demonstrates the example to transfer contract owned ASA
 * from a contract account (lsig) to an owner account.
```

examples/asa/scripts/3-contract-owned-asa.js
```diff
@@ -0,0 +1,33 @@
...
+/**
+* only owner can call - update smart contract to set a new owner
+* only owner can transfer ASA out from the smart contract and maximum payment is 100 ASA.
+
+let's put this example in `examples/asa` and:
+* update README
+* update documentation
+* if needed - let's create a highl level function for this (similar to `opIn*` functions).
+ * @param {*} runtimeEnv
+ * @param {*} deployer
```
robert-zaremba, line 13: Please update the comment. There shouldn't be "let's put this example in ..."

examples/asa/assets/5-contract-owned-asa.teal
```diff
@@ -0,0 +1,51 @@
+#pragma version 2
+// Check if Transaction type is axfer(asset transfer), asset amount is <= 100 and 
+// only owner can transfer ASA out from contract
+// Also check for asset creation transaction.
```
robert-zaremba, line 4:
```suggestion
// Approve only the following transaction: 
// * axfer (asset transfer): only owner can transfer ASA out from the contract, asset amount must be <= 100
// * asset creation transaction.
```

examples/asa/assets/5-contract-owned-asa.teal
```diff
@@ -0,0 +1,51 @@
...
+txn TypeEnum
+int 4
```
robert-zaremba, line 6: Is it possilbe to use contants / enums instead of numbers? (4 above, other numbers below)
If not then please add a comment to what 4 is (and below to what 3 is).

Also, maybe we should limit the Creation of ASA? Is it possible to check that it also has to be signed by the owner? 

+ Please update README and comment in the top of this smart contract.

robert-zaremba (commented):
Left few comments. Notably:
* Also, maybe we should limit the Creation of ASA? Is it possible to check that it also has to be signed by the owner?

examples/asa/assets/5-contract-owned-asa.teal
amityadav0: Yes, we can use `axfer` and `acfg` , adding a comment,

what do you mean by limit creation ? limit per account ?

I tried that by delegating the signature with owner's sign, but we cannot do that, if we do that we can only send ASA from owner's account by validating lsig, and not from lsig account. when i tried to do later one i got logic signature validation failed

OK

examples/asa/scripts/3-contract-owned-asa.js
amityadav0: Yes, removing

examples/asa/assets/5-contract-owned-asa.teal
robert-zaremba: > what do you mean by limit creation ? limit per account ?

Yes - that only one account can create contract ASA.

robert-zaremba: @amityadav0  - try to ask on Discord Algorand channel how to do it.

amityadav0: yes asked on discord and updated example.
but for co-signed we need to group transaction, added task for it

man-of-code (commented):
LGTM. Didn't run this yet.

Also, shouldn't the function `deployContractASA` be documented in `deployer.md` ?

man-of-code (approved):
Pre approving as it is working fine. Just had a question, can you please update the [documentation](https://github.com/scale-it/algo-builder/tree/master/examples/asa#readme).

Also, Is this a general constraint of Algorand Blockchain, that we can only deploy 1 ASA from a Lsig account?

examples/asa/assets/5-contract-owned-asa-approval.py
```diff
@@ -0,0 +1,80 @@
+from pyteal import *
```
ratik21, line 1: Insteal of using *contract-owned-asa* in file name, can we just use *contract-asa* ?

examples/asa/assets/5-contract-owned-asa-approval.py
```diff
@@ -0,0 +1,80 @@
...
+    # contains a rekey
+    rekey_check = And(
+        Gtxn[0].rekey_to() == Global.zero_address(),
+        Gtxn[1].rekey_to() == Global.zero_address()
+    )
+
+    on_asa_create = Seq([
+        Assert(And(
+            App.globalGet(Bytes("Creator")) == Txn.sender(),
+            App.globalGet(Bytes("ASACounter")) == Int(0), # limit no of asa from here: currently set to max 1 ASA.
```
ratik21, line 28: Shouldn't this be `App.globalGet(Bytes("ASACounter")) <= Int(1)` (If max 1 asa creation is allowed) ? If not then please explain this in comment

examples/asa/assets/5-contract-owned-asa-approval.py
```diff
@@ -0,0 +1,80 @@
...
+            Global.group_size() == Int(3),
+            Gtxn[1].type_enum() == TxnType.AssetConfig,
+            Gtxn[2].type_enum() == TxnType.Payment,
+            rekey_check,
+            Gtxn[2].rekey_to() == Global.zero_address(),
+            Gtxn[2].amount() == Int(1000000)
+        )),
+        App.globalPut(Bytes("ASACounter"), Int(1)),
```
ratik21, line 36: If we are keeping a counter here, then shouldn't we do counter++? 
`App.globalPut(Bytes("ASACounter"), App.globalGet(Bytes("ASACounter")) + Int(1))`

examples/asa/assets/5-contract-owned-asa-approval.py
```diff
@@ -0,0 +1,80 @@
...
+            App.globalGet(Bytes("Creator")) == Gtxn[1].asset_receiver(),
+            rekey_check,
+        )),
+        Return(Int(1))
+    ])
+
+    # Change owner of the contract if called by current owner
+    change_owner = Seq([
+        Assert(And(
+            Txn.application_args[0] == Bytes("changeowner"),
```
ratik21, line 52: How about changing the param to "change_owner" as well ?

examples/asa/assets/5-contract-owned-asa-approval.py
```diff
@@ -0,0 +1,80 @@
...
+            App.globalGet(Bytes("Creator")) == Txn.sender()
+        )),
+        App.globalPut(Bytes("Creator"), Txn.application_args[1]),
```
ratik21, line 55: Please add one more asset for `Txn.application_args.length() == Int(2)`

ratik21, line 55: Also let's put in a note that with *TEALv3* we can get address from `appAccounts` as well (Using `Txn.Accounts[n]`). So user can also use appAccounts as well, instead of passing account address in appArgs.

examples/asa/assets/5-contract-owned-asa-approval.py
```diff
@@ -0,0 +1,80 @@
...
+
+    # Check if current transaction is asset creation of asset transfer
+    on_call = Cond(
+        [Global.group_size() == Int(3), on_asa_create],
+        [Global.group_size() == Int(2), on_pay]
+    )
+
+    program = Cond(
+        [Txn.application_id() == Int(0), on_creation],
+        [Txn.application_args.length() == Int(2), change_owner],
```
ratik21, line 67: I think for branching we should check for the first param, and assert for length in that branch. That way seems more intuitive. What  do you think?

examples/asa/assets/5-contract-owned-asa.py
```diff
@@ -0,0 +1,62 @@
+# Escrow Account Contract
```
ratik21, line 1: Can we rename/add a suffix to file name `-stateless` as well? Eg. *5-contract-asa-stateless.py*

examples/asa/assets/5-contract-owned-asa.py
```diff
@@ -0,0 +1,62 @@
...
+        # The first transaction must be 
+        # an ApplicationCall (ie call stateful smart contract)
+        Gtxn[0].type_enum() == TxnType.ApplicationCall,
+
+        # The specific App ID must be called
+        # This should be changed after creation
+        Gtxn[0].application_id() == Int(app_id),
+
+        Gtxn[1].type_enum() == TxnType.AssetTransfer,
+        Gtxn[1].amount() <= Int(100),
```
ratik21, line 29: So this is an inforced condition as well (along with max 1 asa creation) ?  Could you please add in comment at the top of file (like we do in approval.py)?

examples/asa/scripts/3-contract-owned-asa.js
```diff
@@ -0,0 +1,77 @@
...
+      lsig: statelessAccount,
+      payFlags: {}
+    },
+    // Payment of 1 algo signed by alice
+    {
+      type: types.TransactionType.TransferAlgo,
+      sign: types.SignType.SecretKey,
+      fromAccount: alice,
+      toAccountAddr: statelessAccount.address(),
+      amountMicroAlgos: 1000000,
```
ratik21, line 61:
```suggestion
      amountMicroAlgos: 1e6,
```

examples/asa/scripts/3-contract-owned-asa.js
```diff
@@ -0,0 +1,77 @@
...
+      sign: types.SignType.SecretKey,
+      fromAccount: alice,
+      appId: sscInfo.appID,
+      payFlags: {}
+    },
+    // Asset creation
+    {
+      type: types.TransactionType.DeployASA,
+      sign: types.SignType.LogicSignature,
+      fromAccount: { addr: statelessAccount.address() },
```
ratik21, line 50: this should be `fromAccountAddr: statelessAccount.address()` now. Could you please pull from latest master & update?

examples/asa/scripts/3-contract-owned-asa.js
```diff
@@ -0,0 +1,77 @@
+/**
+ * Description:
+ * This file demonstrates how to create ASA owned by
+ * smart contract account(stateless).
```
ratik21, line 4: Please also add steps here. Eg;
```
1. Deploy SSC (controls asa)
2. Create contract account (with ssc app_id embedded/passed as a template param)
3. Deploy ASA
```

examples/asa/scripts/transfer/contract-to-alice.js
```diff
@@ -0,0 +1,79 @@
...
+    // It should fail
+    txGroup[0].fromAccount = bob;
+    await executeTransaction(deployer, txGroup);
+  } catch (e) {
+    console.log(e);
+  }
+
+  try {
+    // try to send asset directly without calling stateful
+    // It should fail
```
ratik21, line 63:
```suggestion
    // tx FAIL: trying to send asset directly without calling stateful smart contract
```

examples/asa/scripts/transfer/contract-to-alice.js
```diff
@@ -0,0 +1,79 @@
...
+    }
+  ];
+
+  await executeTransaction(deployer, txGroup);
+  // print assetHolding of alice
+  await balanceOf(deployer, alice.addr, assetID);
+
+  try {
+    // try to receice asset from another account
+    // It should fail
```
ratik21, line 54:
```suggestion
    // tx FAIL: trying to receive asset from another account
```

examples/asa/scripts/transfer/contract-to-alice.js
```diff
@@ -0,0 +1,79 @@
...
+    txGroup[0].fromAccount = bob;
```
ratik21, line 55: So that means that asset can be received only from contract account(creator) ?

ratik21: What if i do a simple asset transfer transaction (using my `sk`)? That will pass right (If yes then please add a note here) ?

examples/asa/assets/5-contract-owned-asa-approval.py
amityadav0: No, because if you dry run this code, on first iteration it will create Asa therefore global will be 1 ( even if you add a counter + 1) , in next iteration it will still pass according to your logic, therefore we have created 2 ASAs now.

examples/asa/assets/5-contract-owned-asa-approval.py
amityadav0: No, for max 1 Asa, both works

examples/asa/assets/5-contract-owned-asa-approval.py
amityadav0: I think it is same, you can check whichever condition first.

examples/asa/assets/5-contract-owned-asa-approval.py
amityadav0: Already added

examples/asa/scripts/transfer/contract-to-alice.js
amityadav0: What do you mean ?

examples/asa/assets/5-contract-owned-asa-approval.py
amityadav0: i think this file is written in TEALv2

examples/asa/assets/5-contract-owned-asa-approval.py
amityadav0: yes

examples/asa/scripts/3-contract-owned-asa.js
amityadav0: ok

examples/asa/assets/5-contract-owned-asa-approval.py
ratik21: Then can we do it this way:
```py
max_assets = Int(1)
...
on_asa_create = Seq([
  App.globalGet(Bytes("ASACounter")) == max_assets - Int(1) // explain here why we subtract 1 here (as ASA is already deployed)
  ..
])
```
If this way is possible then tt's easier to understand, and is more flexible (user can just update max_assets to Int(2) if (s)he wishes limit more than 1 ASA)

examples/asa/assets/5-contract-owned-asa-approval.py
ratik21: Why do we call it a counter if it will always be max 1 Asa? Say if user wants max 2 asa, this will again update counter to 1, on 2nd asa deploy

ratik21: Why do we call it a counter if it will always be max 1 Asa? Say if user wants max 2 asa, this will again update counter to 1, on 2nd asa deploy

examples/asa/assets/5-contract-owned-asa-approval.py
ratik21: Agree that it's same, i just think it's more intuitive to do a jump on checking `txn.arg[0] == "some_arg"`, rather than jump on args.length()

examples/asa/assets/5-contract-owned-asa-approval.py
ratik21: To write in *TEALv3*, we can just pass version = 3, on `compileTeal(..)`. But since we do not support it here i just suggested adding a note.

examples/asa/assets/5-contract-owned-asa-approval.py
amityadav0: then user can update this counter, but this example is for 1 ASA. but can update if you want

ratik21: I think it's better to use the suggested code. Currently it looks ambigous

amityadav0:
- Updated Readme
- ASA Creation Limit is set
- Co-signed by owner as well

PR #276, opened 2021-03-31
[issue] AssetHolding doesn't work properly

### LINKED ISSUE #274: AssetHolding doesn't work properly
Error is `Error: RUNTIME_ERR1000: Length of stack is less than min length required for current op at line 21` during Runtime test, during normal deploy the error in the transaction is `TEAL runtime encountered err opcode` and the code I'm testing is 
```teal
#pragma version 2
txn ApplicationID
int 0
==
bnz l0
txn OnCompletion
int CloseOut
==
bnz l1
txn OnCompletion
int OptIn
==
bnz l2
err
l0:
int 1
txna ApplicationArgs 1
btoi
asset_holding_get AssetBalance
store 0
store 1
txn NumAppArgs
int 2
==
load 0
&&
bnz l6
err
l6:
byte "dispenser_holding"
load 1
app_global_put
b l5
l1:
int 1
b l5
l2:
int 0
byte "balance"
int 0
app_local_put
int 1
b l5
l5:
``` 
The code was tested in goal and it works properly, it seems to drop on the 
```
store 0
store 1
```

ratik21:
- fix `runtime.addApp`
- Update Asset ops according to docs

Closes #274

### DIFF
--- a/packages/runtime/src/runtime.ts
+++ b/packages/runtime/src/runtime.ts
@@ -428,21 +428,13 @@ export class Runtime {
     if (clearProgram === "") {
       throw new RuntimeError(RUNTIME_ERRORS.GENERAL.INVALID_CLEAR_PROGRAM);
     }
+
+    // prepare context
+    this.ctx.state = cloneDeep(this.store);
     // create app with id = 0 in globalApps for teal execution
     const app = senderAcc.addApp(0, flags, approvalProgram, clearProgram);
     this.ctx.state.accounts.set(senderAcc.address, senderAcc);
     this.ctx.state.globalApps.set(app.id, senderAcc.address);
-
-    for (const [address, account] of this.store.accounts) {
-      this.ctx.state.accounts.set(address, account);
-      for (const appId of account.createdApps.keys()) {
-        this.ctx.state.globalApps.set(appId, account.address);
-      }
-
-      for (const assetId of account.createdAssets.keys()) {
-        this.ctx.state.assetDefs.set(assetId, account.address);
-      }
-    }
 
     this.addCtxAppCreateTxn(flags, payFlags);
     this.run(approvalProgram, ExecutionMode.STATEFUL); // execute TEAL code with appId = 0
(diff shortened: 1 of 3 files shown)


packages/runtime/src/runtime.ts
```diff
@@ -433,6 +433,17 @@ export class Runtime {
     this.ctx.state.accounts.set(senderAcc.address, senderAcc);
     this.ctx.state.globalApps.set(app.id, senderAcc.address);
 
+    for (const [address, account] of this.store.accounts) {
```
amityadav0, line 436: why do we need it here ?

packages/runtime/src/runtime.ts
```diff
@@ -433,6 +433,17 @@ export class Runtime {
...
+      this.ctx.state.accounts.set(address, account);
+      for (const appId of account.createdApps.keys()) {
+        this.ctx.state.globalApps.set(appId, account.address);
+      }
+
+      for (const assetId of account.createdAssets.keys()) {
+        this.store.assetDefs.set(assetId, account.address);
```
amityadav0, line 443: why are we are operating on `store` before running code ?

packages/runtime/src/runtime.ts
amityadav0: we already have this in initializeAccounts or is there any other reason ?

ratik21: > why do we need it here ?

Because earlier we were only preparing `ctx` in senderAcc. But we should encapsulate all created assets/apps before run. Because if an opcode during deploy requires information from external account, then that account should be present in `ctx` as well. Please let me know if that makes sense.

> we already have this in initializeAccounts or is there any other reason ?

In `initializeAccounts` we are preparing store, here we are preparing `ctx`(before run).

packages/runtime/src/runtime.ts
ratik21: missed this one. It should be `ctx`

packages/runtime/src/runtime.ts
amityadav0: how about removing lines 436 to 446.
and add `this.ctx.state = cloneDeep(this.store);` at 430.
This does the same thing and is easily understandable.

Checked on my system test run properly.

ratik21: yes, good idea

amityadav0 (approved):

PR #283, opened 2021-04-08
add deployer.signMultiSig for logic signature

ratik21:
- add function to append signature to multisig-lsig OR create a new lsig.multisig


packages/algob/src/internal/deployer.ts
```diff
@@ -142,7 +142,25 @@ class DeployerBasicMode {
...
-   * Description: Opt-In to ASA for a single account. The opt-in transaction is
+   * Appends signature (using signer's sk) to multi-signed logic signature. If multisig is not found
+   * then new multisig is created
+   * eg. appending own signature to a signed lsig (received from multisignature account address network)
+   * @param lsig Logic Sig object
+   * @param signer: Signer Account which will sign the smart contract
+   * @param mparams: passed when signing a new multisig
+   * @returns multi signed logic signature (with appended signature from signer's sk)
+   */
+  signMultiSig (lsig: LogicSig, signer: rtypes.Account, mparams?: MultisigMetadata): LogicSig {
```
man-of-code, line 153: This function only signs stateless smart contracts. We should change this name to something more appropriate. 

This name can be misleading, as we might be adding a similar function for general transactions too.

packages/types-algosdk/types/index.d.ts
```diff
@@ -195,10 +195,25 @@ declare module 'algosdk' {
     v: number;
   }
 
-  export interface MultiSigAccount {
+  /**
+   * Src: https://github.com/algorand/js-algorand-sdk/blob/develop/src/types/multisig.ts#L6
+   */
+  // TODO: use latest typings from SDK
+  export interface MultisigMetadata {
```
man-of-code, line 202: There is a very similiar interface `MultiSigAccount`. 

Also, if we absolutely need this interface then maybe we can change this to `MultiSigParams` and also use this at lines 362, 469, etc.

ratik21: > There is a very similiar interface MultiSigAccount.

Replaced this because SDK uses `MultiSigAccount`. We'll soon be migrating to ts-sdk.

packages/algob/src/internal/deployer.ts
ratik21: Could you please suggest something?

examples/multisig/scripts/multisig_goal_sc.js
```diff
@@ -37,6 +37,10 @@ async function run (runtimeEnv, deployer) {
 
   const lsig = await deployer.loadMultiSig('sample-raw-asc.blsig');
 
+  // use below line to append bob signature to loaded multisig from file.
+  // note: not required in this case as loaded lsig has 2 signtatures & msig threshold is also 2
+  // deployer.signMultiSig(lsig, john);
+
```
amityadav0, line 43:
```suggestion
```

packages/algob/src/internal/deployer.ts
```diff
@@ -142,7 +142,25 @@ class DeployerBasicMode {
...
+    if (lsig.msig === undefined) { // if multisig not found, create new msig
+      lsig.sign(signer.sk, mparams);
```
amityadav0, line 155: what will happen if `mparams` are not passed ?

amityadav0: and lsig is also not defined

packages/algob/src/internal/deployer.ts
```diff
@@ -347,7 +368,9 @@ export class DeployerDeployMode extends DeployerBasicMode implements Deployer {
   }
 }
 
-// This class is what user interacts with in run task
+/**
+ * This class is what user interacts with in run task
```
amityadav0, line 372:
```suggestion
 * This class is what user interacts with in run task mode
```

packages/types-algosdk/types/index.d.ts
```diff
@@ -195,10 +195,25 @@ declare module 'algosdk' {
...
+    /**
+     * Multisig version
+     */
     version: number;
+
```
amityadav0, line 207:
```suggestion
```

packages/types-algosdk/types/index.d.ts
```diff
@@ -195,10 +195,25 @@ declare module 'algosdk' {
...
+    /**
+     * Multisig threshold value. Authorization requires a subset of signatures,
+     * equal to or greater than the threshold value.
+     */
     threshold: number;
-    // array of base32 encoded addresses
+
```
amityadav0, line 213:
```suggestion
```

packages/types-algosdk/types/index.d.ts
```diff
@@ -195,10 +195,25 @@ declare module 'algosdk' {
...
+    /**
+     * A list of Algorand addresses representing possible signers for this multisig. Order is important.
```
amityadav0, line 215: here you have mentioned order is important but above in threshold, it is mentioned we require a subset of signatures, 
why order here is important ?

packages/algob/src/internal/deployer.ts
amityadav0: maybe `signLogicSigMultiSig` or `signLsigMultiSig`

packages/types-algosdk/types/index.d.ts
ratik21: [https://developer.algorand.org/docs/features/accounts/create/#multisignature](https://developer.algorand.org/docs/features/accounts/create/#multisignature)

packages/algob/src/internal/deployer.ts
ratik21: adding check for `mparams`. 
I would expect lsig to be passed (it is required), as what will user sign if lsig is not defined. Should we check for lsig as well?

amityadav0: Sorry, I meant lsig.msig

ratik21: if `msig` is undefined then it means no account from multisig hash has signed this lsig yet. So "create" the msig in this case (hence we require the *mparams* in order). 

If lsig is already signed (by atleast one account of multisig) then msig should be present and we "append" our own signature to that msig (using `lsig.appendToMultisig(..)`).

packages/algob/src/internal/deployer.ts
ratik21: `signLogicSigMultiSig` works

man-of-code (approved):
Working fine.

PR #287, opened 2021-04-12
Update ASA deployment to pass custom parameters

ratik21:


docs/guide/deployer.md
```diff
@@ -27,6 +27,22 @@ You can write deployment tasks synchronously and they'll be executed in the corr
 
 To deploy an ASA you must have `asa.yaml` file in `assets` folder. Head to the [ASA Definition File Spec](https://paper.dropbox.com/doc/Algorand-builder-specs-Vcdp0XNngizChyUWvFXfs#:uid=077585002872354521007982&h2=ASA-Definition-File) to learn more.
 
+#### Passing custom ASA parameters
```
robert-zaremba, line 30:
```suggestion
#### Deploying ASA with custom parameters
```

docs/guide/deployer.md
```diff
@@ -27,6 +27,22 @@ You can write deployment tasks synchronously and they'll be executed in the corr
...
+
+User can also pass custom ASA params to `deployASA` function. This will override the fields in `assets/asa.yaml`. Eg. If user wants to use a multisig address from 3 accounts: `alice, bob, john` as the Asset Reserve. We can create a multisig address in `algob` script first and then pass this address as a custom asaParams.
```
robert-zaremba, line 32:
```suggestion
User can also override the fields in `assets/asa.yaml` when deploying ASA. Eg. If user wants to use a multisig address from 3 accounts: `alice, bob, john` as the Asset Reserve. We can create a multisig address in `algob` script first and then pass this address as a custom asaParams.
```

packages/runtime/src/lib/asa.ts
```diff
@@ -14,57 +14,81 @@ const ASSETS_DIR = "assets";
...
     }
     throw e;
   }
 }
 
+/**
+ * Override & validate ASA definition in asa.yaml using custom params passed via deployASA
+ * @param accounts accounts by name
+ * @param sourceDef source asset definition (in asa.yaml)
+ * @param customDef custom asset def params (passed during ASA deployment)
```
robert-zaremba, line 67: sourceDef and customDef  doesn't sound very clear. How about renaming to: `origDef` and `newDef`? Or maybe something else?

robert-zaremba (approved):
pre approving. Left few suggestions.

robert-zaremba (commented):
-

robert-zaremba (changes requested):
I forgot about one thing. Let's say we:

    Deploy ASA with overwritten fields.
    load ASA to perform an action: eg: transfer from my account to asa.manager account. What asa.manager account will be used? The original one or the overwritten one?

We must use the one from checkpoint - with overwritten fields. Please make tests for it.

packages/runtime/src/lib/asa.ts
ratik21: `origDef` & `newDef` sounds more clear.

docs/guide/deployer.md
```diff
@@ -27,6 +27,22 @@ You can write deployment tasks synchronously and they'll be executed in the corr
 
 To deploy an ASA you must have `asa.yaml` file in `assets` folder. Head to the [ASA Definition File Spec](https://paper.dropbox.com/doc/Algorand-builder-specs-Vcdp0XNngizChyUWvFXfs#:uid=077585002872354521007982&h2=ASA-Definition-File) to learn more.
 
+#### Deploying ASA with custom parameters
+
+User can also override the fields in `assets/asa.yaml` when deploying ASA. Eg. If user wants to use a multisig address from 3 accounts: `alice, bob, john` as the Asset Reserve. We can create a multisig address in `algob` script first and then pass this address as a custom asaParams.
```
amityadav0, line 32: are we overwriting fields in .yaml ? because that's the current deployed asa.
Please update `override`

docs/guide/deployer.md
```diff
@@ -27,6 +27,22 @@ You can write deployment tasks synchronously and they'll be executed in the corr
...
+
+```javascript
+const { createMsigAddress } = require('@algo-builder/algob');
+
+const addrs = [alice.addr, bob.addr, john.addr];
+const [mparams, multsigaddr] = createMsigAddress(1, 2, addrs); // version = 1, threshold = 2
+
+// while deploying ASA pass custom asa param
+await deployer.deployASA("ASA-2", {...}, { reserve: multsigaddr }); // this will overwrite reserve field in assets/asa.yaml
```
amityadav0, line 41: how about merging it with ASADeployment flags ? this seems as extra variable here, adding them in ASADeployment flags seems clear to me

packages/algob/src/internal/deployer.ts
```diff
@@ -224,8 +224,11 @@ export class DeployerDeployMode extends DeployerBasicMode implements Deployer {
     }
   }
 
-  async deployASA (name: string, flags: rtypes.ASADeploymentFlags): Promise<ASAInfo> {
-    if (this.loadedAsaDefs[name] === undefined) {
+  async deployASA (name: string, flags: rtypes.ASADeploymentFlags,
+    asaParams?: Partial<rtypes.ASADef>): Promise<ASAInfo> {
+    const asaDef = overrideASADef(this.accountsByName, this.loadedAsaDefs[name], asaParams);
```
amityadav0, line 229: how about overwrite here ? because now ASADEF will give you last deployed ASA information

packages/runtime/test/src/lib/asa.ts
```diff
@@ -95,7 +95,7 @@ describe("ASA parser", () => {
         total: 0xFFFFFFFFFFFFFFFFn + 5n,
         decimals: 0
       }
-    };
+    } as any;
```
amityadav0, line 98: please update this to appropriate type

packages/algob/src/internal/deployer.ts
amityadav0, line 229: how about overwrite here ? because now ASADEF will give you last deployed ASA information

docs/guide/deployer.md
ratik21: I think overwriting means that we overwrite .yaml as well. Override means that we override the input from .yaml file with custom `asaParams`. So not sure if this should be overwrite

docs/guide/deployer.md
ratik21: I think passing it as a separate field is better. As `asaDef` has many fields so passing separately looks better to me. But if that doesn't better we can pass in deployment  flags only. let me know.

packages/algob/src/internal/deployer.ts
ratik21: Yes here we can use overwrite (and actually "overwrite" `loadedAsaDefs`). Currently i just return a copy.

ratik21: same comment add twice

packages/runtime/test/src/lib/asa.ts
ratik21: This is fine. It's a test. Actually this was added because in previous implementation were using `Object` as type which is not good so i updated it - [https://github.com/scale-it/algo-builder/pull/287/files#diff-b7dea8e4c54f55543c05d4ed625c45ae0f9df28113e9406df19d814ba25c82bbL46](https://github.com/scale-it/algo-builder/pull/287/files#diff-b7dea8e4c54f55543c05d4ed625c45ae0f9df28113e9406df19d814ba25c82bbL46).

That's why had to add type `any` here. In this test we are validating the field the "total". If to use appropriate type we will need to add all other fields in the definition (which aren't related to test). Let me know if you think we should do that.

docs/guide/deployer.md
amityadav0: That's why I mentioned to overwrite

packages/algob/src/internal/deployer.ts
amityadav0: What I meant here is update yaml file to be overwritten by updated value. Because override doesn't make sense here.
If user wants to know what was deployed he would know by looking at asadef

docs/guide/deployer.md
ratik21: You're suggesting to update the implementation to "overwrite" asa.yaml as well. OK.
By your previous comment it seams you suggest to update the readme only.

docs/guide/deployer.md
robert-zaremba, line 41:
```suggestion
await deployer.deployASA("ASA-2", {...}, { reserve: multsigaddr }); // this will overwrite reserve field from assets/asa.yaml
```

examples/asa/scripts/0-gold-asa.js
```diff
@@ -34,6 +34,10 @@ async function run (runtimeEnv, deployer) {
     // feePerByte: 100,
     // firstValid: 10,
     // validRounds: 1002
+  }, {
+    reserve: bob.addr // pass asa fields here to override asa def in asa.yaml
```
robert-zaremba, line 38:
```suggestion
    reserve: bob.addr // override default value set in asa.yaml
```

packages/algob/src/internal/deployer.ts
```diff
@@ -224,19 +235,22 @@ export class DeployerDeployMode extends DeployerBasicMode implements Deployer {
...
+    if (asaDef === undefined) {
       persistCheckpoint(this.txWriter.scriptName, this.cpData.strippedCP);
       throw new BuilderError(
         ERRORS.BUILTIN_TASKS.DEPLOYER_ASA_DEF_NOT_FOUND, {
           asaName: name
         });
     }
     this.assertNoAsset(name);
-    let asaInfo = {} as any;
+    let asaInfo = {} as ASAInfo;
```
robert-zaremba, line 250: :+1:

packages/algob/test/internal/deployer.ts
```diff
@@ -112,18 +114,76 @@ describe("DeployerDeployMode", () => {
...
+      freeze: fixedAccount.addr
+    };
+
+    assert.deepEqual(asaInfo, {
+      creator: "addr-1-get-address-dry-run",
+      txId: "tx-id-dry-run",
+      confirmedRound: -1,
+      assetIndex: -1,
+      assetDef: expectedASADef
+    });
```
robert-zaremba, line 177: Let's add in this test the following check
* recreate `deployer`
* check that `deployer.asa.get("MY_ASA")` equals to `expectedASADef`.

robert-zaremba (approved):
pre-approving - left one test request.

packages/algob/test/internal/deployer.ts
ratik21: Added.

PR #288, opened 2021-04-12
Changed transaction file extension from .tx to .txn

man-of-code:


amityadav0 (commented):
I think in discord Robert mentioned that we should be able to load any type file (.txt or .txn)
this should be independent of that

man-of-code:
> I think in discord Robert mentioned that we should be able to load any type file (.txt or .txn)
> this should be independent of that

Yes. You are right. But this PR is different. 
Robert mentioned that the command `sign-multisg` should accept any extension for the input transaction file. This PR is just changing the already existing functions which deal with `.tx` files.

amityadav0:
Function you have updated in this PR is sending transaction to network, please update it to use any file extension. Let's discuss on discord if you have questions

examples/signed-txn/README.md
```diff
@@ -14,17 +14,16 @@ Please follow the [setup](../README.md) instructions to install dependencies and
...
-goal clerk send -a <amount> -f <from_account> -t <to_account> -o out.tx -d ~/<data_directory>
+goal clerk send -a <amount> -f <from_account> -t <to_account> -o out.txn -d ~/<data_directory>
 ```
 Sign transaction (signed by `<from_account>`)
 ```bash
 goal clerk sign -i [input file] -o [output file] -d ~/<data_directory>
 ```
 
 You need to ***save the signed transaction file in `examples/multisig/assets` directory***.
- The file must have ***.tx*** extension.
```
ratik21, line 27: Let's add a note here, that extension doesn't necessarily have to be *.txn*. Only that file should have a valid signed tx.

packages/algob/src/lib/tx.ts
```diff
@@ -182,7 +182,7 @@ export async function executeTransaction (
  * firstValid and lastValid which might not be equal to the
  * current network's blockchain block height.
  * @param deployer Deployer
- * @param fileName raw signed txn .tx file
+ * @param fileName raw signed txn file
```
ratik21, line 185:
```suggestion
 * @param fileName raw(encoded) signed txn file
```

ratik21 (approved):
Please check comments before merging.

PR #290, opened 2021-04-15
Security token

ratik21:
- Added few commented code in python scripts (which requires *TEALv3*)
- Added some TODO's. Will handle these in other PR's & update accordingly.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -11,8 +11,3 @@ pyyaml = "*"
 
 [requires]
 python_version = "3.8"
-
-[packages.scratch_var]
-git = "https://github.com/algorand/pyteal.git"
-editable = true
-ref = "master"
--- a/examples/security-token/algob.config.js
+++ b/examples/security-token/algob.config.js
@@ -1,0 +1,14 @@
+const fs = require("fs");
+
+// check if local config in /examples exists if yes then use it, otherwise use a template
+// config provided by this repository.
+
+let config = "../algob.config-local.js";
+try{
+  fs.accessSync(config, fs.constants.F_OK);
+} catch {
+  config = "../algob.config-template.js";
+}
+console.log("config file: ", config);
+
+module.exports = require(config);
--- a/examples/security-token/assets/asa.yaml
+++ b/examples/security-token/assets/asa.yaml
@@ -1,5 +1,5 @@
 gold:
-  total: 1000000000 # (10 billions)
+  total: 10000000000 # (10 billions)
   decimals: 0
   defaultFrozen: true
   unitName: "GLD"
--- a/examples/security-token/assets/asa.yaml
+++ b/examples/security-token/assets/asa.yaml
@@ -1,0 +1,15 @@
+gold:
+  total: 1000000000 # (10 billions)
+  decimals: 0
+  defaultFrozen: true
+  unitName: "GLD"
+  url: "ipfs.io/ipfs/<hash>" # https://ipfs.io/ipfs/Qme7ss3ARVgxv6rXqVPiikMJ8u2NLgmgszg13pYrDKEoiu
+  # User may get "signature validation failed" from node if shorter hash is used.
+  metadataHash: "12312442142141241244444411111133"
+  note: "note"
+  noteb64: "noteb64"
+  # alice-account is the manager and reserve of this asset
+  manager: "EDXG4GGBEHFLNX6A7FGT3F6Z3TQGIU6WVVJNOXGYLVNTLWDOCEJJ35LWJY"
+  reserve: "EDXG4GGBEHFLNX6A7FGT3F6Z3TQGIU6WVVJNOXGYLVNTLWDOCEJJ35LWJY"
+  freeze: ""
+  clawback: "EDXG4GGBEHFLNX6A7FGT3F6Z3TQGIU6WVVJNOXGYLVNTLWDOCEJJ35LWJY"
--- a/examples/security-token/assets/controller.py
+++ b/examples/security-token/assets/controller.py
@@ -1,0 +1,222 @@
+from pyteal import *
+
+def approval_program():
+    """
+    This smart contract acts as a controller of the token. It
+    ensures that the permissions smart contract is called upon each
+    token transfer between non-reserve accounts. The permissions smart contract
+    validates the transfer with some rule checks (eg. asset_amount <= 100).
+    Controller application id is set in clawback-escrow (as a template parameter),
+    to ensure there is a call to controller during a token-transfer.
+    It also handles kill_status of the token. If token is killed then no token transfers
+    are allowed. User can only optOut of the asset.
+    """
+
+    var_is_killed = Bytes("is_token_killed")
+    var_total = Bytes("total_rules")
+    token_id = Bytes("token_id")
+    permissions_manager = Bytes("permissions_manager")	# permissions manager (asa.manager by default)
+    permission_id = Bytes("permissions_sc_app_id") # permissions smart contract application index
+
+    # Always verify that the RekeyTo property of any transaction is set to the ZeroAddress
+    # unless the contract is specifically involved ina rekeying operation.Btoi
+    no_rekey_addr = Txn.rekey_to() == Global.zero_address()
+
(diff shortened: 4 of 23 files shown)


examples/security-token/assets/controller.py
```diff
@@ -0,0 +1,222 @@
...
+    # retreive asset manager from Txn.ForeignAssets[0]
+    assetManager = AssetParam.manager(Int(0))
+
+    on_deployment = Seq([
+        assetManager, # load asset manager from store
+		Assert(And(
+			Txn.application_args.length() == Int(1),
+			# Txn.assets.length() == Int(1), [TEALv3]
+			no_rekey_addr,
+
```
amityadav0, line 34: Please check the indentation here

examples/security-token/assets/controller.py
```diff
@@ -0,0 +1,222 @@
...
+def approval_program():
+    """
+    This smart contract acts as a controller of the token. It
+    ensures that the permissions smart contract is called upon each
+    token transfer between non-reserve accounts. The permissions smart contract
+    validates the transfer with some rule checks (eg. asset_amount <= 100).
+    Controller application id is set in clawback-escrow (as a template parameter),
+    to ensure there is a call to controller during a token-transfer.
+    It also handles kill_status of the token. If token is killed then no token transfers
+    are allowed. User can only optOut of the asset.
```
amityadav0, line 12: except one ?

examples/security-token/assets/controller.py
```diff
@@ -0,0 +1,222 @@
...
+    """
+
+    var_is_killed = Bytes("is_token_killed")
+    var_total = Bytes("total_rules")
+    token_id = Bytes("token_id")
+    permissions_manager = Bytes("permissions_manager")	# permissions manager (asa.manager by default)
+    permission_id = Bytes("permissions_sc_app_id") # permissions smart contract application index
+
+    # Always verify that the RekeyTo property of any transaction is set to the ZeroAddress
+    # unless the contract is specifically involved ina rekeying operation.Btoi
```
amityadav0, line 22:
```suggestion
    # unless the contract is specifically involved in a rekeying operation.
```

examples/security-token/assets/controller.py
```diff
@@ -0,0 +1,222 @@
...
+    # retreive asset reserve from Txn.ForeignAssets[0]
+    assetReserve = AssetParam.reserve(Int(0))
+    issue_token = Seq([
+        assetReserve,
+        Assert(And(
+			Txn.application_args.length() == Int(1),
+			# Txn.assets.length() == Int(1), [TEALv3]
+            Gtxn[1].type_enum() == TxnType.AssetTransfer,
+            Gtxn[1].xfer_asset() == App.globalGet(token_id), # verify if token index is correct
+
```
amityadav0, line 60: Please check the indentation here

examples/security-token/assets/controller.py
```diff
@@ -0,0 +1,222 @@
...
+        Return(Int(1))
+    ])
+
+    # kill a token, after verifying transaction set is_killed to true(Int(1))
+    assetManager = AssetParam.manager(Int(0))
+    kill_token = Seq([
+        assetManager, # load asset_manager of Txn.ForeignAssets[0]
+        Assert(And(
+			Txn.application_args.length() == Int(1),
+			# Txn.assets.length() == Int(1), [TEALv3],
```
amityadav0, line 78: Please check the indentation here

examples/security-token/assets/controller.py
```diff
@@ -0,0 +1,222 @@
...
+
+        # update permissions manager address to the first account address passed in appAccounts
+        App.globalPut(permissions_manager, Txn.accounts[1]),
+        Return(Int(1))
+    ])
+
+	# check properties of txGroup passed
+    group_tx_checks = And(
+        # Ensure 3 basic calls + 1 rules contract call is present in group
+        Global.group_size() == Add(Int(3), App.globalGet(var_total)),
```
amityadav0, line 131: what happens if var_total is > 1 ?
here we are checking four transactions.

examples/security-token/assets/controller.py
```diff
@@ -0,0 +1,222 @@
...
+
+            assetManager.hasValue(),
+			Txn.sender() == assetManager.value(),
+        )),
+
+        # Update total rules counter (= 1 in this case) as we only have a single
+        # permissions contract
+        App.globalPut(var_total, App.globalGet(var_total) + Int(1)),
+
+        # Add permissions(rules) smart contract config in global state (id, manager)
```
amityadav0, line 106: please add what is `id` here ? is it smart contract id ?

examples/security-token/assets/controller.py
```diff
@@ -0,0 +1,222 @@
...
+        assetManager,
+        Assert(And(
+            Txn.application_args.length() == Int(3),
+            # Txn.assets[0] == App.globalGet(token_id), [TEALv3]
+
+            assetManager.hasValue(),
+			Txn.sender() == assetManager.value(),
+        )),
+
+        # Update total rules counter (= 1 in this case) as we only have a single
```
amityadav0, line 102: what are the cases when we have multiple ?
does this contract handles it ?
or it will fail ?

Pipfile
```diff
@@ -10,9 +10,4 @@ pyyaml = "*"
 [dev-packages]
 
 [requires]
-python_version = "3.8"
-
-[packages.scratch_var]
```
amityadav0, line 15: why were we using this before ?

examples/security-token/algob.config.js
```diff
@@ -0,0 +1,14 @@
+const fs = require("fs");
+
+// check if local config in /examples exists if yes then use it, otherwise use a template
+// config provided by this repository.
+
```
amityadav0, line 5:
```suggestion
```

examples/security-token/assets/asa.yaml
```diff
@@ -0,0 +1,15 @@
+gold:
+  total: 1000000000 # (10 billions)
```
amityadav0, line 2:
```suggestion
  total: 1000000000 # (1 billion)
```

examples/security-token/assets/controller.py
ratik21: Yes, on vscode it looks fine on my system. But here it looks incorrect. Checking on another editor.

examples/security-token/assets/controller.py
ratik21: ?

Pipfile
ratik21: Scratch var is used in NFT example project when it wasn't released. So we decided to explicitly add & update pipfile then

examples/security-token/assets/controller.py
ratik21: `permissions_id`, `permissions_manager`. updated.

examples/security-token/assets/controller.py
ratik21: No currently we only handle a single rules contract (hence we have 4 transactions in transfer.js). So in this version, the contract would not fail but overwrite the existing rules config. 
I was discussing this with @robert-zaremba a while ago. That in the future we may update this to handle more than 1 rules contract (that's why kept a rules counter here). I am not sure if we could handle multiple rules smart contract currently with TEAL though.

ratik21: We can keep rules_counter commented if you want

examples/security-token/assets/controller.py
ratik21: if var_total = 2 then transactions = 5 (1 more to call the new rules contract). But in this version we only have single rules contract support.

examples/security-token/assets/controller.py
amityadav0: In spec it is mentioned to kill transfer of asset for everyone except to reserve account, i think that's why we are using this extra variable. right ?

If not why not directly use Asset Freeze feature ?

ratik21: Asset is already frozen, we are using clawback to move assets.
Now if the token is killed, we want to prevent clawback to move token as well.

PR #291, opened 2021-04-19
Support Group transactions with deploy and Opt-In functions

amityadav0:
- Fix group checkpoint persist bug
- Group Support
- Handle Failures
- Handle checkpoints storage


packages/algob/src/internal/deployer.ts
```diff
@@ -214,14 +237,26 @@ export class DeployerDeployMode extends DeployerBasicMode implements Deployer {
...
-        ERRORS.BUILTIN_TASKS.DEPLOYER_ASSET_ALREADY_PRESENT, {
-          assetName: name
-        });
-    }
+  /**
+   * Persist checkpoint till current call.
+   */
+  storeCheckpoint (): void {
+    persistCheckpoint(this.txWriter.scriptName, this.cpData.strippedCP);
+  }
```
ratik21, line 245: Also, can we use this function in other places as well then (eg. in `deployASA`) ?

packages/algob/src/internal/deployer.ts
```diff
@@ -68,6 +69,28 @@ class DeployerBasicMode {
...
+   */
+  assertNoAsset (name: string): void {
+    if (this.isDefined(name)) {
+      persistCheckpoint(this.txWriter.scriptName, this.cpData.strippedCP);
+      throw new BuilderError(
+        ERRORS.BUILTIN_TASKS.DEPLOYER_ASSET_ALREADY_PRESENT, {
+          assetName: name
+        });
+    }
+  }
```
ratik21, line 84: Could you please explain why we moved this function from *Deploy Mode* to *Basic Mode* ?

packages/algob/src/internal/deployer.ts
```diff
@@ -68,6 +69,28 @@ class DeployerBasicMode {
     return found;
   }
 
+  /**
+   * Asserts if asset is present in checkpoint
+   * @param name Asset name
+   */
+  assertNoAsset (name: string): void {
+    if (this.isDefined(name)) {
+      persistCheckpoint(this.txWriter.scriptName, this.cpData.strippedCP);
```
ratik21, line 78:
```suggestion
      this.persistCheckpoint();
```

packages/algob/src/lib/tx.ts
```diff
@@ -141,6 +142,42 @@ async function sendAndWait (
...
+/**
+ * Make transaction parameters and update ASA and SSC params
+ * @param deployer Deployer object
+ * @param t Execution parameters
+ * @param index index of current execParam
+ * @param txIDMap Map for index to name
+ */
+async function mkTx (
+  deployer: Deployer,
+  t: rtypes.ExecParams,
```
ratik21, line 154: `t` looks too cryptic in a separate function. Can we rename this to `tx`/`txn`/`currentTx`/`singleTx` ?

packages/algob/src/lib/tx.ts
```diff
@@ -141,6 +142,42 @@ async function sendAndWait (
...
+  index: number,
+  txIDMap: Map<number, string>
```
ratik21, line 156: Please rename `txIDMap` to something else. From the first look it seams that string here in map is `txID()` of a transaction

packages/algob/src/lib/tx.ts
```diff
@@ -149,31 +186,80 @@ async function sendAndWait (
...
-    signedTxn = signTransaction(txn, execParams);
-    deployer.log(`Signed transaction:`, signedTxn);
+  try {
+    let signedTxn;
+    let txns: Transaction[] = [];
+    const txIDMap = new Map<number, string>();
+    if (Array.isArray(execParams)) {
+      if (execParams.length > 16) { throw new Error("Maximum size of an atomic transfer group is 16"); }
+
+      for (const [index, t] of execParams.entries()) {
```
ratik21, line 196: Let's use `[i, txn]`/`[i, tx]` like you did in line 235

packages/algob/src/lib/tx.ts
```diff
@@ -149,31 +186,80 @@ async function sendAndWait (
...
+      signedTxn = txns.map((txn: Transaction, index: number) => {
+        const signed = signTransaction(txn, execParams[index]);
+        deployer.log(`Signed transaction ${index}`, signed);
+        return signed;
+      });
+    } else {
+      const txn = await mkTx(deployer, execParams, 0, txIDMap);
+      signedTxn = signTransaction(txn, execParams);
+      deployer.log(`Signed transaction:`, signedTxn);
+      txns = [txn];
```
ratik21, line 210: If we're already treating a single txn as a group of 1 tx, then why not do `execParams = [execParams]` above (i think we do this in runtime as well) & remove the else at line 206 ? We are already passing index 0 in line 207. 
Is it because of the *console.log* ?

packages/algob/src/lib/tx.ts
```diff
@@ -149,31 +186,80 @@ async function sendAndWait (
...
+  }
+}
+
+/**
+ * Register checkpoints for ASA and SSC
+ * @param deployer Deployer object
+ * @param txns transaction array
+ * @param txIDMap transaction map index to name
+ */
+async function registerCheckpoints (
```
ratik21, line 230: How about moving this function to `lib/script-checkpoints.ts` ?

packages/algob/src/lib/tx.ts
```diff
@@ -149,31 +186,80 @@ async function sendAndWait (
...
+        const sscInfo: SSCInfo = {
+          creator: encodeAddress(txn.from.publicKey),
+          txId: txn.txID(),
+          appID: txConfirmation['application-index'],
+          confirmedRound: txConfirmation['confirmed-round']
+        };
+        if (name) deployer.registerSSCInfo(name, sscInfo);
+        break;
+      }
+    }
```
ratik21, line 261: I had a question. In `deployer.deployASA(..)` & `deployer.deploySSC(..)` we are also pushing *txWriter* logs (using `deployer.txWriter.push(..)` ).

Are we doing it here as well?

packages/types-algosdk/types/index.d.ts
```diff
@@ -18,7 +18,7 @@ declare module 'algosdk' {
     compile(source: string): Action<CompileOut>;
     status(): Action<any>;
 
-    sendRawTransaction(rawSignedTxn: TxnBytes | TxnBytes[]): Action<TxResult>;
+    sendRawTransaction(rawSignedTxn: TxnBytes | TxnBytes[]): Action<TxResult>;;
```
ratik21, line 21: Why 2 `;;` ?

packages/runtime/src/types.ts
```diff
@@ -235,14 +237,54 @@ export enum TransactionType {
...
   sign: SignType
   lsig?: LogicSig
 }
 
+export interface DeployASAParam extends Sign {
+  type: TransactionType.DeployASA
+  fromAccount: AccountSDK
+  asaName: string
+  asaDef?: ASADef
+  payFlags: TxParams
```
ratik21, line 257: All transaction parameters have `payFlags`, `fromAccount` common, and also each of them *extends Sign*. 

How about moving these common fields to `Sign` interface? What do you think?

ratik21 (commented):
Left few comments and suggestions.
Also i think we missed adding `updateSSC` to `rtypes.TransactionType`. Could you please add that as well ?

packages/algob/src/internal/deployer.ts
amityadav0: so that we can use it both modes.
user can also check if checkpoint has particular asset in run mode

packages/algob/src/internal/deployer.ts
amityadav0: yes

packages/algob/src/internal/deployer.ts
```diff
@@ -214,14 +237,26 @@ export class DeployerDeployMode extends DeployerBasicMode implements Deployer {
...
+   */
+  storeCheckpoint (): void {
+    persistCheckpoint(this.txWriter.scriptName, this.cpData.strippedCP);
+  }
+
+  /**
+   * Register ASA Info in checkpoints
+   */
+  registerASAInfo (name: string, asaInfo: ASAInfo): void {
+    console.log("REGISTERED....");
```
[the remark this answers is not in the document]
amityadav0: ok

packages/algob/src/lib/tx.ts
amityadav0: yes

packages/algob/src/lib/tx.ts
amityadav0: could you please suggest something ?

packages/algob/src/lib/tx.ts
amityadav0: [idx, txn]

packages/algob/src/lib/tx.ts
amityadav0: because of `txns = algosdk.assignGroupID(txns);`

packages/algob/src/lib/tx.ts
amityadav0: yes good idea

packages/algob/src/lib/tx.ts
amityadav0: no, we are not, thanks for catching will add it here

packages/runtime/src/types.ts
amityadav0: i think sign is good, because payFlags seems unrelated

packages/types-algosdk/types/index.d.ts
amityadav0: typo

amityadav0:
yes, adding a task for it, will add it. UpdateSSC in group

packages/algob/src/lib/tx.ts
amityadav0: txIdxMap

packages/algob/src/lib/tx.ts
```diff
@@ -141,6 +143,42 @@ async function sendAndWait (
   return await deployer.waitForConfirmation(txInfo.txId);
 }
 
+/**
+ * Make transaction parameters and update ASA and SSC params
+ * @param deployer Deployer object
+ * @param t Execution parameters
```
ratik21, line 149:
```suggestion
 * @param txn Execution parameters
```

packages/algob/src/lib/tx.ts
ratik21: ok

packages/algob/src/lib/tx.ts
```diff
@@ -149,31 +187,39 @@ async function sendAndWait (
...
+        return signed;
+      });
+    } else {
+      const txn = await mkTx(deployer, execParams, 0, txIdxMap);
+      signedTxn = signTransaction(txn, execParams);
+      deployer.log(`Signed transaction:`, signedTxn);
+      txns = [txn];
+    }
+    const confirmedTx = await sendAndWait(deployer, signedTxn);
+    console.log(confirmedTx);
```
ratik21, line 214: This logs the first transaction confirmation in the *txGroup*(with transactions > 1), right ?

packages/runtime/src/types.ts
ratik21: I was thinking something like:
```javascript
interface BasicParams extends Sign {
  fromAccount: AccountSDK, // keep common keys here
  payFlags: TxParams
}
```
And all transaction parameters will extend *BasicParams*. Just a thought.

packages/algob/src/internal/deployer.ts
```diff
@@ -214,19 +223,51 @@ export class DeployerDeployMode extends DeployerBasicMode implements Deployer {
     this.cpData.putMetadata(this.networkName, key, value);
   }
 
-  private assertNoAsset (name: string): void {
+  /**
+   * Asserts if asset is present in checkpoint
```
ratik21, line 227:
```suggestion
   * Asserts if asset is not already present in checkpoint
```

packages/algob/src/lib/tx.ts
```diff
@@ -141,6 +143,42 @@ async function sendAndWait (
...
+ */
+async function mkTx (
+  deployer: Deployer,
+  txn: rtypes.ExecParams,
+  index: number,
+  txIdxMap: Map<number, string>
+): Promise<Transaction> {
+  switch (txn.type) {
+    case TransactionType.DeployASA: {
+      deployer.assertNoAsset(txn.asaName);
```
ratik21, line 161: just to confirm - this is where the tx will fail if i used `deployASA`/`deploySSC` in *RUN* mode, right? With error message of cannot use assertNoAsset in run mode.. ?

packages/algob/src/lib/tx.ts
```diff
@@ -141,6 +143,42 @@ async function sendAndWait (
...
+      txn.asaDef = deployer.getASADef(txn.asaName);
```
ratik21, line 162: Please update this to handle if user itself passes values in `execParams.asaDef` (i think we'll need to use `overrideASADef()` here). 
Something like:
```javascript
txn.asaDef  = overrideASADef(deployer.getaccountsByName(), deployer.getASADef(txn.asaName), txn.asaDef);
```

packages/runtime/src/types.ts
```diff
@@ -235,14 +237,54 @@ export enum TransactionType {
...
 export interface Sign {
   sign: SignType
   lsig?: LogicSig
 }
 
+export interface DeployASAParam extends Sign {
+  type: TransactionType.DeployASA
+  fromAccount: AccountSDK
+  asaName: string
+  asaDef?: ASADef
```
ratik21, line 256: This should be `Partial<ASADef>`. User can pass only few params (which he wants to override) in execParams

ratik21 (approved):
Pre-approving. Left one update request.

packages/algob/src/lib/tx.ts
amityadav0: yes

packages/algob/src/lib/tx.ts
amityadav0: yes

packages/algob/src/lib/tx.ts
amityadav0: this depends if overrideASADef PR is merged

packages/runtime/src/types.ts
amityadav0: same

packages/runtime/src/types.ts
amityadav0: good thought

PR #292, opened 2021-04-20
Permissioned token

ratik21:
- Added few commented code in python scripts (which requires *TEALv3*)


examples/security-token/README.md
```diff
@@ -0,0 +1,91 @@
...
+a) call to controller smart contract (this ensures all rules are checked)
+b) token transfer transaction from sender -> receiver using clawback (clawback contract ensures controller smart contract is called).
+c) payment transaction from sender -> clawback-escrow (to cover tx fee of above transaction)
+d) call to permissions smart contract (to check rules: max token holding by receiver < 100, both from & to are whitelisted)
+
+5. *Kill Token* (`scripts/permissions/kill.js`): Token manager can kill the token. If the token is killed then all issuance and token transfer transactions are rejected. User can only opt-out from the token to remove his holding and decrease his account's minimum balance.
+a) To kill the token, execute an application call tx to the controller smart contract with appArg: 'kill'. Only ASA mangager can do that. Controller updates the global state key (*is_token_killed*) to true.
```
robert-zaremba, line 72:
```suggestion

    a) To kill the token, execute an application call tx to the controller smart contract with appArg: 'kill'. Only ASA mangager can do that. Controller updates the global state key (*is_token_killed*) to true.
```

examples/security-token/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+6. *Whitelist User* (`scripts/permissions/whitelist.js`): Permissions manager can add a new user to the whitelist which enables that user to receive or send tokens. Only permissions manager (which is stored in the global state of controller smart contract) can add a user to whitelist.
```
robert-zaremba, line 74:
```suggestion
6. *Whitelist User* (`scripts/permissions/whitelist.js`): Only a permissions manager (whose address is stored in the controller smart contract) can add a new user to the whitelist which enables that user to receive or send tokens.
```

examples/security-token/README.md
```diff
@@ -0,0 +1,91 @@
...
+For executing this tx, permissions manager calls the permissions smart contract with application arg "add_whitelist", and passes the account he wishes to white list in AppAccount array. `token_id` is also passed in the ForeignAsset array to verify token. If tx is successful, then permission smart contract updates the local state of `Txn.accounts[1]` & set *whitelisted = true (Int(1))*.
```
robert-zaremba, line 75:
```suggestion
For executing this tx, permissions manager calls the permissions smart contract with application arg "add_whitelist", and passes the account he wishes to whitelist in AppAccount array. `token_id` is also passed in the ForeignAsset array to verify token. If tx is successful, then permission smart contract updates the local state of `Txn.accounts[1]` & set *whitelisted = true (Int(1))*.
```

examples/security-token/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+7. *Change Permissions Manager* (`scripts/permissions/change-manager.js`): Token manager can change the permissions manager via an application call to the controller smart contract. If the parameters are correct, tx is accepted and controller smart contract updates the `permission_manager` in it's global state.
+
+### Other Solution
+
+The solution in this template extends Jason's original idea - using asset clawback as escrow for security token.
```
robert-zaremba, line 81:
```suggestion
The solution in this template extends Algorand Dev Office Hour idea presented by Jason - using asset clawback as escrow for security token.
```

robert-zaremba: Please include a link to the webinar

robert-zaremba (changes requested):
First review on the Readme and the organization.

examples/security-token/README.md
```diff
@@ -0,0 +1,91 @@
+# Security Token
```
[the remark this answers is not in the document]
ratik21: ok

examples/security-token/README.md
```diff
@@ -0,0 +1,91 @@
...
+The Security Token Template outlines a design of security tokens with emphasis for shares and equity.
+
+## Design Overview
+
+We want to enforce permissioned checks on all token transfer between non-reserve accounts. We will use a stateless TEAL as a clawback, which will execute all transfers. So if user A wants to transfer to user B, it will be a clawback which will transfer from account A to account B.
+Steps:
+
+1. Create assets
+2. Distribute to initial users
+3. Freeze all assets OR create the asset default-frozen (clawback can still move frozen tokens).
```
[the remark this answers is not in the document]
ratik21: Either freeze all assets in step 3, OR create the asset by setting default-frozen `true`. If asset is default frozen then we wouldn't need step 3.

examples/security-token/README.md
```diff
@@ -0,0 +1,91 @@
...
+4. Create a stateful contract with rules/permissions (`P`) and another contract: controller (`C`). P will store the C address (which can be modified).
+5. Create a stateless teal (escrow contract `clawback-escrow`) which will check that the C and P is called as well.
```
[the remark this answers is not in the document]
ratik21: ok

examples/security-token/README.md
```diff
@@ -0,0 +1,91 @@
...
+`assets/asa.yaml` - This the the asset definition file.
+
+### Setup
+
+Deployment scripts are placed directly in `/scripts`:
+
+**NOTE:** For transactions involving a multisig account (eg. token can be created and issued by a multisig), user will control one of the multisig accounts and he will receive a signed transaction i.e a signed `deployASA` tx by accounts in msig <= threshold. User can append his/her using `algob sign-multisg` command & then use `executeSignedTxFromFile` function to successfully send transaction to network (eg. deploy token).
+
+1. *0-setup-token.js*: In this script we deploy the token (as an Algorand Standard Asset) using `deployer.deployASA(..)` function. Check note above if creator is a multisig address.
+NOTE: In the template
```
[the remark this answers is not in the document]
ratik21: Sorry, this will be removed.

examples/security-token/README.md
```diff
@@ -0,0 +1,91 @@
...
+NOTE: If asset creator and reserve are different address, then we will need to transfer all tokens from creator to reserve as well (to take those tokens out of circulation/increase total supply).
+
+4. *Asset transfer* (`scripts/transfer/transfer.js`): A non-reserve account can hold few tokens (via issuance) and may wish to transfer some tokens to another non-reserve account. This should be in complaince with rules checks (`permissions.py`). For token transfer we need a group of "atleast" 4 transactions (and more if there are more than 1 rules contracts):
+a) call to controller smart contract (this ensures all rules are checked)
+b) token transfer transaction from sender -> receiver using clawback (clawback contract ensures controller smart contract is called).
+c) payment transaction from sender -> clawback-escrow (to cover tx fee of above transaction)
+d) call to permissions smart contract (to check rules: max token holding by receiver < 100, both from & to are whitelisted)
```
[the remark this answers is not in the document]
ratik21: yes

examples/security-token/README.md
ratik21: ok

examples/security-token/README.md
```diff
@@ -0,0 +1,91 @@
+# Security Token
+
+Security tokens are permissioned tokens that carry complex compliance requirements, such as investor whitelisting and transfer restrictions. For example, token holders outside the U.S. can be prevented from transferring a token to a U.S. individual.
+
+## Motivation
+Specifically, in Security Tokens, we want to:
```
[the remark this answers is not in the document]
ratik21: This will be renamed to permissioned tokens.

examples/security-token/README.md
```diff
@@ -0,0 +1,91 @@
...
+### Execution
+
+1. *Issuance* (`/scripts/issuance/issue.js`): Issuer can issue/mint few tokens to an account. Receiver will need to opt-in to the token first. For issuance tx, we need a transaction group of two transactions :-
+a) Call to the controller smart contract - it checks if token is not killed (as issuance is not allowed if token is killed).
+b) Asset transfer transaction from issuer -> receiver using clawback. `clawback-escrow.py` does the basic sanity checks for issuance_tx.
+We don't need additional rule checks as issuer sets the rules himself.
+
+2. *Update Asset Reserve* (`/scripts/issuance/update-reserve.js`): Say if reserve is a multisig address and we want to kick out one address from the multisig, then we will need to update the asset reserve address. Also, if we update the asset reserve, we will need to move all tokens from previous reserve -> new reserve. This can be done in an atomic transaction group of 2 txns:
+a) Asset transfer transaction using clawback escrow where we move all tokens from previous reserve to new one.
+b) Asset Modification transfer where we update the asset reserve address to new one.
```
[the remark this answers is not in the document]
ratik21: Yes, added support for rekying in #296. Will update this PR after it is merged. 
However, i would like to keep this approach as well (maybe commented), because with rekeying we update an account itself (rather than a property of asset).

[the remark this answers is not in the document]
robert-zaremba: OK. Then let's implement one and add a note for the alternative solution (we can even implement both).

Also, in the text above, do we need word "transfer" in "Asset Modification transfer"?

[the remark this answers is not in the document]
ratik21: > OK. Then let's implement one and add a note for the alternative solution (we can even implement both).

I implemented both approaches in a separate functions - [https://github.com/scale-it/algo-builder/pull/292/files#diff-906cbbe7662016ad6886d22745796565b21389c606b587ac4f559bb352a366abR137](https://github.com/scale-it/algo-builder/pull/292/files#diff-906cbbe7662016ad6886d22745796565b21389c606b587ac4f559bb352a366abR137)

> Also, in the text above, do we need word "transfer" in "Asset Modification transfer"?

No, it was a typo. Updating to "Asset Modification transaction"

examples/permissioned-token/assets/permissions.py
```diff
@@ -0,0 +1,129 @@
...
+            Return(Int(1)) # else return
+        ),
+
+        # finally update txn.accounts[1] local state to set whitelisted status as true(1)
+        App.localPut(Int(1), Bytes("whitelisted"), Int(1)),
+        Return(Int(1))
+    ])
+
+    asset_balance = AssetHolding.balance(Int(1), Gtxn[1].xfer_asset())
+    transfer_token = Seq([
```
man-of-code, line 79: Can you explain why we are checking for this case in every contract?

examples/permissioned-token/README.md
```diff
@@ -0,0 +1,94 @@
...
+  - Both `from` and `to` accounts must be whitelisted
+- *clear_state_program.py*: clears app state (returns 1)
+
+`assets/asa.yaml` - This the the asset definition file.
+
+### Setup
+
+Deployment scripts are placed directly in `/scripts`:
+
+**NOTE:** In a real world scenario, transactions involving a multisig account (eg. token can be created and issued by a multisig) will involve an interaction of many users. A user will receive a signed transaction i.e a signed `deployASA` tx by for a multisig accounts. User can append his/her signature using `algob sign-multisg` command & then use `executeSignedTxFromFile` function to successfully send transaction to network (eg. deploy token).
```
man-of-code, line 46:
```suggestion
**NOTE:** In a real world scenario, transactions involving a multisig account (eg. token can be created and issued by a multisig) will involve an interaction of many users. A user will receive a signed transaction/create an unsigned transaction i.e a `deployASA` tx. User can sign the transaction using `algob sign-multisig` command or by using the `signMultiSig` function  & then use the `executeSignedTxFromFile` function to successfully send transaction to network (eg. deploy token).
```

examples/permissioned-token/scripts/0-setup-token.js
```diff
@@ -0,0 +1,34 @@
+const { fundAccount, totalSupply } = require('./common/common');
+
+/**
+ * NOTE: this function is for demonstration purpose only (if ASA creator, manager are single accounts)
+ * If asset creator is a multisig address, then user should have a signed tx file, decoded tx fetched
```
man-of-code, line 5:
```suggestion
 * If asset creator is a multisig address, then user should have a valid tx file, add his 
```

examples/permissioned-token/scripts/0-setup-token.js
```diff
@@ -0,0 +1,34 @@
...
+ * from that file, append his own signature & send it to network.
```
man-of-code, line 6:
```suggestion
 * own signature & send it to the network.
```

examples/permissioned-token/scripts/0-setup-token.js
```diff
@@ -0,0 +1,34 @@
...
+ *  - Use `algob.executeSignedTxnFromFile` to execute tx from file
```
man-of-code, line 7:
```suggestion
 *  - Use `algob.signMultiSig` or cli command `sign-multisig` to add the signature to multi-sig transaction file
 *  - Use `algob.executeSignedTxnFromFile` to execute tx from file
```

examples/permissioned-token/scripts/2-asset-clawback.js
```diff
@@ -0,0 +1,56 @@
+const {
+  executeTransaction
+} = require('@algo-builder/algob');
+const { types } = require('@algo-builder/runtime');
+
+/**
+ * Compile clawback and set template parameters
+ * - token_id : deployed asa index
+ * - controller_app_id: application index of controller smart contract. This makes sure that
```
man-of-code, line 9:
```suggestion
 * - controller_app_id: controller SSC application index. This makes sure that
```

examples/permissioned-token/scripts/2-asset-clawback.js
```diff
@@ -0,0 +1,56 @@
...
+ *   rules smart contract is called in any token transfer b/w non-reserve accounts
+ * - token_reserve: asa.reserve address (stores all unminted tokens). Clawback checks during
```
man-of-code, line 11:
```suggestion
 * - token_reserve: asa.reserve address (owner of all non-minted tokens). Clawback checks during
```

man-of-code (changes requested):

examples/permissioned-token/scripts/permissions/kill.js
```diff
@@ -0,0 +1,65 @@
+const { types } = require('@algo-builder/runtime');
+const { issue } = require('../issuance/issue');
+const { executeTransaction, fundAccount } = require('../common/common');
+
+/**
+* NOTE: this function is for demonstration purpose only (if ASA manager is a single account)
```
man-of-code, line 6: Can you please correct this note everywhere. 

The transaction file doesn't need to be signed and the user can use 2 methods to add his sign:
+ use CLI command `algob sign-multisig`
+ use exported function `signMultiSig`

man-of-code (changes requested):

examples/permissioned-token/assets/permissions.py
```diff
@@ -0,0 +1,129 @@
...
+
+def approval_program():
+    """
+    This smart contract is the permissions or rules smart contract.
+    If configured in controller, then this sc will ensure 2 rules are
+    followed during a token transfer:
+        a) receiver asset balance should be <=100 after transfer
+        b) both [from, to] accounts must be whitelisted
+    Permissions manager manages the contract permission.
+    It can be changed by the asset manager
```
robert-zaremba, line 11:
```suggestion
    Th Permissions smart contract defines transfer rules.
    Here, we implement two rules:
        a) receiver asset balance should be <=100 after transfer
        b) both [from, to] accounts must be whitelisted
    The Permissions smart contract must be called with the Controller to validate
    transfers. The Controller smart contract defines a manager account, which
    is used here to update user attributes and smart contract state.
```

examples/permissioned-token/assets/permissions.py
```diff
@@ -0,0 +1,129 @@
...
+    """
+
+    true = Int(1)
+    max_tokens = Bytes("max_tokens") # maximum token transfer rule
+    whitelist_count = Bytes("whitelist_count") # whitelisted accounts counter
+    controller_app_id = Bytes("controller_app_id") # token's controller application index
+
+    on_deployment = Seq([
+	Assert(And(
```
robert-zaremba, line 20:
```suggestion
        Assert(And(
```

robert-zaremba: BTW, in Python we use only spaces, and we should never use tabs.

robert-zaremba: I see that in this file tabs and spaces are mixed.

examples/permissioned-token/assets/permissions.py
```diff
@@ -0,0 +1,129 @@
...
+		# verify from global state whether controller application passed is the valid one
+		# note: with tealv3 we can just use txn.applications[0] instead of passing an an appArg
+		Btoi(Txn.application_args[1]) == App.globalGet(controller_app_id),
+
+		# first verify correct token value is passed
+		# token_id.hasValue(),
+			# Txn.assets[0] == token_id.value(), [TEALv3]
+
+		# then verify txn sender is the token manager
+		permission_manager.hasValue(),
```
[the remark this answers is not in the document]
robert-zaremba: We will need to add tests.

examples/permissioned-token/assets/permissions.py
```diff
@@ -0,0 +1,129 @@
...
+        # Save controller application index in global state
+        App.globalPut(controller_app_id, Btoi(Txn.application_args[0])),
+        Return(Int(1))
+    ])
+
+    # fetch permissions manager, token_id from controller's global state (using foreignApps)
+    permission_manager = App.globalGetEx(Int(1), Bytes("permissions_manager"))
+    token_id = App.globalGetEx(Int(1), Bytes("token_id"))
+
+    # add an account to be whitelisted (token transfer is rejected if account is not whitelisted)
```
robert-zaremba, line 43: let's extend comments to have description of parameters for each sub-routine.

robert-zaremba, line 43:
```suggestion
    # whitelist an account
```

examples/permissioned-token/assets/permissions.py
```diff
@@ -0,0 +1,129 @@
...
+    add_whitelist = Seq([
+        permission_manager,
+        token_id,
+        Assert(And(
+		Txn.application_args.length() == Int(2),
+		Txn.accounts.length() == Int(1),
+		Txn.rekey_to() == Global.zero_address(),
```
robert-zaremba, line 50: can we move it to a "common checks (2 lines above)

examples/permissioned-token/assets/permissions.py
```diff
@@ -0,0 +1,129 @@
...
+		#Txn.applications.length() ==  Int(1), [TEALv3]
+
+		# verify from global state whether controller application passed is the valid one
+		# note: with tealv3 we can just use txn.applications[0] instead of passing an an appArg
+		Btoi(Txn.application_args[1]) == App.globalGet(controller_app_id),
+
+		# first verify correct token value is passed
+		# token_id.hasValue(),
+			# Txn.assets[0] == token_id.value(), [TEALv3]
```
robert-zaremba, line 59: I think this is not needed even with TEALv3

examples/permissioned-token/assets/permissions.py
robert-zaremba: This is the most important check. Please read the README file with the mechanism description.

examples/permissioned-token/assets/permissions.py
```diff
@@ -0,0 +1,129 @@
...
+        Return(Int(1))
+    ])
+
+    asset_balance = AssetHolding.balance(Int(1), Gtxn[1].xfer_asset())
+    transfer_token = Seq([
+        asset_balance, # load asset_balance of asset_receiver from store
+        Assert(And(
+		Txn.application_args.length() == Int(1),
+		Global.group_size() >= Int(3),
+		Txn.sender() == Gtxn[1].asset_sender(),
```
robert-zaremba, line 84: again, I think we only need to check that the `Gtxn[1] == clawback`

examples/permissioned-token/assets/permissions.py
```diff
@@ -0,0 +1,129 @@
...
+
+		# verify tx.accounts[1] of current_tx should be same as asset receiver
+		Txn.accounts[1] == Gtxn[1].asset_receiver(),
+
+		# rule 1 - check balance of receiver after receiving token <= 100(max_tokens)
+		asset_balance.hasValue(),
```
robert-zaremba, line 90: do we need this line? Isn't the next line enough?

examples/permissioned-token/assets/permissions.py
```diff
@@ -0,0 +1,129 @@
...
+		asset_balance.value() <= App.globalGet(max_tokens),
+
+		# rule 2 - [from, to] accounts must be whitelisted
+		# NOTE: Int(0) == Txn.Sender(), Int(1) == Txn.accounts[1]
+		App.localGet(Int(0), Bytes("whitelisted")) == true, # from account must be whitelisted
+		App.localGet(Int(1), Bytes("whitelisted")) == true  # to account must be whitelisted
```
robert-zaremba, line 96: What if a user will use a different address here (different than a "true" recipient of the transfer)?
can we extract the `to` address from the `Gtxn[1]`?

examples/permissioned-token/assets/permissions.py
```diff
@@ -0,0 +1,129 @@
...
+    # Verfies that the application_id is 0, jumps to on_deployment.
+    # Verifies that DeleteApplication is used and blocks that call
+    # Verifies that UpdateApplication is used and blocks that call (unsafe for production use).
+    # Verifies that closeOut is used and approves the tx.
+    # Verifies that OptInApplication is used and jumps to handle_optin
+    # Verifies that first argument is "add_whitelist" and jumps to add_whitelist.
+    # Verifies that first argument is "transfer" and jumps to transfer_token.
+    program = Cond(
+        [Txn.application_id() == Int(0), on_deployment],
+        [Txn.on_completion() == OnComplete.UpdateApplication, Return(Int(0))], # block update
```
robert-zaremba, line 118: Updates are OK, we want to enable upadates to add more rules if needed in the future.
Let's make it clear here (using a comment) and in the README.

robert-zaremba (commented):
I will do next round of review once we clarify the existing comments.

examples/permissioned-token/assets/clawback.py
```diff
@@ -0,0 +1,116 @@
...
+        Gtxn[0].asset_receiver() == Gtxn[1].config_asset_reserve(),
+
+        Gtxn[1].type_enum() == TxnType.AssetConfig,
+        Gtxn[1].config_asset() == Int(TOKEN_ID), # verify token index for asset_config
+
+        # ensure no rekeying
+        Gtxn[0].rekey_to() == Global.zero_address(),
+        Gtxn[1].rekey_to() == Global.zero_address()
+    )
+
```
robert-zaremba, line 99: we need to add rule for `force_transfer` from the [spec](https://paper.dropbox.com/doc/Algob-Security-Token-Template--BKDPSFg1YRt3TQQjWHyScz4QAg-FR2LXhVg3edevYPBQZw6F#:uid=481217053399840855734627&h2=Issuance-Functions).  (I've renamed the `clawback` function to `force_transfer` to not confuse with the clawback asset attribute).

examples/permissioned-token/assets/clawback.py
```diff
@@ -0,0 +1,116 @@
...
+        # imp: check asset receiver first tx (move funds) is the new reserve
+        Gtxn[0].asset_receiver() == Gtxn[1].config_asset_reserve(),
+
+        Gtxn[1].type_enum() == TxnType.AssetConfig,
+        Gtxn[1].config_asset() == Int(TOKEN_ID), # verify token index for asset_config
+
+        # ensure no rekeying
+        Gtxn[0].rekey_to() == Global.zero_address(),
+        Gtxn[1].rekey_to() == Global.zero_address()
+    )
```
robert-zaremba, line 98: let's remove this condition. I think the better way to do it is by:
a) rekey - with this we will keep the reserve address - so users won't be confused that the issuer address changed. (the reserve address is used to indicate token issuance)
b) force_transfer - this is something we still need to add -> see my [comment](https://github.com/scale-it/algo-builder/pull/292/files/7bf3c0c34f54f5bdd3a3cd5dc3ff398c26392441#r624949587) below.

So, let's remove the code-case above, and update documentation (README) explaining how to change the reserve in the future using the two listed methods above.

examples/permissioned-token/scripts/2-asset-clawback.js
```diff
@@ -0,0 +1,56 @@
+const {
+  executeTransaction
+} = require('@algo-builder/algob');
+const { types } = require('@algo-builder/runtime');
+
+/**
+ * Compile clawback and set template parameters
```
robert-zaremba, line 7:
```suggestion
 * Compile and set clawback logic sig (escrow) with template parameters:
```

examples/permissioned-token/README.md
```diff
@@ -0,0 +1,94 @@
...
+The permissioned token Template outlines a design of permissioned tokens with emphasis for shares and equity.
+
+## Design Overview
+
+We want to enforce permissioned checks on all token transfer between non-reserve accounts. We will use a stateless TEAL as a clawback, which will execute all transfers. So if user A wants to transfer to user B, (s)he will call a clawback stateless smart contract to transfer from account A to account B.
+We implement a Permissioned Token template, which executes the following steps:
+
+1. Create ASA
+2. Distribute to initial users
+3. Freeze all assets OR create the asset default-frozen (clawback can still move frozen tokens).
```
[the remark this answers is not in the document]
ratik21: I think *reminding* should be *remaining*. Also, what do you mean by remaining assets? Should this be move "some/all" assets?

examples/permissioned-token/assets/controller.py
```diff
@@ -0,0 +1,222 @@
...
+    # Always verify that the RekeyTo property of any transaction is set to the ZeroAddress
+    # unless the contract is specifically involved in a rekeying operation.
+    no_rekey_addr = Txn.rekey_to() == Global.zero_address()
+
+    # retreive asset manager from Txn.ForeignAssets[0]
+    assetManager = AssetParam.manager(Int(0))
+
+    on_deployment = Seq([
+        assetManager, # load asset manager from store
+	Assert(And(
```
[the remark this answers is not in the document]
ratik21: done

examples/permissioned-token/assets/controller.py
```diff
@@ -0,0 +1,222 @@
...
+		assetManager.hasValue(),
+		Txn.sender() == assetManager.value(),
+        )),
+
+        # Update total rules counter (= 1 in this case) as we only have a single
+        # permissions contract
+        App.globalPut(var_total, App.globalGet(var_total) + Int(1)),
+
+        # Add permissions(rules) smart contract config in global state (permission_id, permissions_manager)
+        App.globalPut(permission_id, Btoi(Txn.application_args[1])),
```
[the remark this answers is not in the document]
ratik21: yes, in that case we will have list of permission id(s), manager(s)

examples/permissioned-token/assets/permissions.py
ratik21: fixed

examples/permissioned-token/scripts/0-setup-token.js
ratik21: file will be signed already (by few of multisig accounts)

examples/permissioned-token/scripts/issuance/issue.js
```diff
@@ -0,0 +1,99 @@
...
+    }
+  ];
+  console.log(`* Issuing ${amount} tokens to ${account.name}:${account.addr} *`);
+  await executeTransaction(deployer, issuanceParams);
+
+  console.log(`* ${account.name} asset holding: *`);
+  await balanceOf(deployer, account.addr, asaInfo.assetIndex); // print asset holding
+
+  const supply = await totalSupply(deployer, asaInfo.assetIndex);
+  console.log(`Total Supply of token 'gold': ${supply}`);
```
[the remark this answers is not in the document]
ratik21: After issuance/minting of tokens, supply is increased. We just log that number here

examples/permissioned-token/scripts/transfer/transfer.js
```diff
@@ -0,0 +1,179 @@
...
+
+  // opt-in accounts to asa 'gold' (so they can receive it)
+  await Promise.all([
+    deployer.optInAcountToASA('gold', elon.name, {}),
+    deployer.optInAcountToASA('gold', bob.name, {}),
+    deployer.optInAcountToASA('gold', john.name, {})
+  ]);
+
+  // note: if reserve is multisig, then user will use executeSignedTxnFromFile function
+  await issue(deployer, bob, 300); // issue(mint) 300 tokens to bob from reserve
```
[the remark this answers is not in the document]
ratik21: issuance doesn't go through permission checks. Please read the README/spec

examples/permissioned-token/assets/permissions.py
```diff
@@ -0,0 +1,129 @@
...
+        # Initialize whitelisted accounts counter to 0
+        App.globalPut(whitelist_count, Int(0)),
+
+        # Save controller application index in global state
+        App.globalPut(controller_app_id, Btoi(Txn.application_args[0])),
+        Return(Int(1))
+    ])
+
+    # fetch permissions manager, token_id from controller's global state (using foreignApps)
+    permission_manager = App.globalGetEx(Int(1), Bytes("permissions_manager"))
```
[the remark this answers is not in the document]
ratik21: done

examples/permissioned-token/README.md
```diff
@@ -0,0 +1,94 @@
...
+### Other Solution
+
+The solution in this template extends [Algorand Dev Office Hour](https://register.gotowebinar.com/recording/recordingView?webinarKey=1651582324861824270&registrantEmail=ratikjindal21%40gmail.com) idea presented by Jason - using asset clawback as escrow for permissioned token.
+In this template we add a new contract - `controller.py` which essentially "controls" the asset/token. This controls the general properties of the asset and also ensures that rule(s) smart contract(s) are called.
+
+### Links
+
+Developer hour links:
+- *video:* [https://www.youtube.com/watch?v=aMDZamxtR14](https://www.youtube.com/watch?v=aMDZamxtR14)
+- *article:* [https://developer.algorand.org/solutions/assets-and-custom-transfer-logic](https://developer.algorand.org/solutions/assets-and-custom-transfer-logic/)
```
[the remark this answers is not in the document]
ratik21: line 91 doesn't seem correct. Updating to:
```md
- Developer Office Hours | Assets and Custom Transfer Logic Using Algorand Smart Contracts: [video](https://www.youtube.com/watch?v=aMDZamxtR14)
- [Assets and Custom Transfer Logic](https://developer.algorand.org/solutions/assets-and-custom-transfer-logic/)
```

examples/permissioned-token/assets/controller.py
```diff
@@ -0,0 +1,222 @@
...
+    token_id = Bytes("token_id")
+    permissions_manager = Bytes("permissions_manager")	# permissions manager (asa.manager by default)
+    permission_id = Bytes("permissions_sc_app_id") # permissions smart contract application index
+
+    # Always verify that the RekeyTo property of any transaction is set to the ZeroAddress
+    # unless the contract is specifically involved in a rekeying operation.
+    no_rekey_addr = Txn.rekey_to() == Global.zero_address()
+
+    # retreive asset manager from Txn.ForeignAssets[0]
+    assetManager = AssetParam.manager(Int(0))
```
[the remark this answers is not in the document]
ratik21: > what happens if no asset is linked in the transaction call?

invalid ForeignAssets index Error:
```js
'{"message":"TransactionPool.Remember: transaction GNAINMYQFCARUO7TOUVFVSAYRXCQ7UPKKE4N4SBO7A4IZJNEGE2Q: logic eval error: invalid ForeignAssets index 0"}\n'
```

> Where this will be stored? In scratch space?

Yes, in `Store`. In the subroutine we load it from store.

examples/permissioned-token/assets/controller.py
```diff
@@ -0,0 +1,222 @@
...
+		Txn.sender() == assetManager.value(),
+        )),
+
+        # finally set is_killed to true
+        App.globalPut(var_is_killed, Int(1)),
+        Return(Int(1))
+    ])
+
+    add_new_permission = Seq([
+        assetManager,
```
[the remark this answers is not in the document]
ratik21: No, in line 98 it's stored in *Store*. We load from store in this line.

examples/permissioned-token/assets/controller.py
```diff
@@ -0,0 +1,222 @@
...
+    kill_token = Seq([
+        assetManager, # load asset_manager of Txn.ForeignAssets[0]
+        Assert(And(
+		Txn.application_args.length() == Int(1),
+		# Txn.assets.length() == Int(1), [TEALv3],
+		# Txn.assets[0] == App.globalGet(token_id), [TEALv3] # verify if token index is correct
+		Txn.type_enum() == TxnType.ApplicationCall,
+
+		# Only asset manager can kill the token
+		assetManager.hasValue(),
```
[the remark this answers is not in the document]
ratik21: `.hasValue()` ensures that the asset being looked up exists in the network. It returns 1 even if manager is not defined (set to "") - [https://pyteal.readthedocs.io/en/stable/assets.html#asset-parameters](https://pyteal.readthedocs.io/en/stable/assets.html#asset-parameters)

examples/permissioned-token/assets/controller.py
```diff
@@ -0,0 +1,222 @@
...
+		Gtxn[1].asset_sender() == assetReserve.value(),
+
+		# only allow issue if token is not killed
+		App.globalGet(var_is_killed) == Int(0)
+        )),
+        Return(Int(1))
+    ])
+
+    # kill a token, after verifying transaction set is_killed to true(Int(1))
+    assetManager = AssetParam.manager(Int(0))
```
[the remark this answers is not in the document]
ratik21: yes

examples/permissioned-token/assets/controller.py
```diff
@@ -0,0 +1,222 @@
...
+		Txn.application_args.length() == Int(3),
+		# Txn.assets[0] == App.globalGet(token_id), [TEALv3]
+
+		assetManager.hasValue(),
+		Txn.sender() == assetManager.value(),
+        )),
+
+        # Update total rules counter (= 1 in this case) as we only have a single
+        # permissions contract
+        App.globalPut(var_total, App.globalGet(var_total) + Int(1)),
```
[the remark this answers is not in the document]
ratik21: Yes we don't use it currently because we only keep a single permissions contract (so var_total = 1). In case of multiple rules contracts, we may need to keep total_count. Let me know if we should remove it.

examples/permissioned-token/assets/controller.py
```diff
@@ -0,0 +1,222 @@
...
+		assetManager.hasValue(),
+		Txn.sender() == assetManager.value(),
+        )),
+
+        # finally set is_killed to true
+        App.globalPut(var_is_killed, Int(1)),
+        Return(Int(1))
+    ])
+
+    add_new_permission = Seq([
```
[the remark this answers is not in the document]
ratik21: >  Please add a comment that his also changes the permission_manager address.

No, it does not "change" the permissions manager address, it rather adds a new config (rules contract). Currently it's only used a single time (as we don't support adding multiple contracts with current version of TEAL). 
To change the current permissions_manager address, we do it in a separate branch (`change_permissions_manager`)

examples/permissioned-token/assets/controller.py
```diff
@@ -0,0 +1,222 @@
...
+    ])
+
+    # kill a token, after verifying transaction set is_killed to true(Int(1))
+    assetManager = AssetParam.manager(Int(0))
+    kill_token = Seq([
+        assetManager, # load asset_manager of Txn.ForeignAssets[0]
+        Assert(And(
+		Txn.application_args.length() == Int(1),
+		# Txn.assets.length() == Int(1), [TEALv3],
+		# Txn.assets[0] == App.globalGet(token_id), [TEALv3] # verify if token index is correct
```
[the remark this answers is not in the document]
ratik21: Removing line 77 check. 
Assets.length(), Assets[0] check we need (to ensure correct token_id is passed)

examples/permissioned-token/assets/controller.py
```diff
@@ -0,0 +1,222 @@
...
+        # finally set is_killed to true
+        App.globalPut(var_is_killed, Int(1)),
+        Return(Int(1))
+    ])
+
+    add_new_permission = Seq([
+        assetManager,
+        Assert(And(
+		Txn.application_args.length() == Int(3),
+		# Txn.assets[0] == App.globalGet(token_id), [TEALv3]
```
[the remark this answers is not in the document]
ratik21: It will give an invalid index error (if no args are passed & we try access Txn.application_args[1]). 
But line 95 is a basic check (if we're using `tx.args[0,1..n-1]` then we do an initial assert of `tx.args.length() == n`), and then we access it. It looks more intuitive. 
Let me know if you think we should remove.

examples/permissioned-token/assets/permissions.py
ratik21: moved rekey check to common var

examples/permissioned-token/assets/permissions.py
ratik21: I think it's an important check. It ensures that correct *token_id* is passed in `foreignAssets` array. If we don't check it user (not permissions_manager) can pass a fake id as well. Because we fetch id, manager from controller itself.

examples/permissioned-token/assets/permissions.py
ratik21: as explained above, `hasValue()` checks that asset exists in network, then we fetch the value in next line. But otherwise (if you think asset will always exist & checking doesn't make sense), then we can remove it. I tested and it works. Let me know.

examples/permissioned-token/assets/permissions.py
ratik21: > What if a user will use a different address here (different than a "true" recipient of the transfer)?

We are verifying the `to` address to be asset_receiver before whitelisting: [https://github.com/scale-it/algo-builder/pull/292/files/7bf3c0c34f54f5bdd3a3cd5dc3ff398c26392441#diff-ce9fa1b7e02d1a7a20ef56e646823bb78a55dd88724bad14ecd27e19297b3917R87](https://github.com/scale-it/algo-builder/pull/292/files/7bf3c0c34f54f5bdd3a3cd5dc3ff398c26392441#diff-ce9fa1b7e02d1a7a20ef56e646823bb78a55dd88724bad14ecd27e19297b3917R87) 

> can we extract the to address from the Gtxn[1]?

`App.localGet(..)` expects first arg as index in *AppAccounts* array.

examples/permissioned-token/assets/permissions.py
ratik21: Just to confirm: 
+ updating permissions smart contract means updating an existing set of rules/add new rule in current asc
+ adding new rule means add a new rules contract (which we don't support yet, having multiple rule(s) contract(s))

examples/permissioned-token/scripts/0-setup-token.js
man-of-code: It can be, it's not compulsory.

examples/permissioned-token/assets/controller.py
```diff
@@ -0,0 +1,222 @@
...
+
+    # retreive asset reserve from Txn.ForeignAssets[0]
+    assetReserve = AssetParam.reserve(Int(0))
+    issue_token = Seq([
+        assetReserve,
+        Assert(And(
+		Txn.application_args.length() == Int(1),
+		# Txn.assets.length() == Int(1), [TEALv3]
+		Gtxn[1].type_enum() == TxnType.AssetTransfer,
+		Gtxn[1].xfer_asset() == App.globalGet(token_id), # verify if token index is correct
```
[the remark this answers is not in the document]
ratik21: ok

examples/permissioned-token/assets/controller.py
```diff
@@ -0,0 +1,222 @@
...
+    ])
+
+	# check properties of txGroup passed
+    group_tx_checks = And(
+        # Ensure 3 basic calls + 1 rules contract call is present in group
+        Global.group_size() == Add(Int(3), App.globalGet(var_total)),
+        Gtxn[0].type_enum() == TxnType.ApplicationCall, # call to controller smart contract
+        Gtxn[1].type_enum() == TxnType.AssetTransfer,
+        Gtxn[2].type_enum() == TxnType.Payment, # paying fees of escrow
+        Gtxn[3].type_enum() == TxnType.ApplicationCall, # call to permissions contract
```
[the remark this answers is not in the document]
ratik21: We need to check for the permissions call in the controller itself (as permissions smart contract could be variable). For `Gtxn[0, ..2]` we can do in clawback itself

examples/permissioned-token/assets/controller.py
```diff
@@ -0,0 +1,222 @@
...
+        Gtxn[3].asset_close_to() == Global.zero_address()
+    )
+
+    all_transaction_checks = And(
+        # verify first transaction
+        # call to controller smart contract - signed by asset sender
+        Gtxn[0].application_id() == Global.current_application_id(),
+        Gtxn[0].sender() == Gtxn[2].sender(),
+        Gtxn[0].sender() == Gtxn[3].sender(),
+        Gtxn[0].sender() == Gtxn[1].asset_sender(),
```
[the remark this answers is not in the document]
ratik21: yes, removed from here

examples/permissioned-token/assets/permissions.py
ratik21: removed.

examples/permissioned-token/assets/clawback.py
ratik21: yes, doing `force_transfer`

examples/permissioned-token/assets/permissions.py
ratik21: ok

examples/permissioned-token/README.md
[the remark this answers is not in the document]
robert-zaremba: correct, sorry.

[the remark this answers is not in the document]
robert-zaremba: I was thinking that we will do an initial distribution before we will freez assets and change the clawback address.

examples/permissioned-token/README.md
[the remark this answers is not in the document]
robert-zaremba: correct, thanks.

examples/permissioned-token/assets/permissions.py
robert-zaremba: I think we can  safely remove it - we will save some gas.

examples/permissioned-token/assets/permissions.py
robert-zaremba: yes, but we should be able to deploy a new version of the contract with updated rules.

examples/permissioned-token/assets/controller.py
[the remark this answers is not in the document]
robert-zaremba: let's remove it if we don't use it.

examples/permissioned-token/assets/controller.py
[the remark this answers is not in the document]
robert-zaremba: If `Gtxn[1].sender` is sender, then the clawback logic will be called to check the group transaction composition, right?

examples/permissioned-token/assets/permissions.py
robert-zaremba: But this is already checked in the clawback. So we only need to check that clawback tx is in Gtxn

examples/permissioned-token/assets/controller.py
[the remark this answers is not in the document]
robert-zaremba: ok, we can keep it - let's add a comment what are the expected args.

examples/permissioned-token/assets/controller.py
```diff
@@ -0,0 +1,222 @@
...
+        Gtxn[3].application_id() == App.globalGet(permission_id),
+        Gtxn[3].sender() == Gtxn[1].asset_sender()
+    )
+
+    # transfer token from A -> B. Both A, B are non-reserve accounts
+    transfer_token = Seq([
+        Assert(And(
+            Txn.application_args.length() == Int(1),
+            App.globalGet(var_is_killed) == Int(0), # check token is not killed
+            common_fields,
```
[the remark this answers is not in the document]
ratik21: removed

examples/permissioned-token/README.md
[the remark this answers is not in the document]
ratik21: ah, okay

examples/permissioned-token/README.md
```diff
@@ -0,0 +1,94 @@
+# Permissioned token
+
+In permissioned tokens carry additional requirements and permission mechanism on who, when and how can use that tokens. Securities may use the permissioned token mechanism to define complex compliance requirements, such as investor whitelisting and transfer restrictions. For example, only a verified, KYCed users can exchange tokens.
+
+## Permissioned Token Motivation
+Specifically, for Permissioned Tokens, we want to:
+- Control the number of tokens minted for distribution and also allow flexibility for the administrator to issue more tokens in the future.
+- Create whitelists of individuals who are approved to buy/hold the tokens and segregated sublists, as needed, to support transfer restrictions (e.g., U.S. vs international).
+- Manage lock-ups as needed to ensure there are not transfers before allowed.
```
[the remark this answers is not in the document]
ratik21: yes

examples/permissioned-token/assets/permissions.py
ratik21: ok, removing

examples/permissioned-token/assets/permissions.py
ratik21: ok, thanks for confirming

examples/permissioned-token/assets/controller.py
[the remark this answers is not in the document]
ratik21: ok

examples/permissioned-token/assets/controller.py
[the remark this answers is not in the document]
ratik21: ok

examples/permissioned-token/assets/permissions.py
ratik21: What do you mean? We're in `add_whitelist` branch, how is the clawback involved in whitelisting account?
*permissions_manager* can whitelist an account with a single call to *permissionsSSC* (passing address in `appAccounts`)

robert-zaremba: ah, sorry. You are right - this part is called without calling the clawback.

examples/permissioned-token/assets/clawback.py
ratik21: done.

examples/permissioned-token/assets/clawback.py
ratik21: yes, done.

examples/permissioned-token/assets/controller.py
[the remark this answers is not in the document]
ratik21: removed.

examples/permissioned-token/assets/controller.py
```diff
@@ -0,0 +1,254 @@
...
+    verify_perm_call = And(
+        # verify rules call (ensure permissions smart contract is being called in 4th tx)
+        Gtxn[3].type_enum() == TxnType.ApplicationCall,
+        Gtxn[3].application_id() == App.globalGet(permission_id),
+    )
+
+    # verifies that asset sender in 2nd tx
+    # - calls the controller smart contract
+    # - is also the sender of the payment tx (to pay fees of clawback escrow)
+    # - calls the permissions smart contract (ensures rules check)
```
robert-zaremba, line 163: let's move the comments to the statements below:
```
# asset sender is also the payment tx sender (to pay fees of clawback escrow)
Gtxn[0].sender() == Gtxn[2].sender(),
...
```

examples/permissioned-token/assets/controller.py
```diff
@@ -0,0 +1,254 @@
...
+    """
+    Transfer token from accA -> accB. Both A, B are non-reserve accounts.
+    Only accepted if token is not killed.
+    """
+    transfer_token = Seq([
+        Assert(And(
+            App.globalGet(var_is_killed) == Int(0), # check token is not killed
+
+            # Ensure 3 basic calls + 1 rules contract call is present in group
+            Global.group_size() == Add(Int(3), App.globalGet(var_total)),
```
robert-zaremba, line 179: do we need this check? `verify_perm_call` already makes sure that the `perm` contract is 3rd. We don't do any action here, only check if perm is called, so in fact we don't care what are the other transactions - this contract can be called even with some other random transactions as long as it won't mutate a state.

examples/permissioned-token/assets/permissions.py
```diff
@@ -0,0 +1,156 @@
...
+    rekey_check = Txn.rekey_to() == Global.zero_address()
+
+    """
+    Handles Permissions SSC deployment. Expectes 1 argument
+    * controller_app_id: controller application index
+    During deployment
+    * max_tokens is set to 100
+    * whitelist_counter is intialized to 0
+    * controller application index is saved in global state
+    """
```
robert-zaremba, line 30: as above, we should use the `#` comment for things which are not function or class doc string.

examples/permissioned-token/assets/permissions.py
```diff
@@ -0,0 +1,156 @@
...
+    whitelist_count = Bytes("whitelist_count") # whitelisted accounts counter
+    controller_app_id = Bytes("controller_app_id") # token's controller application index
+
+    # Always verify that the RekeyTo property of any transaction is set to the ZeroAddress
+    # unless the contract is specifically involved ina rekeying operation.
+    rekey_check = Txn.rekey_to() == Global.zero_address()
+
+    """
+    Handles Permissions SSC deployment. Expectes 1 argument
+    * controller_app_id: controller application index
```
robert-zaremba, line 25: we don't need any argument. We can use a template argument and fix it in the smart contract code, rather than global state.

examples/permissioned-token/assets/permissions.py
```diff
@@ -0,0 +1,156 @@
...
+        token_id,
+        Assert(And(
+    		Txn.application_args.length() == Int(2),
+    		Txn.accounts.length() == Int(1),
+    		rekey_check,
+    		# Txn.applications.length() ==  Int(1), [TEALv3]
+
+    		# verify from global state whether controller application passed is the valid one
+    		# note: with tealv3 we can just use txn.applications[0] instead of passing an an appArg
+    		Btoi(Txn.application_args[1]) == App.globalGet(controller_app_id),
```
robert-zaremba, line 70: why do we need to verify a controller_app_id here? I think this "branch" can be called without calling controller.

examples/permissioned-token/assets/permissions.py
```diff
@@ -0,0 +1,156 @@
...
+
+    		# first verify correct token value is passed
+    		# Txn.assets[0] == token_id.value(), [TEALv3]
```
robert-zaremba, line 73: I think we also don't need to do a token check here.

examples/permissioned-token/assets/permissions.py
```diff
@@ -0,0 +1,156 @@
...
+    		# verify tx.accounts[1] of current_tx should be same as asset receiver
+    		Txn.accounts[1] == Gtxn[1].asset_receiver(),
+
+    		# rule 1 - check balance of receiver after receiving token <= 100(max_tokens)
+    		asset_balance.value() <= App.globalGet(max_tokens),
+
+    		# rule 2 - [from, to] accounts must be whitelisted
+    		# NOTE: Int(0) == Txn.Sender(), Int(1) == Txn.accounts[1]
+    		App.localGet(Int(0), Bytes("whitelisted")) == true, # from account must be whitelisted
+    		App.localGet(Int(1), Bytes("whitelisted")) == true  # to account must be whitelisted
```
robert-zaremba, line 112: wrong indentation - each level should be 4 spaces.

examples/permissioned-token/scripts/0-setup-token.js
```diff
@@ -0,0 +1,35 @@
...
+  // just for tutorial purpose (use `executeSignedTxnFromFile` if using multisig account)
+  const asaInfo = await setupASA(deployer);
+  console.log(asaInfo);
+  console.log('total Supply: ', await totalSupply(deployer, asaInfo.assetIndex));
+
+  /**
+   * If using msig address as asa creator or manager, then realistically user will receive a signed tx
+   * file from accounts <= threshold in a multisig group.
+   * - After receiving file, place it in /assets
+   * - Use `algob sign-multisig <account>` to append signature of your account
```
robert-zaremba, line 29:
```suggestion
   * - Use `algob sign-multisig <account> filename` to append signature of your account
```

examples/permissioned-token/scripts/1-setup-apps.js
```diff
@@ -0,0 +1,105 @@
...
+  await setupControllerSSC(asaInfo, deployer);
+
+  const controllerSSCInfo = deployer.getSSC('controller.py', clearStateProgram);
+  await setupPermissionsSSC(asaInfo, controllerSSCInfo, deployer);
+
+  /* Use below function to deploy SSC's if you receive a txn file from a shared network */
+  // executeSignedTxnFromFile(deployer, 'ssc_file_out.tx');
+}
+
+module.exports = { default: run };
```
robert-zaremba, line 105: the order should be:
1. Deploy ASA
2. Deploy controller
3. Deploy clawback
4. Deploy permissions
5. Update controller to set a `P` appID

examples/permissioned-token/scripts/issuance/issue.js
```diff
@@ -0,0 +1,99 @@
+const {
+  balanceOf
+} = require('@algo-builder/algob');
+const { types } = require('@algo-builder/runtime');
+const { executeTransaction, fundAccount, totalSupply } = require('../common/common');
+
+/**
+ * NOTE: this function is for demonstration purpose only (if ASA creator, manager are single accounts)
```
robert-zaremba, line 8:
```suggestion
 * NOTE: this function is for demonstration purpose only.
```

examples/permissioned-token/scripts/issuance/issue.js
```diff
@@ -0,0 +1,99 @@
...
+ * If asset creator is a multisig address, then user should have a signed tx file, decoded tx fetched
+ * from that file, append his own signature & send it to network.
+ *  - Use `algob.executeSignedTxnFromFile` to execute tx from file
+ *  - In below function we assume creator & reserve is a single account (alice)
```
robert-zaremba, line 12:
```suggestion
 *  - In the function below we assume creator & reserve is a single account (alice)
```

examples/permissioned-token/scripts/issuance/issue.js
```diff
@@ -0,0 +1,99 @@
...
+ */
+async function issue (deployer, account, amount) {
+  const asaReserve = deployer.accountsByName.get('alice'); // asset reserve is the issuer
+
+  const asaInfo = deployer.asa.get('gold');
```
robert-zaremba, line 17: let's rename `asaInfo` to `gold`.

examples/permissioned-token/scripts/issuance/issue.js
```diff
@@ -0,0 +1,99 @@
...
+  await fundAccount(deployer, elon); // fund elon
+
+  // opt-in asa to account first (skip if opted-in already)
+  console.log(`* Opt-In ASA gold for ${elon.name} *`);
+  await deployer.optInAcountToASA('gold', elon.name, {});
+  console.log('* Opt-In Successful *');
+
+  /**
+   * If using msig address as asa token reserve, then realistically user will receive a tx
+   * file signed by accounts <= threshold in a multisig group.
```
robert-zaremba, line 81:
```suggestion
   * file signed by accounts >= threshold in a multisig group.
```

examples/permissioned-token/scripts/issuance/issue.js
```diff
@@ -0,0 +1,99 @@
...
+   * - After receiving file, place it in /assets
+   * - Use `algob sign-multisig <account>` to append signature of your account
+   * - Then use below function to issue new tokens (send tx to network)
```
robert-zaremba, line 84:
```suggestion
   * - Then use the function below to issue new tokens and send tx to a network
```

examples/permissioned-token/assets/clawback.py
```diff
@@ -0,0 +1,91 @@
...
+    # check app_id passed through params
+    first_transaction_check = Gtxn[0].application_id() == Int(CONTROLLER_APP_ID)
+
+    # verify second transaction
+    # tx 1 - clawback transactions that moves the frozen asset from sender to receiver - signed by clawback
+    # check asset_id passed through params
+    second_transaction_check = Gtxn[1].xfer_asset() == Int(TOKEN_ID)
+
+    # verify third transaction
+    # tx 2 - payment transaction from sender to clawback to pay for the fee of the clawback
```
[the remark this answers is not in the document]
ratik21: yes

examples/permissioned-token/assets/clawback.py
```diff
@@ -0,0 +1,91 @@
...
+        first_transaction_check,
+        second_transaction_check,
+        third_transaction_checks
+    )
+
+    # issue new token (asset transfer tx from asset reserve to receiver)
+    # since issuer creates the rules, this can be a simple asset transfer transaction (without rule checks)
+    issuance_tx = And(
+        Gtxn[0].type_enum() == TxnType.ApplicationCall,
+        Gtxn[0].application_id() == Int(CONTROLLER_APP_ID),
```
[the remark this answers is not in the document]
ratik21: * To verify sender is the asset reserve
* To verify token is not killed

examples/permissioned-token/scripts/issuance/issue.js
ratik21: No. Current user also appends his signature, so it should be signed by <= threshold. As current user is also one of the accounts of multisig

examples/permissioned-token/assets/controller.py
```diff
@@ -0,0 +1,254 @@
...
+        App.globalPut(permissions_manager, Txn.accounts[1]),
+        Return(Int(1))
+    ])
+
+    # Check basic tx (call to controller, clawback tx)
+    verify_basic_calls = And(
+        # verify first transaction
+        # call to controller smart contract - signed by asset sender
+        Txn.group_index() == Int(0), # this tx (call to controller) should be 1st in group
+        Gtxn[0].application_id() == Global.current_application_id(),
```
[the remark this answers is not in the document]
ratik21: you are right, sorry.

examples/permissioned-token/assets/controller.py
```diff
@@ -0,0 +1,254 @@
...
+
+        # verify 2nd tx
+        Gtxn[1].type_enum() == TxnType.AssetTransfer, # this should be clawback call (to transfer asset)
+        Gtxn[1].xfer_asset() == App.globalGet(token_id), # check asset_id passed through params
```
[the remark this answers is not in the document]
ratik21: good idea

examples/permissioned-token/assets/controller.py
```diff
@@ -0,0 +1,254 @@
...
+
+    # verifies that asset sender in 2nd tx
+    # - calls the controller smart contract
+    # - is also the sender of the payment tx (to pay fees of clawback escrow)
+    # - calls the permissions smart contract (ensures rules check)
+    verify_sender = And(
+        Gtxn[0].sender() == Gtxn[2].sender(),
+        Gtxn[0].sender() == Gtxn[1].asset_sender(),
+        Gtxn[3].sender() == Gtxn[1].asset_sender()
+    )
```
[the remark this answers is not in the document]
ratik21: I'm not sure that we can. Because now we're doing 2 transfers - one normal transfer & one force transfer. So in clawback we cannot check force_transfer (as there we need to ensure that sender is the token manager). That's why we check for the sender here, & clawback ensures that the controller is called.
Another way to do this is to hardcode ASSET_MANAGER in `clawback.py`.

examples/permissioned-token/assets/controller.py
ratik21: alright

examples/permissioned-token/assets/permissions.py
ratik21: yes, noted.

examples/permissioned-token/assets/controller.py
ratik21: Ok, if we don't care what are the other transactions then let's remove.

examples/permissioned-token/scripts/issuance/issue.js
ratik21: ok

examples/permissioned-token/scripts/1-setup-apps.js
ratik21: updated

examples/permissioned-token/assets/permissions.py
ratik21: ok, will remove this check

examples/permissioned-token/assets/permissions.py
ratik21: We need this check because we fetch *permissions_manager* from controller's global state. And `Txn.sender()` should be permissions_manager

examples/permissioned-token/assets/permissions.py
ratik21: ok

examples/permissioned-token/assets/permissions.py
ratik21: It looks correct on my editor, and while viewing file. In diff it looks incorrect

ratik21: it was an issue with my editor

examples/permissioned-token/assets/clawback.py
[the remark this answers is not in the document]
robert-zaremba: The first condition can be verified here. 
Let's add a comment to the case what conditions we need to check overall.

examples/permissioned-token/scripts/issuance/issue.js
robert-zaremba: ah, right, I read it too fast. I thought it's to sign and send.

examples/permissioned-token/assets/permissions.py
robert-zaremba: It's good to use some extension in editor to mark tabs. For, example, I mark tabs with a low contrast `>` symbol: 

image

examples/permissioned-token/assets/controller.py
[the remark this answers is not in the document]
robert-zaremba: So, let's think:
1. for all transfers, controller must be always called with permissions - this we can check here (in controller)
2. For a simple transfer, we need to have a payment transaction in a group - but it doesn't matter who is paying, as long as there is enough payment. For example, you want to transfer your assets, but I will agree to pay for it, so I will attach my signature to the group with a payment transaction from my account.
3. For checking that the permissions contract was called by the asset sender, we can add this line to persmissions 

```
Txn.accounts[0] == Gtxn[1].asset_sender()
````

Ideally, the escrow contract checks a right transaction composition. Controller checks that it's called with the token transfer and that the right permissions contract is called. Permissions contract just check the sender and receiver (it can verify it by using the first transactoin asset sender / asset receiver). 

What do you think about this approach?
Permissions contract should still work if it's called alone. It's not really important who is calling it as long as we don't save data during a transfer. But even then we could

**Review by robert-zaremba — COMMENTED** [review]
responding on comments. I think it looks good - let's have  a final chat tomorrow.

File: examples/permissioned-token/assets/permissions.py (hunk 57 above)
Comment (robert-zaremba) [reply examples/permissioned-token/assets/permissions.py]: ah, correct. Please add this explanation to the comment in the code.

File: examples/permissioned-token/assets/controller.py [hunk 72]
```diff
@@ -0,0 +1,237 @@
...
+    permission_id = Bytes("perm_app") # permissions smart contract application index
+
+    # Always verify that the RekeyTo property of any transaction is set to the ZeroAddress
+    # unless the contract is specifically involved in a rekeying operation.
+    no_rekey_addr = Txn.rekey_to() == Global.zero_address()
+
+    # retreive asset manager from Txn.ForeignAssets[0]
+    assetManager = AssetParam.manager(Int(0))
+
+    # Handles Controller SSC deployment. Only accepted if sender is asset manager.
```
robert-zaremba, line 31:
```suggestion
    # Handles Controller SSC deployment. Only accepted if sender is asset manager.
    # Must be called with the TOKEN_ID asset.
```

examples/permissioned-token/assets/controller.py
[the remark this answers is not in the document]
robert-zaremba: 1. let's update the conditions 
2. and document better this branch - what are other transactions (Gtx[3] and Gtx[1]) and what's the index of "this" transaction.

robert-zaremba (approved):
Thanks for having this many cycles. Approving with comments as we discussed.

man-of-code (approved):
LGTM. The new changes make the template more user-friendly.

examples/permissioned-token/assets/permissions.py
ratik21: thanks for info!

PR #293, opened 2021-04-23
update ModifyAsset transaction to pass optional params

ratik21:


examples/restricted-assets/scripts/1-assetConfig.js
```diff
@@ -22,21 +22,14 @@ async function run (runtimeEnv, deployer) {
...
-
   /** Update clawback address to escrow **/
   console.log('* Updating asset clawback to escrow *');
   const assetConfigParams = {
     type: types.TransactionType.ModifyAsset,
     sign: types.SignType.SecretKey,
     fromAccount: creator,
     assetID: assetInfo.assetIndex,
-    fields: assetModFields,
+    fields: { clawback: escrowAddress }, // only pass the field you want to update
```
man-of-code, line 32: Instead of removing `assetModFields`, can we change it to the hardcoded object here.

robert-zaremba: the idea is that we update only the `clawback` address, all rest stays as it's defined in`asa.yaml`

packages/algob/src/internal/deployer.ts
```diff
@@ -96,6 +96,15 @@ class DeployerBasicMode {
     return await this.algoOp.waitForConfirmation(txId);
   }
 
+  /**
+   * Queries blockchain using algodv2 for asset information by index
+   * @param assetIndex asset index
+   * @returns asset info from network
+   */
+  async getAssetByID (assetIndex: number): Promise<algosdk.AssetInfo> {
```
robert-zaremba, line 104: the type should be `number | BigInt`

packages/algob/src/lib/algo-operator.ts
```diff
@@ -44,6 +44,7 @@ export interface AlgoOperator {
     txWriter: txWriter,
     scTmplParams?: SCParams) => Promise<SSCInfo>
   waitForConfirmation: (txId: string) => Promise<algosdk.ConfirmedTxInfo>
+  getAssetByID: (assetIndex: number) => Promise<algosdk.AssetInfo>
```
robert-zaremba, line 47: ditto

packages/algob/src/lib/tx.ts
```diff
@@ -174,7 +174,25 @@ async function mkTx (
       txIdxMap.set(index, [name, { total: 1, decimals: 1, unitName: "MOCK" }]);
       break;
     }
+    case TransactionType.ModifyAsset: {
+      // fetch asset mutable properties from network and set them (if they are not passed)
+      // before modifying asset
+      const assetInfo = await deployer.getAssetByID(txn.assetID);
+      txn.fields.manager =
+        txn.fields.manager !== "" ? txn.fields.manager ?? assetInfo.params.manager : undefined;
```
robert-zaremba, line 182: It's more clear if we put a condition when to update. How about this:
```suggestion
      if (txn.fields.manager === "" ) txn.fields.manager = undefined;
```

packages/algob/src/types.ts
```diff
@@ -493,6 +493,10 @@ export interface Deployer {
    * Queries blockchain for a given transaction and waits until it will be processed. */
   waitForConfirmation: (txId: string) => Promise<algosdk.ConfirmedTxInfo>
 
+  /**
+   * Queries blockchain using algodv2 for asset information by index  */
+  getAssetByID: (assetIndex: number) => Promise<algosdk.AssetInfo>
```
robert-zaremba, line 498: type should be updated

robert-zaremba (approved):
pre-approving.

packages/algob/src/internal/deployer.ts
ratik21: yes

packages/algob/src/lib/algo-operator.ts
ratik21: updated

packages/algob/src/types.ts
ratik21: done

packages/algob/src/lib/tx.ts
ratik21: yes, thanks for suggestion. Updating.

PR #294, opened 2021-04-26
Add updateSSC in group tx

amityadav0:


robert-zaremba (commented):
Shall we add a test?

amityadav0:
To test this transaction we will have to add example because this test is possible on network only.

robert-zaremba (approved):

PR #295, opened 2021-04-26
Support asa ID in `deployer.OptIn*`

man-of-code:


amityadav0:
Please add a test, showing we can retreive asset info from checkpoint using asa id. You can refer to some Tests in checkpoints test section.

man-of-code:
> Please add a test, showing we can retreive asset info from checkpoint using asa id. You can refer to some Tests in checkpoints test section.

@amityadav0  My understanding is that this task was to provide support for ASA not deployed by `deployer` (not existing in checkpoints). @ratik21 Has added a function in this [PR](https://github.com/scale-it/algo-builder/pull/293) with the same functionality you are talking about.

amityadav0:
@man-of-code right, we may have overestimated this task.

packages/algob/src/internal/deployer.ts
```diff
@@ -184,27 +184,51 @@ class DeployerBasicMode {
...
    * @param flags Transaction flags
    */
-  async optInLsigToASA (asaName: string, lsig: LogicSig, flags: rtypes.TxParams): Promise<void> {
-    await this.algoOp.optInLsigToASA(asaName, this._getASAInfo(asaName).assetIndex, lsig, flags);
+  async optInLsigToASA (asa: string, lsig: LogicSig, flags: rtypes.TxParams): Promise<void> {
+    try {
+      const asaId = this._getASAInfo(asa).assetIndex;
+      await this.algoOp.optInLsigToASA(asa, asaId, lsig, flags);
+    } catch (error) {
+      console.log(error);
```
amityadav0, line 225: let's change this log here, maybe to asset not found in checkpoints proceeding to process it as asset ID

packages/algob/src/internal/deployer.ts
```diff
@@ -184,27 +184,51 @@ class DeployerBasicMode {
...
+      if (!Number(asa)) {
+        throw Error("Please provide a valid Number to be used as ASA ID");
```
amityadav0, line 227: This error should show both cases - 1. asset not in checkpoints also asset is not valid ASA ID

packages/algob/src/internal/deployer.ts
```diff
@@ -184,27 +184,51 @@ class DeployerBasicMode {
...
+        this._getAccount(accountName),
+        flags);
+    }
   }
 
   /**
    * Description: Opt-In to ASA for a contract account (represented by logic signture).
    * The opt-in transaction is signed by the logic signature
-   * @param asaName ASA name
+   * @param asa ASA (name/ID)
```
amityadav0, line 216: Please add here ID can be used for ASA not existing in checkpoints

packages/algob/src/internal/deployer.ts
```diff
@@ -184,27 +184,51 @@ class DeployerBasicMode {
...
+  async optInAcountToASA (asa: string, accountName: string, flags: rtypes.TxParams): Promise<void> {
+    try {
+      const asaId = this._getASAInfo(asa).assetIndex;
+      await this.algoOp.optInAcountToASA(
+        asa,
+        asaId,
+        this._getAccount(accountName),
+        flags);
+    } catch (error) {
+      console.log(error);
```
amityadav0, line 200: Please update this function as per suggestions in Lsig function

amityadav0 (changes requested):
I don't think we support this in `executeTransaction`. Please add these checks there.

man-of-code:
> I don't think we support this in `executeTransaction`. Please add these checks there.

@amityadav0 while executing any asset related transaction using `executeTransaction` we are already sending AssetID.

amityadav0 (approved):

PR #296, opened 2021-04-27
Update transaction execution parameters & upgrade pyteal version

ratik21:
- Updated type structure of *execParams* (includes support for transaction from rekeyed account as well)
- add support for rekeying
- upgrade pyteal

### DIFF
--- a/docs/guide/testing-teal.md
+++ b/docs/guide/testing-teal.md
@@ -85,7 +85,7 @@ Let's try to execute a transaction where a user (say `john`) can withdraw funds 
   const txnParams: ExecParams = {
     type: TransactionType.TransferAlgo, // payment
     sign: SignType.LogicSignature;
-    fromAccount: escrow.account,
+    fromAccountAddr: escrow.address,
     toAccountAddr: john.address,
     amountMicroAlgos: 100,
     payFlags: { totalFee: 1000 },
@@ -141,7 +141,7 @@ Let's try to execute a transaction where a user (say `john`) will use delegated 
     const txnParams: ExecParams = {
       type: TransactionType.TransferAlgo, // payment
       sign: SignType.LogicSignature,
-      fromAccount: john.account,
+      fromAccountAddr: john.address,
       toAccountAddr: bob.address,
       amountMicroAlgos: 100,
       payFlags: { totalFee: 1000 }
--- a/examples/asa/scripts/transfer/gold-delegated-lsig.js
+++ b/examples/asa/scripts/transfer/gold-delegated-lsig.js
@@ -18,7 +18,7 @@ async function run (runtimeEnv, deployer) {
   let txnParam = {
     type: types.TransactionType.TransferAsset,
     sign: types.SignType.LogicSignature,
-    fromAccount: goldOwner,
+    fromAccountAddr: goldOwner.addr,
     toAccountAddr: john.addr,
     amount: 500,
     assetID: assetID,
@@ -45,7 +45,7 @@ async function run (runtimeEnv, deployer) {
   txnParam = {
     type: types.TransactionType.TransferAlgo,
     sign: types.SignType.LogicSignature,
-    fromAccount: goldOwner,
+    fromAccountAddr: goldOwner.addr,
     toAccountAddr: bob.addr,
     amountMicroAlgos: 58,
     lsig: logicSignature,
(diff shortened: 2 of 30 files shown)


docs/guide/testing-teal.md
```diff
@@ -85,7 +85,7 @@ Let's try to execute a transaction where a user (say `john`) can withdraw funds
   const txnParams: ExecParams = {
     type: TransactionType.TransferAlgo, // payment
     sign: SignType.LogicSignature;
-    fromAccount: escrow.account,
+    fromAccountAddr: escrow.address,
```
robert-zaremba, line 88: :+1:

examples/asa/scripts/transfer/gold-delegated-lsig.js
```diff
@@ -18,7 +18,7 @@ async function run (runtimeEnv, deployer) {
   let txnParam = {
     type: types.TransactionType.TransferAsset,
     sign: types.SignType.LogicSignature,
-    fromAccount: goldOwner,
+    fromAccountAddr: goldOwner.addr,
```
robert-zaremba, line 21: Why sometimes we have `.address` and other times `.addr`?

robert-zaremba (approved):
Looks good. Let's check if we can unify `.addr` and `.address`.

man-of-code (approved):
Working fine.

packages/runtime/test/src/runtime.ts
```diff
@@ -48,26 +48,36 @@ describe("Logic Signature Transaction in Runtime", function () {
...
 
   it("should not verify signature because alice sent it", () => {
-    txnParam.fromAccount = alice.account;
+    const invalidParams: ExecParams = {
+      ...txnParam,
+      sign: SignType.LogicSignature,
+      fromAccountAddr: alice.account.addr,
+      lsig: lsig
+    };
+    // txnParam.fromAccount = alice.account;
```
amityadav0, line 57: please remove this comment if not needed

packages/runtime/test/integration/close-clear-ssc.ts
```diff
@@ -92,16 +92,20 @@ describe("ASC - CloseOut from Application and Clear State", function () {
...
+      fromAccount: alice.account, // sending txn sender other than creator (john), so txn should be rejected
+      appId: appId,
+      payFlags: {}
+    };
 
     expectRuntimeError(
-      () => runtime.executeTx(closeOutParams),
-      RUNTIME_ERRORS.TEAL.REJECTED_BY_LOGIC
+      () => runtime.executeTx(invalidParams),
+      RUNTIME_ERRORS.GENERAL.APP_NOT_FOUND
```
amityadav0, line 108: This should be rejected right ?, but i think in this case app is being deleted

packages/runtime/src/types.ts
```diff
@@ -270,62 +280,62 @@ export interface DeploySSCParam extends BasicParams, SSCOptionalFlags {
   globalBytes: number
   approvalProg?: Uint8Array
   clearProg?: Uint8Array
-}
+};
 
-export interface OptInSSCParam extends BasicParams, SSCOptionalFlags {
+export type OptInSSCParam = BasicParams & SSCOptionalFlags & {
```
amityadav0, line 285: what is the difference between interface and types ? why we are using type here ?

packages/runtime/test/src/runtime.ts
ratik21: ok

examples/asa/scripts/transfer/gold-delegated-lsig.js
ratik21: In runtime's *AccountStore*, we have acc.account (which is the account with `addr`, `sk`) & acc.address as well. 

`.address` is being used in runtime tests, and otherwise with deployer we are using `account.addr` (eg. goldOwner.addr)

packages/runtime/test/integration/close-clear-ssc.ts
ratik21: nice catch

packages/runtime/src/types.ts
ratik21: With interface i wanted to do something like: 
```js
interface BasicParams extends (SignWithLsig | SignWithSk) { .. }
```
But it's not possible with interfaces (extending unions). And this is the correct way because *lsig* should not be optional, it should be a required param if SignType is logic signature. Also we don't need to take complete account in this case, only [`from`, `to`] addresses. 

Read more about the difference here - [https://medium.com/@martin_hotell/interface-vs-type-alias-in-typescript-2-7-2a8f1777af4c](https://medium.com/@martin_hotell/interface-vs-type-alias-in-typescript-2-7-2a8f1777af4c). Picked `type` because of last points in this article.

robert-zaremba: With type you create a type variable (like `var`, `const` ....) - so you need assignment. 
With interface you create a named type.

PR #298, opened 2021-04-30
Change the way how we assign smart contract arguments

ratik21:


packages/runtime/src/types.ts
```diff
@@ -257,6 +258,7 @@ interface SignWithLsig {
   sign: SignType.LogicSignature
   fromAccountAddr: AccountAddress
   lsig: LogicSig
+  args?: LogicSigArgs // state smart contract args (used when calling smart contract)
```
amityadav0, line 261:
```suggestion
  args?: LogicSigArgs // stateless smart contract args
```

packages/algob/src/lib/tx.ts
```diff
@@ -119,6 +119,9 @@ function signTransaction (txn: Transaction, execParams: rtypes.ExecParams): Uint
       return txn.signTxn(execParams.fromAccount.sk);
     }
     case rtypes.SignType.LogicSignature: {
+      const logicSig = execParams.lsig;
+      // Object.assign(logicSig.args, execParams.args ?? []);
+      logicSig.args = execParams.args ?? [];
```
amityadav0, line 124: how about using `execParams.lsig.args = execParams.args ?? []` ?
this seems more `intuitive`

amityadav0 (approved):

PR #303, opened 2021-05-03
Fix Bug

man-of-code:


packages/runtime/src/interpreter/opcode-list.ts
```diff
@@ -2043,6 +2044,10 @@ export class GetAssetDef extends Op {
           break;
         default:
           value = stringToBytes(AssetDefinition[s] as string);
+          addr = Buffer.from(value as any).toString("utf-8");
+          if (isValidAddress(addr)) {
+            value = decodeAddress(addr).publicKey;
+          }
```
ratik21, line 2050: No let's not do it this way. Let do it like:
```js
const value = AssetDefinition[s];
if (typeof value === "string") {
 ..check for algorand address
} else {
 value = stringToBytes(AssetDefinition[s] as string);
}
```

man-of-code: okay

ratik21 (approved):

PR #305, opened 2021-05-04
Fix Asset Modify Txn

man-of-code:
Currently, AssetModify Txn is not being handled properly. I have created a helper function for this.

### DIFF
--- a/packages/runtime/src/account.ts
+++ b/packages/runtime/src/account.ts
@@ -7,6 +7,7 @@ import { generateAccount } from "algosdk";
 
 import { RUNTIME_ERRORS } from "./errors/errors-list";
 import { RuntimeError } from "./errors/runtime-errors";
+import { checkAndSetASAFields } from "./lib/asa";
 import {
   ALGORAND_ACCOUNT_MIN_BALANCE, APPLICATION_BASE_FEE, ASSET_CREATION_FEE, MAX_ALGORAND_ACCOUNT_APPS,
   MAX_ALGORAND_ACCOUNT_ASSETS,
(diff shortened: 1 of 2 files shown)


packages/runtime/src/account.ts
```diff
@@ -24,6 +24,20 @@ const StateMap = "key-value";
...
+
+  if (!(fields.reserve !== "" && asset.reserve === "") && fields.reserve !== undefined) asset.reserve = fields.reserve;
+  else if (fields.reserve !== undefined) throw new RuntimeError(RUNTIME_ERRORS.ASA.BLANK_ADDRESS_ERROR);
+
+  if (!(fields.freeze !== "" && asset.freeze === "") && fields.freeze !== undefined) asset.freeze = fields.freeze;
+  else if (fields.freeze !== undefined) throw new RuntimeError(RUNTIME_ERRORS.ASA.BLANK_ADDRESS_ERROR);
+
+  if (!(fields.clawback !== "" && asset.clawback === "") && fields.clawback !== undefined) asset.clawback = fields.clawback;
+  else if (fields.clawback !== undefined) throw new RuntimeError(RUNTIME_ERRORS.ASA.BLANK_ADDRESS_ERROR);
+}
```
robert-zaremba, line 39: maybe we can use loop?
```
for (x of ['manager', 'reserve', 'freeze', 'clawback'])
  if ....
```

man-of-code: Sure.

robert-zaremba: Also, let's the conditions:
```
if (fields.clawback !== "" && fields.clawback !== undefined && asset.clawback === "") throw ...
else if (fields.clawback !== "") assets.clawback = fields.clawback
```

man-of-code: The `!` operator is just for the conditions `(fields.clawback !== "" && asset.clawback === "")` and not the whole line. Also, the loop implementation isn't working.

robert-zaremba: What's the condition to throw an error? I think it's the one I put above.

man-of-code: This implementation fails when `fields.clawback` is undefined.

robert-zaremba: So, maybe we can remove the first component in the if: `fields.clawback !== ""`. What do you think?
I prefer to firstly have a condition for throw, and then to have a condition to update.

robert-zaremba: can we use the loop instead of copy-pasting the `if-else`

packages/runtime/src/account.ts
```diff
@@ -24,6 +24,37 @@ const StateMap = "key-value";
 const globalState = "global-state";
 const localStateSchema = "local-state-schema";
 
+function isSetOrDefined (value: string | undefined): boolean {
```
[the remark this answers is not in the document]
man-of-code: Yes. That would make more sense. I'll move it.

packages/runtime/src/account.ts
```diff
@@ -24,6 +24,37 @@ const StateMap = "key-value";
...
+  if (value !== undefined && value !== "") return true;
+  return false;
+}
+
+function checkAndSetAssetFields (fields: AssetModFields, asset: AssetDef): void {
```
ratik21, line 32: Please add a doc string for this function

packages/runtime/src/account.ts
ratik21: @man-of-code could you check if this works:
```js
for (const x of ['manager', 'reserve', 'freeze', 'clawback']) {
  const asaField = fields[x as keyof AssetModFields];
  if (isSetOrDefined(asaField) && !isSetOrDefined(asaField)) {
    throw new RuntimeError(RUNTIME_ERRORS.ASA.BLANK_ADDRESS_ERROR);
  } else if (asaField !== undefined && isSetOrDefined(asaField)) {
    asset[x as keyof AssetModFields] = asaField;
  }
}
```

packages/runtime/src/lib/asa.ts
```diff
@@ -96,3 +97,27 @@ export function loadASAFile (accounts: AccountMap | RuntimeAccountMap): ASADefs
...
+ * Check and Change ASA fields
+ * @param fields Custom ASA fields
+ * @param asset Defined ASA fields
+ */
+export function checkAndSetASAFields (fields: AssetModFields, asset: AssetDef): void {
+  for (const x of ['manager', 'reserve', 'freeze', 'clawback']) {
+    const customField = fields[x as keyof AssetModFields];
+    const asaField = asset[x as keyof AssetModFields];
+    if (isDefined(customField) && !isDefined(asaField)) {
+      // Check if custom field is set and defined and ASA field is blank field
```
robert-zaremba, line 116: this comment should be above `if`

robert-zaremba (approved):
pre approving. left one more commment.

Issue #302, opened 2021-05-03
In grouped transaction: Txn refers to Gtxn 0


gidonkatten:
This doesn't seem to be a problem when using one stateful contract and one stateless contract although further investigation may be needed.

When executing a grouped transaction where there is a stateful contract call in position 0 and a stateful contract call in postion x (where x > 0) then at both positions `txn` seems to be referencing `txn` at postion 0. This contradicts the documentation https://developer.algorand.org/docs/reference/teal/opcodes/#txn-f which suggests `txn` refers to the current transaction. 

I believe this problem impacts all scenarios where `txn` is accessed whether implicitly or explicity eg. `asset_params_get` and `txn accounts`.

To recreate:
```
# test1.py
from pyteal import *

def contract():
    program = Cond(
        [Txn.application_id() == Int(0), Int(1)],  # on creation
        [Txn.on_completion() == OnComplete.DeleteApplication, Int(0)],
        [Txn.on_completion() == OnComplete.UpdateApplication, Int(0)],
        [Txn.on_completion() == OnComplete.CloseOut, Int(0)],
        [Txn.on_completion() == OnComplete.OptIn, Int(0)],
        [Txn.on_completion() == OnComplete.NoOp, Txn.group_index() == Int(0)]
    )
    return program

if __name__ == "__main__":
    print(compileTeal(contract(), Mode.Application))
```

```
# test2.py
from pyteal import *

def contract():
    program = Cond(
        [Txn.application_id() == Int(0), Int(1)],  # on creation
        [Txn.on_completion() == OnComplete.DeleteApplication, Int(0)],
        [Txn.on_completion() == OnComplete.UpdateApplication, Int(0)],
        [Txn.on_completion() == OnComplete.CloseOut, Int(0)],
        [Txn.on_completion() == OnComplete.OptIn, Int(0)],
        [Txn.on_completion() == OnComplete.NoOp, Txn.group_index() == Int(1)]
    )
    return program

if __name__ == "__main__":
    print(compileTeal(contract(), Mode.Application))

```

```
# test file
    it('test', () => {
      const test1 = getProgram('test1.py');
      const test2 = getProgram('test2.py');

      const appId1 = runtime.addApp(
        {
          sender: master.account,
          localInts: 0,
          localBytes: 0,
          globalInts: 0,
          globalBytes: 0,
          appArgs: []
        },
        {},
        test1,
        clearProgram
      );

      const appId2 = runtime.addApp(
        {
          sender: master.account,
[...]

PR #306, opened 2021-05-04
Fix group idx opcode and txn context

amityadav0:
Closes: #302

### DIFF
--- a/packages/runtime/src/ctx.ts
+++ b/packages/runtime/src/ctx.ts
@@ -304,17 +304,20 @@ export class Ctx implements Context {
           break;
         }
         case TransactionType.CallNoOpSSC: {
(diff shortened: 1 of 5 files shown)


packages/runtime/src/ctx.ts
```diff
@@ -304,17 +304,20 @@ export class Ctx implements Context {
           break;
         }
         case TransactionType.CallNoOpSSC: {
+          this.tx = this.gtxs[idx]; // update current tx to index of stateful
```
robert-zaremba, line 307:
```suggestion
          this.tx = this.gtxs[idx]; // update current tx to the requested index
```

robert-zaremba (approved):
Preapproving. Let's update the comment and add a test.

PR #312, opened 2021-05-10
Permissioned token tests

man-of-code:


man-of-code:
Runtime Bugs Found:
+ In `runtime/src/ctx.ts`, skipped asset frozen check if to and from address are same. (This was making the optOut txn fail)
+ In `runtime/src/lib/txn.ts`, transactions of type other than `acfg` (Asset Config txn) were being parsed as `acfg` txn.

examples/permissioned-token/test/common.js
```diff
@@ -0,0 +1,325 @@
...
+  runtime.optIntoASA(assetIndex, address, {});
+}
+
+// Opt-In address to Permissions SSC
+function optInToPermissions (runtime, address, permissionsAppId) {
+  runtime.optInToApp(address, permissionsAppId, {}, {});
+}
+
+// Issue some tokens to the passed account
+function issue (runtime, manager, acc, amount, controllerAppID, assetIndex, lsig) {
```
amityadav0, line 48: Please add doc string here, manager: Asset manager, acc: ?, amount: Issue amount,.
maybe change `acc` to `receiver`.

examples/permissioned-token/test/failing-paths.js
```diff
@@ -0,0 +1,171 @@
...
+
+  it('Token Issuance', () => {
+    // Cannot issue before opting-in
+    assert.throws(() => issue(runtime, alice, elon, 20, controllerAppID, assetIndex, lsig), 'RUNTIME_ERR1404');
+
+    // Opting-in
+    optInToASA(runtime, elon.address, assetIndex);
+    syncAccounts();
+
+    // Cannot issue after killing token
```
amityadav0, line 51: is it possible to transfer tokens after killing ?

examples/permissioned-token/test/happy-paths.js
```diff
@@ -0,0 +1,188 @@
...
+      accounts: [elon.address]
+    };
+
+    // works as fromAccount is the current SSC manager
+    runtime.executeTx(txn);
+    syncAccounts();
+    whitelist(runtime, elon, bob.address, assetIndex, controllerAppID, permissionsAppId);
+  });
+
+  it('Force Transfer', () => {
```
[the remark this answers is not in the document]
amityadav0: also when assert_reserve = Gtxn[1].reserve

packages/runtime/src/ctx.ts
```diff
@@ -150,8 +150,14 @@ export class Ctx implements Context {
...
-    this.assertAssetNotFrozen(txnParam.assetID, fromAccountAddr);
-    this.assertAssetNotFrozen(txnParam.assetID, txnParam.toAccountAddr);
+    if (txnParam.payFlags.closeRemainderTo !== undefined) {
+      this.assertAssetNotFrozen(txnParam.assetID, txnParam.payFlags.closeRemainderTo);
+    }
+    if (txnParam.amount > 0) {
+      this.assertAssetNotFrozen(txnParam.assetID, fromAccountAddr);
+      this.assertAssetNotFrozen(txnParam.assetID, txnParam.toAccountAddr);
+    }
+
```
amityadav0, line 160: is this one also verified from algorand docs ? Please add a link above.

examples/permissioned-token/test/happy-paths.js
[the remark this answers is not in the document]
man-of-code: Okay

packages/runtime/src/lib/txn.ts
```diff
@@ -52,7 +52,8 @@ export function txnSpecbyField (txField: string, tx: Txn, gtxns: Txn[], tealVers
   if (assetTxnFields.has(txField)) {
     const s = TxnFields[tealVersion][txField];
     const assetMetaData = tx.apar;
-    result = assetMetaData[s as keyof AssetDefEnc];
+    if (assetMetaData === undefined) result = undefined;
+    else result = assetMetaData[s as keyof AssetDefEnc];
```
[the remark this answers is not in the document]
man-of-code: I am only returning the `GlobalZeroAddress` here. I couldn't find anything related to this bug on Algorand Docs.

examples/permissioned-token/test/failing-paths.js
amityadav0: please add a test for it.
does this mean balance of each user which has token becomes zero ?

amityadav0: or only transfer is prohibited

packages/runtime/src/ctx.ts
man-of-code: @amityadav0 Couldn't find any proper documentation on this. I think the checks I have made make pretty good sense.

packages/runtime/src/lib/txn.ts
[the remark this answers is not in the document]
man-of-code: I tried running the similar transactions on the network and since it also processes this property as undefined so it must be returning something(probably undefined or GlobalZeroAddress(Which is the default value I'm returning here)) to make the transaction not fail.

packages/runtime/src/ctx.ts
amityadav0: okay, please add comments describing the checks.

examples/permissioned-token/package.json
```diff
@@ -10,7 +10,7 @@
     "algob": "algob",
     "lint": "eslint --ext .js,.ts scripts",
     "lint:fix": "eslint --fix --ext .js,.ts scripts",
-    "test": "echo ok",
+    "test": "algob test",
```
ratik21, line 13: I think this be `mocha`

examples/permissioned-token/test/common.js
```diff
@@ -0,0 +1,380 @@
...
+let assetIndex;
+let lsig;
+let controllerAppID, permissionsAppId;
+
+let CLAWBACK_PROGRAM;
+let CONTROLLER_PROGRAM;
+let PERMISSIONS_PROGRAM;
+const CLEAR_STATE_PROGRAM = getProgram(CLEAR_STATE);
+
+const TRANSFER = 'str:transfer';
```
ratik21, line 27:
```suggestion
const TRANSFER_ARG = 'str:transfer';
```

examples/permissioned-token/test/common.js
```diff
@@ -0,0 +1,380 @@
...
+function syncAccounts (runtime) {
+  master = runtime.getAccount(master.address);
+  alice = runtime.getAccount(alice.address);
+  bob = runtime.getAccount(bob.address);
+  elon = runtime.getAccount(elon.address);
+}
+
+// Opt-In account to ASA
+function optInToASA (runtime, address, assetIndex) {
+  runtime.optIntoASA(assetIndex, address, {});
```
ratik21, line 39: For a single line, do we need to a separate function?

examples/permissioned-token/test/common.js
```diff
@@ -0,0 +1,380 @@
...
+  runtime.executeTx({
+    type: types.TransactionType.CallNoOpSSC,
+    sign: types.SignType.SecretKey,
+    fromAccount: manager.account,
+    appId: permissionsAppId,
+    payFlags: { totalFee: 1000 },
+    appArgs: ['str:add_whitelist'],
+    accounts: [address],
+    foreignAssets: [assetIndex],
+    foreignApps: [controllerAppID]
```
ratik21, line 121: I don't think we pass these two

examples/permissioned-token/test/common.js
```diff
@@ -0,0 +1,380 @@
...
+  const txGroup = [
+    {
+      type: types.TransactionType.CallNoOpSSC,
+      sign: types.SignType.SecretKey,
+      fromAccount: from.account,
+      appId: controllerAppID,
+      payFlags: { totalFee: 1000 },
+      appArgs: [TRANSFER],
+      accounts: [to.address]
+
```
ratik21, line 163:
```suggestion
```

examples/permissioned-token/test/common.js
```diff
@@ -0,0 +1,380 @@
...
+    },
+    {
+      type: types.TransactionType.CallNoOpSSC,
+      sign: types.SignType.SecretKey,
+      fromAccount: manager.account,
+      appId: permissionsAppId,
+      payFlags: { totalFee: 1000 },
+      appArgs: [TRANSFER],
+      accounts: [from.address, to.address]
+    }
```
ratik21, line 269: We should also add a test where we don't pass permissions app call in `force_transfer`

examples/permissioned-token/test/common.js
```diff
@@ -0,0 +1,380 @@
...
+  const asaDef = runtime.getAssetDef(assetIndex);
+
+  // Deploy Controller SSC
+  let sscFlags = {
+    sender: alice.account,
+    localInts: 0,
+    localBytes: 0,
+    globalInts: 2,
+    globalBytes: 0,
+    appArgs: [`int:${assetIndex}`],
```
ratik21, line 292: I don't think we pass appArgs while setting up controller

examples/permissioned-token/test/common.js
```diff
@@ -0,0 +1,380 @@
...
+    sscFlags, {}, CONTROLLER_PROGRAM, CLEAR_STATE_PROGRAM
+  );
+
+  // Deploy Clawback Lsig and Modify Asset
+  CLAWBACK_PROGRAM = getProgram(CLAWBACK, {
+    TOKEN_ID: assetIndex,
+    CONTROLLER_APP_ID: controllerAppID
+  });
+  lsig = runtime.getLogicSig(CLAWBACK_PROGRAM, []);
+  fund(runtime, master, lsig.address());
```
ratik21, line 306: Why do we "fund" an account here? Aren't we intializing these accounts with some `minBalance`?

examples/permissioned-token/test/failing-paths.js
```diff
@@ -0,0 +1,178 @@
...
+  whitelist,
+  fund,
+  killToken,
+  transfer,
+  optOut,
+  forceTransfer,
+  setupEnv
+} = require('./common');
+
+describe('Permissioned Token Tests -- Failing Paths', function () {
```
ratik21, line 16:
```suggestion
describe('Permissioned Token Tests - Failing Paths', function () {
```

examples/permissioned-token/test/failing-paths.js
```diff
@@ -0,0 +1,178 @@
...
+      elon,
+      assetIndex,
+      controllerAppID,
+      lsig,
+      permissionsAppId
+    ] = setupEnv();
+  });
+
+  it('Token Issuance', () => {
+    // Cannot issue before opting-in
```
ratik21, line 44:
```suggestion
    // Cannot issue before opt-in to ASA by receiver
```

examples/permissioned-token/test/failing-paths.js
```diff
@@ -0,0 +1,178 @@
...
+    assert.throws(() => issue(runtime, alice, elon, 20, controllerAppID, assetIndex, lsig), 'RUNTIME_ERR1404');
```
ratik21, line 45: `RUNTIME_ERR1404` doesn't look very intuitive, maybe we could be more specfic with err message here?

examples/permissioned-token/test/failing-paths.js
```diff
@@ -0,0 +1,178 @@
...
+    optInToASA(runtime, elon.address, assetIndex);
+    syncAccounts();
+
+    // Cannot issue after killing token
+    killToken(runtime, alice, controllerAppID);
+    assert.throws(() => issue(runtime, alice, elon, 20, controllerAppID, assetIndex, lsig), 'RUNTIME_ERR1009');
+  });
+
+  it('Kill Token', () => {
+    // Fails as wrong ASA manager is passed
```
ratik21, line 57: this comment should be in the test description

examples/permissioned-token/test/failing-paths.js
```diff
@@ -0,0 +1,178 @@
...
+    assert.throws(() => killToken(runtime, bob, controllerAppID), 'RUNTIME_ERR1009');
+  });
+
+  it('WhiteListing', () => {
+    // Only asset manager can whitelist
```
ratik21, line 62: No, permissions_manager can whitelist, it doesn't have to be asset.manager

examples/permissioned-token/test/failing-paths.js
```diff
@@ -0,0 +1,178 @@
...
+      sign: types.SignType.SecretKey,
+      fromAccount: bob.account, // Bob is not the asset manager
+      appId: permissionsAppId,
+      payFlags: { totalFee: 1000 },
+      appArgs: ['str:add_whitelist', `int:${controllerAppID}`],
+      accounts: [address],
+      foreignAssets: [assetIndex],
+      foreignApps: [controllerAppID]
+    }), 'RUNTIME_ERR1009');
+  });
```
ratik21, line 80: I think we'll need to update it, params look incorrect

examples/permissioned-token/test/failing-paths.js
```diff
@@ -0,0 +1,178 @@
...
+
+  it('Opt Out', () => {
+    // Opt-out wont work without opting in
+    assert.throws(() => optOut(runtime, alice, elon, assetIndex), 'RUNTIME_ERR1404');
```
ratik21, line 84: If we add optOut params here, then i think it will be more clear, as it is only a single transaction. What do you think?

examples/permissioned-token/test/failing-paths.js
```diff
@@ -0,0 +1,178 @@
...
+    // Opt-In to ASA
+    optInToASA(runtime, bob.address, assetIndex);
+    optInToASA(runtime, elon.address, assetIndex);
+    syncAccounts();
+
+    // Issue some tokens to sender
+    issue(runtime, alice, bob, 150, controllerAppID, assetIndex, lsig);
+    syncAccounts();
+
+    // Fails as only manager can perform force transfer
```
ratik21, line 125: What is the asset.manager here (in this test)?

examples/permissioned-token/test/failing-paths.js
```diff
@@ -0,0 +1,178 @@
...
+    );
+
+    // Fails as reciever will have more than 100 tokens post succesfull transaction
+    assert.throws(() =>
+      forceTransfer(runtime, bob, elon, 101, assetIndex, controllerAppID, permissionsAppId, lsig, alice),
+    'RUNTIME_ERR1009'
+    );
+  });
+
+  it('Token Transfer', () => {
```
ratik21, line 138: Headings/titles should be in `describe(..)`. `it(..)` should explain the specific test we are executing here

examples/permissioned-token/test/failing-paths.js
```diff
@@ -0,0 +1,178 @@
...
+
+    // Issue some tokens to Bob and Elon
+    optInToASA(runtime, elon.address, assetIndex);
+    optInToASA(runtime, bob.address, assetIndex);
+    syncAccounts();
+    issue(runtime, alice, elon, 50, controllerAppID, assetIndex, lsig);
+    issue(runtime, alice, bob, 50, controllerAppID, assetIndex, lsig);
+    syncAccounts();
+
+    // Cannot transfer directly
```
ratik21, line 154: What do we mean by "Cannot transfer directly"?

examples/permissioned-token/test/failing-paths.js
```diff
@@ -0,0 +1,178 @@
...
+      type: types.TransactionType.TransferAsset,
+      sign: types.SignType.SecretKey,
+      fromAccount: bob.account,
+      toAccountAddr: elon.address,
+      amount: amount,
+      assetID: assetIndex,
+      payFlags: {}
+    }), 'RUNTIME_ERR1505');
+
+    // Cannot transfer before being whitelisted
```
ratik21, line 165: We need to assert that the account is not whitelisted as well.

examples/permissioned-token/test/failing-paths.js
```diff
@@ -0,0 +1,178 @@
...
+    'RUNTIME_ERR1009'
+    );
+
+    // Both sender and receiver needs to be whitelisted before transferring tokens
+    whitelist(runtime, alice, bob.address, assetIndex, controllerAppID, permissionsAppId);
+    assert.throws(() =>
+      transfer(runtime, bob, elon, amount, assetIndex, controllerAppID, permissionsAppId, lsig),
+    'RUNTIME_ERR1009'
+    );
+  });
```
ratik21, line 177: - Please also add a failing test - regarding receiver balance (> 100).
- We should also add a test where we don't include call to contoller, or call to permissions (both should fail). Maybe change sender of clawback & controller (it should also fail)

examples/permissioned-token/test/happy-paths.js
```diff
@@ -0,0 +1,205 @@
...
+
+  it('Kill Token', () => {
+    // Works as correct ASA manager is passed
+    killToken(runtime, alice, controllerAppID);
+    assert.throws(() => issue(runtime, alice, elon, 20, controllerAppID, assetIndex, lsig), 'RUNTIME_ERR1009');
+  });
+
+  it('WhiteListing', () => {
+    // Only asset manager can whitelist
+    const address = elon.address;
```
ratik21, line 60:
```suggestion
    const elonAddr = elon.address;
```

examples/permissioned-token/test/happy-paths.js
```diff
@@ -0,0 +1,205 @@
...
+    optInToPermissions(runtime, address, permissionsAppId);
+    syncAccounts();
+    assert.isDefined(elon.getAppFromLocal(permissionsAppId));
+
+    // Works with correct ASA manager
```
ratik21, line 65: But we should "assert" that alice is the ASA.manager.
You can use `runtime.getASADef(..)`

examples/permissioned-token/test/happy-paths.js
```diff
@@ -0,0 +1,205 @@
...
+
+  it('Opt Out', () => {
+    // Opt-In
+    optInToASA(runtime, elon.address, assetIndex);
+    syncAccounts();
+    issue(runtime, alice, elon, 20, controllerAppID, assetIndex, lsig);
+    syncAccounts();
+    assert.equal(
+      Number(runtime.getAssetHolding(assetIndex, elon.address).amount),
+      20
```
ratik21, line 82: I think it should be:
```js
assert.equal(
  runtime.getAssetHolding(assetIndex, elon.address).amount,
  20n
);
```
Because if amount is bigger than 2^52, then current assertion would fail due to being maxed_out

examples/permissioned-token/test/happy-paths.js
```diff
@@ -0,0 +1,205 @@
...
+    );
+
+    // Works because opted-in
+    optOut(runtime, alice, elon, assetIndex);
+    syncAccounts();
+    assert.equal(
+      Number(runtime.getAssetHolding(assetIndex, elon.address).amount),
```
ratik21, line 89: Same as above, we should assert with bigints

packages/runtime/src/ctx.ts
```diff
@@ -150,8 +150,16 @@ export class Ctx implements Context {
     const toAssetHolding = this.getAssetHolding(txnParam.assetID, txnParam.toAccountAddr);
     txnParam.amount = BigInt(txnParam.amount);
 
-    this.assertAssetNotFrozen(txnParam.assetID, fromAccountAddr);
-    this.assertAssetNotFrozen(txnParam.assetID, txnParam.toAccountAddr);
+    // Check if `closeRemainderTo` is set then if it's ASA holding isn't frozen
+    if (txnParam.payFlags.closeRemainderTo !== undefined) {
+      this.assertAssetNotFrozen(txnParam.assetID, txnParam.payFlags.closeRemainderTo);
```
ratik21, line 155: We could just move this line to condition after line 172

packages/runtime/src/lib/txn.ts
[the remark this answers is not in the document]
ratik21: I think we need to check if this is verified in the docs or not.
Regarding the implementation, can we just do:
```js
result = assetMetaData?[s as keyof AssetDefEnc];
```
?

examples/permissioned-token/test/common.js
man-of-code: I thought this will be a bit more friendly.

packages/runtime/src/lib/txn.ts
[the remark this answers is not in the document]
man-of-code: I'll change the implementation, but I didn't find anything related to this on Algorand docs.

examples/permissioned-token/test/common.js
man-of-code: I tried making a force transfer where the receiver is asset reserve address, but the compiled TEAL is throwing error in that case.

examples/permissioned-token/test/common.js
man-of-code: We are funding lsig here.

examples/permissioned-token/test/failing-paths.js
man-of-code: We are just checking for the error thrown here. It is not being shown in the logs. I can add a comment here describing the error.

examples/permissioned-token/test/failing-paths.js
man-of-code: Okay..

examples/permissioned-token/test/failing-paths.js
man-of-code: The helper function is being used at several instances, if we directly make the transaction here it will add redundant code.

examples/permissioned-token/test/failing-paths.js
man-of-code: I'll add a comment. (Alice is the manager)

examples/permissioned-token/test/failing-paths.js
man-of-code: Direct ASA transfer. We can only transfer by making a clawback transaction.

examples/permissioned-token/test/failing-paths.js
man-of-code: > Please also add a failing test - regarding receiver balance (> 100).
I have added this. 

I don't think the second test is required, but I have added that too.

ratik21:
Added more tests and updated existing tests in #334.
Closing this PR.

PR #313, opened 2021-05-10
update package dev dependencies

robert-zaremba:
Some dependencies had important security releases, so I updated most of them.

NOTE: with this update, the linter is issuing more warnings. I've check on master, and running `yarn lint` also creates warnings. Warnings are accepted by linter hooks and huskey, however it seams we have too many of them - I will create a task in pivotal to clean that warnings in a next sprint.


ratik21:
> NOTE: with this update, the linter is issuing more warnings.

Yes, checked on my system as well. Linter is issuing warnings for imported types in all files (as they're not being used a value according to `no-unused-vars` rule). 
We can add plugin *plugin:@typescript-eslint/recommended* to remove those warnings, but it adds few extra warnings - `no-explicit-any`, `no-var`..etc - we can handle it in a separate task.

robert-zaremba:
Yes, I created the task last night. Thanks for review @ratik21

PR #315, opened 2021-05-10
Mute logs in tests

robert-zaremba:
Files to watch:
* examples/*/package.json
* packages/*/.mocharc.json

### DIFF
--- a/packages/algob/test/internal/core/config/config-loading.ts
+++ b/packages/algob/test/internal/core/config/config-loading.ts
@@ -205,14 +205,15 @@ describe("config loading", function () {
     });
 
     it("should detect conflict of account names", async function () {
-      const spy = sinon.spy(console, 'warn');
+      // console is mocked in package.json mocha options
+      let stub = console.warn as sinon.SinonStub;
+      stub.reset();
 
       kmdOp.addKmdAccount({ name: net.accounts[0].name, addr: "some-addr-1", sk: new Uint8Array(kmdOp.skArray) });
       await loadKMDAccounts(net, kmdOp);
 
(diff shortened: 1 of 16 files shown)


packages/algob/test/internal/core/config/config-loading.ts
```diff
@@ -205,14 +205,15 @@ describe("config loading", function () {
     });
 
     it("should detect conflict of account names", async function () {
-      const spy = sinon.spy(console, 'warn');
+      // console is mocked in package.json mocha options
+      let stub = console.warn as sinon.SinonStub;
```
ratik21, line 209: curious - what's the difference between using `spy` and `stub`? Is using `stub` better ?

test/setup.js
```diff
@@ -0,0 +1,20 @@
...
+beforeEach(() => {
+  for (let l of levels)
+    console[l].reset();
+})
+
+
+after(() => {
+  for (let l of levels)
+    console[l].restore();
+})
```
ratik21, line 20: could you  please add a doc string (maybe at the top of file), regarding suppressing console logs in setup?

ratik21 (approved):
left one comment

packages/algob/test/internal/core/config/config-loading.ts
robert-zaremba: I was usually using stubs, but it seams it's better to use spy here. I will update to spy. Thanks for asking.
here is an overview: https://www.javatpoint.com/mock-vs-stub-vs-spy

test/setup.js
robert-zaremba: yes ,good idea.

PR #320, opened 2021-05-11
update loadLogic in contract owned asa example

ratik21:


amityadav0:
Please update the asset receiver condition as well, that anybody can receive assets

ratik21:
> could you please update the asset receiver condition as well, that anybody can receive assets ?

Done

robert-zaremba (approved):

PR #321, opened 2021-05-12
Should not be able to close asset using clawback 

### LINKED ISSUE #311: Should not be able to close asset using clawback
An asset clawback transaction with a `closeRemainderTo` payFlag should throw an error (which is the behaviour of algorand network). Clawbacks cannot close assets from an address - only the actual owner can.

ratik21:
- Should not be able to close asset using clawback, added test
- Also checked that asset transfer is only account if both asset sender and receiver are not `frozen`, but if amount is 0 then txn is accepted. Added test

closes #311

### DIFF
--- a/packages/runtime/src/errors/errors-list.ts
+++ b/packages/runtime/src/errors/errors-list.ts
@@ -385,6 +385,12 @@ const runtimeAsaErrors = {
     message: "All of the created assets should be in creator's account",
     title: "Asset Total Error",
     description: "Asset Total Error"
+  },
+  CANNOT_CLOSE_ASSET_BY_CLAWBACK: {
+    number: 1509,
+    message: `Assetholding of account cannot be closed by clawback`,
+    title: "Close asset by clawback error",
+    description: "Cannot close asset by clawback"
   }
 };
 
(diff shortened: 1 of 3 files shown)


packages/runtime/src/errors/errors-list.ts
```diff
@@ -385,6 +385,12 @@ const runtimeAsaErrors = {
     message: "All of the created assets should be in creator's account",
     title: "Asset Total Error",
     description: "Asset Total Error"
+  },
+  CANNOT_CLOSE_ASSET_BY_CLAWBACK: {
+    number: 1509,
+    message: `Assetholding of account cannot be closed by clawback`,
+    title: "Close asset by clawback error",
+    description: "Cannot close asset by clawback"
```
robert-zaremba, line 393: please add more details into description. Maybe a link to documentation.

robert-zaremba (approved):
pre-approving

packages/runtime/src/errors/errors-list.ts
ratik21: I couldn't find in documentation regarding this behaviour. I got to know it by asking on discord. In documentation they mention `AssetCloseTo` param in clawback transaction reference - [https://developer.algorand.org/docs/reference/transactions/?query=clawback#asset-clawback-transaction](https://developer.algorand.org/docs/reference/transactions/?query=clawback#asset-clawback-transaction), which led to confusion

ratik21: I will add a link for general asset clawback behaviour

PR #322, opened 2021-05-12
update links for user and api docs

ratik21:


amityadav0:
link

Link is not working

ratik21:
> Link is not working

It's because they're not deployed yet. [https://scale-it.github.io/algo-builder](https://scale-it.github.io/algo-builder/) still points to the algob api docs, not the home page.

robert-zaremba (approved):

PR #327, opened 2021-05-13
add new command: algob sign-lsig

ratik21:
added new command `algob sign-lsig --account <acc> --file <input> --out <out-file>` to sign logic signature fetched from file


packages/algob/src/lib/lsig.ts
```diff
@@ -27,8 +28,30 @@ export async function getLsig (
...
 }
+
+/**
+ * Appends signature (using signer's sk) to multi-signed logic signature. If multisig is not found
+ * then new multisig is created
+ * eg. appending own signature to a signed lsig (received from multisignature account address network)
+ * @param lsig Logic Sig object
+ * @param signer: Signer Account which will sign the smart contract
+ * @param mparams: passed when signing a new multisig
+ * @returns multi signed logic signature (with appended signature from signer's sk)
```
robert-zaremba, line 44:
```suggestion
 * @returns multi signed logic signature (with appended signature using signer's sk)
```

robert-zaremba (commented):
This is great. Test is missing - would you be able to add a test? I would like to include this functionality in the Release.

robert-zaremba (approved):
pre-approving - but we still need a test.

ratik21:
Added tests.

PR #328, opened 2021-05-13
fix error message & updateSSC

amityadav0:
Previous message looks like this, even after using `--show-stack-traces` it is not clear why are we getting error
er3

Update error message
er1

UpdateSSC function doesn't update checkpoints, continuing on discord.


robert-zaremba (commented):
Please rebase on the latest `master` and add an entry to Changelog.

amityadav0:
> Please rebase on the latest `master` and add an entry to Changelog.

done

robert-zaremba (approved):

PR #332, opened 2021-05-17
Add a test of a group tx which  will execute a same app update

amityadav0:
- https://www.pivotaltracker.com/story/show/178116575

### DIFF
--- a/packages/runtime/test/fixtures/app-update/assets/approval_program.py
+++ b/packages/runtime/test/fixtures/app-update/assets/approval_program.py
@@ -1,0 +1,54 @@
+from pyteal import *
+
+def app():
+    '''
+    - An app with 2 global variables: `total: int` and `counter: int`.
+        - (on app creation), `counter = 0, n = 10`
+    - The app update accepts exactly one argument: `n: int`
+    - On update call the app verifies that: `n >= 1`,  increments `app.counter += 1` and:
+        * when `app.counter % 2 == 0` then update `app.total += n`
+        * otherwise, subtract: `app.total -= n`
+    '''
+    total = Bytes("total")
+    counter = Bytes("counter")
+
+    on_creation = Seq([
+        App.globalPut(counter, Int(0)),
+        App.globalPut(total, Int(10)),
+        Return(Int(1))
+    ])
+
+    common_checks = Assert(And(
+        Txn.application_args.length() == Int(1),
+        Btoi(Txn.application_args[0]) >= Int(1)
+    ))
+
+    on_success = Seq([
+        App.globalPut(total, App.globalGet(total) + Btoi(Txn.application_args[0])),
+        App.globalPut(counter, App.globalGet(counter) + Int(1)),
+        common_checks
+    ])
+
+    on_fail = Seq([
+        App.globalPut(total, App.globalGet(total) - Btoi(Txn.application_args[0])),
+        App.globalPut(counter, App.globalGet(counter) + Int(1)),
+        common_checks
+    ])
+
+    on_update = Cond(
+        [(App.globalGet(counter) % Int(2)) == Int(0), on_success],
+        [(App.globalGet(counter) % Int(2)) != Int(0), on_fail]
+    )
+
+    program = Cond(
+        [Txn.application_id() == Int(0), on_creation],  # on creation
+        [Txn.on_completion() == OnComplete.DeleteApplication, Return(Int(1))],
+        [Txn.on_completion() == OnComplete.UpdateApplication, on_update],
+        [Txn.on_completion() == OnComplete.CloseOut, Return(Int(1))],
+        [Txn.on_completion() == OnComplete.OptIn, Return(Int(1))],
+        [Txn.on_completion() == OnComplete.NoOp, Return(Int(1))]
+    )
+    return program
+
+if __name__ == "__main__":
+    print(compileTeal(app(), Mode.Application))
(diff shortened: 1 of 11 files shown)


packages/runtime/test/fixtures/app-update/assets/approval_program.py
```diff
@@ -0,0 +1,54 @@
+from pyteal import *
+
+def app():
+    '''
+    - An app with 2 global variables: `total: int` and `counter: int`.
+        - (on app creation), `counter = 0, n = 10`
```
ratik21, line 6: so `n` represents total here?

packages/runtime/test/fixtures/app-update/assets/approval_program.py
```diff
@@ -0,0 +1,54 @@
...
+
+    common_checks = Assert(And(
+        Txn.application_args.length() == Int(1),
+        Btoi(Txn.application_args[0]) >= Int(1)
+    ))
+
+    on_success = Seq([
+        App.globalPut(total, App.globalGet(total) + Btoi(Txn.application_args[0])),
+        App.globalPut(counter, App.globalGet(counter) + Int(1)),
+        common_checks
```
ratik21, line 29: I think we should do `common_checks` before `App.globalPut(..)`

packages/runtime/test/integration/app-update.ts
```diff
@@ -0,0 +1,119 @@
...
+        appID: appId,
+        newApprovalProgram: approvalProgram,
+        newClearProgram: clearProgram,
+        payFlags: {},
+        appArgs: ['int:5']
+      }
+    ];
+  });
+
+  it("First case", function () {
```
ratik21, line 71: Can we be a little more specific here?

packages/runtime/test/integration/app-update.ts
```diff
@@ -0,0 +1,119 @@
...
+    runtime.executeTx(groupTx);
+
+    const globalCounter = runtime.getGlobalState(appId, "counter");
+    const total = runtime.getGlobalState(appId, "total");
+    assert(globalCounter === 2n, "failed counter");
+    assert(total === 7n, "failed total");
+  });
+
+  it("Second case", function () {
+    groupTx[0].appArgs = ['int:5'];
```
ratik21, line 81: Maybe we should create a copy first, instead of directly updating `groupTx` array?

packages/runtime/test/integration/app-update.ts
```diff
@@ -0,0 +1,119 @@
...
+    groupTx[1].appArgs = ['int:5'];
+
+    expectRuntimeError(
+      function () {
+        for (let i = 0; i < 1000; ++i) {
+          runtime.executeTx(groupTx);
+        }
+      },
+      RUNTIME_ERRORS.TEAL.UINT64_UNDERFLOW
+    );
```
ratik21, line 103: Please add a comment here (or update test description), that why we get `UINT64_UNDERFLOW`

packages/runtime/test/fixtures/app-update/assets/approval_program.py
amityadav0: yes, updating

packages/runtime/test/fixtures/app-update/assets/approval_program.py
amityadav0: i think this will get reverted if checks don't pass right ?
but it will be little faster, will update

packages/runtime/test/integration/app-update.ts
amityadav0: added comments at top file, will move them to each test case

packages/runtime/test/integration/app-update.ts
amityadav0: we are using beforeEach so it will be updated each time anyway

packages/runtime/test/integration/app-update.ts
amityadav0: ok

packages/runtime/test/integration/app-update.ts
ratik21: ok

packages/runtime/test/integration/app-update.ts
ratik21: I am thinking we should update the test description as well, because when i run `yarn run test`, then it displays
```
First Case
Second Case
..and so on
```
I think we can just update `First Case:app_update(n=2), app_update(n=5)`

ratik21 (approved):

packages/runtime/test/integration/app-update.ts
```diff
@@ -0,0 +1,123 @@
...
+  /**
+   * Create 2 transactions in a group: `app_update(n=2), app_update(n=5)`.
+   * Check the expected `app.counter == 2, app.total=7`
+   */
+  it("First case", function () {
+    runtime.executeTx(groupTx);
+
+    const globalCounter = runtime.getGlobalState(appId, "counter");
+    const total = runtime.getGlobalState(appId, "total");
+    assert(globalCounter === 2n, "failed counter");
```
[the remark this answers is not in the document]
amityadav0: the main difference is we get the failing message so we can directly search it in file, it is easier to pinpoint.

packages/runtime/test/integration/app-update.ts
amityadav0: ok

amityadav0: updated to `Third case: (app_update(n=2) + app_update(n=5)) * 1000` so it looks like a group tx

PR #335, opened 2021-05-18
Permissioned token tests

ratik21:
- Happy Paths
    ✓ should issue token if sender is token reserve
    ✓ should kill token if sender is token manager
    ✓ should whitelist account if sender is permissions manager
    ✓ should opt-out of token successfully (using closeRemainderTo)
    ✓ should change Permissions SSC Manager if sender is current_permissions_manager
    ✓ should force transfer tokens between non reserve accounts successfully if sender is token manager
    ✓ should force transfer tokens without permission checks if receiver is asset reserve
    ✓ should transfer tokens between non reserve accounts successfully
    ✓ should update asset reserve account to another address if sender is asset manager

- Failing Paths
    - *Token Issuance*
      ✓ should not issue token if receiver is not opted in
      ✓ should not issue if token is killed
      ✓ should reject issuance tx if sender is not token reserve
      ✓ should reject issuance tx if trying to send asset using secret key instead of clawback
   - *Kill Token*
      ✓ should reject tx to kill token if sender is not token manager
   - *WhiteListing*
      ✓ should reject whitelist if account is not opted in to permissions
      ✓ should not whitelist account if sender is not current permissions manager
   - *Opt Out*
      ✓ should reject tx if user not opted-in
  -  *Change Permissions Manager*
      ✓ should fail if sender is not current permissions manager
  -  *Force Transfer(Clawback)*
      ✓ Should fail on force transfer if transaction group is not valid
      ✓ should reject transfer if transaction group is valid but controller.sender is not equal to asset.sender
      ✓ Should reject transfer is sender is not token manager
      ✓ Should reject force transfer if accounts are not whitelisted
      ✓ Should reject force transfer if accounts are whitelisted but receiver balance becomes > 100
      ✓ Should reject force transfer if token is killed
  -  *Token Transfer*
      ✓ Should fail if transaction group is not valid
      ✓ should reject transfer if transaction group is valid but controller.sender is not equal to asset.sender
      ✓ should reject transfer if sender trying to transfer token using secret key instead of clawback
      ✓ Should reject transfer if accounts are not whitelisted
      ✓ Should reject force transfer if accounts are whitelisted but receiver balance becomes > 100
      ✓ should reject transfer if rules are followed, but token is killed
  -  *Update token reserve*
      ✓ Should fail if transaction group is not valid
      ✓ should reject update if controller.sender is not asset manager
      ✓ should reject update if sender of asset config tx is not asset manager

### DIFF
--- a/examples/permissioned-token/assets/controller.py
+++ b/examples/permissioned-token/assets/controller.py
@@ -160,7 +160,7 @@ def approval_program(TOKEN_ID):
                     # If the receiver is the reserve address - old, or the new one being updated in the assset config tx (Gtxn[3]),
                     # then it can bypass permission checks
                     Gtxn[1].asset_receiver() == assetReserve.value(),
-                    Gtxn[1].asset_receiver() == Gtxn[3].config_asset_reserve()
+                    If(Global.group_size() > Int(3), Gtxn[1].asset_receiver() == Gtxn[3].config_asset_reserve(), Int(0))
                 ),
                 Int(1),
                 # else verify that permissions ssc is called
--- a/examples/permissioned-token/test/common.js
+++ b/examples/permissioned-token/test/common.js
@@ -1,379 +1,492 @@
-const {
-  getProgram
-} = require('@algo-builder/algob');
-const { Runtime, AccountStore, types } = require('@algo-builder/runtime');
+// const {
+//   getProgram
+// } = require('@algo-builder/algob');
+// const { Runtime, AccountStore, types } = require('@algo-builder/runtime');
 
-const minBalance = 20e6; // 20 ALGOs
+// const minBalance = 20e6; // 20 ALGOs
 
-const CLAWBACK = 'clawback.py';
-const CONTROLLER = 'controller.py';
-const PERMISSIONS = 'permissions.py';
-const CLEAR_STATE = 'clear_state_program.py';
+// const CLAWBACK_STATELESS_PROGRAM = 'clawback.py';
+// const CONTROLLER_APPROVAL_PROGRAM = 'controller.py';
+// const PERMISSIONS_APPROVAL_PROGRAM = 'permissions.py';
+// const CLEAR_STATE_PROGRAM = 'clear_state_program.py';
 
-const ALICE_ADDRESS = 'EDXG4GGBEHFLNX6A7FGT3F6Z3TQGIU6WVVJNOXGYLVNTLWDOCEJJ35LWJY';
+// const ALICE_ADDRESS = 'EDXG4GGBEHFLNX6A7FGT3F6Z3TQGIU6WVVJNOXGYLVNTLWDOCEJJ35LWJY';
 
-let master = new AccountStore(10000e6);
-let alice, bob, elon;
-let runtime;
-let assetIndex;
-let lsig;
-let controllerAppID, permissionsAppId;
+// let master = new AccountStore(10000e6);
+// let alice, bob, elon;
+// let runtime;
+// let assetIndex;
+// let lsig;
+// let controllerAppID, permissionsAppId;
 
-let CLAWBACK_PROGRAM;
-let CONTROLLER_PROGRAM;
-let PERMISSIONS_PROGRAM;
-const CLEAR_STATE_PROGRAM = getProgram(CLEAR_STATE);
+// let CLAWBACK_PROGRAM;
+// let CONTROLLER_PROGRAM;
+// let PERMISSIONS_PROGRAM;
 
-const TRANSFER_ARG = 'str:transfer';
+// const TRANSFER_ARG = 'str:transfer';
 
-// Sync Accounts
-function syncAccounts (runtime) {
-  master = runtime.getAccount(master.address);
-  alice = runtime.getAccount(alice.address);
-  bob = runtime.getAccount(bob.address);
-  elon = runtime.getAccount(elon.address);
-}
+// // Sync Accounts
+// function syncAccounts (runtime) {
+//   master = runtime.getAccount(master.address);
+//   alice = runtime.getAccount(alice.address);
+//   bob = runtime.getAccount(bob.address);
+//   elon = runtime.getAccount(elon.address);
+// }
 
-/**
- * - Setup token (ASA gold)
- * - Setup Controller asc
- * - Setup asset clawback + update asset clawback to escrow contract
- * - Setup permissions smart contract
- * NOTE: During setup - ASA.reserve, ASA.manager & current_permissions_manager is set as alice.address
- */
-function setupEnv () {
-  // Create Accounts and Env
-  alice = new AccountStore(minBalance, { addr: ALICE_ADDRESS, sk: new Uint8Array(0) });
-  bob = new AccountStore(minBalance);
-  elon = new AccountStore(minBalance);
-  runtime = new Runtime([master, alice, bob, elon]);
+// function deployASA(runtime, account) {
+//   return runtime.addAsset('gold', { creator: account });
+// }
 
-  // Deploy ASA
-  assetIndex = runtime.addAsset('gold', { creator: { ...alice.account, name: 'alice' } });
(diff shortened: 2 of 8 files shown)


examples/permissioned-token/assets/controller.py
```diff
@@ -160,7 +160,7 @@ def approval_program(TOKEN_ID):
                     # If the receiver is the reserve address - old, or the new one being updated in the assset config tx (Gtxn[3]),
                     # then it can bypass permission checks
                     Gtxn[1].asset_receiver() == assetReserve.value(),
-                    Gtxn[1].asset_receiver() == Gtxn[3].config_asset_reserve()
+                    If(Global.group_size() > Int(3), Gtxn[1].asset_receiver() == Gtxn[3].config_asset_reserve(), Int(0))
```
robert-zaremba, line 163: good we have tests :+1:

examples/permissioned-token/test/common.js
```diff
@@ -0,0 +1,379 @@
...
+    localInts: 0,
+    localBytes: 0,
+    globalInts: 2,
+    globalBytes: 0,
+    foreignAssets: [assetIndex]
+  };
+  CONTROLLER_PROGRAM = getProgram(CONTROLLER, { TOKEN_ID: assetIndex });
+  controllerAppID = runtime.addApp(
+    sscFlags, {}, CONTROLLER_PROGRAM, CLEAR_STATE_PROGRAM
+  );
```
robert-zaremba, line 67: let's move it to a separate functions:
* deployController
* deployClawback
* ....

examples/permissioned-token/test/common.js
```diff
@@ -0,0 +1,379 @@
...
+    runtime,
+    master,
+    alice,
+    bob,
+    elon,
+    assetIndex,
+    controllerAppID,
+    lsig,
+    permissionsAppId
+  ];
```
robert-zaremba, line 135: I think we should have a class  for this:
```
class Context:
  let runtime: Runtime
  let alice: ...

  constructor(runtime, alice, ...) ...
  ... 

  syncAccounts( { .... } )
}
```
This function can create a new Context and return it.
This way we don't have a flying global variables, and we encapsulate the runtime and the required accounts.

examples/permissioned-token/test/failing-paths.js
```diff
@@ -0,0 +1,627 @@
...
+  let runtime, master, alice, bob, elon;
+  let lsig, assetIndex, controllerAppID, permissionsAppId;
+  let asaDef, asaReserve, asaManager;
+
+  function syncAccounts () {
+    master = runtime.getAccount(master.address);
+    alice = runtime.getAccount(alice.address);
+    bob = runtime.getAccount(bob.address);
+    elon = runtime.getAccount(elon.address);
+  }
```
robert-zaremba, line 31: let's use context, instead of hack with double `syncAccounts` hack in two files.

examples/permissioned-token/test/failing-paths.js
```diff
@@ -0,0 +1,627 @@
...
+
+      assert.throws(() =>
+        issue(runtime, bob.account, elon, 20, controllerAppID, assetIndex, lsig),
+      RUNTIME_ERR1009);
+    });
+
+    it('should reject issuance tx if trying to send asset using secret key instead of clawback', () => {
+      // Opt-in to ASA by receiver
+      optInToASA(runtime, elon.address, assetIndex);
+      syncAccounts();
```
robert-zaremba, line 86: I think we don't need to call syncAccounts here (above as well)

examples/permissioned-token/test/failing-paths.js
```diff
@@ -0,0 +1,627 @@
...
+        sign: types.SignType.SecretKey,
+        fromAccount: permManager.account,
+        appId: permissionsAppId,
+        payFlags: { totalFee: 1000 },
+        appArgs: ['str:add_whitelist'],
+        accounts: [elon.address]
+      };
+    });
+
+    it('should reject whitelist if account is not opted in to permissions', () => {
```
robert-zaremba, line 133:
```suggestion
    it('should reject account whitelist if the account doesn't opt-in to permissions app', () => {
```

robert-zaremba (commented):
let's create a context for handling accounts and runtime.

examples/permissioned-token/test/common.js
ratik21: ok

examples/permissioned-token/assets/controller.py
ratik21: yes, agree

examples/permissioned-token/test/common.js
ratik21: done

examples/permissioned-token/test/common.js
ratik21: done

examples/permissioned-token/test/failing-paths.js
ratik21: yes, checked. Removed `syncAccounts()` after optInToASA

examples/permissioned-token/test/failing-paths.js
ratik21: yes

examples/permissioned-token/test/common.js
```diff
@@ -0,0 +1,411 @@
...
+  runtime.executeTx({
+    type: types.TransactionType.CallNoOpSSC,
+    sign: types.SignType.SecretKey,
+    fromAccount: asaManager,
+    appId: controllerAppID,
+    payFlags: { totalFee: 1000 },
+    appArgs: ['str:kill'],
+    foreignAssets: [assetIndex]
+  });
+}
```
robert-zaremba, line 242: I wonder if we could have same API and use it in scripts with algob and in tests with runtime - this  way we can define this functions once. Each of that functions will have a `txExecutor` or deployer (we would need a deployer test mode) to post transactions.

robert-zaremba (approved):

examples/permissioned-token/test/common.js
```diff
@@ -0,0 +1,411 @@
...
+   * - master;
+   * - alice;
+   * - bob;
+   * - elon;
+   * - runtime;
+   * - assetIndex;
+   * - lsig;
+   * - controllerAppID;
+   * - permissionsAppId;
+   */
```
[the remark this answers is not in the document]
ratik21: yes, i think so. Because not all properties are initialized in the constructor, so it would be good adding a comment here

examples/permissioned-token/test/common.js
```diff
@@ -0,0 +1,411 @@
...
+    this.alice = alice;
+    this.bob = bob;
+    this.elon = elon;
+    this.runtime = new Runtime([master, alice, bob, elon]);
+    this.deployASA('gold', { ...alice.account, name: 'alice' });
+    this.deployController(alice, CONTROLLER_APPROVAL_PROGRAM, CLEAR_STATE_PROGRAM);
+    this.syncAccounts();
+
+    this.deployClawback(alice, CLAWBACK_STATELESS_PROGRAM);
+    this.syncAccounts();
```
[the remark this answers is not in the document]
ratik21: ah, you're right. We're modifying asset clawback so i thought we needed it. 
Added a single `syncAccount()` after all deployments.

examples/permissioned-token/test/common.js
```diff
@@ -0,0 +1,411 @@
...
+  }
+
+  // refresh state
+  syncAccounts () {
+    this.alice = this.getAccount(this.alice.address);
+    this.bob = this.getAccount(this.bob.address);
+    this.elon = this.getAccount(this.elon.address);
+  }
+
+  deployASA (name, account) {
```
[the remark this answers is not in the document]
ratik21: ok

examples/permissioned-token/test/common.js
ratik21: We can think about it - i like the idea. It would be good for the user as well (using same high level API in scripts and in test).

PR #336, opened 2021-05-18
replace find-up with findup-sync in algob

ratik21:


robert-zaremba:
Please merge the `master` and remove `find-up` from dependencies. We need to check if find-up is used in any other place

ratik21:
> Please merge the `master` and remove `find-up` from dependencies. We need to check if find-up is used in any other place

done.

robert-zaremba (approved):

PR #337, opened 2021-05-18
Update SSC checkpoints storage

amityadav0:
- Move updateSSC to deployer
- change checkpoints for ssc
- change to nested `programs-name => timestamp => checkpoint`
- Add tests


packages/algob/src/internal/deployer.ts
```diff
@@ -132,7 +132,26 @@ class DeployerBasicMode {
...
   getSSC (nameApproval: string, nameClear: string): SSCInfo | undefined {
     const resultMap = this.cpData.precedingCP[this.networkName]?.ssc ?? new Map();
-    return resultMap.get(nameApproval + "-" + nameClear);
+    const nestedMap = resultMap.get(nameApproval + "-" + nameClear);
+    if (nestedMap) {
+      // return last pushed element in the map(latest timestamp value)
+      return [...nestedMap][nestedMap.size - 1][1];
+    } else {
+      return undefined;
+    }
```
ratik21, line 141: Can we do `return getSSCFromKey(nameApproval + "-" + nameClear);` ? If possible then we won't need to repeat code

packages/algob/src/internal/deployer.ts
```diff
@@ -132,7 +132,26 @@ class DeployerBasicMode {
...
+  }
+
+  /**
+   * Queries a stateful smart contract info from checkpoint using key. */
+  getSSCFromKey (key: string): SSCInfo | undefined {
```
ratik21, line 146: Please add a doc string explaining @param key

packages/algob/src/internal/deployer.ts
```diff
@@ -465,6 +491,41 @@ export class DeployerDeployMode extends DeployerBasicMode implements Deployer {
...
+        newApprovalProgram, newClearProgram, flags, this.txWriter);
+    } catch (error) {
+      this.persistCP();
+
+      console.log(error);
+      throw error;
+    }
+
+    this.registerUpdatedSSCInfo(name, sscInfo);
+
```
ratik21, line 526:
```suggestion
```

packages/algob/src/internal/deployer.ts
```diff
@@ -465,6 +491,41 @@ export class DeployerDeployMode extends DeployerBasicMode implements Deployer {
...
+   */
+  async updateSSC (
+    sender: algosdk.Account,
+    payFlags: rtypes.TxParams,
+    appID: number,
+    newApprovalProgram: string,
+    newClearProgram: string,
+    flags: rtypes.SSCOptionalFlags
+  ): Promise<SSCInfo> {
+    const name = newApprovalProgram + "-" + newClearProgram;
```
ratik21, line 512:
```suggestion
    const cpKey = newApprovalProgram + "-" + newClearProgram;
```

packages/algob/src/internal/deployer.ts
```diff
@@ -541,4 +608,17 @@ export class DeployerRunMode extends DeployerBasicMode implements Deployer {
...
+    payFlags: rtypes.TxParams,
+    appId: number,
+    newApprovalProgram: string,
+    newClearProgram: string,
+    flags: rtypes.SSCOptionalFlags
+  ): Promise<SSCInfo> {
+    throw new BuilderError(ERRORS.BUILTIN_TASKS.DEPLOYER_EDIT_OUTSIDE_DEPLOY, {
+      methodName: "updateSSC"
+    });
+  }
```
ratik21, line 623: I think user should be able to update SSC outside *DEPLOY* mode as well

packages/algob/src/lib/algo-operator.ts
```diff
@@ -1,4 +1,5 @@
 import { encodeNote, mkTransaction, types as rtypes } from "@algo-builder/runtime";
+import { ExecParams } from "@algo-builder/runtime/build/types";
```
ratik21, line 2: We're importing from build here

packages/algob/src/lib/algo-operator.ts
```diff
@@ -344,7 +354,73 @@ export class AlgoOperatorImpl implements AlgoOperator {
...
       txId: txInfo.txId,
       confirmedRound: confirmedTxInfo[confirmedRound],
-      appID: appId
+      appID: appId,
+      timestamp: Math.round(+new Date() / 1000)
+    };
+  }
+
+  /**
+   * Update programs for a contract.
```
ratik21, line 363:
```suggestion
   * Update programs (approval, clear) for a stateful smart contract.
```

packages/algob/src/lib/algo-operator.ts
```diff
@@ -344,7 +354,73 @@ export class AlgoOperatorImpl implements AlgoOperator {
...
+   * @param sender Account from which call needs to be made
+   * @param payFlags Transaction Flags
+   * @param appId ID of the application being configured or empty if creating
```
ratik21, line 366:
```suggestion
   * @param appId index of the application being configured or empty if creating
```

ratik21: also, what do we mean by "empty if creating"? Can we update ssc during creation?

packages/algob/src/lib/algo-operator.ts
```diff
@@ -344,7 +354,73 @@ export class AlgoOperatorImpl implements AlgoOperator {
...
+      note: flags.note,
+      lease: flags.lease
+    };
+
+    const txn = mkTransaction(execParam, params);
+    const txId = txn.txID().toString();
+    const signedTxn = txn.signTxn(sender.sk);
+
+    const txInfo = await this.algodClient.sendRawTransaction(signedTxn).do();
+    const confirmedTxInfo = await this.waitForConfirmation(txId);
```
ratik21, line 410: Can we use `signTransaction`, `algoOp.sendAndWait` functions here?

packages/algob/src/lib/algo-operator.ts
```diff
@@ -344,7 +354,73 @@ export class AlgoOperatorImpl implements AlgoOperator {
...
+
+    const appId = confirmedTxInfo['application-index'];
+    const message = `Signed transaction with txID: ${txId}\nUpdated app-id: ${appId}`; // eslint-disable-line @typescript-eslint/restrict-template-expressions
```
ratik21, line 413:
```suggestion
    const message = `Signed transaction with txID: ${txId}\nUpdated app: ${appId}`; // eslint-disable-line @typescript-eslint/restrict-template-expressions
```

packages/algob/src/lib/script-checkpoints.ts
```diff
@@ -38,14 +38,15 @@ export class CheckpointImpl implements Checkpoint {
...
-  ssc: Map<string, SSCInfo>;
+  ssc: Map<string, Map<number, SSCInfo>>;
   dLsig: Map<string, LsigInfo>;
 
   constructor (metadata?: Map<string, string>) {
     this.timestamp = +new Date();
     this.metadata = (metadata === undefined ? new Map<string, string>() : metadata);
     this.asa = new Map<string, ASAInfo>();
-    this.ssc = new Map<string, SSCInfo>();
+    const mp = new Map<number, SSCInfo>();
```
ratik21, line 48: Maybe we can add a type `type TimeStamp = number`, and use that type here for more clarity?
```js
const mp = new Map<TimeStamp, SSCInfo>();
```

packages/algob/src/types.ts
```diff
@@ -484,6 +488,24 @@ export interface Deployer {
     payFlags: rtypes.TxParams,
     scTmplParams?: SCParams) => Promise<SSCInfo>
 
+  /**
+   * Update programs for a contract.
```
ratik21, line 492:
```suggestion
   * Update programs(approval, clear) for a stateful smart contract.
```

packages/algob/src/types.ts
```diff
@@ -552,6 +574,10 @@ export interface Deployer {
    * Queries a stateful smart contract info from checkpoint. */
   getSSC: (nameApproval: string, nameClear: string) => SSCInfo | undefined
 
+  /**
+   * Queries a stateful smart contract info from checkpoint using key. */
+  getSSCFromKey: (key: string) => SSCInfo | undefined
```
ratik21, line 579: shouldn't this be a private method? User will use `getSSC` to retreive ssc info.

packages/algob/src/internal/deployer.ts
amityadav0: yes

packages/algob/src/internal/deployer.ts
amityadav0: ok/

packages/algob/src/internal/deployer.ts
amityadav0: updating is same as deploying. why user can update SSC outside DEPLOY using algob ?

packages/algob/src/lib/script-checkpoints.ts
amityadav0: yes, good idea

packages/algob/src/lib/algo-operator.ts
amityadav0: this must be wrong then, i moved this function here, do you think it is possible ?

packages/algob/src/lib/algo-operator.ts
amityadav0: yes

packages/algob/src/lib/algo-operator.ts
amityadav0: removed

packages/algob/src/internal/deployer.ts
ratik21: Why not? I think updating is not same as setup/deploying.. It could happen that after a while user decides to update the application, so he could add a script in *RUN* mode. Say if i wanted to update existing rules (eg. adding a third rule) in security token, then maybe user does so in run mode? What do you think?

If not, could you please check [https://github.com/scale-it/algo-builder/pull/337/files#diff-84170b6018e0919812f4a60f9c179f1e9d86a04d789d02ac02e2bba865cc23bbR9](https://github.com/scale-it/algo-builder/pull/337/files#diff-84170b6018e0919812f4a60f9c179f1e9d86a04d789d02ac02e2bba865cc23bbR9)? It's using 
 *updateSSC* in *RUN* mode

packages/algob/src/lib/algo-operator.ts
ratik21: ok

packages/algob/src/lib/algo-operator.ts
ratik21: fair enough. suggested because it seemed like we updated application index (which we didn't). That's why renamed `app-id` to `app`

packages/algob/src/types.ts
ratik21: I was saying from the user perspective, having both functions (`getSSC`, `getSSCFromKey`) looks confusing. yes, using `key.split("-")` could be wrong in some cases. Maybe we could move `registerCheckpoints` to deployer? If not, could we rename these functions ?

amityadav0: I think renaming will be better, could you please suggest something ?
From user perspective I think it won't be confusing because getsscfromkey takes only one argument.

packages/algob/src/internal/deployer.ts
amityadav0: Then what do we do with checkpoints ?
App will be updated but checkpoint will be old, this is not an interaction only function, it can completely change the code, which is same as deploying.

Do you think we should allow deploy in run mode ? According to your comment we should

amityadav0: In previous code updatessc function was independent and wasn't associated with checkpoints but once we have it in deployer class, we have to think about checkpoints as well

amityadav0: Updating a application is part of re-setup, in most of the cases there will be no application which is modified at each call.

It will be required to be updated once in a while.

amityadav0: yes, you are right about that script.
but what do we do with checkpoints ?
`algob deploy scripts/interaction/update_application.js` maybe we can support this command ?

ratik21: > Then what do we do with checkpoints ? but what do we do with checkpoints ?

This is what i was discussing on discord. We make it clear in the documentation that in this case checkpoint is not created. This is a user problem, but i don't think we should take control away from the user. User can do anything in run mode which is not part of "setup". transfer, opt-in, udpate.. 

> Do you think we should allow deploy in run mode ? According to your comment we should

No. User will only deploy an app once, but it could be updated multiple times (maybe dynamically in a script using template parameter). Updating is different than deploying. 

> App will be updated but checkpoint will be old, this is not an interaction only function, it can completely change the code, which is same as deploying.

We can add a function in deployer to query the blockchain to retreive app info (in case user updates an application in *RUN* mode), what do you think? [https://github.com/algorand/js-algorand-sdk/blob/develop/src/client/v2/algod/getApplicationByID.ts](https://github.com/algorand/js-algorand-sdk/blob/develop/src/client/v2/algod/getApplicationByID.ts)

amityadav0: I am not not sure about moving update ssc to readmode.

packages/algob/src/types.ts
amityadav0: how about `getSSCfromCPKey` or `getFromCPKey` ?

packages/algob/src/internal/deployer.ts
```diff
@@ -331,6 +346,13 @@ export class DeployerDeployMode extends DeployerBasicMode implements Deployer {
     this.cpData.registerSSC(this.networkName, sscName, sscInfo);
   }
 
+  /**
+   * Register UpdatedSSC Info in checkpoints
+   */
+  registerUpdatedSSCInfo (sscName: string, sscInfo: SSCInfo): void {
+    this.cpData.registerUpdatedSSC(this.networkName, sscName, sscInfo);
+  }
```
ratik21, line 354: Is it possible to use `registerSSCInfo` while updating SSC as well? I think we create a checkpoint even when we update an ssc, so maybe we should reuse registerSSCInfo?

packages/algob/src/types.ts
ratik21: `getSSCfromCPKey` works

packages/algob/src/internal/deployer.ts
amityadav0: i thought so, but having two different functions seems clear and easily understandable to developers, I can change to 
combining into functions if you want. also reuse part is only setting values in map.

ratik21: yes, then i think we could make this method private (`_registerUpdatedSSCInfo`), and move it's usage to *registerSSCInfo*. So during deploy/update ssc, only *registerSSCInfo* is used (which internally uses *this._registerUpdatedSSCInfo*, during app update). So it would be understandable as well

CHANGELOG.md
```diff
@@ -20,6 +21,8 @@
 ### Improvements
 * Update how error is displayed to a user
 * Add Update stateful smart contracts using execute transaction in runtime
+* Store checkpoints in nested form for SSC
+* Add tests for checkpoints
```
ratik21, line 25:
```suggestion
* Store checkpoints in nested form for SSC, added tests.
```

docs/guide/execution-checkpoints.md
```diff
@@ -62,3 +62,12 @@ Currently it's possible to edit these files by hand but it may be decided to dis
...
 Make sure that your local `algob` version matches the version from the link.
+
+## Checkpoint storage structure for Stateful Smart Contracts(SSC)
+
+We can update SSC on algorand blockchain, therefore we store checkpoints for SSC in a nested form.
+structure for the same is:
+        `program-name-key` -> `timestamp` -> `SSCInfo`.
+Consider a case when SSC is updated with same programs, this is the case where we append to checkpoint with different timestamp, txConfirmation but same `program-name-key`.
+
+When retreiving from checkpoints only the lastest timestamp SSCInfo is used.
```
ratik21, line 73:
```suggestion
While retreiving sscInfo from checkpoints only the lastest timestamp SSCInfo is used.
```

ratik21 (approved):
lgtm

PR #339, opened 2021-05-19
Add tests to PR 323 (Add asaName support in executeTransaction) 

amityadav0:


packages/algob/test/lib/tx.ts
```diff
@@ -22,3 +33,89 @@ describe("Note in TxParams", () => {
...
+  });
+
+  it("should opt-in to asa using asset id as number", async () => {
+    const res = await executeTransaction(deployer, execParams);
+
+    assert.deepEqual(res, expected);
+  });
+
+  it("Should fail if asset name is passed but not found in checkpoints", async () => {
+    execParams.assetID = "gold";
```
robert-zaremba, line 105:
```suggestion
    execParams.assetID = "unknown";
```

packages/algob/test/lib/tx.ts
```diff
@@ -22,3 +33,89 @@ describe("Note in TxParams", () => {
...
+
+    await expectBuilderErrorAsync(
+      async () => await executeTransaction(deployer, execParams),
+      ERRORS.BUILTIN_TASKS.DEPLOYER_ASA_NOT_DEFINED,
+      "gold"
```
robert-zaremba, line 110:
```suggestion
      "unknown"
```

packages/algob/test/lib/tx.ts
robert-zaremba: let's make it clear :)

packages/algob/test/lib/tx.ts
```diff
@@ -22,3 +33,89 @@ describe("Note in TxParams", () => {
...
+  };
+
+  const account = {
+    addr: 'UDF7DS5QXECBUEDF3GZVHHLXDRJOVTGR7EORYGDBPJ2FNB5D5T636QMWZY',
+    sk: new Uint8Array([28, 45, 45, 15, 70, 188, 57, 228, 18, 21, 42,
+      228, 33, 187, 222, 162, 89, 15, 22, 52, 143, 171, 182, 17, 168,
+      238, 96, 177, 12, 163, 243, 231, 160, 203, 241, 203, 176, 185,
+      4, 26, 16, 101, 217, 179, 83, 157, 119, 28, 82, 234, 204,
+      209, 249, 29, 28, 24, 97, 122, 116, 86, 135, 163, 236, 253])
+  };
```
robert-zaremba, line 63: don't we have a common account we use for tests? If not, let's create a test package with test data (we can put SuggestedParams there as well).

amityadav0: ok

amityadav0: done created tx file mocks folder.

packages/algob/test/lib/tx.ts
```diff
@@ -22,3 +35,72 @@ describe("Note in TxParams", () => {
...
+      'confirmed-round': 1,
+      "asset-index": 1,
+      'application-index': 1,
+      'global-state-delta': "string",
+      'local-state-delta': "string"
+    };
+  });
+
+  afterEach(() => {
+    fn.restore();
```
robert-zaremba, line 80:
```suggestion
    algod.algodClient.restore();
```
an you can remove fn assignment in line 68 and declaration in line 52

robert-zaremba (approved):
added one more small comment.

packages/algob/test/lib/tx.ts
amityadav0: i don't think we can do this `Property 'restore' does not exist on type 'Algodv2'`

robert-zaremba: hmm, it should work:
```
> sinon.stub(console, 'warn')    // NOTE - we don't to store the returned value
> console.warn("abc")            // NOTE - nothing is printed
> console.warn.restore()
> console.warn("abc")   // after restore we see printed "abc"
abc
```

PR #340, opened 2021-05-19
API docs various improvements

ratik21:
Types (defined in runtime) are clickable, but after exporting them in algob `index.ts`
image

### DIFF
--- a/packages/algob/src/runtime.ts
+++ b/packages/algob/src/runtime.ts
@@ -1,0 +1,15 @@
+import {
+  addressToPk, mkTransaction,
+  parseSSCAppArgs, stringToBytes,
+  types,
+  uint64ToBigEndian
+} from "@algo-builder/runtime";
+
+export {
+  addressToPk,
+  mkTransaction,
+  parseSSCAppArgs,
+  stringToBytes,
+  uint64ToBigEndian,
+  types
+};
--- a/packages/runtime/src/lib/asa.ts
+++ b/packages/runtime/src/lib/asa.ts
@@ -81,6 +81,14 @@ export function overrideASADef (
   return origDef;
 }
 
+/**
+ * Parses and validates asset defs fetched from yaml file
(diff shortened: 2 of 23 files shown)


packages/algob/src/runtime.ts
```diff
@@ -0,0 +1,15 @@
...
+  uint64ToBigEndian
+} from "@algo-builder/runtime";
+
+export {
+  addressToPk,
+  mkTransaction,
+  parseSSCAppArgs,
+  stringToBytes,
+  uint64ToBigEndian,
+  types
```
ratik21, line 14: I wasn't able to export some specific types (eg. `ExecParams`..), using
```
export {
 ..
 types: { ExecParams, .. }
 }
 ```
because then we're using type as a value (gives type assertion error). 
 
Types declared in runtime api docs, are clickable in algob api docs only when i exported all types (like done above). Looks like this in algob api docs:
image

File: packages/runtime/src/lib/asa.ts [hunk 2]
```diff
@@ -81,6 +81,14 @@ export function overrideASADef (
   return origDef;
 }
 
+/**
+ * Parses and validates asset defs fetched from yaml file
```
robert-zaremba, line 85:
```suggestion
 * Parses, overrides and validates asset defs map. Filaname parameter is used to
   indicate an ASA definition source when reporting errors.  
```

packages/runtime/src/lib/asa.ts
```diff
@@ -90,6 +98,12 @@ export function validateASADefs (
   return asaDefs;
 }
 
+/**
+ * Loads, validates and returns asset definition from asa.yaml file
```
robert-zaremba, line 102:
```suggestion
 * Loads, validates and returns asset definitions from the assets/asa.yaml file
```

robert-zaremba (approved):
left few suggestions

PR #344, opened 2021-05-21
Rename parseArgs to parse_params

amityadav0:


robert-zaremba (approved):
Please add a changelog entry in _API Breaking_ section.

PR #346, opened 2021-05-23
Add missing tests for asset fields in execute transaction

amityadav0:


packages/algob/test/lib/tx.ts
```diff
@@ -103,3 +104,56 @@ describe("Opt-In to ASA", () => {
...
+   * @param rawTxns rawTxns Signed transactions in Uint8Array
+   */
+  function checkTx (rawTxns: Uint8Array | Uint8Array[]): Promise<ConfirmedTxInfo> {
+    if (isArray(rawTxns)) {
+      // verify here if group tx
+    } else {
+      const tx: Transaction = decodeSignedTransaction(rawTxns).txn;
+      // Verify if fields are set correctly
+      assert.equal(tx.assetManager, undefined);
+      assert.equal(tx.assetReserve, undefined);
```
ratik21, line 140:
```suggestion
      assert.isUndefined(tx.assetManager);
      assert.isUndefined(tx.assetReserve);
```

packages/algob/test/stubs/algo-operator.ts
```diff
@@ -22,25 +23,44 @@ export class AlgoOperatorDryRunImpl implements AlgoOperator {
...
     return new Promise((resolve, reject) => {
-      resolve(res);
+      resolve(mockConfirmedTx);
     });
   }
 
+  /* eslint-disable sonarjs/no-identical-functions */
   waitForConfirmation (txId: string): Promise<import("algosdk").ConfirmedTxInfo> {
-    throw new Error("Not implemented");
+    return new Promise((resolve, reject) => {
```
ratik21, line 61:
```suggestion
    return new Promise((resolve, _) => {
```

packages/algob/test/stubs/algo-operator.ts
```diff
@@ -22,25 +23,44 @@ export class AlgoOperatorDryRunImpl implements AlgoOperator {
...
+          reserve: undefined,
+          freeze: bobAcc.addr,
+          clawback: undefined
+        }
+      };
+      return new Promise((resolve, reject) => {
+        resolve(res);
+      });
+    } else {
+      throw new Error("Not implemented");
```
ratik21, line 49: If we use `promise.resolve(res)` here, shouldn't we use `promise.reject("Not implemented")` here ? I don't exactly know the reject mechanism, does it not throw error, & simply return?

packages/algob/test/stubs/algo-operator.ts
```diff
@@ -22,25 +23,44 @@ export class AlgoOperatorDryRunImpl implements AlgoOperator {
...
 
   sendAndWait (rawTxns: Uint8Array | Uint8Array[]): Promise<ConfirmedTxInfo> {
-    const res: ConfirmedTxInfo = {
-      'confirmed-round': 1,
-      "asset-index": 1,
-      'application-index': 1,
-      'global-state-delta': "string",
-      'local-state-delta': "string"
-    };
     return new Promise((resolve, reject) => {
```
ratik21, line 54:
```suggestion
    return new Promise((resolve, _) => {
```

packages/algob/test/stubs/algo-operator.ts
```diff
@@ -22,25 +23,44 @@ export class AlgoOperatorDryRunImpl implements AlgoOperator {
...
+          decimals: 8,
+          'default-frozen': false,
+          'unit-name': "TKN",
+          name: "ASA-1",
+          url: "link",
+          'metadata-hash': "12312442142141241244444411111133",
+          manager: bobAcc.addr,
+          reserve: undefined,
+          freeze: bobAcc.addr,
+          clawback: undefined
```
ratik21, line 42: Let's add this `asaDef` in `/mock` and import here. That way we won't need to hardcode assertions in `checkTx` as well

packages/algob/test/stubs/algo-operator.ts
amityadav0: it will throw error, because function is not going to resolve, Before resolving it throws an error.

packages/algob/test/stubs/algo-operator.ts
amityadav0: Yes good idea but we will have to hardcode because that's a Transaction object and this one is asadef.

amityadav0: also we are not checking these values,
we check for values which are sent to network.

ratik21: I was thinking if we could do assertions for tx.fields based on their values fetched (either from network/custom params), instead of directly hardcoding them. Eg.
```js
import { mockAsaDef } from "/mock/..";
const fields = { manager: "", clawback: bobAcc.addr }; // used in test (custom fields to overwrite network values)

function checkTx(..) {
 ..
 assert.equal(tx.assetManager, mockAsaDef.manager); // assert with network field as this wasn't overwritten
 assert.equal(tx.assetClawback, fields.clawback); // assert with custom field as the network value should be overwritten
 ..
}
```

packages/algob/test/stubs/algo-operator.ts
ratik21: Ok so i thought this could be:
```js
return new Promise((resolve, reject) => {   
 assetIndex === 1 ? resolve(res); : reject("Not implemented");      
});

amityadav0: will check this one.

packages/algob/test/stubs/algo-operator.ts
amityadav0: actually we are using values we want to update and not what is already there.
Consider an example when we execute transaction we pass those values as asset fields. We are checking for those values, If you look at the code we are using values passed to network(they are not hardcoded) 
We need to match them with given values , and not the values that are in checkpoint

amityadav0: will update accordingly

ratik21 (approved):

PR #347, opened 2021-05-24
fix types for download-git-repo

amityadav0:
- Types were missing


ratik21 (approved):
I wonder why we removed these in [https://github.com/scale-it/algo-builder/commit/fd395bdf17e14e83afc735d93024fe41720dae48](https://github.com/scale-it/algo-builder/commit/fd395bdf17e14e83afc735d93024fe41720dae48)

PR #348, opened 2021-05-25
Update runtime to process execParams.deployASA, deploySSC, OptInToASA, OptIntoSSC

amityadav0:
- Update runtime to process execParams.deployASA, deploySSC, OptInToASA, OptIntoSSC
- Add functions to get app and asset ids using string
- Add tests


packages/runtime/src/runtime.ts
```diff
@@ -608,15 +628,28 @@ export class Runtime {
...
+    for (const txn of txnParameters) {
+      switch (txn.type) {
+        case TransactionType.DeployASA: {
+          txn.asaDef = this.loadedAssetsDefs[txn.asaName];
+          break;
+        }
+        case TransactionType.DeploySSC: {
+          txn.approvalProg = new Uint8Array(32); // mock approval program
+          txn.clearProg = new Uint8Array(32); // mock clear program
+          break;
```
ratik21, line 641: could you please explain why we set mock values, instead of `approval`, `clear` programs while deploying ssc? Is it because it's not the compiled teal code?

packages/runtime/test/integration/execute-transaction.ts
```diff
@@ -0,0 +1,209 @@
...
+    // verify app doesn't exist in map
+    const res = runtime.getAppIdFromName(approvalProgram, clearProgram);
+    assert.isUndefined(res);
+  });
+
+  it("Should opt-in to app, through execute transaction", () => {
+    setupApp();
+    syncAccounts();
+    const appId = runtime.getAppIdFromName(approvalProgram, clearProgram);
+    if (appId) {
```
ratik21, line 195: I think we should `assert.isDefined(appId)` here

packages/runtime/src/runtime.ts
```diff
@@ -29,19 +28,21 @@ export class Runtime {
...
 
   constructor (accounts: AccountStoreI[]) {
     // runtime store
     this.store = {
       accounts: new Map<AccountAddress, AccountStoreI>(), // string represents account address
       globalApps: new Map<number, AccountAddress>(), // map of {appId: accountAddress}
-      assetDefs: new Map<number, AccountAddress>() // number represents assetId
+      assetDefs: new Map<number, AccountAddress>(), // number represents assetId
+      assetNameId: new Map<string, number>(),
+      appNameId: new Map<string, number>(),
```
ratik21, line 43: maybe we should keep similar names as `deployer` (`sscInfo`, `ASAInfo`)? note: means we return app/asset info/undefined

packages/runtime/src/runtime.ts
```diff
@@ -234,6 +233,25 @@ export class Runtime {
...
+
+  /**
+   * Queries app id by app name from global state.
+   * Returns undefined if app is not found.
+   * @param approval
+   * @param clear
+   */
+  getAppIdFromName (approval: string, clear: string): number | undefined {
+    return this.store.appNameId.get(approval + "-" + clear);
+  }
```
ratik21, line 253: Can we return app/asset info here? As while loading from checkpoints we do something similar, so user can have same experience in tests as well. What do you think?
Note: i think app/asset info has their id's (along with other metadata)

packages/runtime/src/runtime.ts
amityadav0: yes, and we don't need it while testing

packages/runtime/test/integration/execute-transaction.ts
amityadav0: ok

packages/runtime/src/runtime.ts
amityadav0: yes

packages/runtime/src/ctx.ts
```diff
@@ -403,6 +429,85 @@ export class Ctx implements Context {
...
+            note: txnParam.note,
+            lease: txnParam.lease
+          };
+          this.runtime.addCtxOptInTx(fromAccountAddr, txnParam.appID, txnParam.payFlags, flags);
+          const account = this.getAccount(fromAccountAddr);
+          account.optInToApp(txnParam.appID, appParams);
+
+          this.runtime.run(appParams[approvalProgram], ExecutionMode.STATEFUL);
+          break;
+        }
```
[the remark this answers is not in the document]
amityadav0: ok, i will add a task for it.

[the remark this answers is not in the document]
ratik21: Please add a TODO to code along with a link to task in pivotal tracker

ratik21 (approved):
Please update the changelog before merging

PR #351, opened 2021-05-27
TEALv3 - version handle

ratik21:


packages/runtime/src/interpreter/opcode-list.ts
```diff
@@ -38,7 +38,7 @@ export class Pragma extends Op {
     if (this.line > 1) {
       throw new RuntimeError(RUNTIME_ERRORS.TEAL.PRAGMA_NOT_AT_FIRST_LINE, { line: line });
     }
-    if (args[0] === "version" && (args[1] === '1' || args[1] === '2')) {
+    if (args[0] === "version" && Number(args[1]) <= MaxLogicSigVersion) {
```
robert-zaremba, line 41: How about renaming `MaxLogicSigVersion` to `MaxTEALVersion`?

packages/runtime/src/lib/constants.ts
```diff
@@ -28,14 +31,18 @@ const zeroAddress = new Uint8Array(32);
...
 // https://developer.algorand.org/docs/reference/teal/opcodes/#txn
 // transaction fields supported by teal v1
-export const TxnFields: {[key: number]: {[key: string]: any}} = {
+export const TxnFields: {[key: number]: {[key: string]: keyOfEncTx | null }} = {
   1: {
     Sender: 'snd',
     Fee: 'fee',
     FirstValid: 'fv',
-    FirstValidTime: '',
+    FirstValidTime: null,
```
robert-zaremba, line 45: should we use `null` or undefined? I think undefined is better.

packages/runtime/src/lib/constants.ts
```diff
@@ -154,7 +181,7 @@ export const AssetParamMap: {[key: string]: string} = {
 export const reDigit = /^\d+$/;
 
 /** is Base64 regex
- * * ^                          # Start of input
+ * * ^                        # Start of input
```
robert-zaremba, line 184:
```suggestion
 * ^                          # Start of input
```

packages/runtime/src/lib/constants.ts
```diff
@@ -187,10 +214,16 @@ export const GlobalFields: {[key: number]: {[key: string]: any}} = { // teal ver
 // round and timestamp
 GlobalFields[2] = {
   ...GlobalFields[1],
-  LogicSigVersion: 2, // LogicSigVersion >= 2
+  LogicSigVersion: MaxLogicSigVersion, // LogicSigVersion >= 2
```
robert-zaremba, line 217:
```suggestion
  LogicSigVersion: MaxLogicSigVersion,
```

robert-zaremba (approved):
pre-approving . left few comments.

packages/runtime/src/interpreter/opcode-list.ts
ratik21: makes sense.

packages/runtime/src/lib/constants.ts
ratik21: Actually we can't use undefined here, because in the `op.execute(..)` we want to assert that we tx field passed by user is a valid one (eg. *FirstValidTime* is correct, but *firstvalidtime* isn't).
We check this by `txn[tealv][field] !== undefined`, so if we can't keep `undefined` as a value in our original maps

PR #353, opened 2021-05-27
load nested maps from checkpoints

amityadav0:


packages/algob/src/lib/script-checkpoints.ts
```diff
@@ -253,9 +253,23 @@ export function toMap <T> (obj: {[name: string]: T}): Map<string, T> {
   return mp;
 };
 
+// converts objects loaded from yaml file to nested map
+export function toSSCMap <T> (obj: {[name: string]: {[timestamp: string]: T}}): Map<string, Map<number, T>> {
```
ratik21, line 257:
```suggestion
export function toSSCMap <T> (obj: {[name: string]: {[timestamp: string]: T}}): Map<string, Map<Timestamp, T>> {
```

ratik21 (approved):
Thanks for the update

PR #354, opened 2021-05-28
Tealv3 opcodes: assert, pushint, pushbytes, swap

ratik21:


packages/runtime/src/interpreter/opcode-list.ts
```diff
@@ -2146,3 +2146,116 @@ export class Addr extends Op {
     stack.push(addr.publicKey);
   }
 }
+
+/* TEALv3 Ops */
+
+// immediately fail unless value top is a non-zero number
```
amityadav0, line 2152: you mean top value ?

packages/runtime/src/interpreter/opcode-list.ts
```diff
@@ -2146,3 +2146,116 @@ export class Addr extends Op {
...
+  constructor (args: string[], line: number) {
+    super();
+    this.line = line;
+    assertLen(args.length, 0, line);
+  };
+
+  execute (stack: TEALStack): void {
+    this.assertMinStackLen(stack, 1, this.line);
+    const top = this.assertBigInt(stack.pop(), this.line);
+    if (top <= 0n) {
```
amityadav0, line 2170:
```suggestion
    if (top == 0n) {
```

packages/runtime/src/interpreter/opcode-list.ts
```diff
@@ -2146,3 +2146,116 @@ export class Addr extends Op {
...
+}
+
+// push immediate UINT to the stack as an integer
+// push to stack: [...stack, uint64]
+export class PushInt extends Op {
+  /**
+   * NOTE: in runtime this class is similar to Int, but from tealv3 perspective this is optimized
+   * because pushint args are not added to the intcblock during assembly processes
+   */
+  readonly uint64: bigint;
```
amityadav0, line 2183: if that is the case i think we can reuse Int class instead of repeating code

packages/runtime/src/interpreter/opcode-list.ts
```diff
@@ -2146,3 +2146,116 @@ export class Addr extends Op {
...
+   */
+  constructor (args: string[], line: number) {
+    super();
+    this.line = line;
+    assertLen(args.length, 1, line);
+    assertOnlyDigits(args[0], line);
+
+    this.checkOverflow(BigInt(args[0]), line);
+    this.uint64 = BigInt(args[0]);
+  }
```
amityadav0, line 2199: In Int we have support for txType map.
`const intConst = TxnOnComplete[args[0] as keyof typeof TxnOnComplete] ||
      TxnType[args[0] as keyof typeof TxnType];`
 do we need it here as well ?

packages/runtime/src/interpreter/opcode-list.ts
```diff
@@ -2146,3 +2146,116 @@ export class Addr extends Op {
...
+}
+
+// push bytes to stack
+// push to stack [...stack, converted data]
+export class PushBytes extends Op {
+  /**
+   * NOTE: in runtime this class is similar to Byte, but from tealv3 perspective this is optimized
+   * because pushbytes args are not added to the bytecblock during assembly processes
+   */
+  readonly str: string;
```
amityadav0, line 2213: i think we can resuse

packages/runtime/test/src/interpreter/opcode-list.ts
```diff
@@ -3098,4 +3098,117 @@ describe("Teal Opcodes", function () {
...
+    });
+  });
+
+  describe("PushBytes", () => {
+    let stack: Stack<StackElem>;
+    this.beforeEach(() => { stack = new Stack<StackElem>(); });
+
+    it("should push bytes to stack", () => {
+      const str = "\"Algorand\"";
+      const bytes = new Uint8Array(Buffer.from("Algorand"));
```
amityadav0, line 3121:
```suggestion
      const expected_bytes = new Uint8Array(Buffer.from("Algorand"));
```

packages/runtime/test/src/interpreter/opcode-list.ts
```diff
@@ -3098,4 +3098,117 @@ describe("Teal Opcodes", function () {
...
+    let stack: Stack<StackElem>;
+    this.beforeEach(() => { stack = new Stack<StackElem>(); });
+
+    it("should not panic if top of stack is non zero uint64", () => {
+      const op = new Assert([], 0);
+      stack.push(55n);
+      assert.doesNotThrow(() => { op.execute(stack); });
+    });
+
+    it("should panic if top of stack is not valid", () => {
```
amityadav0, line 3141:
```suggestion
    it("should panic if top of stack is zero or bytes", () => {
```

packages/runtime/test/src/interpreter/opcode-list.ts
```diff
@@ -3098,4 +3098,117 @@ describe("Teal Opcodes", function () {
...
+      const op = new Assert([], 0);
+      stack.push(0n);
+
+      expectRuntimeError(
+        () => op.execute(stack),
+        RUNTIME_ERRORS.TEAL.TEAL_ENCOUNTERED_ERR
+      );
+
+      stack.push(stringToBytes("HelloWorld"));
```
amityadav0, line 3150: is this case verified by running on node ?

packages/runtime/test/src/parser/parser.ts
```diff
@@ -656,6 +656,67 @@ describe("Parser", function () {
         RUNTIME_ERRORS.TEAL.ASSERT_LENGTH
       );
     });
+
+    describe("should return correct opcodes for tealv3 ops", () => {
+      it("assert op", () => {
```
amityadav0, line 661:
```suggestion
      it("assert", () => {
```

packages/runtime/test/src/parser/parser.ts
```diff
@@ -656,6 +656,67 @@ describe("Parser", function () {
...
+        const expected = new Assert([], 1);
+        assert.deepEqual(res, expected);
+
+        expectRuntimeError(
+          () => opcodeFromSentence(["assert", "1234"], 1, interpreter),
+          RUNTIME_ERRORS.TEAL.ASSERT_LENGTH
+        );
+      });
+
+      it("pushint op", () => {
```
amityadav0, line 672:
```suggestion
      it("pushint", () => {
```

packages/runtime/test/src/parser/parser.ts
```diff
@@ -656,6 +656,67 @@ describe("Parser", function () {
...
+          RUNTIME_ERRORS.TEAL.INVALID_TYPE
+        );
+
+        expectRuntimeError(
+          () => opcodeFromSentence(["pushint", (MAX_UINT64 + 10n).toString()], 1, interpreter),
+          RUNTIME_ERRORS.TEAL.UINT64_OVERFLOW
+        );
+      });
+
+      it("pushbytes op", () => {
```
amityadav0, line 693:
```suggestion
      it("pushbytes", () => {
```

packages/runtime/test/src/parser/parser.ts
```diff
@@ -656,6 +656,67 @@ describe("Parser", function () {
...
+          RUNTIME_ERRORS.TEAL.ASSERT_LENGTH
+        );
+
+        expectRuntimeError(
+          () => opcodeFromSentence(["pushbytes", `0x250001000192CD0000002F6D6E742F72`], 1, interpreter),
+          RUNTIME_ERRORS.TEAL.UNKOWN_DECODE_TYPE
+        );
+      });
+
+      it("swap op", () => {
```
amityadav0, line 709:
```suggestion
      it("swap", () => {
```

packages/runtime/src/interpreter/opcode-list.ts
ratik21: No its not. It is similar but it's not the same. `pushInt` & `pushbytes` immediately push only direct `uint64`, `string literal (eg. "hello"). But `Int` supports string enums (NoOp, ..etc) as well, and `pushbytes` support different encoded types (base64, base32..etc). Tested on network.

packages/runtime/src/interpreter/opcode-list.ts
ratik21: no

packages/runtime/src/interpreter/opcode-list.ts
ratik21: yes

packages/runtime/test/src/interpreter/opcode-list.ts
ratik21: I think it should be `expectedBytes`, not `expected_bytes`

packages/runtime/test/src/interpreter/opcode-list.ts
ratik21: yes. It's also mentioned here that `assert` pops `unit64` from stack: [https://developer.algorand.org/docs/reference/teal/opcodes/#assert](https://developer.algorand.org/docs/reference/teal/opcodes/#assert)

packages/runtime/src/interpreter/opcode-list.ts
amityadav0: ok

amityadav0 (approved):
Please update the change log.

PR #358, opened 2021-06-01
Tealv3 Txn & Global fields

ratik21:
- Added support for the following txn fields:
   + `ForeignAssets`
   + `NumForeignAssets`
   + `ForeignApps`
   + `NumForeignApps`
   + `GlobalStateInts`
   + `GlobalStateByteslices`
   + `LocalStateInts`
   + `LocalStateByteslices`

- Added support for new global field: `CreatorAddress`
- Added tests


packages/runtime/test/src/interpreter/opcode-list.ts
```diff
@@ -1988,12 +1992,87 @@ describe("Teal Opcodes", function () {
...
+        assert.equal(1, stack.length());
+        assert.equal(BigInt(TXN_OBJ.apls.nbs), stack.pop());
+      });
     });
 
     describe("Gtxn", function () {
       before(function () {
         const tx = interpreter.runtime.ctx.tx;
-        const tx2 = { ...tx, fee: 2222 };
+        const tx2 = { ...tx, fee: 2222, apas: [3033, 4044], apfa: [5005, 6006, 7077] };
```
amityadav0, line 2075: could you please add here what are these values ?

packages/runtime/test/mocks/txn.ts
```diff
@@ -60,12 +60,12 @@ export const TXN_OBJ = {
...
   apls: {
-    nui: 2,
+    nui: 1,
     nbs: 2
   },
   apgs: {
-    nui: 2,
-    nbs: 2
+    nui: 3,
+    nbs: 4
```
amityadav0, line 68: Please add a comment here about values

packages/runtime/test/src/interpreter/opcode-list.ts
ratik21: yes

packages/runtime/test/mocks/txn.ts
ratik21: since the entire object is encoded, i will add a link for referencing encoded tx fields at the top of object

amityadav0 (approved):
LGTM

PR #360, opened 2021-06-02
TEALv3 - implement dig, select, gtxns, gtxnsa opcodes

ratik21:
- added `dig n`, `select`, `gtxns f`, `gtxnsa f i` ops
- added tests

NOTE: created another task for handling `min_balance` opcode as currently we are not decreasing an account's minimum balance while opting out from/destroying app or asset


packages/runtime/src/interpreter/opcode-list.ts
```diff
@@ -2259,3 +2260,123 @@ export class Swap extends Op {
...
+    assertOnlyDigits(args[0], line);
+
+    this.assertUint8(BigInt(args[0]), line);
+    this.depth = Number(args[0]);
+  };
+
+  execute (stack: TEALStack): void {
+    this.assertMinStackLen(stack, this.depth + 1, this.line);
+    const tempStack = new Stack<StackElem>(this.depth + 1); // depth = 2 means 3rd slot from top of stack
+    let copy;
```
amityadav0, line 2290: how about renaming to `target` ?

packages/runtime/src/interpreter/opcode-list.ts
```diff
@@ -2259,3 +2260,123 @@ export class Swap extends Op {
...
+    for (let i = 0; i <= this.depth; i++) {
+      const top = stack.pop();
+      tempStack.push(top);
+      if (i === this.depth) { copy = top; }
```
amityadav0, line 2294: how about this ?
```
for(int i = 0; i <= this.depth; ++i) {
    copy = stack.pop();
    tempStack.push(copy);
}
```

packages/runtime/src/interpreter/opcode-list.ts
```diff
@@ -2259,3 +2260,123 @@ export class Swap extends Op {
...
+    }
+    while (tempStack.length()) { stack.push(tempStack.pop()); }
+    stack.push(copy as StackElem);
```
amityadav0, line 2297: i don't think we will need `as StackElem` here if we use the above code.

packages/runtime/src/interpreter/opcode-list.ts
```diff
@@ -2259,3 +2260,123 @@ export class Swap extends Op {
...
+    if (toCheck !== 0n) { stack.push(notZeroSelection); } else { stack.push(isZeroSelection); }
+  }
+}
+
+/**
+ * push field F of the Ath transaction (A = top of stack) in the current group
+ * pops from stack: [...stack, uint64]
+ * pushes to stack: [...stack, transaction field]
+ * NOTE: "gtxns field" is equivalent to "gtxn _i_ field" (where _i_ is fetched from stack).
+ * gtxns exists so that i can be calculated, often based on the index of the current transaction.
```
amityadav0, line 2333: does it mean `i` is index in group ?
NOTE: is not quite clear.

packages/runtime/src/interpreter/opcode-list.ts
```diff
@@ -2259,3 +2260,123 @@ export class Swap extends Op {
...
+  /**
+   * Sets `field`, `txIdx` values according to arguments passed.
+   * @param args Expected arguments: [transaction field]
+   * // Note: Transaction field is expected as string instead of number.
+   * For ex: `Fee` is expected and `0` is not expected.
+   * @param line line number in TEAL file
+   * @param interpreter interpreter object
+   */
+  constructor (args: string[], line: number, interpreter: Interpreter) {
+    super(["100", ...args], line, interpreter);
```
amityadav0, line 2345: why are we passing `100` here ?

packages/runtime/src/interpreter/opcode-list.ts
```diff
@@ -2259,3 +2260,123 @@ export class Swap extends Op {
...
+  /**
+   * Sets `field`(Transaction Field), `idx`(Array Index) values according to arguments passed.
+   * @param args Expected arguments: [transaction field(F), transaction field array index(I)]
+   * // Note: Transaction field is expected as string instead of number.
+   * For ex: `Fee` is expected and `0` is not expected.
+   * @param line line number in TEAL file
+   * @param interpreter interpreter object
+   */
+  constructor (args: string[], line: number, interpreter: Interpreter) {
+    super(["100", ...args], line, interpreter);
```
amityadav0, line 2372: Please add a comment here for `100`.

packages/runtime/test/src/interpreter/opcode-list.ts
```diff
@@ -3211,4 +3211,247 @@ describe("Teal Opcodes", function () {
...
+
+    it("Gtxnsa: should push value of txfieldArr[index] from tx in group", () => {
+      stack.push(0n);
+      let op = new Gtxnsa(["ApplicationArgs", "1"], 1, interpreter);
+      op.execute(stack);
+      assert.equal(1, stack.length());
+      assert.deepEqual(stringToBytes("arg2"), stack.pop()); // args from tx0
+
+      stack.push(1n);
+      op = new Gtxnsa(["ApplicationArgs", "0"], 1, interpreter);
```
amityadav0, line 3425: this seems same as `Gtxns` case. what is the difference ?

packages/runtime/src/interpreter/opcode-list.ts
ratik21: yes

packages/runtime/src/interpreter/opcode-list.ts
ratik21: It's a mock value. didn't want to repeat code & reuse `Gtxn`. will add a comment

packages/runtime/test/src/interpreter/opcode-list.ts
ratik21: no difference, array fields are supported in both `Gtxns` & `Gtxnsa`, so added test for both.

packages/runtime/src/interpreter/opcode-list.ts
ratik21: no, we still need it.

amityadav0 (approved):

PR #362, opened 2021-06-04
TEALv3: Min balance

ratik21:
- compute minimum balance
  - When creating or opting into an app, the minimum balance grows before the app code runs
  - When deleting or closing out, the minimum balance decreases after the app executes.
  - after increasing minBalance while creating asset/app, we should assert `account minimum balance < account.balance()`
  - updated few tests to assert minimum balance as well

- add `min_balance` opcode + tests


packages/runtime/src/ctx.ts
```diff
@@ -142,11 +144,12 @@ export class Ctx implements Context {
   addAsset (name: string, fromAccountAddr: AccountAddress, flags: ASADeploymentFlags): number {
     const senderAcc = this.getAccount(fromAccountAddr);
 
-    // create asset
+    // create asset(with holding) in sender account
     const asset = senderAcc.addAsset(
       ++this.state.assetCounter, name,
       this.runtime.loadedAssetsDefs[name]
     );
+    this.assertMinBalance(0n, fromAccountAddr);
```
amityadav0, line 152: this is not quite clear, i think it is verifying if current balance is greater than minBalance, Please update the function to be more clear

packages/runtime/src/ctx.ts
```diff
@@ -181,32 +171,34 @@ export class Ctx implements Context {
...
       mockSuggestedParams(flags, this.runtime.getRound()));
 
     const assetHolding: AssetHoldingM = {
-      amount: address === creatorAddr ? BigInt(assetDef.total) : 0n, // for creator opt-in amount is total assets
+      amount: 0n,
       'asset-id': assetIndex,
-      creator: creatorAddr,
-      'is-frozen': address === creatorAddr ? false : assetDef.defaultFrozen
+      creator: assetDef.creator,
+      'is-frozen': assetDef.defaultFrozen
```
amityadav0, line 182: i think this was discussed before, assets are not frozen for creator. right ?
Please add a link here about this assumption

packages/runtime/src/ctx.ts
```diff
@@ -237,7 +230,7 @@ export class Ctx implements Context {
       {
         creator: senderAcc.address,
         appID: this.state.appCounter,
-        txId: this.tx.txID,
+        txId: "tx-id",
```
amityadav0, line 233: why are we updating this ?

packages/runtime/src/ctx.ts
```diff
@@ -181,32 +171,34 @@ export class Ctx implements Context {
...
    * @param approvalProgram application approval program
    * @param clearProgram application clear program
-   * NOTE - approval and clear program must be the TEAL code as string (not compiled code)
+   * NOTE:
+   * - approval and clear program must be the TEAL code as string (not compiled code)
+   * - When creating or opting into an app, the minimum balance grows before the app code runs
    */
   addApp (
-    fromAccountAddr: AccountAddress, flags: SSCDeploymentFlags,
+    fromAccountAddr: string, flags: SSCDeploymentFlags,
```
amityadav0, line 201:
```suggestion
    fromAccountAddr: AccountAddress, flags: SSCDeploymentFlags,
```

packages/runtime/src/ctx.ts
```diff
@@ -181,32 +171,34 @@ export class Ctx implements Context {
...
 
     const account = this.getAccount(address);
     account.optInToASA(assetIndex, assetHolding);
+    this.assertMinBalance(0n, address);
   }
 
   /**
    * creates new application and returns application id
-   * @param fromAccountAddr creator account address
    * @param flags SSCDeployment flags
```
amityadav0, line 192: Params are incorrect

packages/runtime/src/ctx.ts
```diff
@@ -250,12 +243,16 @@ export class Ctx implements Context {
    * Account address opt-in for application Id
    * @param accountAddr Account address
    * @param appId Application Id
+   * @param flags Stateful smart contract transaction optional parameters (accounts, args..)
+   * @param payFlags Transaction Parameters
+   * NOTE: When creating or opting into an app, the minimum balance grows before the app code runs
    */
-  optInToApp (accountAddr: AccountAddress, appID: number): void {
+  optInToApp (accountAddr: string, appID: number): void {
```
amityadav0, line 250:
```suggestion
  optInToApp (accountAddr: AccountAddress, appID: number): void {
```

packages/runtime/src/ctx.ts
```diff
@@ -302,7 +301,9 @@ export class Ctx implements Context {
         txnParam.assetID as number, closeToAddr);
 
       closeRemToAssetHolding.amount += fromAssetHolding.amount; // transfer assets of sender to closeRemTo account
-      this.getAccount(fromAccountAddr).assets.delete(txnParam.assetID as number); // remove asset holding from sender account
+      const fromAccount = this.getAccount(fromAccountAddr);
+      fromAccount.minBalance -= ASSET_CREATION_FEE;
+      fromAccount.assets.delete(txnParam.assetID as number); // remove asset holding from sender account
```
amityadav0, line 306: Please add a link to docs here.

packages/runtime/src/ctx.ts
```diff
@@ -302,7 +301,9 @@ export class Ctx implements Context {
         txnParam.assetID as number, closeToAddr);
 
       closeRemToAssetHolding.amount += fromAssetHolding.amount; // transfer assets of sender to closeRemTo account
-      this.getAccount(fromAccountAddr).assets.delete(txnParam.assetID as number); // remove asset holding from sender account
+      const fromAccount = this.getAccount(fromAccountAddr);
```
amityadav0, line 304: let's move `opt-out` from asset part to account class and reuse it here.

packages/runtime/test/integration/close-clear-ssc.ts
```diff
@@ -90,10 +96,13 @@ describe("ASC - CloseOut from Application and Clear State", function () {
   });
 
   it("should not delete application on CloseOut call if logic is rejected", function () {
-    // create app
     const appId = runtime.addApp(flags, {}, approvalProgram, clearProgram);
+    const initialJohnMinBalance = runtime.getAccount(john.address).minBalance;
+
```
amityadav0, line 101:
```suggestion
```

packages/runtime/test/integration/close-clear-ssc.ts
```diff
@@ -111,6 +120,10 @@ describe("ASC - CloseOut from Application and Clear State", function () {
     // verify app is not deleted from account's local state (as tx is rejected)
     const res = john.getAppFromLocal(appId);
     assert.isDefined(res);
+
```
amityadav0, line 123:
```suggestion
```

packages/runtime/src/ctx.ts
ratik21: i didn't. will check

packages/runtime/src/ctx.ts
ratik21: ok, let's create another function for this

packages/runtime/src/ctx.ts
ratik21: The concept of Opting In for creator does not make sense, because the create created the asset, and therefore it cannot opt-out of the ASA as well. 
While creating an asset,  asset holding is added in creator's account will all assets & default frozen false. Means that this code (`optInToASA`) will never be executed for asset creator.

I will add a link about ASA's in `account.addApp` where we create asset holding for creator

packages/runtime/src/account.ts
```diff
@@ -199,6 +212,18 @@ export class AccountStore implements AccountStoreI {
...
+  /**
+   * removes asset holding from account
+   * @param assetId asset index
+   */
+  closeAsset (assetId: number): void {
+    if (this.assets.has(assetId)) {
+      this.minBalance -= ASSET_CREATION_FEE;
+      // https://developer.algorand.org/docs/reference/transactions/#asset-transfer-transaction
+      this.assets.delete(assetId); // remove asset holding from account
+    }
```
amityadav0, line 224: should we throw error if asset is not present in account ? or a warning ?

packages/runtime/src/ctx.ts
```diff
@@ -184,29 +174,31 @@ export class Ctx implements Context {
...
 
     const assetHolding: AssetHoldingM = {
-      amount: address === creatorAddr ? BigInt(assetDef.total) : 0n, // for creator opt-in amount is total assets
+      amount: 0n,
       'asset-id': assetIndex,
-      creator: creatorAddr,
-      'is-frozen': address === creatorAddr ? false : assetDef.defaultFrozen
+      creator: assetDef.creator,
+      'is-frozen': assetDef.defaultFrozen
     };
```
amityadav0, line 187:
```suggestion
```

packages/runtime/src/ctx.ts
amityadav0: `fromAccountAddr` is still missing

packages/runtime/src/ctx.ts
```diff
@@ -184,29 +174,31 @@ export class Ctx implements Context {
...
     const account = this.getAccount(address);
     account.optInToASA(assetIndex, assetHolding);
+    this.assertAccBalAboveMin(address);
   }
 
   /**
    * creates new application and returns application id
-   * @param fromAccountAddr creator account address
    * @param flags SSCDeployment flags
+   * @param payFlags Transaction parameters
```
amityadav0, line 196: param is incorrect

packages/runtime/src/ctx.ts
```diff
@@ -253,12 +246,16 @@ export class Ctx implements Context {
    * Account address opt-in for application Id
    * @param accountAddr Account address
    * @param appId Application Id
+   * @param flags Stateful smart contract transaction optional parameters (accounts, args..)
+   * @param payFlags Transaction Parameters
+   * NOTE: When creating or opting into an app, the minimum balance grows before the app code runs
```
amityadav0, line 251: doc string is incorrect, please update

packages/runtime/test/integration/close-clear-ssc.ts
```diff
@@ -54,11 +54,13 @@ describe("ASC - CloseOut from Application and Clear State", function () {
...
   it("should successfully closeOut from app and update state according to asc", function () {
     const appId = runtime.addApp(flags, {}, approvalProgram, clearProgram); // create app
-    closeOutParams.appId = appId;
-    runtime.optInToApp(john.address, appId, {}, {}); // opt-in to app (set new local state)
+    const initialJohnMinBalance = runtime.getAccount(john.address).minBalance;
 
-    runtime.executeTx(closeOutParams);
+    runtime.optInToApp(john.address, appId, {}, {}); // opt-in to app (set new local state)
+    syncAccount();
+    assert.isAbove(john.minBalance, initialJohnMinBalance); // verify minimum balance raised after optIn
```
amityadav0, line 61: please update it to check how much it is increased.

packages/runtime/test/integration/deleteApp.ts
```diff
@@ -48,20 +48,28 @@ describe("Algorand Smart Contracts - Delete Application", function () {
...
   });
 
   it("should not delete application if logic is rejected", function () {
-    // create app
-    const appId = runtime.addApp(flags, {}, approvalProgram, clearProgram);
+    const initialMinBalance = john.minBalance;
+    const appId = runtime.addApp(flags, {}, approvalProgram, clearProgram); // create app
+
+    const minBalanceAfterAddApp = runtime.getAccount(john.address).minBalance;
+    assert.isAbove(minBalanceAfterAddApp, initialMinBalance); // min balance should increase
```
amityadav0, line 71: please add how much balance is increased, it is different for opt-in and creation. right ?

packages/runtime/test/src/interpreter/opcode-list.ts
```diff
@@ -3992,4 +3994,79 @@ describe("Teal Opcodes", function () {
...
+
+      op = new MinBalance([], 1, interpreter);
+      stack.push(1n); // push sender id
+
+      op.execute(stack);
+      top = stack.pop();
+      assert.equal(top, BigInt(ALGORAND_ACCOUNT_MIN_BALANCE));
+    });
+
+    it("should push raised min_balance to stack after creating asset", () => {
```
amityadav0, line 4039: i think this test should be in flow/integration tests, because here we check operations related to opcode-list only.
min_balance should push `min_balance`. this needs to be tested here

packages/runtime/test/src/interpreter/opcode-list.ts
```diff
@@ -3992,4 +3994,79 @@ describe("Teal Opcodes", function () {
...
+        () => op.execute(stack),
+        RUNTIME_ERRORS.GENERAL.ACCOUNT_DOES_NOT_EXIST
+      );
+    });
+
+    it("should throw index out of bound error", () => {
+      const op = new Balance([], 1, interpreter);
+      stack.push(8n);
+
+      expectRuntimeError(
```
amityadav0, line 4066: :+1:

packages/runtime/test/src/runtime.ts
```diff
@@ -205,6 +210,7 @@ describe("Algorand Standard Assets", function () {
     assert.equal(res.reserve, elon.address);
     assert.equal(res.freeze, elon.address);
     assert.equal(res.clawback, elon.address);
+    assert.equal(john.minBalance, initialMinBalance + ASSET_CREATION_FEE);
```
amityadav0, line 213: here we are testing this case, in opcode-list this is repeated

packages/runtime/src/account.ts
ratik21: before closing an asset(removing asset holding) we need to empty it's holding to the `closeRemTo` account (even if amount is zero). So we're verifying asset holding exists in `runtime.transferAsset` (before emptying balance). That's why i kept this code there only. Will add a comment here now.

packages/runtime/test/integration/close-clear-ssc.ts
ratik21: thought so initially, but then realized it doesn't make sense to compare constant increase (as we will be testing increment for exactly same value as we'll be increasing in code (..does that make sense?) ). Didn't do it here because we are testing min_balance in multiple flow (unlike ASA tests), which could be hard to interpret.
The important part is that minimum balance should grow after app/asset creation or opt-in, and reduce (to prev value) after opt-out/deletion.
If you want i can update to check exact value

packages/runtime/test/integration/deleteApp.ts
ratik21: no, same `0.1ALGO`

packages/runtime/test/src/interpreter/opcode-list.ts
ratik21: > min_balance should push min_balance. this needs to be tested here

did that in previous test

> i think this test should be in flow/integration tests, because here we check operations related to opcode-list only.

in integration tests we test flow in smart contracts. Here we just add an asset and asset that `MinBalance` op should push raised balance.

packages/runtime/test/integration/close-clear-ssc.ts
amityadav0: agree this may not make sense now, but at some later point if storage costs are changed in code, then these tests will fail and we will know why

packages/runtime/test/src/runtime.ts
ratik21: no, in opcode list we specifically test execution of `MinBalance` opcode. There is a slight difference. Here we directly test account's min balance. I would prefer to keep it.

packages/runtime/test/integration/deleteApp.ts
amityadav0: This one of for asa, I think for app it differs

packages/runtime/src/ctx.ts
ratik21: done

packages/runtime/test/src/runtime.ts
amityadav0: Ok, your call, better to have more tests

packages/runtime/test/integration/close-clear-ssc.ts
ratik21: makes sense

packages/runtime/test/integration/deleteApp.ts
ratik21: oh yes, thought it was asa. 
For app, cost for creation & optin are different as in creation we increase balance using `global ints/bytes`, while in opt-in we increase/decrease balance referring to `local ints/bytes`. Base fee is same.

amityadav0 (approved):

amityadav0:
Please update the change log before merging

PR #364, opened 2021-06-04
Marking not valid checkpoints ("removal")

amityadav0:
- Handle checkpoint deletion
- Update structure to use a boolean which will be set to true if asset/ssc is deleted

- fix appId to appID
- fix types
- add transaction flow check for deleted cp
- add flow tests


docs/guide/testing-teal.md
```diff
@@ -282,7 +282,7 @@ Now, we will execute a transaction with an app call (stateful TEAL). The app is
     type: TransactionType.CallNoOpSSC,
     sign: SignType.SecretKey,
     fromAccount: john.account,
-    appId: 0,
+    appID: 0,
```
robert-zaremba, line 285: this is API breaking, not sure if this is good at this stage.

packages/algob/src/internal/deployer.ts
```diff
@@ -147,6 +151,36 @@ class DeployerBasicMode {
...
 
+  /**
+   * Returns SSC checkpoint key using application index,
+   * returns undefined if it doesn't exist
+   * @param index Application index
+   */
+  getSSCCPKeyFromId (index: number): string | undefined {
+    const resultMap = this.cpData.precedingCP[this.networkName]?.ssc ?? new Map();
+    for (const [key, nestedMap] of resultMap) {
+      if ([...nestedMap][nestedMap.size - 1][1].appID === index) {
```
robert-zaremba, line 162: we want to get the last object? As far as I know, the order is not predictable in JS, so not sure if the thing above will work. Also there is no need to construct an array.

packages/algob/src/internal/deployer.ts
```diff
@@ -147,6 +151,36 @@ class DeployerBasicMode {
     }
   }
 
+  /**
+   * Returns SSC checkpoint key using application index,
+   * returns undefined if it doesn't exist
+   * @param index Application index
+   */
+  getSSCCPKeyFromId (index: number): string | undefined {
```
robert-zaremba, line 159: this is very cryptic name. Let's try to think about something better. `GetAppIndexCheckpoint`?

packages/algob/src/internal/deployer.ts
```diff
@@ -147,6 +151,36 @@ class DeployerBasicMode {
...
+   */
+  getASAKeyFromId (index: number): string | undefined {
+    const resultMap = this.cpData.precedingCP[this.networkName]?.asa ?? new Map();
+    for (const [key, value] of resultMap) {
+      if (value.assetIndex === index) {
+        return key;
+      }
+    }
+    return undefined;
+  }
```
robert-zaremba, line 182: would be good to have a namespace for checkpoint related functions, then it will be easier to refer this methods.

packages/algob/src/internal/deployer.ts
```diff
@@ -284,6 +348,92 @@ class DeployerBasicMode {
...
     return await this.algoOp.ensureCompiled(name, force, scTmplParams);
   }
+
+  /**
+   * Check if ASA info exist for asset,
+   * if it exists check for delete param,
+   * throw error if it is deleted
+   * @param asset asset index or asset name
+   */
+  private checkIfASAExist (asset: string | number): void {
```
robert-zaremba, line 358: we need to update the documentation. Be clear that we are searching in checkpoints.

Also:
* why we throw an exception instead of returning `bool`?
* if we want to throw an exception, the the function name should start with `assert...` (eg: assertASAExists)

packages/algob/src/internal/deployer.ts
```diff
@@ -284,6 +348,92 @@ class DeployerBasicMode {
...
+    }
+  }
+
+  /**
+   * Check if SSC info exist for app id,
+   * if it exists check for delete param,
+   * throw error if it is deleted
+   * @param appID Application index
+   */
+  private checkIfAppExist (appID: number): void {
```
robert-zaremba, line 380: as above.

robert-zaremba (commented):
first part of the review.

docs/guide/testing-teal.md
amityadav0: this needs to be done either we have to change appID to appId or vice versa, because in some params it is appID and in others it is appId, therefore we cannot group them in one

packages/algob/src/internal/deployer.ts
amityadav0: I did research on this, I found that we cannot directly get last value from a map, the way to do it is to traverse the whole map and choose the last element.
Could you please tell me the other way ?

packages/algob/src/internal/deployer.ts
amityadav0: this returns checkpoint key from given Id,
So maybe  getSSCKeyfromindex or getStatefulKeyFromIndex ?

packages/algob/src/internal/deployer.ts
amityadav0: ok will add more details about cases, changing to assert

packages/algob/src/internal/deployer.ts
amityadav0: i was thinking to put them in checkpoint interface, but because checkpoints are only accessible to deployer,
it couldn't be used in `executeTransaction`. did you mean this ? or something else

packages/algob/src/internal/deployer.ts
amityadav0: how about `getAppCheckpointKeyFromIndex` ?

docs/guide/testing-teal.md
robert-zaremba: Ah, OK.
So let's use `appID`. Please add entry to CHANGELOG

packages/algob/src/internal/deployer.ts
robert-zaremba: What is the last element? What order are you going to use? If it's min / max then you can do it using one loop.

packages/algob/src/internal/deployer.ts
robert-zaremba: OK

packages/algob/src/internal/deployer.ts
robert-zaremba: I was thinking about having a checkpoint operator and use it in deployer, so this way we would need to do:
```
deployer.checkpoint.appKeyFromIndex(...)  // here checkpoint is a new class
// other methods would need to be modified as well.
```
What do you think?

packages/algob/src/internal/deployer.ts
amityadav0: the last element have the latest timestamp

amityadav0: Ok, got it.
 - traverse the map store value whose key is maximum

docs/guide/testing-teal.md
amityadav0: ok

packages/algob/src/internal/deployer.ts
amityadav0: yes, this would be good

packages/algob/src/lib/script-checkpoints.ts
```diff
@@ -209,13 +226,21 @@ export async function registerCheckpoints (
...
+          break;
+        }
         if (res) {
           const asaInfo: rtypes.ASAInfo = {
             creator: encodeAddress(txn.from.publicKey),
             txId: txn.txID(),
             assetIndex: txConfirmation["asset-index"],
             confirmedRound: txConfirmation['confirmed-round'],
-            assetDef: res[1]
+            assetDef: res[1],
```
ratik21, line 242: Can we add a comment about `res[1]` ? That res[0] is name and res[1] is rtypes.ASADef

packages/algob/src/lib/script-checkpoints.ts
```diff
@@ -224,14 +249,24 @@ export async function registerCheckpoints (
...
-          creator: encodeAddress(txn.from.publicKey),
-          txId: txn.txID(),
-          appID: txConfirmation['application-index'],
-          confirmedRound: txConfirmation['confirmed-round'],
-          timestamp: Math.round(+new Date() / 1000)
-        };
+        const key = deployer.getAppCheckpointKeyFromIndex(txn.appIndex);
+        if (key) {
+          const temp: rtypes.SSCInfo | undefined = deployer.getSSCfromCPKey(key);
+          if (txn.appOnComplete === 5 && temp) {
```
ratik21, line 255: instead of 5 can we use `TxnOnComplete.DeleteApplication` (taken from types-algosdk)?

packages/algob/src/lib/script-checkpoints.ts
```diff
@@ -224,14 +249,24 @@ export async function registerCheckpoints (
...
+            temp.deleted = true;
+            deployer.registerSSCInfo(key, temp);
+            break;
```
ratik21, line 258: Shall we `deployer.logTx()` before breaking?

packages/algob/src/types.ts
```diff
@@ -583,6 +598,14 @@ export interface Deployer {
    *     (used only when compiling PyTEAL to TEAL)
    */
   ensureCompiled: (name: string, force?: boolean, scTmplParams?: SCParams) => Promise<ASCCache>
+
+  /**
+   * Checks if checkpoint is deleted for a particular transaction
+   * if checkpoint exist and is marked as deleted,
```
ratik21, line 604:
```suggestion
   * if checkpoint exists and is marked as deleted,
```

packages/algob/test/internal/deployer.ts
```diff
@@ -145,8 +153,9 @@ describe("DeployerDeployMode", () => {
       creator: "addr-1-get-address-dry-run",
       txId: "tx-id-dry-run",
       confirmedRound: -1,
-      appID: -1,
-      timestamp: 1
+      appID: 33,
```
ratik21, line 156: could you please explain why we're changing appId here?

packages/algob/test/mocks/tx.ts
```diff
@@ -38,3 +38,13 @@ export const mockAssetInfo: AssetInfo = {
...
   }
 };
+
+const mockProgram = new Uint8Array([
+  2, 32, 4, 1, 4, 100, 144, 78, 49, 16,
+  34, 18, 49, 16, 35, 18, 17, 49, 8, 36,
+  14, 16, 49, 18, 36, 14, 16, 49, 32, 50,
+  3, 18, 16, 49, 9, 50, 3, 18, 16, 49,
+  1, 37, 14, 16
+]);
```
ratik21, line 48: Do we need to populate this? Could `mockProgram = new Uint8Array(0);` work?

packages/algob/test/stubs/algo-operator.ts
```diff
@@ -110,25 +114,25 @@ export class AlgoOperatorDryRunImpl implements AlgoOperator {
   optInAcountToASA (
     asaName: string, assetIndex: number, account: rtypes.Account,
     params: rtypes.TxParams): Promise<void> {
-    throw new Error("Method not implemented.");
+    return new Promise((resolve, reject) => { resolve(); });
```
ratik21, line 117:
```suggestion
    return new Promise((resolve, _) => { resolve(); });
```

packages/algob/test/stubs/algo-operator.ts
```diff
@@ -110,25 +114,25 @@ export class AlgoOperatorDryRunImpl implements AlgoOperator {
...
   }
 
   optInLsigToASA (
     asaName: string, assetIndex: number, lsig: LogicSig, flags: rtypes.TxParams
   ): Promise<void> {
-    throw new Error("Method not implemented.");
+    return new Promise((resolve, reject) => { resolve(); });
```
ratik21, line 123:
```suggestion
    return new Promise((resolve, _) => { resolve(); });
```

packages/algob/test/stubs/algo-operator.ts
```diff
@@ -110,25 +114,25 @@ export class AlgoOperatorDryRunImpl implements AlgoOperator {
...
   }
 
   optInAccountToSSC (
     sender: rtypes.Account, index: number,
     payFlags: rtypes.TxParams, flags: rtypes.SSCOptionalFlags): Promise<void> {
-    throw new Error("Method not implemented.");
+    return new Promise((resolve, reject) => { resolve(); });
```
ratik21, line 129:
```suggestion
    return new Promise((resolve, _) => { resolve(); });
```

packages/algob/test/stubs/algo-operator.ts
```diff
@@ -110,25 +114,25 @@ export class AlgoOperatorDryRunImpl implements AlgoOperator {
...
   }
 
   optInLsigToSSC (
-    appId: number, lsig: LogicSig,
+    appID: number, lsig: LogicSig,
     payFlags: rtypes.TxParams, flags: rtypes.SSCOptionalFlags): Promise<void> {
-    throw new Error("Method not implemented.");
+    return new Promise((resolve, reject) => { resolve(); });
```
ratik21, line 135:
```suggestion
    return new Promise((resolve, _) => { resolve(); });
```

packages/algob/test/fixture-projects/stateful/artifacts/.cp.yaml
```diff
@@ -0,0 +1,25 @@
+network1:
```
ratik21, line 1: file name is `.cp.yaml` ?

packages/algob/src/types.ts
```diff
@@ -312,6 +312,7 @@ export interface LsigInfo {
   creator: AccountAddress
   contractAddress: string
   lsig: LogicSig
+  deleted: boolean
```
ratik21, line 315: Could you please explain why we add deleted flag in logic signature info ?

packages/runtime/src/types.ts
```diff
@@ -85,6 +85,7 @@ export interface DeployedAssetInfo {
   creator: AccountAddress
   txId: string
   confirmedRound: number
+  deleted: boolean
```
ratik21, line 88: Can we make this flag optional? An app/asset can only be deleted "once". So deleted flag will only be true when asset/app is deleted from network, and afterwards we don't save any info in checkpoints (as after deletion only optOut is allowed, which is not cp related). 
Not sure about having `deleted:false` in artifacts each time after deploying or updating ASA/SSC.
What do you think?

packages/algob/src/internal/deployer.ts
```diff
@@ -284,6 +348,98 @@ class DeployerBasicMode {
...
+   * throw error(except for opt-out transactions), else pass
+   * @param execParams Transaction execution parameters
+   */
+  checkForDeletedCP (execParams: rtypes.ExecParams | rtypes.ExecParams[]): void {
+    if (Array.isArray(execParams)) {
+      for (const txn of execParams) {
+        this._internalDeletedCPCheck(txn);
+      }
+    } else {
+      this._internalDeletedCPCheck(execParams);
```
ratik21, line 440: Why using internal prefix? I think `_assertCpNotDeleted` is clear

packages/algob/test/internal/deployer.ts
amityadav0: because teal doesn't support negative numbers, it is needed when making tx

packages/algob/test/mocks/tx.ts
amityadav0: no it doesn't

packages/algob/test/stubs/algo-operator.ts
amityadav0: it doesn't work like this

packages/algob/src/internal/deployer.ts
amityadav0: not any specific reason, just out of names, this seemed easily understandable

packages/algob/src/lib/script-checkpoints.ts
```diff
@@ -192,12 +192,29 @@ export function persistCheckpoint (scriptName: string, checkpoint: Checkpoints):
...
+ * https://developer.algorand.org/docs/reference/transactions/#asset-configuration-transaction
+ * https://github.com/algorand/js-algorand-sdk/blob/e07d99a2b6bd91c4c19704f107cfca398aeb9619/src/transaction.ts#L528
+ */
+function checkAssetDeletionTx (txn: Transaction): boolean {
+  if (txn.assetClawback || txn.assetFreeze || txn.assetManager || txn.assetReserve) {
+    return false;
+  } else if (txn.assetIndex) {
+    return true;
+  }
+  return false;
```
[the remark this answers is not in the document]
amityadav0: this looks more cryptic to me, can update if you want

packages/algob/src/lib/script-checkpoints.ts
```diff
@@ -209,13 +226,21 @@ export async function registerCheckpoints (
     switch (txn.type) {
       case 'acfg': {
         txConfirmation = await deployer.waitForConfirmation(txn.txID());
+        const key = deployer.getAssetCheckpointKeyFromIndex(txn.assetIndex);
+        if (key && checkAssetDeletionTx(txn)) {
```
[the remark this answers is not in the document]
amityadav0: i prefer `checkAssetDeletionTx`

packages/algob/src/internal/deployer.ts
```diff
@@ -495,6 +651,15 @@ export class DeployerDeployMode extends DeployerBasicMode implements Deployer {
     newClearProgram: string,
     flags: rtypes.SSCOptionalFlags
   ): Promise<rtypes.SSCInfo> {
+    this.checkForDeletedCP({
```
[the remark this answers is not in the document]
amityadav0: ok

packages/algob/src/lib/script-checkpoints.ts
```diff
@@ -192,12 +192,29 @@ export function persistCheckpoint (scriptName: string, checkpoint: Checkpoints):
...
+ * https://github.com/algorand/js-algorand-sdk/blob/e07d99a2b6bd91c4c19704f107cfca398aeb9619/src/transaction.ts#L528
+ */
+function checkAssetDeletionTx (txn: Transaction): boolean {
+  if (txn.assetClawback || txn.assetFreeze || txn.assetManager || txn.assetReserve) {
+    return false;
+  } else if (txn.assetIndex) {
+    return true;
+  }
+  return false;
+}
```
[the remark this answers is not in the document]
amityadav0: yes, seems clear

packages/algob/src/lib/script-checkpoints.ts
```diff
@@ -192,12 +192,29 @@ export function persistCheckpoint (scriptName: string, checkpoint: Checkpoints):
...
+
 /**
  * Register checkpoints for ASA and SSC
  * @param deployer Deployer object
  * @param txns transaction array
  * @param txIdxMap transaction map index to name
```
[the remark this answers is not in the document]
amityadav0: ok

packages/algob/src/lib/script-checkpoints.ts
```diff
@@ -209,13 +226,21 @@ export async function registerCheckpoints (
     switch (txn.type) {
       case 'acfg': {
         txConfirmation = await deployer.waitForConfirmation(txn.txID());
+        const key = deployer.getAssetCheckpointKeyFromIndex(txn.assetIndex);
+        if (key && checkAssetDeletionTx(txn)) {
+          const temp: rtypes.ASAInfo = deployer.getASAInfo(key);
```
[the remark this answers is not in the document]
amityadav0: No it won't throw error, because we also have `key`. key won't be defined if asa doesn;t exist , so it will pass.
adding a test also.

packages/algob/src/lib/script-checkpoints.ts
amityadav0: yes

packages/algob/src/lib/script-checkpoints.ts
amityadav0: yes

packages/algob/src/types.ts
amityadav0: thanks, removed

packages/algob/src/lib/script-checkpoints.ts
amityadav0: asdasdas

packages/runtime/src/types.ts
amityadav0: i think we can keep deleted boolean in checkpoints, we have boolean `true` / `false` user can easily know if it's deleted or not, 
and it also keeps checkpoints consistent, if user wants to use them somewhere else, all attributes will be there.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 + Rename `parseArgs` to `parse_params`
 
 + For External support of parameters user should replace TMPL_ prefix in their smart contracts, and only use it when using pyteal.tmpl(..)
++ Rename `appID` to `appID` in all places.
```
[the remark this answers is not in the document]
ratik21: I also think we can clarify a bit (especially for breaking changes), regarding migration to new change

packages/algob/test/fixture-projects/stateful/artifacts/.cp.yaml
amityadav0: checking, added a .gitignore

packages/algob/test/lib/tx.ts
```diff
@@ -161,3 +167,392 @@ describe("ASA modify fields", () => {
...
+      type: types.TransactionType.DestroyAsset,
+      sign: types.SignType.SecretKey,
+      fromAccount: bobAcc,
+      payFlags: {},
+      assetID: 123
+    };
+
+    await executeTransaction(deployer, execParam);
+  });
+});
```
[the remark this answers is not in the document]
amityadav0: yes

packages/algob/src/internal/deployer.ts
```diff
@@ -284,6 +369,98 @@ class DeployerBasicMode {
...
+          this.assertIfASAExist(txn.assetID);
+        }
+        break;
+      }
+      case rtypes.TransactionType.DeleteSSC:
+      case rtypes.TransactionType.OptInSSC:
+      case rtypes.TransactionType.UpdateSSC:
+      case rtypes.TransactionType.CallNoOpSSC: {
+        this.assertIfAppExist(txn.appID);
+        break;
```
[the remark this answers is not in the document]
amityadav0: yes, we have test for it

packages/algob/src/internal/deployer.ts
```diff
@@ -284,6 +369,98 @@ class DeployerBasicMode {
...
+      case rtypes.TransactionType.RevokeAsset:
+      case rtypes.TransactionType.OptInASA:
+      case rtypes.TransactionType.DestroyAsset: {
+        this.assertIfASAExist(txn.assetID);
+        break;
+      }
+      case rtypes.TransactionType.TransferAsset: {
+        // If transaction is not opt-out check for CP deletion
+        if (txn.payFlags.closeRemainderTo === undefined) {
+          this.assertIfASAExist(txn.assetID);
```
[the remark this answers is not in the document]
amityadav0: i think we should let that part handled by algorand node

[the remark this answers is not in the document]
amityadav0: ok, this is according to docs, will add a link here

packages/algob/src/lib/script-checkpoints.ts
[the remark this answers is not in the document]
amityadav0: ok let's update to `checkIf`

packages/algob/src/lib/script-checkpoints.ts
amityadav0: the code seems to be showing that, or we can add a comment with a link, let's add a object

packages/algob/src/internal/deployer.ts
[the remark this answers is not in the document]
amityadav0: `asset cannot be deleted if it is currently in supply`, for asset to be deleted all of it needs to be present in creator's account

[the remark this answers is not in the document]
amityadav0: found out https://developer.algorand.org/docs/reference/transactions/#asset-transfer-transaction
we have assetCloseTo, though cannot find them in fields, so `closeRemainderTo` is not  for asset

packages/algob/src/internal/deployer.ts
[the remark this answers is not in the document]
amityadav0: checked we cannot closout from deleted app.

packages/algob/src/internal/deployer.ts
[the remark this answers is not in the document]
amityadav0: https://developer.algorand.org/articles/algos-asas/#opting-in-and-out-of-asas

https://developer.algorand.org/docs/reference/transactions/#asset-transfer-transaction

types should be `axfer`, will add the links in code as well.

packages/algob/src/lib/script-checkpoints.ts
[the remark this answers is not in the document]
ratik21: yes, please update

packages/algob/src/internal/deployer.ts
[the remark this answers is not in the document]
ratik21: ok, thanks

packages/algob/src/lib/script-checkpoints.ts
```diff
@@ -1,5 +1,6 @@
 
 import { checkIfAssetDeletionTx, loadFromYamlFileSilent, lsTreeWalk, types as rtypes } from "@algo-builder/runtime";
+import { TxnOnComplete } from "@algo-builder/runtime/build/types";
```
ratik21, line 3: we're importing from build here

packages/algob/test/lib/tx.ts
```diff
@@ -1,5 +1,5 @@
 import { encodeNote, types } from "@algo-builder/runtime";
-import { SSCDeploymentFlags } from "@algo-builder/runtime/build/types";
+import { ExecParams, SSCDeploymentFlags } from "@algo-builder/runtime/build/types";
```
ratik21, line 2: we're importing from build here

packages/algob/src/lib/script-checkpoints.ts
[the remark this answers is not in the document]
amityadav0: cool

packages/algob/src/lib/script-checkpoints.ts
amityadav0: thanks

packages/algob/test/lib/tx.ts
```diff
@@ -555,4 +555,25 @@ describe("Delete ASA and SSC transaction flow(with functions and executeTransact
...
+      {
+        type: types.TransactionType.DeleteSSC,
+        sign: types.SignType.SecretKey,
+        fromAccount: bobAcc,
+        payFlags: {},
+        appID: 12213
+      }
+    ];
+
+    await executeTransaction(deployer, txGroup);
```
ratik21, line 579: how about using `assert.notthrows(func, ERRORS.GENERAL.APP_DELETED)` here

packages/algob/test/lib/tx.ts
```diff
@@ -532,7 +532,10 @@ describe("Delete ASA and SSC transaction flow(with functions and executeTransact
       payFlags: {},
       appID: appID
     };
-    await executeTransaction(deployer, execParam);
+    await expectBuilderErrorAsync(
+      async () => await executeTransaction(deployer, execParam),
+      ERRORS.GENERAL.APP_DELETED
+    );
```
ratik21, line 537: please update test description as well

packages/algob/src/internal/deployer.ts
```diff
@@ -38,12 +40,80 @@ class DeployerBasicMode {
...
+        let res: number = -1;
+        const cmpValue: number = -1;
+        map.forEach((value, key) => {
+          if (key >= cmpValue) {
+            res = value.appID;
+          }
+        });
+        return res;
+      }
+    };
```
robert-zaremba, line 106: can you move it to a class please? 
Creating an inline object is not what we want - it tricks the type system and it's not reusable.

amityadav0: ok

packages/algob/test/lib/tx.ts
amityadav0: OK

packages/algob/test/lib/tx.ts
amityadav0: Ok

packages/algob/test/lib/tx.ts
amityadav0: but that means it could throw another error ?

amityadav0: if that's the case, `await exec.....` is better

packages/algob/src/internal/deployer.ts
```diff
@@ -284,6 +306,101 @@ class DeployerBasicMode {
...
+  /**
+   * Check if ASA info exist for asset id / string,
+   * First: search for ASAInfo in checkpoints
+   * Case 1: If it exist check if that info is deleted or not by checking deleted boolean
+   * If deleted boolean is true throw error
+   * else, pass
+   * Case 2: If it doesn't exist, pass
+   * @param asset asset index or asset name
+   */
+  private assertIfASAExist (asset: string | number): void {
```
robert-zaremba, line 319: let's remove the `If` infix. In all assertion libraries we have `assert<noun><verb>`

packages/algob/src/internal/deployer.ts
```diff
@@ -284,6 +306,101 @@ class DeployerBasicMode {
   async ensureCompiled (name: string, force?: boolean, scTmplParams?: SCParams): Promise<ASCCache> {
     return await this.algoOp.ensureCompiled(name, force, scTmplParams);
   }
+
+  /**
+   * Check if ASA info exist for asset id / string,
```
robert-zaremba, line 311:
```suggestion
   * Asserts ASA is defined in a checkpoint by asset id / string,
```

packages/algob/src/internal/deployer.ts
```diff
@@ -284,6 +306,101 @@ class DeployerBasicMode {
...
+    if (res?.deleted === true) {
+      throw new BuilderError(
+        ERRORS.GENERAL.ASSET_DELETED, {
+          asset: asset
+        });
+    }
+  }
+
+  /**
+   * Check if SSC info exist for app id,
```
robert-zaremba, line 336:
```suggestion
   * Asserts App is defined in a checkpoint by app id.
```

packages/algob/src/internal/deployer.ts
```diff
@@ -284,6 +306,101 @@ class DeployerBasicMode {
...
+   * First: search for SSCInfo in checkpoints
+   * Case 1: If it exist check if that info is deleted or not by checking deleted boolean
+   * If deleted boolean is true throw error
+   * else, pass
+   * Case 2: If it doesn't exist, pass
+   * @param appID Application index
+   */
+  private assertIfAppExist (appID: number): void {
```
robert-zaremba, line 344: let's remove "If" infix

robert-zaremba (approved):

packages/algob/src/internal/deployer.ts
amityadav0: ok

packages/algob/src/internal/deployer.ts
amityadav0: ok

PR #366, opened 2021-06-05
add helper function to print top of stack after each opcode execution

ratik21:


amityadav0:
I think according to task description we need `--debug-stack=x`. ex: `algob test --debug-stack=2`
this will print stack traces. Instead of passing variable each time we call execute function.

ratik21:
> I think according to task description we need --debug-stack=x

I don't think it was a `CLI` based task, otherwise i may have misunderstood (task title is to `add helper function to print ...`). 

> algob test --debug-stack=2

a) curious: can we pass `debug-stack` from `yarn run mocha` (in project root) to the `runtime.interpreter` (while doing `algob test`)? note: user creates a new instance of `runtime` in each test, and while running TEAL code new recreate interpreter each time. The goal is to print `top n` elements of teal stack "after each opcode execution" (in TEAL code).

b) Now if above could be done, it will print stack for all the tests, which will result in a very messy console log (maybe we could log in an `outFile: debug-stack.log`,?). I think user will generally inspect the stack for some specific test, let's say if one test is failing, then it makes sense to log the stack only for that transaction in test by the user. Not sure about logging stack for each `interpreter.run` in each test.

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@
 + Added support for sub directories in assets folder, with tests.
 + Update runtime to process execParams.deployASA, deploySSC, OptInToASA, OptIntoSSC
 + Exported `@algorand-builder/algob`, `@algorand-builder/runtime` error types and make it accessible for API documentation.
++ Add `debugStack` option in `runtime.executeTx()` to print stack (upto depth = debugStack) after each opcode execution.
```
robert-zaremba, line 19:
```suggestion
+ Added `debugStack` option in `runtime.executeTx()` to print stack (upto depth = debugStack) after each opcode execution.
```

docs/guide/testing-teal.md
```diff
@@ -321,6 +321,13 @@ Now, we will execute a transaction with an app call (stateful TEAL). The app is
 
 Please look at [stateful-counter.ts](https://github.com/scale-it/algo-builder/blob/master/packages/runtime/test/integration/stateful-counter.ts) to see the complete integration test suite.
 
+## Debugging
+Based upon input from the community, we have introduced an optional parameter `debugStack` to pass with `runtime.executeTx()`. Since setting up teal debugger(start tealdbg instance, creating a dry-run dump.json) could be a lengthy process, this param is helpful to quickly  inspect the state of stack while executing TEAL code (for each opcode in smart contract in test).
```
robert-zaremba, line 325:
```suggestion
`runtime.executeTx()` function has an optional `debugStack: number` parameter. Setting up teal debugger (start tealdbg instance, creating a dry-run dump.json) could be a lengthy process, this param is helpful to quickly  inspect the state of stack while executing TEAL code (for each opcode in smart contract in test).
```

docs/guide/testing-teal.md
```diff
@@ -321,6 +321,13 @@ Now, we will execute a transaction with an app call (stateful TEAL). The app is
...
+If `debugStack` parameter is passed in `executeTx(..)`, then the TEALStack is printed to console (upto max_depth = `debugStack`) after each opcode execution in TEAL file.
```
robert-zaremba, line 326:
```suggestion
When the `debugStack: number` parameter is set, a TEAL stack is printed to a console (up to max_depth = `debugStack`) after each opcode execution in the TEAL file.
```

packages/runtime/test/src/interpreter/stack.ts
```diff
@@ -15,18 +19,34 @@ describe("Stack", function () {
...
+    stack.push(stringToBytes("pulp"));
+    stack.push(stringToBytes("fiction"));
+
+    let newStack = stack.debug(2);
+    assert.equal(newStack.length, 2);
+    assert.deepEqual(newStack[0], stringToBytes("fiction")); // top
+    assert.deepEqual(newStack[1], stringToBytes("pulp")); // 2nd elem from top
+
+    newStack = stack.debug(200); // should return an array of all stack elements
+    assert.equal(newStack.length, stack.length());
```
robert-zaremba, line 50: :+1:

robert-zaremba (approved):
left few suggestions

robert-zaremba:
When creating the task, I was thinking what @amityadav0 wrote

robert-zaremba:
When creating the task, I was thinking what @amityadav0 wrote in https://github.com/scale-it/algo-builder/pull/366#issuecomment-855515657

However, when looking at the implementation - I think this is a good idea. User can debug only part of the test. It makes sense. Let's merge it and see how others will think.

PR #367, opened 2021-06-05
Update workflow

amityadav0:


.github/workflows/main.yaml
```diff
@@ -30,6 +30,15 @@ jobs:
           restore-keys: |
             ${{ runner.OS }}-node-
             ${{ runner.OS }}-
+      - name: Setup Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: "3.7"
```
robert-zaremba, line 36: let's use 3.8

.github/workflows/main.yaml
```diff
@@ -30,6 +30,15 @@ jobs:
...
+      - name: Install python dependencies
+        run: python -m pip install --upgrade pip
+      - run: pip install pyteal
+      - run: pip install pyyaml
```
robert-zaremba, line 40: we should test our `pipenv` workflow. So install `pipenv`, then `pipenv sync` and finally `pipenv shell`.

packages/algob/src/lib/algo-operator.ts
```diff
@@ -145,7 +145,7 @@ export class AlgoOperatorImpl implements AlgoOperator {
     account: rtypes.Account, params: algosdk.SuggestedParams,
     flags: rtypes.TxParams
   ): Promise<void> {
-    console.log(`ASA ${account.name} opt-in for ASA ${asaName}`);
+    console.log(`ASA ${String(account.name)} opt-in for ASA ${String(asaName)}`);
```
robert-zaremba, line 148: why we need to cast to string? Isn't this already a string?

amityadav0: because of yarn link while checking with github actions, in local system it doesn't give error
but with actions it does, i searched about why this is happening, got answer maybe tslint is missing some edge cases or these values maybe undefined at some point.

.github/workflows/main.yaml
amityadav0: ok

.github/workflows/main.yaml
amityadav0: https://github.com/scale-it/algo-builder/pull/367/checks?check_run_id=2764586424
getting this error

amityadav0: updated to use pipenv

.github/workflows/main.yaml
```diff
@@ -30,10 +30,17 @@ jobs:
...
+        run: python -m pip install pipenv
+      - run: pipenv sync
       - name: Install dependencies
-        run: yarn install
+        run: pipenv run yarn install
       - name: Execute lint
-        run: yarn lint
-      - run: yarn build
+        run: pipenv run yarn lint
+      - run: pipenv run yarn build
```
robert-zaremba, line 44: I think we don't need to prefix lint and build with `pipenv`

requirements.txt
```diff
@@ -0,0 +1,2 @@
+pyteal===0.7.0
+pyyaml===5.4.1
```
robert-zaremba, line 2: we can remove this file.

robert-zaremba (approved):
left 2 comments.

requirements.txt
amityadav0: ok

.github/workflows/main.yaml
amityadav0: ok, checking and removing

PR #374, opened 2021-06-11
Clean yarn linter warnings

ratik21:


.eslintrc.js
```diff
@@ -25,6 +25,7 @@ module.exports = {
     "no-underscore-dangle": 0,
     "simple-import-sort/sort": "warn",
     "sort-imports": "off",
+    "no-unused-vars": "off",
```
amityadav0, line 28: i don't think we should do this, it will miss major linting, when we write code

ratik21: no, in typescript-eslint it's recommended to do it - [https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-unused-vars.md#how-to-use](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-unused-vars.md#how-to-use)

because we're using "@typescript-eslint/no-unused-vars" rules here. Checked in the js-examples as well, works as expected.

amityadav0:
how about having error for unsed variables ?

ratik21:
> how about having error for unused variables ?

I don't have strong preference b/w error or warning. I slightly prefer to have a warning for `no-unused-vars`. Could use another opinion, @robert-zaremba, what do you think?

robert-zaremba (approved):

PR #382, opened 2021-06-16
add test to parsing functions

amityadav0:


packages/runtime/test/src/lib/parsing.ts
```diff
@@ -64,6 +66,30 @@ describe("Convert integer to big endian", () => {
...
+    let initialString = "50";
+    let first = stringToBytes(initialString);
+    let second = decodeUint64(first, DecodingMode.BIGINT);
+    let third = uint64ToBigEndian(second);
+    let final = convertToString(third);
+
+    // These two are not exactly equal
+    // actual - '\u0000\u0000\u0000\u0000\u0000\u000050'
+    // expected - '50'
+    assert.notEqual(final, initialString);
```
ratik21, line 80: I think we should compare `Uint8Arrays` here, as in stack (both in algorand network & runtime), strings are processed in bytes

packages/runtime/test/src/lib/parsing.ts
```diff
@@ -64,6 +66,30 @@ describe("Convert integer to big endian", () => {
...
+  it("should be equal when string in bytes is passed to uint decoder", () => {
+    let initialString = "50";
+    let first = stringToBytes(initialString);
+    let second = decodeUint64(first, DecodingMode.BIGINT);
+    let third = uint64ToBigEndian(second);
+    let final = convertToString(third);
+
+    // These two are not exactly equal
+    // actual - '\u0000\u0000\u0000\u0000\u0000\u000050'
+    // expected - '50'
```
ratik21, line 79: I think expected should be `new Uint8Array([0, 0, 0, 0, 0, 0, 0, 50])` (big endian bytes).

amityadav0: what do you mean ? how ? we are converting string ? (it won't be that value)
this is converting bytes back to string to confirm the identity.(back and forth)

also if you think, first value will be bytes with no padded zeroes, but third value will have padded zeroes
so only way to know if they give same results is to convert them back to string.

ratik21: > what do you mean ? how ? we are converting string ? (it won't be that value)
> so only way to know if they give same results is to convert them back to string.


I don't understand why you're testing against strings here. In algorand network & runtime as well we process strings as `bytes[]/uint8Arrays`. Even in algo-builder as well we feed the application args (`eg. 'str:50'`) after converting to bytes. 

And i'm not sure about encodings of `go-algorand`, but while inspecting the state with `js-sdk`, byte keys are interpreted as base64 encoded strings - [https://github.com/scale-it/algo-builder/blob/master/examples/nft/scripts/transfer/common.js#L15](https://github.com/scale-it/algo-builder/blob/master/examples/nft/scripts/transfer/common.js#L15) (i am assuming here you understand `\u0000\u0000\u0000\u0000\u0000\u000050` is not a proper utf-8 string). 

So, i think we should be testing for padded zeros against bytes (`Uint8Arrays` in javascript). Having strings as `\u0000\u0000\u0000\u0000\u0000\u000050` is very confusing and does not make sense. For a reference check out these tests - [https://github.com/algorand/js-algorand-sdk/blob/develop/tests/2.Encoding.js](https://github.com/algorand/js-algorand-sdk/blob/develop/tests/2.Encoding.js), much better readable & makes sense as well!

amityadav0: > > what do you mean ? how ? we are converting string ? (it won't be that value)
> > so only way to know if they give same results is to convert them back to string.
> 
> I don't understand why you're testing against strings here. In algorand network & runtime as well we process strings as `bytes[]/uint8Arrays`. Even in algo-builder as well we feed the application args (`eg. 'str:50'`) after converting to bytes.
> 
> And i'm not sure about encodings of `go-algorand`, but while inspecting the state with `js-sdk`, byte keys are interpreted as base64 encoded strings - https://github.com/scale-it/algo-builder/blob/master/examples/nft/scripts/transfer/common.js#L15 (i am assuming here you understand `\u0000\u0000\u0000\u0000\u0000\u000050` is not a proper utf-8 string).
> 
> So, i think we should be testing for padded zeros against bytes (`Uint8Arrays` in javascript). Having strings as `\u0000\u0000\u0000\u0000\u0000\u000050` is very confusing and does not make sense. For a reference check out these tests - https://github.com/algorand/js-algorand-sdk/blob/develop/tests/2.Encoding.js, much better readable & makes sense as well!

https://discord.com/channels/754286650540032030/754286650540032033/854296912097968128

amityadav0: actually this task is about testing what code and not tweak code to what we want it to do.
also, Please help me understand difference between `base64` and `utf-8` (i am assuming there is)

amityadav0: ` For a reference check out these tests - https://github.com/algorand/js-algorand-sdk/blob/develop/tests/2.Encoding.js, much better readable & makes sense as well!` have you checked the issue ?
these tests doesn't include that where user sends `str:50` instead of `int:50`.

ratik21: > actually this task is about testing what code does and not tweak code to what we want it to do.

What are we tweaking here?

> Please help me understand difference between base64 and utf-8 (i am assuming there is)

You're assuming that the response is utf-8 encoded string, which is wrong. In the test `convertToString(third);` is blindly converting to utf-8 encoded string, which is logically incorrect. If it's still not clear, try running nft example (or any stateful example) and see if state returns utf-8 / base64 encoded string.

Hope it makes sense now

amityadav0: instead of changing functions, changing test value(check bytes in this case) is tweaking code to make it work.

does that mean `convertToString` is incorrect or tests ?

ratik21: test itself was wrong

amityadav0: hmm, all of us had different cases in our mind
anyway let's update this

ratik21 (approved):

PR #383, opened 2021-06-16
Add docs for parameters passing

amityadav0:


docs/guide/execute-transaction.md
```diff
@@ -0,0 +1,147 @@
...
+
+`executeTransaction` function can be used to perform many transactions. It supports every transaction which is possible in algorand network. Ex: Deploy ASA/SSC, Opt-In, Transfers, Delete, Destroy etc. `executeTransaction` takes `ExecParams` or `ExecParams[]` as parameter.
+If you pass an array of `ExecParams`, it will be considered as `atomic transaction`.
+In below sections we will demonstrate how to pass these parameters.
+
+Note: For each parameter `type` and `sign` fields are mandatory.
+  - `type`: type of transaction
+  - `sign`: sign type
+
+Examples of parameter(`ExecParams`):
```
ratik21, line 15: We should explain common parameters here: `TxParams` (passed in `execParams.payflags`)

docs/guide/execute-transaction.md
```diff
@@ -0,0 +1,147 @@
...
+
+```js
+  {
+    type: TransactionType.OptInASA,
+    sign: SignType.SecretKey,
+    fromAccount: alice,
+    assetID: assetIndex,
+    payFlags: { totalFee: 1000 }
+  }
+```
```
ratik21, line 65: Let's cover both cases: `optInLsigToASA`, `optInAccountToASA`. Same for SSC.

docs/guide/execute-transaction.md
```diff
@@ -0,0 +1,147 @@
+---
+layout: splash
+---
+
+# Execute Transaction
+
+`executeTransaction` function can be used to perform many transactions. It supports every transaction which is possible in algorand network. Ex: Deploy ASA/SSC, Opt-In, Transfers, Delete, Destroy etc. `executeTransaction` takes `ExecParams` or `ExecParams[]` as parameter.
```
ratik21, line 7:
```suggestion
`executeTransaction` is a high level function which can be used to perform transactions on Algorand Network. It supports every transaction (atomic or single) which is possible in network. Ex: Deploy ASA/SSC, Opt-In, Transfers, Delete, Destroy etc. `executeTransaction` takes `ExecParams` or `ExecParams[]` as parameter.
```

docs/guide/execute-transaction.md
amityadav0: No I don't think so, because it will be copy pasting doc string of that interface, main goal of this task is to give clear picture on what to pass, because this part is cryptic.

docs/guide/execute-transaction.md
amityadav0: https://github.com/scale-it/algo-builder/blob/master/docs/guide/deployer.md#optin-to-asa , these are already covered, the main goal here is to `how to pass params`

docs/guide/execute-transaction.md
amityadav0: after looking that we don't have explanation in api docs https://algobuilder.dev/api/algob/interfaces/runtime.types.txparams.html , we should add more comments on code in types and add a link to that in docs.

docs/guide/execute-transaction.md
```diff
@@ -0,0 +1,147 @@
...
+    sign: SignType.SecretKey,
+    fromAccount: john,
+    toAccountAddr: alice.address,
+    amount: 10,
+    assetID: assetId,
+    payFlags: { totalFee: 1000 }
+  }
+```
+
+# Deploy SSC
```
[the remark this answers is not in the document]
amityadav0: let's add these definitions in interfaces and add a link to api docs here, https://algobuilder.dev/api/algob/interfaces/runtime.types.sscoptionalflags.html

docs/guide/execute-transaction.md
ratik21: As you prefer. I just want these explanations to be included in user docs (as user will probably start from here).

docs/guide/execute-transaction.md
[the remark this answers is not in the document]
ratik21: ok

docs/guide/execute-transaction.md
ratik21: Please add this link here, remember these are user docs

docs/guide/execute-transaction.md
amityadav0: yes, api docs are incomplete, added docs

docs/guide/execute-transaction.md
amityadav0: user will start from `deployer`, that's the essential part, then after reading that user will come here which is more advanced than deployer. adding executeTransaction link in deployer is better

docs/guide/execute-transaction.md
robert-zaremba: Let's add a link to API docs in the first line of the section.

docs/guide/execute-transaction.md
```diff
@@ -0,0 +1,147 @@
...
+`executeTransaction` is a high level function which can be used to perform transactions on Algorand Network. It supports every transaction (atomic or single) which is possible in network. Ex: Deploy ASA/SSC, Opt-In, Transfers, Delete, Destroy etc. `executeTransaction` takes `ExecParams` or `ExecParams[]` as parameter.
+If you pass an array of `ExecParams`, it will be considered as `atomic transaction`.
+In below sections we will demonstrate how to pass these parameters.
+
+Note: For each parameter `type` and `sign` fields are mandatory.
+  - `type`: type of transaction
+  - `sign`: sign type
+
+Examples of parameter(`ExecParams`):
+# Transfer Algo using secret key
```
robert-zaremba, line 16:
```suggestion
### Transfer Algo using secret key
```

robert-zaremba: First level header should be use only for title.
Let's update the headers below as well

robert-zaremba (approved):
Please change the header levels.

docs/guide/execute-transaction.md
amityadav0: ok

docs/guide/execute-transaction.md
amityadav0: OK

PR #390, opened 2021-06-24
Migrate types

amityadav0:


examples/htlc-pyteal-ts/package.json
```diff
@@ -4,7 +4,7 @@
   "main": "index.js",
   "license": "MIT",
   "dependencies": {
-    "algosdk": "^1.9.1"
+    "algosdk": "scale-it/js-algorand-sdk.git#f0b9dd9bf45827e0b448a2f6a7dd8c0735317d24"
```
ratik21, line 7: The type changes are now released - [https://github.com/algorand/js-algorand-sdk/releases/tag/v1.10.1](https://github.com/algorand/js-algorand-sdk/releases/tag/v1.10.1). Can we use `^1.10.1` here?

examples/multisig/scripts/multisig_goal_sc.js
```diff
@@ -27,6 +27,7 @@ async function run (runtimeEnv, deployer) {
     amountMicroAlgos: 10000000,
     payFlags: { note: 'Funding multisig account', totalFee: 500 } // totalFee will be converted to 1000 as it is minimum required transaction fee
   };
+  console.log(masterAccount.address, multsigaddr);
```
ratik21, line 30: do we need this comment?

packages/algob/src/internal/deployer.ts
```diff
@@ -101,7 +101,7 @@ class DeployerBasicMode {
     return this.algoOp.algodClient;
   }
 
-  async waitForConfirmation (txId: string): Promise<algosdk.ConfirmedTxInfo> {
+  async waitForConfirmation (txId: string): Promise<algosdk.modelsv2.PendingTransactionResponse> {
```
ratik21, line 104: Can we directly import `modelsv2` first ( `import { modelsv2 } from "algosdk"`) & directly do `modelsv2.<type>` at all places?

packages/algob/src/lib/algo-operator.ts
```diff
@@ -208,7 +211,8 @@ export class AlgoOperatorImpl implements AlgoOperator {
       if (account.addr === creator.addr) {
         throw new BuilderError(ERRORS.SCRIPT.ASA_TRIED_TO_OPT_IN_CREATOR);
       }
-      const accountInfo = await this.algodClient.accountInformation(account.addr).do();
+      const accountInfo = await
+      this.algodClient.accountInformation(account.addr).do() as algosdk.modelsv2.Account;
```
ratik21, line 215: Could you explain why we're adding  `as algosdk.modelsv2.Account` here ? Previously it as not there.

packages/algob/src/lib/script-checkpoints.ts
```diff
@@ -223,9 +223,9 @@ export async function registerCheckpoints (
           const asaInfo: rtypes.ASAInfo = {
             creator: encodeAddress(txn.from.publicKey),
             txId: txn.txID(),
-            assetIndex: txConfirmation["asset-index"],
-            confirmedRound: txConfirmation['confirmed-round'],
-            assetDef: res[1],
+            assetIndex: Number(txConfirmation.assetIndex),
```
ratik21, line 226: Are we doing `Number(assetIndex)` because algosdk type is `number | bigint` ? If yes then can we update `rtypes.ASAInfo.assetIIndex` to `number | bigint` (same as SDK) ?

packages/algob/src/lib/tx.ts
```diff
@@ -70,15 +70,15 @@ export function makeAssetCreateTxn (
...
-    asaDef.decimals,
-    asaDef.defaultFrozen,
-    asaDef.manager,
-    asaDef.reserve,
-    asaDef.freeze,
-    asaDef.clawback,
-    asaDef.unitName,
+    Number(asaDef.decimals),
+    asaDef.defaultFrozen ? asaDef.defaultFrozen : false,
+    asaDef.manager ? asaDef.manager : "",
```
ratik21, line 75: Can we do: `asaDef.manager ?? ""` ?

packages/algob/test/lib/tx.ts
```diff
@@ -143,11 +141,11 @@ describe("ASA modify fields", () => {
    * Verifies correct asset fields are sent to network
    * @param rawTxns rawTxns Signed transactions in Uint8Array
    */
-  function checkTx (rawTxns: Uint8Array | Uint8Array[]): Promise<ConfirmedTxInfo> {
+  function checkTx (rawTxns: Uint8Array | Uint8Array[]): Promise<modelsv2.PendingTransactionResponse> {
     if (isArray(rawTxns)) {
       // verify here if group tx
     } else {
-      const tx: Transaction = decodeSignedTransaction(rawTxns).txn;
+      const tx: any = decodeSignedTransaction(rawTxns).txn;
```
ratik21, line 148: was `Transaction` incorrect here?

packages/algob/test/lib/tx.ts
```diff
@@ -709,7 +707,7 @@ describe("Update transaction test in run mode", () => {
       type: wtypes.TransactionType.UpdateApp,
       sign: wtypes.SignType.SecretKey,
       fromAccount: bobAcc,
-      appID: appInfo["application-index"],
+      appID: appInfo.applicationIndex as number,
```
ratik21, line 710: same as assetIndex.

packages/web/src/types.ts
```diff
@@ -236,3 +236,46 @@ export type AppCallsParam = BasicParams & AppOptionalFlags & {
...
+  // Signs and appends a signature
+  appendToMultisig: (secretKey: Uint8Array) => void
+  // signs and returns program signature, without appending it to this object
+  signProgram: (secretKey: Uint8Array) => Uint8Array
+  singleSignMultisig: (secretKey: Uint8Array, msig: EncodedMultisig) => [Uint8Array, number]
+  // serializes and encodes the LogicSig
+  toByte: () => Uint8Array
+  // deserializes a LogicSig which was serialized using toByte()
+  fromByte: (encoded: Uint8Array) => LogicSig
+}
```
ratik21, line 269: These are not exported in algosdk? Can we request them to export these?

packages/types-algosdk/types/index.d.ts
```diff
@@ -1,770 +0,0 @@
-/* eslint-disable  @typescript-eslint/naming-convention */
```
ratik21, line 1: Should we delete this immediately? If a user is using this package (say via linking), then code will break after release. Maybe we should add a depreciation notice first?

ratik21 (commented):
Could you please update the changelog that we have migrated to new `algosdk.types`, and thus add a depreciation notice that we won't be maintaining our `types-algosdk` ?

packages/types-algosdk/types/index.d.ts
amityadav0: we are in develop branch, so this is OK. we can add this in master before updating.

examples/htlc-pyteal-ts/package.json
amityadav0: no, there is one issue in types of js-sdk, I have updated in scaleit. Will update it here.

packages/algob/src/internal/deployer.ts
amityadav0: yes, it is like that l. Probably missed here. Will update it.

examples/multisig/scripts/multisig_goal_sc.js
amityadav0: No

packages/algob/src/lib/script-checkpoints.ts
amityadav0: the main problem is undefined, it is a compiler issue. We can change it to bigInt or number

packages/algob/test/lib/tx.ts
amityadav0: same as assetIndex. compiler issue with undefined

packages/web/src/types.ts
amityadav0: these are exported as a class, they don't have them as types. Importing class was incompatible that's why I have added this here.

amityadav0: Will check once again I forgot the exact  problem

packages/algob/src/lib/tx.ts
amityadav0: this needs to be updated to undefined.

packages/algob/src/lib/algo-operator.ts
amityadav0: it's a compiler issue, when we load using do() function we get Record type which is used as Account type
Reference error:
`Argument of type 'Record<string, any>' is not assignable to parameter of type 'Account'.`

packages/web/src/types.ts
amityadav0: hk
This was the issue.

packages/algob/test/lib/tx.ts
amityadav0: previoulsy it was, i saw there is updated commit in 1.10.0 , i will push the fixes after this and update scale-it.

CHANGELOG.md
```diff
@@ -33,6 +33,7 @@
 + Import are changed to scoped imports
     + instead of stringToBytes, you can import a `convert` namespace, and then use `convert.stringToBytes`
 + Types imports for `ExecParams`, `TransactionTypes`, `SignType` moved to new package `@algo-builder/web`
++ Migrate to algorand/js-sdk types from types-algosdk.
```
ratik21, line 36:
```suggestion
+ Migrate to algorand/js-sdk types from `@algo-builder/types-algosdk`.
```

packages/algob/src/lib/msig.ts
```diff
@@ -1,5 +1,6 @@
 import { getPathFromDirRecursive } from "@algo-builder/runtime";
-import type { Account, LogicSig, MultiSig, TxSig } from "algosdk";
+import { types as wtypes } from "@algo-builder/web";
+import type { Account, EncodedMultisig, LogicSig } from "algosdk";
```
ratik21, line 3: linter warning: `'wtypes' is defined but never used. Allowed unused vars must match /_/u`

packages/algob/src/lib/tx.ts
```diff
@@ -70,15 +70,15 @@ export function makeAssetCreateTxn (
...
-    asaDef.freeze,
-    asaDef.clawback,
-    asaDef.unitName,
+    Number(asaDef.decimals),
+    asaDef.defaultFrozen ? asaDef.defaultFrozen : false,
+    asaDef.manager !== "" ? asaDef.manager : undefined,
+    asaDef.reserve !== "" ? asaDef.reserve : undefined,
+    asaDef.freeze !== "" ? asaDef.freeze : undefined,
+    asaDef.clawback !== "" ? asaDef.clawback : undefined,
+    asaDef.unitName ? asaDef.unitName : undefined,
```
ratik21, line 79: Can't we directly use `asaDef.unitName` here? 
Isn't `asaDef.unitName !== undefined ? asaDef.unitName : undefined` same as `asaDef.unitName` ? Or do we check for something else here (other than undefined)?

packages/algob/src/lib/tx.ts
```diff
@@ -70,15 +70,15 @@ export function makeAssetCreateTxn (
...
     name,
-    asaDef.url,
+    asaDef.url ? asaDef.url : "",
```
ratik21, line 81:
```suggestion
    asaDef.url ?? "",
```

packages/algob/test/lib/tx.ts
ratik21: will we update this from `any`?

packages/web/src/types.ts
```diff
@@ -236,3 +236,15 @@ export type AppCallsParam = BasicParams & AppOptionalFlags & {
...
+  response?: {
+    statusCode: number
+    text: string
+    body: {
+      message: string
+    }
+    error?: Error
+  }
+  error?: Error
+}
```
ratik21, line 250: I think we have a similar type defined in `algob/types`. Could you please check? We can use this one as common.

ratik21 (approved):

packages/web/src/types.ts
amityadav0: no we don't have it. It was present in types-algosdk. i couldn't find it in js-sdk

packages/algob/src/lib/tx.ts
amityadav0: yes

packages/algob/src/lib/msig.ts
amityadav0: removed

PR #391, opened 2021-06-24
fix example/multisig

amityadav0:
erir


ratik21 (approved):
hmm, so this happens when running this example directly after setting up a fresh network. Which is why we're funding toAccount as well. Will have a look at other examples

amityadav0:
yes, checked and updated other examples as well

PR #398, opened 2021-06-30
Workflow for examples

amityadav0:
## Proposed Changes

+ Add workflow for examples
+ Add deploy scripts
+  Add run scripts
Note: some run scripts are not added because they depend on cases on how user interacts with the example, essential scripts are included for testing.

### DIFF
--- a/.github/workflows/examples.yaml
+++ b/.github/workflows/examples.yaml
@@ -53,28 +53,28 @@ jobs:
       - name: Example ASA
         run: yarn link @algo-builder/algob  && cd ~/.yarn/bin && chmod +x algob
       - run: ls
-      - run: cd examples/asa && yarn algob deploy
+      - run: cd examples/asa && pipenv run yarn algob deploy
 
       - name: Example crowdfunding
-        run: cd examples/crowdfunding && yarn run algob deploy
+        run: cd examples/crowdfunding && pipenv run yarn algob deploy
       
       - name: Example htlc-pyteal-ts
-        run: cd ~/algo-builder/examples/htlc-pyteal-ts && yarn run algob deploy
+        run: cd examples/htlc-pyteal-ts && pipenv run yarn algob deploy
       
       - name: Example multisig
-        run: cd ~/algo-builder/examples/multisig && yarn run algob deploy
+        run: cd examples/multisig && pipenv run yarn algob deploy
       
       - name: Example nft
-        run: cd ~/algo-builder/examples/nft && yarn run algob deploy
+        run: cd examples/nft && pipenv run yarn algob deploy
 
       - name: Example permissioned-token
-        run: cd ~/algo-builder/examples/permissioned-token && yarn run algob deploy
+        run: cd examples/permissioned-token && pipenv run yarn algob deploy
 
       - name: Example permissioned-token-freezing
-        run: cd ~/algo-builder/examples/permissioned-token-freezing && yarn run algob deploy
+        run: cd examples/permissioned-token-freezing && pipenv run yarn algob deploy
 
       - name: Example permissioned-voting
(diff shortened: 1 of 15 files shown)


ratik21 (commented):
each linter warning is being shown 2 times:
image

.github/workflows/examples.yaml
```diff
@@ -0,0 +1,82 @@
...
+        uses: actions/setup-python@v2
+        with:
+          python-version: "3.8"
+      - name: Install python dependencies
+        run: python -m pip install pipenv
+      - run: pipenv sync
+      - name: Install dependencies
+        run: pipenv run yarn install
+      - name: Execute lint
+        run: yarn lint
```
ratik21, line 42: Could it be possible we're seeing multiple lint warnings as we're executing lint in multiple workflows?

.github/workflows/examples.yaml
```diff
@@ -0,0 +1,82 @@
...
+
+      - name: Setup Network
+        run: sudo apt-get update && sudo apt-get install -y gnupg2 curl software-properties-common
+      - run: curl -O https://releases.algorand.com/key.pub && sudo apt-key add key.pub
+      - run: sudo add-apt-repository "deb [arch=amd64] https://releases.algorand.com/deb/ stable main"
+      - run: sudo apt-get update && sudo apt-get install -y algorand
+      - run: cd infrastructure && make setup-private-net
+
+      - name: Example ASA
+        run: yarn link @algo-builder/algob  && cd ~/.yarn/bin && chmod +x algob
```
ratik21, line 54:
```suggestion
        run: yarn link @algo-builder/algob  && chmod +x ~/.yarn/bin/algob
```

ratik21: let's move this out of  `Example ASA` & move to `Setup Algob`

.github/workflows/examples.yaml
amityadav0: maybe, do we need a lint in examples ? i think we can remove this

.github/workflows/examples.yaml
```diff
@@ -0,0 +1,95 @@
...
+      - name: Setup algob
+        run: yarn link @algo-builder/algob  && chmod +x ~/.yarn/bin/algob
+
+      - name: Example ASA 
+        run: cd examples/asa && pipenv run yarn algob deploy
+      - run: cd examples/asa && pipenv run yarn algob run scripts/transfer/gold-contract-sc.js
+      - run: cd examples/asa && pipenv run yarn algob run scripts/transfer/gold-delegated-lsig.js
+      - run: cd examples/asa && pipenv run yarn algob run scripts/transfer/gold-to-john.js
+      - run: cd examples/asa && pipenv run yarn algob run scripts/transfer/master-fund-john.js
+      - run: cd examples/asa && pipenv run yarn algob run scripts/transfer/tesla-to-john.js
```
ratik21, line 60: do we need to `cd` on each run? i was reading this [https://stackoverflow.com/a/58142276](https://stackoverflow.com/a/58142276). Can we use the same concept here?

amityadav0: yes we need it, ok will try this

amityadav0: no it doen;t work, it works only for a line under name section, so you have create more names. i think current one is good.

ratik21: ok

ratik21 (approved):

.github/workflows/examples.yaml
amityadav0: trying run with `|` , this may work

amityadav0: works, thanks

PR #399, opened 2021-06-30
update algob-test

ratik21:
## Proposed Changes

previously `algob test` only ran JS files in `/test`, added support to run tests in typescript project with `algob test`.

### DIFF
--- a/packages/algob/src/builtin-tasks/test.ts
+++ b/packages/algob/src/builtin-tasks/test.ts
@@ -1,3 +1,4 @@
+import findupSync from "findup-sync";
 import Mocha from "mocha";
 
 import { task } from "../internal/core/config/config-env";
@@ -6,10 +7,17 @@ import type { Config } from "../types";
 import { loadFilenames } from "./deploy";
 import { TASK_TEST } from "./task-names";
 
+const TEST_DIR = 'test';
 async function runTests (config: Config): Promise<void> {
   try {
-    const testFiles = loadFilenames(testsDirectory, "test");
-    console.log("Test files:", testFiles);
+    const tsPath = findupSync("tsconfig.json", { cwd: process.cwd() });
+    if (tsPath) {
+      // run tests via ts-mocha, if project is in typescript
+      process.env.TS_NODE_PROJECT = tsPath;
+      require('ts-mocha');
+    }
+
+    const testFiles = loadFilenames(testsDirectory, TEST_DIR);
     const mocha = new Mocha(config.mocha);
     // Adding test files to mocha object
     testFiles.forEach((file) => mocha.addFile(file));
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## unreleased
 
+### Improvements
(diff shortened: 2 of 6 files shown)


packages/algob/src/builtin-tasks/test.ts
```diff
@@ -6,10 +7,17 @@ import type { Config } from "../types";
...
-    const testFiles = loadFilenames(testsDirectory, "test");
-    console.log("Test files:", testFiles);
+    const tsPath = findupSync("tsconfig.json", { cwd: process.cwd() });
+    if (tsPath) {
+      // run tests via ts-mocha, if project is in typescript
+      process.env.TS_NODE_PROJECT = tsPath;
+      require('ts-mocha');
+    }
+
+    const testFiles = loadFilenames(testsDirectory, TEST_DIR);
```
amityadav0, line 20: do we have test for this ?

amityadav0: if not, please add it.

ratik21: this is a common function for deploy and run tasks as well (to load scripts). We have test for it in deploy task - [https://github.com/scale-it/algo-builder/blob/master/packages/algob/test/builtin-tasks/deploy.ts#L15](https://github.com/scale-it/algo-builder/blob/master/packages/algob/test/builtin-tasks/deploy.ts#L15). I think all tests for deploy/run would fail if this function fails.
Let me know if you want to add more tests.

amityadav0: I think this test is already passing. so why are we adding new feature?  if that test passes before and after that means it doesn't test the feature added here

ratik21: > do we have test for this ?

this was confusing. I thought you meant test for `loadFilenames`.

ratik21: will add test for `process.env.TS_NODE_PROJECT`. 

BDW - how do we add test for a command which runs mocha in project root (basically a test for a command which runs all tests in project) ? The feature here is to run mocha on ts-files as well.

packages/algob/src/builtin-tasks/test.ts
```diff
@@ -6,10 +7,17 @@ import type { Config } from "../types";
...
+const TEST_DIR = 'test';
 async function runTests (config: Config): Promise<void> {
   try {
-    const testFiles = loadFilenames(testsDirectory, "test");
-    console.log("Test files:", testFiles);
+    const tsPath = findupSync("tsconfig.json", { cwd: process.cwd() });
+    if (tsPath) {
+      // run tests via ts-mocha, if project is in typescript
+      process.env.TS_NODE_PROJECT = tsPath;
+      require('ts-mocha');
```
robert-zaremba, line 17: good, this should not be imported in a non typescript project because it's a big dependency and takes lot of time.

robert-zaremba (approved):

packages/algob/src/builtin-tasks/test.ts
ratik21: yes

PR #407, opened 2021-07-07
Integrate types

amityadav0:
## Proposed Changes

+ Integrate algo-signer types with @algob/web

### DIFF
--- a/packages/web/src/algo-signer-types.ts
+++ b/packages/web/src/algo-signer-types.ts
@@ -1,3 +1,4 @@
+// https://github.com/PureStake/algosigner/blob/develop/packages/common/src/types.ts
 /* eslint-disable no-unused-vars */
 export enum RequestErrors {
   None,
--- a/packages/web/src/algo-signer-types.ts
+++ b/packages/web/src/algo-signer-types.ts
@@ -1,0 +1,136 @@
+/* eslint-disable no-unused-vars */
+export enum RequestErrors {
+  None,
+  NotAuthorized = '[RequestErrors.NotAuthorized] The extension user does not authorize the request.',
+  InvalidTransactionParams = '[RequestErrors.InvalidTransactionParams] Invalid transaction parameters.',
+  UnsupportedAlgod = '[RequestErrors.UnsupportedAlgod] The provided method is not supported.',
+  UnsupportedLedger = '[RequestErrors.UnsupportedLedger] The provided ledger is not supported.',
+  Undefined = '[RequestErrors.Undefined] An undefined error occurred.',
+}
+
+export type Field<T> = string | number;
+
+export type TAccount = Field<string>;
+export type Note = Field<string>;
+export type Amount = Field<number>;
+
+export interface Transaction {
+  readonly amount: Amount
+  readonly from: TAccount
+  readonly note?: Note
+  readonly to: TAccount
+}
+
+export interface MultisigTransaction {
+  readonly msig: any
+  readonly txn: Transaction
+}
+
+export interface WalletMultisigMetadata {
+  readonly version: number
+  readonly threshold: number
+  readonly addrs: string[]
+}
+
+export interface WalletTransaction {
+  readonly txn: string
+  readonly signers?: string[]
+  readonly message?: string
+  readonly msig?: WalletMultisigMetadata
+  readonly authAddr?: string
+}
+
+export const JSONRPC_VERSION: string = '2.0';
+
+/* eslint-disable no-unused-vars */
+export enum JsonRpcMethod {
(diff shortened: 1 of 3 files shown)


packages/web/src/algo-signer-types.ts
```diff
@@ -0,0 +1,136 @@
+/* eslint-disable no-unused-vars */
```
ratik21, line 1: Can you please add link(s) to top of file, from where we have taken these typings?

packages/web/src/lib/web-mode.ts
```diff
@@ -74,17 +75,17 @@ export class WebMode {
...
+      genesisID: txParams["genesis-id"] as string,
+      firstRound: txParams[LAST_ROUND] as number,
       lastRound: Number(txParams[LAST_ROUND]) + 1000,
       flatFee: false
     };
 
     s.flatFee = userParams.totalFee !== undefined;
-    s.fee = userParams.totalFee ?? userParams.feePerByte ?? txParams["min-fee"];
-    if (s.flatFee) s.fee = Math.max(s.fee, txParams["min-fee"]);
+    s.fee = userParams.totalFee ?? userParams.feePerByte ?? txParams["min-fee"] as number;
```
ratik21, line 87: I think we'll probably need to update these, because if we're using `??` in web, then import still won't work in webapp.

packages/web/src/algo-signer-types.ts
```diff
@@ -0,0 +1,136 @@
...
+
+  algod: (params: JsonPayload, error?: RequestErrors) => Promise<JsonPayload>
+
+  indexer: (params: JsonPayload, error?: RequestErrors) => Promise<JsonPayload>
+
+  signTxn: (
+    transactions: WalletTransaction[],
+    error?: RequestErrors
+  ) => Promise<JsonPayload>
+}
```
ratik21, line 136: So these types are not exported in algosigner? Can we request them to export these typings and add a TODO here?
I think that will be good because then we won't need to maintain their types.

packages/web/src/lib/web-mode.ts
```diff
@@ -128,6 +129,9 @@ export class WebMode {
     }
 
     const txInfo = await this.sendTransaction(signedTxn);
-    return await this.waitForConfirmation(txInfo.txId);
+    if (typeof txInfo.txId === "string") {
+      return await this.waitForConfirmation(txInfo.txId);
+    }
```
ratik21, line 134: Could you please explain this? What could be the value of `txInfo.txId` if not string?

amityadav0: Please check the typings, according to algosigner types we are getting jsonpayload, which could be undefined, number, etc.

packages/web/src/algo-signer-types.ts
amityadav0: I thought that but How will we import those ? Algosigner is a chrome extension, we cannot import it, it is not present in npm, I couldn't find a way to import extension in package,
Please let me know if you find something.

packages/web/src/algo-signer-types.ts
amityadav0: Yes

packages/web/src/lib/web-mode.ts
amityadav0: Thanks for catching

packages/web/src/lib/web-mode.ts
ratik21: my question is wouldn't this code break on `typeof undefined.txId` (cannot fetch property `txId` of `undefined`..). Or is `typeof undefined.key` results in `undefined` also?

ratik21: return types are of `txInfo` right (not txInfo.txId) ?

amityadav0: what do you suggest ?

ratik21: if `typeof undefined.txId` will break then we should use 
`if (typeof txInfo?.txId === "string")` OR
`if (txInfo && typeof txInfo.txId === "string")`

Note: we'll need to use the latter, as we can't use optional chaning. But my comment was really a question, i am not sure myself. Will need to test if it will break or not.

packages/web/src/algo-signer-types.ts
ratik21: hmm, makes sense. I think we can ask once about this on their discord, maybe they'll release their typings under `@types/algosigner`.

amityadav0: ok, will ask

ratik21 (approved):

PR #408, opened 2021-07-08
document `sign-multisig` command in user guide

ratik21:

### DIFF
--- a/docs/guide/README.md
+++ b/docs/guide/README.md
@@ -20,6 +20,8 @@ layout: splash
 + [Test TEAL](./testing-teal.md)
 + [Best Practices](./best-practices.md)
 + [Templates](./templates.md)
++ [Execute Transaction](./execute-transaction.md)
++ [Sign Multisig](./sign-multisig.md)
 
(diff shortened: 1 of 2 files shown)


amityadav0:
could you please add tests for `sign-multisig` ? i am not sure if it works, having tests before docs would be good.

ratik21:
> could you please add tests for sign-multisig ? i am not sure if it works, having tests before docs would be good.

This is just a basic documentation task. I have tested this command and it works, but it still feels incomplete to me. As we need to "create" multisignatures as well, which we're not doing here (we're only appending). Tests are also needed. 
I added one task for test: [https://www.pivotaltracker.com/story/show/178806449](https://www.pivotaltracker.com/story/show/178806449).

amityadav0:
okay, so it is working properly.

docs/guide/README.md
```diff
@@ -20,6 +20,8 @@ layout: splash
 + [Test TEAL](./testing-teal.md)
 + [Best Practices](./best-practices.md)
```
robert-zaremba, line 21: Best Practices should be the last one.

robert-zaremba (approved):
we can create another task to create tests.

docs/guide/README.md
ratik21: Alright

PR #410, opened 2021-07-09
add tests for algob sign-multisig

ratik21:
## Proposed Changes

Added tests for `algob sign-multisig` command. Removed a TODO from previous test.


amityadav0:
do we have test for `signMultiSig` function ?
this doesn't seems to test that.

amityadav0:
with new types i cannot read `msig` property here.
could you please try it with new types ?

ratik21:
I think we discussed that `decodeSignedTransaction` should return `msig` property. In tests with latest `algosdk` (although without types) i am getting the `txn.msig` property (as expected). So it seems like a types issue right (same as [https://github.com/algorand/js-algorand-sdk/issues/389](https://github.com/algorand/js-algorand-sdk/issues/389))?

amityadav0:
> I think we discussed that `decodeSignedTransaction` should return `msig` property. In tests with latest `algosdk` (although without types) i am getting the `txn.msig` property (as expected). So it seems like a types issue right (same as [algorand/js-algorand-sdk#389](https://github.com/algorand/js-algorand-sdk/issues/389))? Msig should be there in new types

I know we discussed, but need to confirm does this work here.

packages/algob/test/builtin-tasks/sign-multisig.ts
```diff
@@ -0,0 +1,129 @@
...
+  });
+
+  it("Append bob's signature to multisigned txn loaded from file", async function () {
+    const encodedTx = loadSignedTxnFromFile(inputFile) as Uint8Array;
+    const txMsig = decodeSignedTransaction(encodedTx).msig;
+
+    // input assertions (verfiy bob pk is present in pre-image, but it is not signed,
+    // i.e signature is not present)
+    assert.isDefined(txMsig);
+    assert.deepInclude(txMsig?.subsig, { pk: bobPk } as any); // msig includes bobPk (but without signature i.e "s" field)
```
amityadav0, line 45: do we need this part ? can we remove it ? or  move to another section.

packages/algob/test/builtin-tasks/sign-multisig.ts
```diff
@@ -0,0 +1,129 @@
...
+      file: inputFile,
+      account: 'bob',
+      out: outFile
+    });
+
+    assert.isTrue(stub.calledWith(
+     `File assets/${outFile} already exists. Aborting. Use --force flag if you want to overwrite it`));
+  });
+
+  it("Should append _out to input file if outfile is not passed", async function () {
```
amityadav0, line 118:
```suggestion
  it("Should append output to input file if outfile is not passed", async function () {
```

packages/algob/test/builtin-tasks/sign-multisig.ts
```diff
@@ -0,0 +1,129 @@
...
+    await this.env.run(TASK_SIGN_MULTISIG, {
+      file: inputFile,
+      account: 'bob'
+    });
+
+    const newOutPath = path.join(ASSETS_DIR, 'multisig-signed_out.txn');
+    assert.isTrue(fs.existsSync(newOutPath)); // outfile path (with appended _out) should exist
+
```
amityadav0, line 126: how about adding assertions here that output file has multisigned bob sig ?

packages/algob/test/builtin-tasks/sign-multisig.ts
ratik21: yes, we need it. We're trying to append bob signature to multisig, so we ensure first it's not present when loading the multisig from file. i.e `pk` is present but `s` isn't.

packages/algob/test/builtin-tasks/sign-multisig.ts
ratik21: NO. this test is that if output file name is not passed, then input filename is appended `_out`. Reverting  the commit

packages/algob/test/builtin-tasks/sign-multisig.ts
ratik21: we're already asserting this in first test - [https://github.com/scale-it/algo-builder/pull/410/files#diff-53b33f6a4c7dc9293cd4572a9584c30ac6e217826adf4406e6a7c9eb2a05eca1R38](https://github.com/scale-it/algo-builder/pull/410/files#diff-53b33f6a4c7dc9293cd4572a9584c30ac6e217826adf4406e6a7c9eb2a05eca1R38). 

This test is to verify input file name is append `_out` str.

packages/algob/test/builtin-tasks/sign-multisig.ts
amityadav0: could you please update the info, that -out is appended to file name ?
It seems from line that output is appended

ratik21: ok

amityadav0 (approved):

PR #413, opened 2021-07-12
Bond token

amityadav0:
## Proposed Changes

+ Bond Token implementation


examples/bond/assets/bond-dapp-stateful.py
```diff
@@ -0,0 +1,206 @@
...
+             And(
+                Txn.sender() == App.globalGet(store_manager),
+                basic_checks
+            )
+        ),
+        App.globalPut(issuer_address, Txn.application_args[1]),
+        Return(Int(1))
+    ])
+
+    # only bon app manager can update issue price
```
robert-zaremba, line 58:
```suggestion
    # only bond app manager can update issue price
```

examples/bond/assets/bond-dapp-stateful.py
```diff
@@ -0,0 +1,206 @@
...
+        # Verfies that the application_id is 0, jumps to on_initialize.
+        [Txn.application_id() == Int(0), on_initialize],
+        # Verifies Update transaction, rejects it.
+        [Txn.on_completion() == OnComplete.UpdateApplication, Return(Int(0))],
+        # Verifies delete transaction, rejects it.
+        [Txn.on_completion() == OnComplete.DeleteApplication, Return(Int(0))],
+        # Verifies closeout transaction, approves it.
+        [Txn.on_completion() == OnComplete.CloseOut, Return(Int(1))],
+        # Verifies opt-in transaction, approves it.
+        [Txn.on_completion() == OnComplete.OptIn, Return(Int(1))],
```
robert-zaremba, line 187: only manager should be able to opt-in the bond app to some other asset (otherwise someone could spam and block the bond app).

examples/bond/assets/bond-dapp-stateful.py
```diff
@@ -0,0 +1,206 @@
...
+        [Txn.application_args[0] == Bytes("create_buyback"), create_buyback],
+        # Verifies exit transaction, jumps to on_exit branch.
+        [Txn.application_args[0] == Bytes("exit"), on_exit],
+        # Verifies redeem coupon transaction, jumps to on_redeem_coupon branch.
+        [Txn.application_args[0] == Bytes("redeem_coupon"), on_redeem_coupon],
+        # Verifies update issue price transaction, jumps to on_update_issue_price branch.
+        [Txn.application_args[0] == Bytes("update_issue_price"), on_update_issue_price],
+        # Verifies create dex transaction, jumps to on_create_dex branch.
+        [Txn.application_args[0] == Bytes("create_dex"), on_create_dex],
+    )
```
robert-zaremba, line 204: should we have a default condition - if it doesn't match anything above then reject?

robert-zaremba (commented):
few initial comments.

examples/bond/assets/bond-dapp-stateful.py
```diff
@@ -0,0 +1,206 @@
...
+    issue price, nominal price, maturity date and coupon value, total amount.
+    Additionally, the App will store a reference to the currently
+    tradeable bond ASA and the current epoch number.
+    """
+
+    store_manager = Bytes("store_manager")
+    issue_price = Bytes("issue_price")
+    nominal_price = Bytes("nominal_price")
+    maturity_date = Bytes("maturity_date")
+    coupon_value = Bytes("coupon_value")
```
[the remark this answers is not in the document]
amityadav0: Ok

examples/bond/assets/bond-dapp-stateful.py
amityadav0: What do you mean by `opt-in the bond app to some other asset` ?
in this condition user can opt-in to the app, this is used when user buys bond tokens. user cannot modify any global variables.

examples/bond/assets/bond-dapp-stateful.py
amityadav0: we have one assert for store manager check. If store manager doesn't call to create Dex, it will be rejected.
Am I understanding your question correctly ?

examples/bond/README.md
```diff
@@ -0,0 +1,42 @@
...
+- We bundle coupons with bonds in a single token - Bond token (ASA). 
+    - There is a different ASA for each epoch: in a single `BondApp` we define family of Bond tokens: `B_0`, `B_1`, `B_2` , … `B_n` .  `B_i` token represent a bond which received a coupon payment at `i-1` epoch but execute a coupon payment at `i` epoch.
+    - Holder of `B_i` token can redeem his token and get coupon payment using DEX lsigs (see below)
+- Set of logical signatures (stateless contracts): `DEX_0`, `DEX_1` ….
+    - `B_i` holder can exchange tokens with `DEX_i` for equal number of `B_{i+1}` tokens and `coupon_value` of ALGOs only when `i < BondApp.epoch`. For example: At epoch 2, user X owns 10 `B_1`  tokens. He can send his tokens to `DEX_1` and in exchange of 10 `B_2` tokens and `10 × BondApp.coupon_value` of ALGOs. NOTE: at epoch 2, user can exchange his `B_0` tokens with `DEX_0` if he didn’t do it yet.
+    -  `B_i` tokens sent to `DEX_i` are locked forever. 
+    - Issuer must create and supply ALGO to `DEX_i` in advance. Otherwise legal consequences can be triggered.
```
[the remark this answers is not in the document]
amityadav0: ok

examples/bond/assets/bond-dapp-stateful.py
robert-zaremba: This condition is used to opt-int `BondApp` to `B_1`, `B_2`, ... ASA, right? If so, only manager should be able to do it.

examples/bond/assets/bond-dapp-stateful.py
robert-zaremba: What happens if someone calls app (noop) with arg: `some-other-argument:str`?

examples/bond/assets/bond-dapp-stateful.py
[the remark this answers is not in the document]
robert-zaremba: oh, I though I remove this comment. Sorry for confusions, in this version we make the coupon constant in all epochs. So let's add a note in the README that for non constant coupon values, they should be part of the `DEX_i` lsig.

examples/bond/assets/bond-dapp-stateful.py
amityadav0: no, this condition if for an address to opt-in to an app, for tracking B_1, B_2,.. we have a global variable `current_bond`.

examples/bond/assets/bond-dapp-stateful.py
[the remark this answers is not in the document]
amityadav0: got it

examples/bond/assets/bond-dapp-stateful.py
amityadav0: it fails with error `logic eval error: TEAL runtime encountered err opcode`

examples/bond/assets/bond-dapp-stateful.py
robert-zaremba: What would be a condition for the BondApp to opt-in to an ASA?

examples/bond/assets/bond-dapp-stateful.py
robert-zaremba: ah, so PyTEAL is adding that. Good.

examples/bond/assets/bond-dapp-stateful.py
amityadav0: BondApp itself cannot opt-in or send, receive assets/algo's becuase it doesn't have address, so we have issuer, dex lsig which opts-in to assets or send or receive them, In lsig we have check for calling specific app ID, which checks for the global variables/owner, 
- got this case, will add one more condition in lsigs `a algo payment with zero amount from storeManager` then only allow opt-in

amityadav0: Done. added this check in all lsigs
```
# allow opt-in transaction
    opt_in = And(
        Gtxn[0].type_enum() == TxnType.Payment,
        Gtxn[0].amount() == Int(0),
        Gtxn[0].sender() == Tmpl.Addr("TMPL_STORE_MANAGER"),
        Gtxn[1].type_enum() == TxnType.AssetTransfer,
        Gtxn[1].asset_amount() == Int(0)
    )
```

examples/bond/assets/bond-dapp-stateful.py
```diff
@@ -0,0 +1,209 @@
...
+    issuer_address = Bytes("issuer_address")
+    epoch = Bytes("epoch")
+    total = Bytes("total")
+    current_bond = Bytes("current_bond")
+    max_amount = Bytes("max_amount")
+    creator = Bytes("creator")
```
ratik21, line 21: Does this represent ASC creator? If yes, can we use `Global.CreatorAddress`?

examples/bond/assets/bond-dapp-stateful.py
```diff
@@ -0,0 +1,209 @@
...
+        # Always verify that the RekeyTo property of any transaction is set to the ZeroAddress
+        # unless the contract is specifically involved ina rekeying operation.
+        Txn.rekey_to() == Global.zero_address(),
+        Txn.close_remainder_to() == Global.zero_address(),
+        Txn.asset_close_to() == Global.zero_address()
+    )
+
+    # initialize variables
+    on_initialize = Seq([
+        App.globalPut(store_manager, Txn.application_args[0]),
```
ratik21, line 34: Is this the bond-token manager or bond-dapp manager? Could you please add some documentation regarding the keys in global state not present in readme? Currently it is not clear

examples/bond/assets/bond-dapp-stateful.py
```diff
@@ -0,0 +1,209 @@
...
+                Txn.sender() == App.globalGet(store_manager),
+                basic_checks
+            )
+        ),
+        App.globalPut(issuer_address, Txn.application_args[1]),
+        Return(Int(1))
+    ])
+
+    # only bond app manager can update issue price
+    on_update_issue_price = Seq([
```
ratik21, line 59: Could you please add in doc strings that what are the expected arguments?

examples/bond/assets/bond-dapp-stateful.py
```diff
@@ -0,0 +1,209 @@
...
+                Gtxn[0].type_enum() == TxnType.Payment,
+                # verify buying amount,
+                Gtxn[0].amount() >= Add(Mul(Gtxn[1].asset_amount(), App.globalGet(issue_price)), Gtxn[1].fee()),
+                # verify ASA transfer
+                Gtxn[1].type_enum() == TxnType.AssetTransfer,
+                Gtxn[1].xfer_asset() == App.globalGet(current_bond),
+                Gtxn[1].asset_receiver() == Gtxn[0].sender()
+            )
+        ),
+        App.globalPut(total, Add(App.globalGet(total), Gtxn[1].asset_amount())),
```
ratik21, line 96: I think we should explain the transaction composition somewhere in docs as well. i.e in this case 2nd transaction in group should be asset_transfer transaction when calling `on_buy`. What do you think?

examples/bond/assets/bond-dapp-stateful.py
```diff
@@ -0,0 +1,209 @@
...
+        Assert(
+            And(
+                basic_checks,
+                # verify payment
+                Gtxn[0].type_enum() == TxnType.Payment,
+                # verify buying amount,
+                Gtxn[0].amount() >= Add(Mul(Gtxn[1].asset_amount(), App.globalGet(issue_price)), Gtxn[1].fee()),
+                # verify ASA transfer
+                Gtxn[1].type_enum() == TxnType.AssetTransfer,
+                Gtxn[1].xfer_asset() == App.globalGet(current_bond),
```
ratik21, line 92: Can we rename `current_bond` to `current_bond_id`/`current_bond_index` (i am assuming it store an assetID)?

examples/bond/assets/bond-dapp-stateful.py
```diff
@@ -0,0 +1,209 @@
...
+                # index 1 of Txn.accounts().
+                asset_balance.value() == Gtxn[1].asset_amount(),
+                # burn `B_i` issuer bonds. send to creator
+                Gtxn[2].type_enum() == TxnType.AssetTransfer,
+                Gtxn[2].asset_receiver() == App.globalGet(creator),
+                asset_balance.value() == Gtxn[2].asset_amount()
+            ),
+        ),
+        # Increment `BondApp.epoch`
+        App.globalPut(epoch, Add(App.globalGet(epoch), Int(1))),
```
ratik21, line 171: So just for use-case purposes, we are using `epoch` as a counter here, right? Or is it used this way only?

examples/bond/assets/buyback-lsig.py
```diff
@@ -0,0 +1,42 @@
...
+        Gtxn[0].rekey_to() == Global.zero_address(),
+        Gtxn[1].rekey_to() == Global.zero_address(),
+        Gtxn[2].rekey_to() == Global.zero_address(),
+        Gtxn[0].close_remainder_to() == Global.zero_address(),
+        Gtxn[1].close_remainder_to() == Global.zero_address(),
+        Gtxn[2].close_remainder_to() == Global.zero_address(),
+        Gtxn[1].asset_close_to() == Global.zero_address(),
+    )
+
+    # verify that buyer deposits required algos
```
ratik21, line 15: is this comment right? I don't see verifying deposit amount here.

examples/bond/assets/dex-lsig.py
```diff
@@ -0,0 +1,56 @@
...
+    )
+
+    # allow opt-in transaction
+    opt_in = And(
+        Gtxn[0].type_enum() == TxnType.Payment,
+        Gtxn[0].amount() == Int(0),
+        Gtxn[0].sender() == Tmpl.Addr("TMPL_STORE_MANAGER"),
+        Gtxn[1].type_enum() == TxnType.AssetTransfer,
+        Gtxn[1].asset_amount() == Int(0)
+    )
```
ratik21, line 44: So here 2nd transaction is optInToASA transaction (type = axfer, amount=0)? Could you please explain this in a doc string: why 1st transaction is a payment transaction with amount == 0?

examples/bond/assets/issuer-lsig.py
```diff
@@ -0,0 +1,67 @@
...
+        Gtxn[2].application_id() == Tmpl.Int("TMPL_APPLICATION_ID"),
+        Gtxn[2].application_args[0] == Bytes("buy"),
+        common_fields
+    )
+
+    # burn tx
+    burn_tx = And(
+        Gtxn[2].type_enum() == TxnType.AssetTransfer,
+        Gtxn[1].type_enum() == TxnType.AssetTransfer,
+        Gtxn[0].application_id() == Tmpl.Int("TMPL_APPLICATION_ID")
```
ratik21, line 27: Should we assert first that `Gtxn[0].type_enum() == TxnType.NoOp` ?

examples/bond/scripts/deploy.js
```diff
@@ -0,0 +1,115 @@
...
+  const masterAccount = deployer.accountsByName.get('master-account');
+  const storeManagerAccount = deployer.accountsByName.get('alice');
+  const creatorAccount = deployer.accountsByName.get('john');
+
+  const algoTxnParams = {
+    type: types.TransactionType.TransferAlgo,
+    sign: types.SignType.SecretKey,
+    fromAccount: masterAccount,
+    toAccountAddr: storeManagerAccount.addr,
+    amountMicroAlgos: 200000000,
```
ratik21, line 16:
```suggestion
    amountMicroAlgos: 200e6,
```

examples/bond/scripts/deploy.js
```diff
@@ -0,0 +1,115 @@
...
+
+  // Create Application
+  const appInfo = await deployer.deployApp(
+    'bond-dapp-stateful.py',
+    'bond-dapp-clear.py', {
+      sender: storeManagerAccount,
+      localInts: 1,
+      localBytes: 1,
+      globalInts: 8,
+      globalBytes: 15,
```
ratik21, line 59: Could you please confirm if we need 23 (8 ints, 15 bytes) of storage?

examples/bond/scripts/deploy.js
```diff
@@ -0,0 +1,115 @@
...
+    maturityDate,
+    couponValue,
+    epoch,
+    currentBond,
+    maxAmount,
+    creator
+  ];
+
+  // Create Application
+  const appInfo = await deployer.deployApp(
```
ratik21, line 52:
```suggestion
  const bondAppInfo = await deployer.deployApp(
```

examples/bond/scripts/deploy.js
```diff
@@ -0,0 +1,115 @@
...
+      localBytes: 1,
+      globalInts: 8,
+      globalBytes: 15,
+      appArgs: appArgs
+    }, {});
+  console.log(appInfo);
+
+  // Initialize issuer lsig with bond-app ID
+  const scInitParam = {
+    TMPL_APPLICATION_ID: appInfo.appID,
```
ratik21, line 66:
```suggestion
    TMPL_APPLICATION_ID: bondAppInfo.appID,
```

examples/bond/scripts/run/bond-functions.js
```diff
@@ -0,0 +1,327 @@
...
+    type: types.TransactionType.TransferAlgo,
+    sign: types.SignType.SecretKey,
+    fromAccount: masterAccount,
+    toAccountAddr: accountAddress,
+    amountMicroAlgos: 200000000,
+    payFlags: {}
+  };
+  await executeTransaction(deployer, algoTxnParams);
+}
+/**
```
ratik21, line 27:
```suggestion
}

/**
```

examples/bond/scripts/run/bond-functions.js
```diff
@@ -0,0 +1,327 @@
...
+
+// fund account using master account
+async function fundAccount (deployer, accountAddress) {
+  const masterAccount = deployer.accountsByName.get('master-account');
+  const algoTxnParams = {
+    type: types.TransactionType.TransferAlgo,
+    sign: types.SignType.SecretKey,
+    fromAccount: masterAccount,
+    toAccountAddr: accountAddress,
+    amountMicroAlgos: 200000000,
```
ratik21, line 22:
```suggestion
    amountMicroAlgos: 200e6,
```

examples/bond/scripts/run/bond-functions.js
```diff
@@ -0,0 +1,327 @@
...
+  ];
+  await executeTransaction(deployer, optInTx);
+
+  console.log('Storing buyback address!');
+  await executeTransaction(deployer, buybackTx);
+  console.log('Buyback address stored successfully!');
+}
+
+// Buyer's exit from bond
+async function exitBuyer (deployer, buyerAccount, storeManagerAccount) {
```
ratik21, line 268: are we using `storeManagerAccount` in this function?

ratik21 (commented):
We can add some documentation

examples/bond/assets/bond-dapp-stateful.py
amityadav0: this is ASA_creator, will change it to `asa_creator` or `bond_token_creator`. what do you think ?

examples/bond/assets/bond-dapp-stateful.py
amityadav0: this is store manager as the name suggests, `store manager`. will add from spec document

examples/bond/assets/bond-dapp-stateful.py
amityadav0: i have added this while we call `buy` in scripts, will copy-paste here as well. for more clarity

examples/bond/assets/bond-dapp-stateful.py
amityadav0: i prefer `current_bond`. id seems extra here

examples/bond/assets/bond-dapp-stateful.py
amityadav0: what do you mean ? we have epoch as constant in this example, we have note for this in readme

examples/bond/assets/buyback-lsig.py
amityadav0: thanks, this should be moved to stateful contract

examples/bond/scripts/deploy.js
amityadav0: yes, we need this much

examples/bond/assets/bond-dapp-stateful.py
amityadav0: yes

examples/bond/assets/issuer-lsig.py
amityadav0: yes

examples/bond/scripts/run/bond-functions.js
amityadav0: removed

examples/bond/scripts/deploy.js
```diff
@@ -0,0 +1,115 @@
+const {
+  executeTransaction, convert
+} = require('@algo-builder/algob');
+const { types } = require('@algo-builder/web');
+
+async function run (runtimeEnv, deployer) {
+  const masterAccount = deployer.accountsByName.get('master-account');
+  const storeManagerAccount = deployer.accountsByName.get('alice');
```
robert-zaremba, line 8: let's rename here as well: storeManagerAccount -> managerAcc

examples/bond/scripts/deploy.js
```diff
@@ -0,0 +1,115 @@
...
+  const creatorAccount = deployer.accountsByName.get('john');
```
robert-zaremba, line 9: these accounts are used in multiple places. Let's create a file `scripts/common/accounts` with a function which will create and return that accounts, eg:
```
export accounts(deployer) {
    return {
        masterAcc: deployer.accountsByName.get('master-account'), ...
    }
}

File: examples/bond/scripts/deploy.js [hunk 24]
```diff
@@ -0,0 +1,115 @@
...
+
+  const algoTxnParams = {
+    type: types.TransactionType.TransferAlgo,
+    sign: types.SignType.SecretKey,
+    fromAccount: masterAccount,
+    toAccountAddr: storeManagerAccount.addr,
+    amountMicroAlgos: 200e6,
+    payFlags: {}
+  };
```
robert-zaremba, line 18: why do we pay 200 ALGO to the manager? It's more then he should need.

examples/bond/scripts/deploy.js
```diff
@@ -0,0 +1,115 @@
...
+
+  // Create B_0 - Bond Token
+  const asaInfo = await deployer.deployASA('bond-token', { creator: creatorAccount });
+  console.log(asaInfo);
+
+  // Bond-Dapp initialization parameters
+  const storeManager = convert.addressToPk(storeManagerAccount.addr);
+  const issuePrice = 'int:1000';
+  const nominalPrice = 'int:1000';
+  const maturityDate = convert.uint64ToBigEndian(Math.round(new Date().getTime() / 1000) + 1000);
```
robert-zaremba, line 31: oh, so maturity only in 1000 seconds?

examples/bond/scripts/deploy.js
```diff
@@ -0,0 +1,115 @@
...
+  const couponValue = 'int:100';
```
robert-zaremba, line 32: that's a bit too much, let's change to 20.

examples/bond/scripts/deploy.js
```diff
@@ -0,0 +1,115 @@
...
+  const maxAmount = convert.uint64ToBigEndian(asset.params.total);
+  const creator = convert.addressToPk(creatorAccount.addr);
+
+  let appArgs = [
+    storeManager,
+    issuePrice,
+    nominalPrice,
+    maturityDate,
+    couponValue,
+    epoch,
```
robert-zaremba, line 45: epoch shouldn't be a part of the initializaiton it will always start from 0.

examples/bond/scripts/deploy.js
```diff
@@ -0,0 +1,115 @@
...
+  // Bond-Dapp initialization parameters
+  const storeManager = convert.addressToPk(storeManagerAccount.addr);
+  const issuePrice = 'int:1000';
+  const nominalPrice = 'int:1000';
+  const maturityDate = convert.uint64ToBigEndian(Math.round(new Date().getTime() / 1000) + 1000);
+  const couponValue = 'int:100';
+  const epoch = 'int:0';
+  const currentBond = convert.uint64ToBigEndian(asaInfo.assetIndex);
+  const asset = await deployer.getAssetByID(asaInfo.assetIndex);
+  const maxAmount = convert.uint64ToBigEndian(asset.params.total);
```
robert-zaremba, line 36:
```suggestion
  const maxIssuance = convert.uint64ToBigEndian(asset.params.total);
```

examples/bond/scripts/deploy.js
```diff
@@ -0,0 +1,115 @@
...
+    TMPL_APPLICATION_ID: bondAppInfo.appID,
+    TMPL_OWNER: creatorAccount.addr,
+    TMPL_STORE_MANAGER: storeManagerAccount.addr
+  };
+  const issuerLsig = await deployer.loadLogic('issuer-lsig.py', scInitParam);
+
+  algoTxnParams.toAccountAddr = issuerLsig.address();
+  await executeTransaction(deployer, algoTxnParams);
+
+  // Only store manager can allow opt-in to ASA for lsig
```
robert-zaremba, line 75:
```suggestion
  // Only app manager can opt-in issueer lsig to ASA
```

examples/bond/scripts/run/bond-functions.js
```diff
@@ -0,0 +1,328 @@
...
+  console.log('Exiting');
+  await executeTransaction(deployer, exitTx);
+  console.log('Exited');
+}
+
+async function run (runtimeEnv, deployer) {
+  const masterAccount = deployer.accountsByName.get('master-account');
+  const creatorAccount = deployer.accountsByName.get('john');
+  const storeManagerAccount = deployer.accountsByName.get('alice');
+  const buyerAccount = deployer.accountsByName.get('bob');
```
robert-zaremba, line 312: let's use addresses from `scripts/common/accounts`

examples/bond/scripts/deploy.js
```diff
@@ -0,0 +1,115 @@
...
+      payFlags: {}
+    },
+    {
+      type: types.TransactionType.OptInASA,
+      sign: types.SignType.LogicSignature,
+      fromAccountAddr: issuerLsig.address(),
+      lsig: issuerLsig,
+      assetID: asaInfo.assetIndex,
+      payFlags: {}
+    }
```
robert-zaremba, line 92: we have a highlevel function to do opt in. we don't need to create the tx manually: https://algobuilder.dev/api/algob/interfaces/types.deployer.html#optinlsigtoasa

examples/bond/scripts/run/bond-functions.js
```diff
@@ -0,0 +1,328 @@
...
+  await executeTransaction(deployer, algoTxnParams);
+}
+
+/**
+ * Creates DEX_i lsig, burn B_i tokens, issue B_i+1 tokens
+ * @param {Account} masterAccount
+ * @param {Account} creatorAccount
+ * @param {Account} storeManagerAccount
+ */
+async function createDex (deployer, masterAccount, creatorAccount, storeManagerAccount) {
```
robert-zaremba, line 34: the function should take a paramter `i` - an index of the bond token to create. In the function we should assert that `i >= 1`

examples/bond/scripts/run/bond-functions.js
```diff
@@ -0,0 +1,328 @@
...
+  };
+  const issuerLsig = await deployer.loadLogic('issuer-lsig.py', scInitParam);
+
+  const deployTx = {
+    type: types.TransactionType.DeployASA,
+    sign: types.SignType.SecretKey,
+    fromAccount: creatorAccount,
+    asaName: 'new-bond-token',
+    payFlags: {}
+  };
```
robert-zaremba, line 50: as noted in `assets.yaml` we should create bond "manually"

examples/bond/scripts/run/bond-functions.js
```diff
@@ -0,0 +1,328 @@
...
+      payFlags: {}
+    },
+    {
+      type: types.TransactionType.OptInASA,
+      sign: types.SignType.LogicSignature,
+      fromAccountAddr: issuerLsig.address(),
+      lsig: issuerLsig,
+      assetID: newIndex,
+      payFlags: {}
+    }
```
robert-zaremba, line 73: this is same transaction as in deploy. Let's export it to a common function and reuse.

examples/bond/scripts/run/bond-functions.js
```diff
@@ -0,0 +1,328 @@
...
+  ];
+  await executeTransaction(deployer, optInTx);
+
+  scInitParam = {
```
robert-zaremba, line 77:
```suggestion
  const lsigParams = {
```

examples/bond/scripts/run/bond-functions.js
```diff
@@ -0,0 +1,328 @@
...
+
+/**
+ * Creates DEX_i lsig, burn B_i tokens, issue B_i+1 tokens
+ * @param {Account} masterAccount
+ * @param {Account} creatorAccount
+ * @param {Account} storeManagerAccount
+ */
+async function createDex (deployer, masterAccount, creatorAccount, storeManagerAccount) {
+  asaInfo = deployer.getASAInfo('bond-token');
+  appInfo = deployer.getApp('bond-dapp-stateful.py', 'bond-dapp-clear.py');
```
robert-zaremba, line 36:
```suggestion
  const asaInfo = deployer.getASAInfo('bond-token');
  const appInfo = deployer.getApp('bond-dapp-stateful.py', 'bond-dapp-clear.py');
```

examples/bond/scripts/run/bond-functions.js
```diff
@@ -0,0 +1,328 @@
...
+  await executeTransaction(deployer, optInTx);
+
+  const globalState = await readGlobalStateSSC(deployer, storeManagerAccount.addr, appInfo.appID);
+  let total = 0;
+  for (const l of globalState) {
+    const key = Buffer.from(l.key, 'base64').toString();
+    if (key === 'total') {
+      total = l.value.uint;
+      break;
+    }
```
robert-zaremba, line 102: can we do `globalState[base64("total")]` instead of doing the for loop? (I used a pseudo function base64)

examples/bond/scripts/run/bond-functions.js
```diff
@@ -0,0 +1,328 @@
...
+  const transferTx = {
+    type: types.TransactionType.TransferAsset,
+    sign: types.SignType.SecretKey,
+    fromAccount: creatorAccount,
+    toAccountAddr: dexLsig.address(),
+    amount: total,
+    assetID: newIndex,
+    payFlags: { totalFee: 1000 }
+  };
+  await executeTransaction(deployer, transferTx);
```
robert-zaremba, line 115: this should be part of a group transaction -> see the spec. (and in fact you have it already in `groupTx`.

examples/bond/scripts/run/bond-functions.js
```diff
@@ -0,0 +1,328 @@
...
+  for (const l of globalState) {
+    const key = Buffer.from(l.key, 'base64').toString();
+    if (key === 'total') {
+      total = l.value.uint;
+      break;
+    }
+  }
+
+  // Transfer total amount to dex lsig
+  const transferTx = {
```
robert-zaremba, line 106:
```suggestion
    // Transfer app.total amount of new Bonds to dex lsig
  const issueNewBond = {
```

examples/bond/scripts/run/bond-functions.js
```diff
@@ -0,0 +1,328 @@
...
+      fromAccountAddr: issuerLsig.address(),
+      lsig: issuerLsig,
+      toAccountAddr: creatorAccount.addr,
+      amount: info.amount,
+      assetID: asaInfo.assetIndex,
+      payFlags: { totalFee: 1000 }
+    }
+  ];
+
+  console.log('Creating dex!');
```
robert-zaremba, line 154:
```suggestion
  console.log('Creating dex %d!', i);
```

robert-zaremba: `i` is the new parameter to this function

examples/bond/scripts/run/bond-functions.js
```diff
@@ -0,0 +1,328 @@
...
+    {
+      type: types.TransactionType.OptInASA,
+      sign: types.SignType.LogicSignature,
+      fromAccountAddr: buybackLsig.address(),
+      lsig: buybackLsig,
+      assetID: newAsaInfo[assetID],
+      payFlags: {}
+    }
+  ];
+  await executeTransaction(deployer, optInTx);
```
robert-zaremba, line 261: ditto - it's copy paste almost - let's make a function out of it.

examples/bond/scripts/run/bond-functions.js
```diff
@@ -0,0 +1,328 @@
...
+
+  console.log('Storing buyback address!');
```
robert-zaremba, line 263:
```suggestion
  console.log('Setting buyback address!');
```

examples/bond/scripts/run/bond-functions.js
```diff
@@ -0,0 +1,328 @@
...
+  await executeTransaction(deployer, buybackTx);
+  console.log('Buyback address stored successfully!');
```
robert-zaremba, line 265:
```suggestion
  console.log('Buyback address set successfully!');
```

examples/bond/scripts/run/bond-functions.js
```diff
@@ -0,0 +1,328 @@
...
+
+  // Redeem coupon_value
+  await redeem(deployer, buyerAccount, storeManagerAccount);
+
+  // create buyback
+  await createBuyback(deployer, storeManagerAccount);
+
+  // exit buyer from bond, buyer can exit only if maturity period is over
+  // currently set to 1000 seconds
+  await exitBuyer(deployer, buyerAccount);
```
robert-zaremba, line 325: let's move `exit` and `createBuyback` to another `run` script (both can be in the same script).

examples/bond/scripts/run/buy.js
```diff
@@ -0,0 +1,71 @@
...
+      payFlags: {},
+      appArgs: ['str:buy']
+    }
+  ];
+
+  console.log('Buying tokens!');
+  await executeTransaction(deployer, groupTx);
+  console.log('Tokens Bought!');
+}
+
```
robert-zaremba, line 70: let's update the _run_ scripts to have the following use case (after deploying the contract):
+ in epoch_0 alice buys 10 bonds
+ in epoch 0 alice sells 2 bonds to charlie (new account) for 2020 ALGO (in a group transaction)
+ manager starts epoch 1 (create dex)
+ alice redeems her bonds
+ alice buys 4 more bonds (so she will have 12 bonds in total)
+ manager starts epoch 2 (create dex)
+ alice redeems her bonds
+ charlie redeems her bonds_1 and then bonds_2 (so going through the previous dex)
+ at maturity (let's change 1000 seconds to 240 seconds == 4 min) manager creates and funds buyback and both alice and charlie exit  all their tokens (12 and 2 respectively).

Please describe this scenario in the README. We should have the following files:
+ deploy.js
+ run/epoch0.js
+ run/epoch1.js
+ run/exit.js
and maybe some common files to reuse in the run scripts.

In test we should also add this scenario and make all checks (total atoms received etc...)

examples/bond/assets/asa.yaml
```diff
@@ -0,0 +1,25 @@
+bond-token:
```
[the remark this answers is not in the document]
amityadav0: ok

examples/bond/assets/asa.yaml
```diff
@@ -0,0 +1,25 @@
...
+    unitName: "BOND"
+    url: "url"
+    # User may get "signature validation failed" from node if shorter hash is used.
+    metadataHash: "12312442142141241244444411111133"
+    noteb64: "noteb64"
+    manager: "WWYNX3TKQYVEREVSW6QQP3SXSFOCE3SKUSEIVJ7YAGUPEACNI5UGI4DZCE"
+    reserve: "WWYNX3TKQYVEREVSW6QQP3SXSFOCE3SKUSEIVJ7YAGUPEACNI5UGI4DZCE"
+    freeze: "WWYNX3TKQYVEREVSW6QQP3SXSFOCE3SKUSEIVJ7YAGUPEACNI5UGI4DZCE"
+    optInAccNames: ["alice"]
+new-bond-token:
```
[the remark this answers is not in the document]
amityadav0: i was thinking to do that but when we deploy asa we pass name of asa present in asa.yaml, and we cannot pass asa.yaml config directly therefore even if we change the object using script we still need to write it in asa.yaml.

examples/bond/assets/bond-dapp-stateful.py
```diff
@@ -0,0 +1,255 @@
...
+def approval_program():
+    """
+    A stateful smart contract which will store the bond parameters:
+    issue price, nominal price, maturity date and coupon value, total amount.
+    Additionally, the App will store a reference to the currently
+    tradeable bond ASA and the current epoch number.
+    """
+
+    # Store manager manages bond-dapp contract
+    store_manager = Bytes("store_manager")
```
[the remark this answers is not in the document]
amityadav0: ok

examples/bond/assets/bond-dapp-stateful.py
```diff
@@ -0,0 +1,255 @@
...
+    # Price at which bonds are sold by the issuer.
+    issue_price = Bytes("issue_price")
+    # Bond nominal price.
+    nominal_price = Bytes("nominal_price")
+    # Date when a bond holder can redeem bonds to the issuer.
+    maturity_date = Bytes("maturity_date")
```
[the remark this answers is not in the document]
amityadav0: these are fixed here at the time of initialization, this should not be part of global variables of bond-dapp ?
what do you mean by part of contract template ? lsig ?

examples/bond/assets/bond-dapp-stateful.py
```diff
@@ -0,0 +1,255 @@
...
+                basic_checks,
+                # verify payment
+                Gtxn[0].type_enum() == TxnType.Payment,
+                # verify buying amount,
+                Gtxn[0].amount() >= Add(Mul(Gtxn[1].asset_amount(), App.globalGet(issue_price)), Gtxn[1].fee()),
+                # verify ASA transfer
+                Gtxn[1].type_enum() == TxnType.AssetTransfer,
+                # verify current bond is being transferred
+                Gtxn[1].xfer_asset() == App.globalGet(current_bond),
+                Gtxn[1].asset_receiver() == Gtxn[0].sender()
```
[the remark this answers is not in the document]
amityadav0: there are three transactions in buy operation:
1. algo transfer from buyer to issuer (fees paid by buyer because buyer is transferring algo)
    - this amount includes fee of tx2 + amount required to buy tokens
2. tokens transfer from issuer to buyer(`fee deducted from amount issuer got which already has fees`)
3. call to bond-dapp (by buyer)
`Buyer pays all the fees in above tranactions.`

examples/bond/assets/bond-dapp-stateful.py
```diff
@@ -0,0 +1,255 @@
...
+            )
+        ),
+        App.globalPut(total, Add(App.globalGet(total), Gtxn[1].asset_amount())),
+        Return(Int(1))
+    ])
+
+    # Create buyback address transaction
+    # Expected arguments: [Bytes("create_buyback"), buyback address]
+    create_buyback = Seq([
```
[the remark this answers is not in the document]
amityadav0: ok

examples/bond/assets/bond-dapp-stateful.py
```diff
@@ -0,0 +1,255 @@
...
+            And(
+                basic_checks,
+                # User sends `B_i` to `DEX_i` lsig.
+                Gtxn[0].type_enum() == TxnType.AssetTransfer,
+                # `DEX_i` sends `B_{i+1}` to the user.
+                Gtxn[1].type_enum() == TxnType.AssetTransfer,
+                Gtxn[0].asset_amount() == Gtxn[1].asset_amount(),
+                # `Dex_i` sends coupon value to user
+                Gtxn[2].type_enum() == TxnType.Payment,
+                Gtxn[2].receiver() == Txn.sender(),
```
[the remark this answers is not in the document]
amityadav0: Txn.sender() is same Gtxn[0].sender() in this case because redeemer called this stateful contract. therfore according to current transaction address redeemer address == Gtxn[0].sender()

examples/bond/assets/bond-dapp-stateful.py
```diff
@@ -0,0 +1,255 @@
...
+                # verify coupon amount
+                Gtxn[2].amount() == Mul(Gtxn[0].asset_amount(), App.globalGet(coupon_value))
+            )
+        ),
+        Return(Int(1))
+    ])
+
+    # fetch asset_holding.balance from Txn.accounts[0]
```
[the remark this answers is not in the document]
amityadav0: specifying Int(1) in AssetHolding.balance() is accounts[0] that's why i added it. I will add that it expects it to be issuer

examples/bond/assets/bond-dapp-stateful.py
```diff
@@ -0,0 +1,255 @@
...
+
+    # Create dex transaction
+    # Expected arguments: [Bytes("create_dex")]
+    on_create_dex = Seq([
+        asset_balance, # load asset_balance from store
+        Assert(
+            And(
+                # verify sender is store manager
+                Txn.sender() == App.globalGet(store_manager),
+                basic_checks,
```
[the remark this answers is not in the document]
amityadav0: ok

examples/bond/assets/bond-dapp-stateful.py
```diff
@@ -0,0 +1,255 @@
...
+                Gtxn[1].type_enum() == TxnType.AssetTransfer,
```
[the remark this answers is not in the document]
amityadav0: ok

examples/bond/assets/buyback-lsig.py
```diff
@@ -0,0 +1,50 @@
...
+        Gtxn[0].xfer_asset() == Tmpl.Int("TMPL_BOND"),
+        Gtxn[1].type_enum() == TxnType.Payment,
+        # verify call to bond-dapp
+        Gtxn[2].type_enum() == TxnType.ApplicationCall,
+        Gtxn[2].application_id() == Tmpl.Int("TMPL_APPLICATION_ID"),
+        # verify first argument is `exit`
+        Gtxn[2].application_args[0] == Bytes("exit"),
+        common_fields
+    )
+
```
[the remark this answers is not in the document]
amityadav0: we already have this in `on_create_dex` branch.

examples/bond/assets/buyback-lsig.py
```diff
@@ -0,0 +1,50 @@
...
+    # Opt-in transaction
+    # Note: we are checking that first transaction is payment with amount 0
+    # and sent by store manager, because we don't want another
+    # user to opt-in too many asa/app and block this address
+    opt_in = And(
+        Gtxn[0].type_enum() == TxnType.Payment,
+        Gtxn[0].amount() == Int(0),
+        Gtxn[0].sender() == Tmpl.Addr("TMPL_STORE_MANAGER"),
```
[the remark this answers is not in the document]
amityadav0: ok

examples/bond/assets/dex-lsig.py
```diff
@@ -0,0 +1,63 @@
...
+        Gtxn[0].asset_amount() == Gtxn[1].asset_amount()
+    )
+
+    # verify dex sends new bond tokens to buyer 
+    second_tx = And(
+        Gtxn[1].type_enum() == TxnType.AssetTransfer,
+        Gtxn[1].xfer_asset() == Tmpl.Int("TMPL_NEW_BOND"),
+    )
+
+    # verify dex pays coupon value to buyer
```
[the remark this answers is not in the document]
amityadav0: ok

examples/bond/assets/issuer-lsig.py
amityadav0: added and ordered

examples/bond/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "algorand-bond",
+  "version": "0.0.1",
+  "main": "index.js",
+  "license": "Apache-2.0",
+  "dependencies": {
+    "algosdk": "^1.9.1"
```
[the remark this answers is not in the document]
amityadav0: ok

examples/bond/scripts/deploy.js
amityadav0: ok

examples/bond/scripts/deploy.js
amityadav0: ok

examples/bond/scripts/deploy.js
amityadav0: ok

examples/bond/scripts/deploy.js
amityadav0: we have this function, but this only works for a single opt-in transaction, here we need to add a transaction from app manager so that malicious user cannot spam issuer opt-in limit.

examples/bond/assets/bond-dapp-stateful.py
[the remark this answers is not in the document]
ratik21: > this amount includes fee of tx2 + amount required to buy tokens

ok, so this is checked at line 124 right? Could you please update the comment, that we're verifying fee of `tx1` is being checked as well, that it's paid by `tx0`.

BDW, @amityadav0  @robert-zaremba  we also have pooled transaction fees support now, we can utilize that as well. Buyer don't need to explicitly send amount of fees in algos to the issuer. So in this case for example:
```py
fees(tx0 OR tx2) = 2000 (to cover fee of tx1 as well)
fee(tx1) = 0 // issuer doesn't need to pay fee (during token transfer)
```
And in contract we can just assert `Gtxn[1].fee == 0` (verify issuer doesn't need to pay the fee). Just a thought, current implementation works as well.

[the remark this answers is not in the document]
amityadav0: good idea. yes, i have already updated them and added checks that issuer doesn't pay fee, Gtxn[1].fee == Int(0).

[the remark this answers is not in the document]
amityadav0: pooled fees is better also, because attacker can easily bypass our scripts, send only the amount required this will result in deduction in fee from issuer.

examples/bond/assets/bond-dapp-stateful.py
```diff
@@ -0,0 +1,255 @@
...
+                # verify second transaction is payment
+                Gtxn[1].type_enum() == TxnType.Payment,
+                # verify sender of first transaction is buyback address
+                Gtxn[1].sender() == App.globalGet(Bytes("buyback")),
+                # verify amount of algo received by buyer
+                Gtxn[1].amount() == Minus(Mul(Gtxn[0].asset_amount(), App.globalGet(nominal_price)), Gtxn[1].fee()),
+                Gtxn[1].receiver() == Txn.sender(),
+                # verify maturity date is passed
+                Global.latest_timestamp() > App.globalGet(maturity_date)
+            )
```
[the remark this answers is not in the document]
amityadav0: Updated to pooled transaction, so added check Gtxn[1].fee == Int(0)

examples/bond/assets/bond-dapp-stateful.py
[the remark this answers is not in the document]
amityadav0: here Txn.sender() and Gtxn[0].sender() is same, Gtxn[0].sender() is more self explanatory, updated to this.

examples/bond/assets/bond-dapp-stateful.py
```diff
@@ -0,0 +1,255 @@
...
+                # User sends `B_i` to `DEX_i` lsig.
+                Gtxn[0].type_enum() == TxnType.AssetTransfer,
+                # `DEX_i` sends `B_{i+1}` to the user.
+                Gtxn[1].type_enum() == TxnType.AssetTransfer,
+                Gtxn[0].asset_amount() == Gtxn[1].asset_amount(),
+                # `Dex_i` sends coupon value to user
+                Gtxn[2].type_enum() == TxnType.Payment,
+                Gtxn[2].receiver() == Txn.sender(),
+                # verify coupon amount
+                Gtxn[2].amount() == Mul(Gtxn[0].asset_amount(), App.globalGet(coupon_value))
```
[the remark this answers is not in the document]
amityadav0: redeemer will pay the fees, updated to pooled fees transaction

examples/bond/assets/buyback-lsig.py
```diff
@@ -0,0 +1,50 @@
...
+        Gtxn[2].close_remainder_to() == Global.zero_address(),
+        Gtxn[1].asset_close_to() == Global.zero_address(),
+    )
+
+    # verify that buyer deposits required algos(Not we are checking this in stateful contract)
+    # Here we are verifying if the call is made to bond-dapp stateful contract
+    verify_tx = And(
+        # verify TMPL_BOND is being transferred
+        Gtxn[0].type_enum() == TxnType.AssetTransfer,
+        Gtxn[0].xfer_asset() == Tmpl.Int("TMPL_BOND"),
```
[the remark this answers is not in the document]
amityadav0: I think we can because buyback lsig is only specific to a particular TMPL_BOND, if we change it with epoch address of buyback lsig will be changed , therefore it will be considered different contract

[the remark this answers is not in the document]
amityadav0: for each epoch there will be different buyback address if creator wants to do that, else i think it is meant for last bond token.

examples/bond/assets/asa.yaml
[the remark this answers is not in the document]
robert-zaremba: we need to be able to deploy ASA programmatically, without declaring it in asa.yaml.

examples/bond/assets/bond-dapp-stateful.py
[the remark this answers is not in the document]
robert-zaremba: this should be template variable (so hardcoded during initialization).

examples/bond/assets/asa.yaml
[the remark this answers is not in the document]
amityadav0: This needs to be done in algob package, i will add a task for it.

examples/bond/assets/bond-dapp-stateful.py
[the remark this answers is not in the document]
amityadav0: we need to compare maturity date and nominal price while exiting operation, so we need it when we do that, template variables are used in stateless

examples/bond/assets/bond-dapp-stateful.py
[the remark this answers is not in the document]
robert-zaremba: Hmm, I'm not sure if pooled fee will work there. If user will not attach more fee, then the fee will be charged from the lsig account for the transaction where we send tokens from issuer to the user.
We need to investigate it and add appropriate comments in the code and scripts.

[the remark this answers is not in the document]
amityadav0: I have added checks for that `Gtxn[1].fee == Int(0)` for issuer, dexLsig transactions, so it is not possible for user to not attach less fees, If user does that the transaction will always fail. we ensure this in stateful contract that issuer, dex don't pay fees when a user interacts with contract. Adding appropriate comments.

examples/bond/assets/bond-dapp-stateful.py
```diff
@@ -0,0 +1,255 @@
...
+                Txn.sender() == App.globalGet(store_manager),
+                basic_checks,
+                Gtxn[1].type_enum() == TxnType.AssetTransfer,
+                # transfer `balanceOf(issuer, B_i)`  of `B_{i+1}` from the bond_token_creator to the `issuer`.
+                # index 1 of Txn.accounts().
+                asset_balance.value() == Gtxn[1].asset_amount(),
+                # burn `B_i` issuer bonds. send to bond_token_creator
+                Gtxn[2].type_enum() == TxnType.AssetTransfer,
+                Gtxn[2].asset_receiver() == App.globalGet(bond_token_creator),
+                asset_balance.value() == Gtxn[2].asset_amount()
```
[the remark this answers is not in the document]
amityadav0: added.

examples/bond/assets/bond-dapp-stateful.py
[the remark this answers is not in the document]
amityadav0: ok

examples/bond/assets/bond-dapp-stateful.py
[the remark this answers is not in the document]
robert-zaremba: see other comment above - not sure if pooled transaction will work with lsig as we want.

examples/bond/assets/bond-dapp-stateful.py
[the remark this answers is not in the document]
robert-zaremba: let's add a comment.

examples/bond/assets/buyback-lsig.py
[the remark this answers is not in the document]
robert-zaremba: right, I initially I thought this is `lsig` code (didn't read the buyback prefix) and then I realized it's buyback but forgot to remove the comment.

examples/bond/assets/buyback-lsig.py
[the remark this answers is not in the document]
robert-zaremba: yes, but it should be part of the atomic group.

examples/bond/scripts/deploy.js
robert-zaremba: it would be good to have a high level function which will return the a TX config for the opt-in to not do it manually each time. It will be good for other users as well.

examples/bond/assets/bond-dapp-stateful.py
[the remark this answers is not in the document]
amityadav0: yes, i investigated further and found that if we add checks that issuer can never pay fees when user interacts with it, It should work, I have also tested this in local node, i will also add tests for this in token-tests.

examples/bond/assets/bond-dapp-stateful.py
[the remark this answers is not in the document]
amityadav0: ok

examples/bond/assets/buyback-lsig.py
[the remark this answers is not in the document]
amityadav0: yes, it is a part of atomic group, this is where we check it https://github.com/scale-it/algo-builder/blob/11ed05a59ba8083094a09f764e9278b107e9583c/examples/bond/assets/bond-dapp-stateful.py#L216

examples/bond/scripts/deploy.js
amityadav0: ok, i will create one function in common, where user can simply pass addresses, assetId and runtime,  That will opt-in using Tx config.

examples/bond/scripts/run/bond-functions.js
amityadav0: ok

examples/bond/scripts/run/bond-functions.js
amityadav0: added a task for it, this change needs to be done in algob

examples/bond/scripts/run/bond-functions.js
amityadav0: done

examples/bond/scripts/run/bond-functions.js
amityadav0: ok

examples/bond/assets/asa.yaml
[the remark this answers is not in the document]
robert-zaremba: OK, so let's firstly add this to the algob, and then rebase this template (before merging this PR).

examples/bond/assets/bond-dapp-stateful.py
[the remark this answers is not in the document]
robert-zaremba: OK, `Gtxn[1].fee == Int(0)` is a good solution. Please review other places where where a user is interacting with the contract - to make sure he will pay the fees.

examples/bond/scripts/deploy.js
robert-zaremba: ok, thanks.

examples/bond/assets/asa.yaml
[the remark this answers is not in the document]
amityadav0: Got it

examples/bond/scripts/deploy.js
amityadav0: ok

examples/bond/assets/bond-dapp-stateful.py
robert-zaremba: this is not done - `opt_in` condition should be added here.

examples/bond/assets/asa.yaml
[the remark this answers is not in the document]
amityadav0: https://github.com/scale-it/algo-builder/pull/434

examples/bond/assets/bond-dapp-stateful.py
robert-zaremba: Ah, my mistake again. App can't `opt-in` to another thing.

examples/bond/scripts/run/bond-functions.js
amityadav0: done

examples/bond/scripts/run/bond-functions.js
amityadav0: ok

examples/bond/scripts/deploy.js
amityadav0: ok, reducing it

examples/bond/scripts/run/bond-functions.js
amityadav0: done

examples/bond/scripts/run/buy.js
amityadav0: done, describing in readme now

examples/bond/scripts/run/bond-functions.js
amityadav0: done

examples/bond/scripts/run/bond-functions.js
amityadav0: no, it gives undefined

examples/bond/assets/bond-dapp-stateful.py
```diff
@@ -0,0 +1,255 @@
...
+        Return(Int(1))
+    ])
+
+    # fetch asset_holding.balance from Txn.accounts[0]
+    asset_balance = AssetHolding.balance(Int(1), Gtxn[2].xfer_asset())
+
+    # Create dex transaction
+    # Expected arguments: [Bytes("create_dex")]
+    on_create_dex = Seq([
+        asset_balance, # load asset_balance from store
```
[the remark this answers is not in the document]
amityadav0: we are saving it in scratch space above, and then loading here from scratch space

examples/bond/README.md
```diff
@@ -0,0 +1,62 @@
+# Algorand bond token
+
+Bonds are units of corporate debt issued by companies and securitized as tradeable assets. A bond is referred to as a fixed-income instrument since bonds traditionally paid a fixed interest rate (coupon) to debt holders. Variable or floating interest rates are also now quite common. Bond prices are inversely correlated with interest rates: when rates go up, bond prices fall and vice-versa. 
+
+In this template, we use ALGO to express the nominal and trading values (investors will buy, sell and exchange bonds using ALGO). It's easy to change this and use any other ASA.
+
+When creating a bond the following information must be recorded:
+
+- `issue_price` — price at which bonds are sold by the issuer.
```
robert-zaremba, line 9:
```suggestion
- `issue_price` — price of a bonds sold by the issuer.
```

examples/bond/README.md
```diff
@@ -0,0 +1,62 @@
...
+We have deploy script in `scripts/deploy`, This script deploys initial bond token, deploys bond dapp application and updates issuer address in bond-dapp contract.
+
+## Run script
+
+In `scripts/run` folder we have scripts with the following files:
+
+ - `run/run.js`: It controls all the scripts and maintain order of execution.
+ - `run/issue.js`: This is first script, in this script tokens are issued to issuer from token creator.
+ - `run/epoch0.js`: In this script, `elon` buys 10 bond tokens and sell 2 bond tokens to `bob` for 2020 Algos.
+ - `run/createDex.js`: This script creates ith Dex, burn B_i bond tokens, issue B_i+1 bond tokens.
```
robert-zaremba, line 45:
```suggestion
 - `run/createDex.js`: This script creates i-th Dex, burns `B_i` bond tokens, issues `B_i+1` bond tokens.
```

examples/bond/README.md
```diff
@@ -0,0 +1,62 @@
...
+ - `run/redeem.js`: This script redeems tokens from a given dex id.
```
robert-zaremba, line 46:
```suggestion
 - `run/redeem.js`: This script exchanges tokens and redeems coupon.
```

examples/bond/README.md
```diff
@@ -0,0 +1,62 @@
...
+ - `run/epoch1.js`: `elon` redeems 8 bond tokens and buys 4 more from dex.
+ - `run/createBuyback.js`: Creates a buyback logic signature address.
+ - `run/exit.js`: Buyer can exit their bond tokens using this script.
+ - `run/common/accounts.js`: Loads accounts
+ - `run/common/common.js`: contains common functions and constants.
+ 
+ - In run script we have the following scenario:
+    - Issue initial bond tokens to issuer
```
robert-zaremba, line 54:
```suggestion
    - Issue initial bond tokens to the issuer
```

examples/bond/README.md
```diff
@@ -0,0 +1,62 @@
...
+ - `run/epoch0.js`: In this script, `elon` buys 10 bond tokens and sell 2 bond tokens to `bob` for 2020 Algos.
+ - `run/createDex.js`: This script creates ith Dex, burn B_i bond tokens, issue B_i+1 bond tokens.
+ - `run/redeem.js`: This script redeems tokens from a given dex id.
+ - `run/epoch1.js`: `elon` redeems 8 bond tokens and buys 4 more from dex.
+ - `run/createBuyback.js`: Creates a buyback logic signature address.
+ - `run/exit.js`: Buyer can exit their bond tokens using this script.
+ - `run/common/accounts.js`: Loads accounts
+ - `run/common/common.js`: contains common functions and constants.
+ 
+ - In run script we have the following scenario:
```
robert-zaremba, line 53:
```suggestion
 - In a run script example we present the following scenario:
```

examples/bond/README.md
```diff
@@ -0,0 +1,62 @@
...
+    - Issue initial bond tokens to issuer
+    - In epoch_0 elon buys 10 bonds
+    - In epoch 0 elon sells 2 bonds to bob for 2020 ALGO (in a group transaction)
+    - Manager creates dex 1
+    - Elon redeems her bonds(8), Elon buys 4 more bonds (so he will have 12 bonds in total)
```
robert-zaremba, line 58:
```suggestion
    - Elon redeems his bonds (8), Elon buys 4 more bonds (so he will have 12 bonds in total)
```

examples/bond/README.md
```diff
@@ -0,0 +1,62 @@
...
+    - Manager creates dex 2
+    - Elon redeems all this bonds.
```
robert-zaremba, line 60:
```suggestion
    - Elon redeems all his bonds.
```

examples/bond/README.md
```diff
@@ -0,0 +1,62 @@
...
+    - Bob redeems his bonds_1 and then bonds_2 (so going through the previous dex)
```
robert-zaremba, line 61:
```suggestion
    - Bob redeems his bonds from epoch 0 and 1
```

examples/bond/README.md
```diff
@@ -0,0 +1,62 @@
...
+    - Maturity period is set to 240 seconds(4 min) from the time of deployment of contract. At maturity manager creates and funds buyback and both elon and bob exit all their tokens (12 and 2 respectively).
```
robert-zaremba, line 62:
```suggestion
    - Maturity period is set to 240 seconds(4 min) after the contract deployment. At maturity, manager creates and funds buyback and both elon and bob can exit all their tokens (12 and 2 respectively).
```

examples/bond/assets/asa.yaml
```diff
@@ -0,0 +1,14 @@
...
+    defaultFrozen: false
+    unitName: "BOND"
+    url: "url"
+    # User may get "signature validation failed" from node if shorter hash is used.
+    metadataHash: "12312442142141241244444411111133"
+    noteb64: "noteb64"
+    manager: "WWYNX3TKQYVEREVSW6QQP3SXSFOCE3SKUSEIVJ7YAGUPEACNI5UGI4DZCE"
+    reserve: "WWYNX3TKQYVEREVSW6QQP3SXSFOCE3SKUSEIVJ7YAGUPEACNI5UGI4DZCE"
+    freeze: "WWYNX3TKQYVEREVSW6QQP3SXSFOCE3SKUSEIVJ7YAGUPEACNI5UGI4DZCE"
+    optInAccNames: ["alice"]
```
robert-zaremba, line 13:
```suggestion
    optInAccNames: ["alice"]  # alice is a manager
```

robert-zaremba (commented):
few style comments. Will finish the rest later.

robert-zaremba (approved):
pre approving. Left few comments. Moreover:
- [x] to each scripts/run/ ... file - please add a comment in the header explaining what it is doing

We will need to make comprehensive tests.

PR #426, opened 2021-07-22
Add accounts name support in runtime

### LINKED ISSUE #423: Runtime, optInAccNames problem w/ Asa definition
## Describe the bug
In runtime, I've to opt in a couple of accounts in ASA. I do that directly in `Assets/asa.yaml` with `optInAccNames: ["alice", "bob"]`. when I run the test, an error returns to me as below.

Schermata 2021-07-20 alle 15 31 37
Schermata 2021-07-20 alle 15 31 25

## Steps to Reproduce

1. 
2. 

### Environment

- OS:  iOs 11.4
- Node.js version: 14.17
- algob version: 1.1.1

amityadav0:
Closes #423 

## Proposed Changes

+ Add accounts name support while opting in to assets
+ Add tests


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 + Updated `algob init` to initialize a typescript project as well by passing `--typescript` flag. Usage: `algob init <location> --typescript`.
 + Support pooled transaction fees in algob and runtime. Added tests
 + Add `flatFee` in `TxParams`
++ User can use accounts name in runtime(similar to algob)('john', 'bob' etc)
```
ratik21, line 15:
```suggestion
+ User can initialize & use accounts by name in runtime, similar to algob ('john', 'bob' etc)
```

packages/runtime/src/account.ts
```diff
@@ -39,8 +39,12 @@ export class AccountStore implements AccountStoreI {
...
-    if (account) {
+  constructor (balance: number | bigint, account?: RuntimeAccount | string) {
+    if (typeof account === 'string') {
+      this.account = generateAccount();
+      this.account.name = account;
+      this.address = this.account.addr;
+    } else if (account) {
       // set config if account is passed by user
       this.account = account;
       this.address = account.addr;
```
ratik21, line 50: Could you please add `this.account.name = account.name` in `else if` as well? User can pass whole account including name as well, as we're using type `RuntimeAccount`

packages/runtime/src/lib/asa.ts
```diff
@@ -7,7 +7,7 @@ import * as z from 'zod';
 import { RUNTIME_ERRORS } from "../errors/errors-list";
 import { RuntimeError } from "../errors/runtime-errors";
 import { parseZodError } from "../errors/validation-errors";
-import { AccountMap, RuntimeAccountMap } from "../types";
+import { AccountMap, RuntimeAccount, RuntimeAccountMap } from "../types";
```
ratik21, line 10: Linter warning: `'RuntimeAccount' is defined but never used. Allowed unused vars must match /_/u`

packages/runtime/test/integration/asa-load-fail.ts
```diff
@@ -0,0 +1,32 @@
...
+      RUNTIME_ERRORS.ASA.PARAM_ERROR_NO_NAMED_OPT_IN_ACCOUNT
+    );
+  });
+
+  it("Should load runtime if all accounts exist", () => {
+    const elon = new AccountStore(minBalance, 'elon');
+    bob = new AccountStore(minBalance, 'bob');
+    alice = new AccountStore(minBalance, 'alice');
+
+    runtime = new Runtime([elon, bob, alice]);
```
ratik21, line 30:
```suggestion
    new Runtime([elon, bob, alice]);
```

packages/runtime/test/integration/deploy-asa.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  // helper function
+  function syncAccounts (): void {
+    john = runtime.getAccount(john.address);
+    bob = runtime.getAccount(bob.address);
+    alice = runtime.getAccount(alice.address);
+  }
+
+  it("Should opt-in to multiple accounts mentioned in asa.yaml", () => {
+    const asset = runtime.addAsset('asa', { creator: { ...john.account, name: "john" } });
```
ratik21, line 30: if `john.account` includes name(=john) as well, then can we directly use `{ creator: john.account }` here?

packages/runtime/test/integration/deploy-asa.ts
```diff
@@ -0,0 +1,36 @@
...
+
+    syncAccounts();
+    assert.isDefined(bob.getAssetHolding(asset));
+    assert.isDefined(alice.getAssetHolding(asset));
+  });
+});
```
ratik21, line 36: could you please clear this linter warning
image

ratik21 (commented):
Please add a note in `docs/guide` as well, regarding initializing accounts by name

packages/runtime/src/account.ts
amityadav0: that's not required because this.account's type is `RuntimeAcccount`, so if account has name in it, it will be stored

packages/runtime/test/integration/deploy-asa.ts
amityadav0: no, because name is optional in john.account but while adding a asset it is required

packages/runtime/src/lib/asa.ts
amityadav0: removed

packages/runtime/test/integration/deploy-asa.ts
amityadav0: removed

packages/runtime/test/integration/deploy-asa.ts
ratik21: hmm, so this is a types issue. Can we use `{ creator: john.account as Account }` ? Because we're already initializing account by name, so passing name again seems extra here. Anyway, not a big deal.

ratik21 (approved):

packages/runtime/test/integration/deploy-asa.ts
amityadav0: the problem here is that name is optional, so we will have to use a type where name is not optional.

ratik21: [https://github.com/scale-it/algo-builder/blob/master/packages/runtime/src/types.ts#L435](https://github.com/scale-it/algo-builder/blob/master/packages/runtime/src/types.ts#L435)

PR #431, opened 2021-07-29
update `make start-indexer` to reset db before starting

ratik21:


infrastructure/Makefile
```diff
@@ -139,9 +139,13 @@ setup-postgresql:
...
 	@sudo -u postgres psql -c "\l"
 
+reset-db:
+	@sudo -u postgres psql -c "DROP DATABASE IF EXISTS $(POSTGRES_DBNAME);"
+	@sudo -u postgres createdb $(POSTGRES_DBNAME)
+	@sudo -u postgres psql -c "grant all privileges on database $(POSTGRES_DBNAME) to $(POSTGRES_UNAME)"
 
 # Sets up indexer in local mode (to be used with private-net)
-start-indexer:
+start-indexer: reset-db
```
robert-zaremba, line 148: why it will reset the DB? I think it's better to continue and have similar commands as we have with the private net for recreating: `setup-*` and `recreate-*`

ratik21: This is recreation only. It will drop the database and create a new one (before starting indexer each time). While i was doing [this](https://www.pivotaltracker.com/story/show/178687692) task using `algorand-indexer`, with each start the indexer wasn't able to catch up properly so i was doing this manually. 

Shall i rename this to `recreate-db`, or close this PR?

robert-zaremba: yes, let's rename to `recreate-db`

ratik21: ok

infrastructure/Makefile
```diff
@@ -152,7 +156,8 @@ start-indexer:
 # Start indexer by providing database connection info and path to algod node
 	$(INDEXER_PATH)/algorand-indexer_linux_amd64_2.5.0/algorand-indexer daemon -P "host=localhost port=5432 user=$(POSTGRES_UNAME) password=$(POSTGRES_PASS) dbname=$(POSTGRES_DBNAME) sslmode=disable" --algod=./node_data/PrimaryNode
 
+recreate-indexer: recreate-indexer-db start-indexer
```
robert-zaremba, line 159: I've added `recreate-indexer` job.

robert-zaremba (approved):

infrastructure/Makefile
ratik21: OK

PR #436, opened 2021-08-02
Test watcher

robert-zaremba:
## Proposed Changes

+ Added test watch option for all tests
+ Updated contribution guide with note about running tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
     * Support for signing in a group transaction (loaded from file).
 + Add docs for web package
 + Handle group transaction in web package
++ Added `yarn run test:watch` command. NOTE: it will spawn multiple process in the same terminal session. So if you want to stop the all processes the best solution is to kill the terminal session.
 
 ### Infrastructure
 * Added new make commands:
--- a/README.md
+++ b/README.md
@@ -231,7 +231,7 @@ To execute tests in all workspaces, run the following from the root directory:
 yarn run test
 ```
 
-To execute and watch tests in all workspaces, run the following from the root directory. Note: it will spawn multiple process in the same terminal session. So if you want to stop the all processes you can either call `pkill -f mocha` or kill the terminal session.
+To execute and watch tests in all workspaces, run the following from the root directory. Note: it will spawn multiple processes in the same terminal session. So if you want to stop the all processes you can either call `pkill -f mocha` or kill the terminal session.
 ```
 yarn run test:watch
 ```
--- a/README.md
+++ b/README.md
@@ -231,7 +231,9 @@ To execute tests in all workspaces, run the following from the root directory:
 yarn run test
 ```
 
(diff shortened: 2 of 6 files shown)


robert-zaremba:
I didn't find a better option. Once `yarn-2` will be widely adopted then we can use `yarn workspaces foreach -p`: https://next.yarnpkg.com/cli/workspaces/foreach

amityadav0:
This works for the first time, when i make changes in algob, test run but some tests are failing.
After making more changes tests start failing.

robert-zaremba:
Hmm, I was testing it and it worked. I did changes in both algob and runtime to make tests falling and then rollback and everything worked.

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@
     * Support for signing in a group transaction (loaded from file).
 + Add docs for web package
 + Handle group transaction in web package
++ Added `yarn run test:watch` command. NOTE: it will spawn multiple process in the same terminal session. So if you want to stop the all processes the best solution is to kill the terminal session.
```
amityadav0, line 22:
```suggestion
+ Added `yarn run test:watch` command. NOTE: It will spawn multiple processes in the same terminal session. So if you want to stop all processes the best solution is to kill the terminal session.
```

README.md
```diff
@@ -207,3 +207,31 @@ We use **yarn workspaces** to manage all sub packages. here is a list of command
...
+cd packages/runtime
+yarn run test -w
+```
+
+To execute tests in all workspaces, run the following from the root directory:
+```
+yarn run test
+```
+
+To execute and watch tests in all workspaces, run the following from the root directory. Note: it will spawn multiple process in the same terminal session. So if you want to stop the all processes the best solution is to kill the terminal session.
```
amityadav0, line 234:
```suggestion
To execute and watch tests in all workspaces, run the following from the root directory. Note: It will spawn multiple processes in the same terminal session. So if you want to stop all processes the best solution is to kill the terminal session.
```

amityadav0 (approved):

README.md
```diff
@@ -231,7 +231,9 @@ To execute tests in all workspaces, run the following from the root directory:
 yarn run test
 ```
 
-To execute and watch tests in all workspaces, run the following from the root directory. Note: it will spawn multiple process in the same terminal session. So if you want to stop the all processes the best solution is to kill the terminal session.
+To execute and watch tests in all workspaces, run the following from the root directory. Note: it will spawn multiple process in the same terminal session. So if you want to stop the all processes you can either call `pkill -f mocha` or kill the terminal session.
 ```
 yarn run test:watch
 ```
+
+NOTE: for the moment test watching in `packages/algob` not stable because of tear down issues in some test suites. We advise to not use test watcher in `packages/algob`.
```
ratik21, line 239:
```suggestion
NOTE: For the moment test watching in `packages/algob` is not stable because of tear down issues in some test suites. We advise to **not use** test watcher in `packages/algob`.
```

ratik21 (approved):

Issue #433, opened 2021-08-01
Unable to install with npm or yarn: missing file/module


plasmatech8:
## Describe the bug

I have tried to install in an Ubuntu docker container and on my Pop-OS desktop machine.

When I try to install using npm, it gives an error 'no such file or directory' ... 'This is related to npm not being able to find a file.'. 

When I try to install using yarn, it installs but on usage gives: 'Error: Cannot find module '@nodelib/fs.walk''

#### Expected Behavior
I expect the package to install without stating that a file or module is missing.

#### Current Behavior
On `npm install -g @algo-builder/algob`
```bash
npm ERR! code ENOENT
npm ERR! syscall chmod
npm ERR! path /usr/local/lib/node_modules/@algo-builder/algob/node_modules/@algo-builder/runtime/build/algod.js
npm ERR! errno -2
npm ERR! enoent ENOENT: no such file or directory, chmod '/usr/local/lib/node_modules/@algo-builder/algob/node_modules/@algo-builder/runtime/build/algod.js'
npm ERR! enoent This is related to npm not being able to find a file.
npm ERR! enoent 

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/mark/.npm/_logs/2021-08-01T05_08_22_306Z-debug.log
```
On `yarn global add @algo-builder/algob`, and then `algob`:
```
Error: Cannot find module '@nodelib/fs.walk'
Require stack:
- /usr/local/share/.config/yarn/global/node_modules/@algo-builder/runtime/build/lib/files.js
- /usr/local/share/.config/yarn/global/node_modules/@algo-builder/runtime/build/lib/asa.js
- /usr/local/share/.config/yarn/global/node_modules/@algo-builder/runtime/build/account.js
- /usr/local/share/.config/yarn/global/node_modules/@algo-builder/runtime/build/index.js
- /usr/local/share/.config/yarn/global/node_modules/@algo-builder/algob/build/errors/errors.js
- /usr/local/share/.config/yarn/global/node_modules/@algo-builder/algob/build/internal/cli/cli.js
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:889:15)
    at Function.Module._load (internal/modules/cjs/loader.js:745:27)
    at Module.require (internal/modules/cjs/loader.js:961:19)
    at require (internal/modules/cjs/helpers.js:92:18)
[...]

PR #437, opened 2021-08-03
Fix: yarn add @algo-builder/algob gives dependency issues

ratik21:
Closes: #433

image

After fixing the above:
Screenshot from 2021-08-03 04-22-16

## Proposed Changes

Added these dependencies in `package.json`. Testing by locally publishing these pacakges (using `yarn pack`).
NOTE: `yarn link` is working fine. `yarn add @algo-builder/algob` is giving issues.


packages/algob/package.json
```diff
@@ -99,6 +99,7 @@
     "source-map-support": "^0.5.19",
     "tmp": "^0.2.1",
     "yaml": "^1.10.0",
+    "mocha": "^9.0.3",
```
robert-zaremba, line 102: I this should be devDependency

robert-zaremba (approved):
pre approving. Left one comment about mocha

packages/algob/package.json
ratik21: Yes, but it is required in /src while executing `algob test` command. So putting this as a dev dependency would fail

PR #442, opened 2021-08-10
Add dependency zod in `web/package.json`

ratik21:
`yarn add @algo-builder/web` in webapp:
Screenshot from 2021-08-10 23-28-28


amityadav0 (approved):

robert-zaremba:
I think we already were adding it in the past.

robert-zaremba (approved):

PR #443, opened 2021-08-12
TEALv4: Sharing data between contracts in runtime

amityadav0:
## Proposed Changes

+ Add shared scratch space support in runtime
+ Add tealv4 constants
+ Add `gload` and `gloads` opcodes
+ Add tests

## Potential followups
+ Add integration test with two stateful contracts (https://www.pivotaltracker.com/story/show/179202771)

### DIFF
--- a/packages/runtime/src/ctx.ts
+++ b/packages/runtime/src/ctx.ts
@@ -12,7 +12,7 @@ import {
   AppDeploymentFlags,
   ASADeploymentFlags, AssetHoldingM,
   Context, ExecutionMode,
-  SSCAttributesM, State, Txn
+  SSCAttributesM, StackElem, State, Txn
 } from "./types";
 
 const APPROVAL_PROGRAM = "approval-program";
@@ -24,6 +24,7 @@ export class Ctx implements Context {
   args: Uint8Array[];
   runtime: Runtime;
   debugStack?: number; //  max number of top elements from the stack to print after each opcode execution.
+  sharedScratchSpace: Map<number, StackElem[]>;
 
   constructor (state: State, tx: Txn, gtxs: Txn[], args: Uint8Array[],
     runtime: Runtime, debugStack?: number) {
@@ -33,6 +34,7 @@ export class Ctx implements Context {
     this.args = args;
     this.runtime = runtime;
     this.debugStack = debugStack;
+    this.sharedScratchSpace = new Map<number, StackElem[]>();
   }
 
   // verify account's balance is above minimum required balance
@@ -225,7 +227,7 @@ export class Ctx implements Context {
    */
   addApp (
     fromAccountAddr: AccountAddress, flags: AppDeploymentFlags,
-    approvalProgram: string, clearProgram: string
+    approvalProgram: string, clearProgram: string, idx?: number
   ): number {
     const senderAcc = this.getAccount(fromAccountAddr);
 
@@ -242,7 +244,9 @@ export class Ctx implements Context {
     this.state.accounts.set(senderAcc.address, senderAcc);
     this.state.globalApps.set(app.id, senderAcc.address);
 
-    this.runtime.run(approvalProgram, ExecutionMode.APPLICATION, this.debugStack); // execute TEAL code with appID = 0
+    this.runtime.run(
+      approvalProgram, ExecutionMode.APPLICATION, idx ?? 0, this.debugStack
+    ); // execute TEAL code with appID = 0
 
     // create new application in globalApps map
     this.state.globalApps.set(++this.state.appCounter, senderAcc.address);
@@ -272,13 +276,13 @@ export class Ctx implements Context {
    * @param appID Application index
    * NOTE: When creating or opting into an app, the minimum balance grows before the app code runs
    */
-  optInToApp (accountAddr: AccountAddress, appID: number): void {
+  optInToApp (accountAddr: AccountAddress, appID: number, idx?: number): void {
     const appParams = this.getApp(appID);
 
     const account = this.getAccount(accountAddr);
     account.optInToApp(appID, appParams);
     this.assertAccBalAboveMin(accountAddr);
-    this.runtime.run(appParams[APPROVAL_PROGRAM], ExecutionMode.APPLICATION, this.debugStack);
+    this.runtime.run(appParams[APPROVAL_PROGRAM], ExecutionMode.APPLICATION, idx ?? 0, this.debugStack);
   }
 
   /**
@@ -474,7 +478,8 @@ export class Ctx implements Context {
   updateApp (
     appID: number,
     approvalProgram: string,
-    clearProgram: string
+    clearProgram: string,
+    idx?: number
   ): void {
     if (approvalProgram === "") {
       throw new RuntimeError(RUNTIME_ERRORS.GENERAL.INVALID_APPROVAL_PROGRAM);
@@ -484,7 +489,7 @@ export class Ctx implements Context {
     }
 
     const appParams = this.getApp(appID);
-    this.runtime.run(appParams[APPROVAL_PROGRAM], ExecutionMode.APPLICATION, this.debugStack);
+    this.runtime.run(appParams[APPROVAL_PROGRAM], ExecutionMode.APPLICATION, idx ?? 0, this.debugStack);
 
     const updatedApp = this.getApp(appID);
     updatedApp[APPROVAL_PROGRAM] = approvalProgram;
@@ -526,13 +531,13 @@ export class Ctx implements Context {
         case types.TransactionType.CallNoOpSSC: {
           this.tx = this.gtxs[idx]; // update current tx to the requested index
           const appParams = this.getApp(txnParam.appID);
-          this.runtime.run(appParams[APPROVAL_PROGRAM], ExecutionMode.APPLICATION, this.debugStack);
+          this.runtime.run(appParams[APPROVAL_PROGRAM], ExecutionMode.APPLICATION, idx, this.debugStack);
           break;
         }
         case types.TransactionType.CloseApp: {
           this.tx = this.gtxs[idx]; // update current tx to the requested index
           const appParams = this.getApp(txnParam.appID);
-          this.runtime.run(appParams[APPROVAL_PROGRAM], ExecutionMode.APPLICATION, this.debugStack);
+          this.runtime.run(appParams[APPROVAL_PROGRAM], ExecutionMode.APPLICATION, idx, this.debugStack);
           this.closeApp(fromAccountAddr, txnParam.appID);
           break;
         }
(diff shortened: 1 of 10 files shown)


packages/runtime/src/ctx.ts
```diff
@@ -24,6 +24,7 @@ export class Ctx implements Context {
   args: Uint8Array[];
   runtime: Runtime;
   debugStack?: number; //  max number of top elements from the stack to print after each opcode execution.
+  sharedScratchSpace: Map<number, StackElem[]>;
```
ratik21, line 27: Please add a comment what number represents here. Or create a type `type TransactionIndex = number;`

packages/runtime/src/ctx.ts
```diff
@@ -225,7 +227,7 @@ export class Ctx implements Context {
    */
   addApp (
     fromAccountAddr: AccountAddress, flags: AppDeploymentFlags,
-    approvalProgram: string, clearProgram: string
+    approvalProgram: string, clearProgram: string, idx?: number
```
ratik21, line 230: please update description

packages/runtime/src/ctx.ts
```diff
@@ -272,13 +276,13 @@ export class Ctx implements Context {
    * @param appID Application index
    * NOTE: When creating or opting into an app, the minimum balance grows before the app code runs
    */
-  optInToApp (accountAddr: AccountAddress, appID: number): void {
+  optInToApp (accountAddr: AccountAddress, appID: number, idx?: number): void {
```
ratik21, line 279: please update description

packages/runtime/src/ctx.ts
```diff
@@ -484,7 +489,7 @@ export class Ctx implements Context {
     }
 
     const appParams = this.getApp(appID);
-    this.runtime.run(appParams[APPROVAL_PROGRAM], ExecutionMode.APPLICATION, this.debugStack);
+    this.runtime.run(appParams[APPROVAL_PROGRAM], ExecutionMode.APPLICATION, idx ?? 0, this.debugStack);
```
ratik21, line 492: When can `idx` be `undefined`?

packages/runtime/src/errors/errors-list.ts
```diff
@@ -237,6 +237,12 @@ by an index that does not exist.`
     message: "set bit index %index% is beyond bytes length. [error-line: %line%]",
     title: 'set bit index error',
     description: `set bit index error`
+  },
+  SCRATCH_EXIST_ERROR: {
+    number: 1035,
+    message: "scratch space doesn't exist for index: %index%. [error-line: %line%]",
+    title: 'scratch space error',
+    description: `fails maybe because the requested transaction is an ApplicationCall or T < GroupIndex.`
```
ratik21, line 245: message and description seems completely different here. Could you please add a reason in message as well?

packages/runtime/src/errors/errors-list.ts
```diff
@@ -237,6 +237,12 @@ by an index that does not exist.`
     message: "set bit index %index% is beyond bytes length. [error-line: %line%]",
     title: 'set bit index error',
     description: `set bit index error`
+  },
+  SCRATCH_EXIST_ERROR: {
+    number: 1035,
+    message: "scratch space doesn't exist for index: %index%. [error-line: %line%]",
```
ratik21, line 243:
```suggestion
    message: "scratch space doesn't exist for transaction index: %index%. [error-line: %line%]",
```

packages/runtime/src/errors/errors-list.ts
```diff
@@ -237,6 +237,12 @@ by an index that does not exist.`
...
+    title: 'scratch space error',
```
ratik21, line 244: I think title can be more intuitive. This is related to the shared scratch space here, right?

packages/runtime/src/interpreter/opcode-list.ts
```diff
@@ -2591,3 +2591,84 @@ export class MinBalance extends Op {
...
+  }
+
+  execute (stack: TEALStack): void {
+    const scratch = this.interpreter.runtime.ctx.sharedScratchSpace.get(this.txIndex);
+    if (scratch === undefined) {
+      throw new RuntimeError(
+        RUNTIME_ERRORS.TEAL.SCRATCH_EXIST_ERROR,
+        { index: this.txIndex, line: this.line }
+      );
+    }
```
ratik21, line 2631: just to confirm, Is this validated on node?

packages/runtime/src/interpreter/opcode-list.ts
```diff
@@ -2591,3 +2591,84 @@ export class MinBalance extends Op {
...
+    const txIndex = Number(this.assertBigInt(stack.pop(), this.line));
+    const scratch = this.interpreter.runtime.ctx.sharedScratchSpace.get(txIndex);
+    if (scratch === undefined) {
+      throw new RuntimeError(
+        RUNTIME_ERRORS.TEAL.SCRATCH_EXIST_ERROR,
+        { index: txIndex, line: this.line }
+      );
+    }
+    this.checkIndexBound(this.scratchIndex, scratch, this.line);
+    stack.push(scratch[this.scratchIndex]);
```
ratik21, line 2672: This is the same code as `Gload` (except getting `txIndex`). Can we inherit from Gload (`class Gloads is Gload)`?

packages/runtime/src/lib/constants.ts
```diff
@@ -235,6 +240,12 @@ GlobalFields[3] = {
   CreatorAddress: null
 };
 
+// global fields supported by tealv3
```
ratik21, line 243:
```suggestion
// global fields supported by tealv4
```

packages/runtime/src/lib/constants.ts
```diff
@@ -235,6 +240,12 @@ GlobalFields[3] = {
...
+GlobalFields[4] = {
+  ...GlobalFields[3],
+  CreatorAddress: null
+};
```
ratik21, line 247:
```suggestion
GlobalFields[4] = {
  ...GlobalFields[3]
};
```

packages/runtime/src/lib/constants.ts
```diff
@@ -259,3 +270,9 @@ OpGasCost[2] = {
  * All other opcodes have cost 1
  */
 OpGasCost[3] = { ...OpGasCost[2] };
+
+/**
+ * In tealv4, cost of crypto opcodes are same as v2.
+ * All other opcodes have cost 1
+ */
+OpGasCost[4] = { ...OpGasCost[3] };
```
ratik21, line 278:
```suggestion
/*
 * tealv4
 */
OpGasCost[4] = { ...OpGasCost[3] };
```

packages/runtime/src/runtime.ts
```diff
@@ -646,8 +647,11 @@ export class Runtime {
    * each opcode execution (upto depth = debugStack)
    * NOTE: Application mode is only supported in TEALv > 1
    */
-  run (program: string, executionMode: ExecutionMode, debugStack?: number): void {
+  run (program: string, executionMode: ExecutionMode, indexInGroup: number, debugStack?: number): void {
     const interpreter = new Interpreter();
     interpreter.execute(program, executionMode, this, debugStack);
+    if (executionMode === ExecutionMode.APPLICATION) {
+      this.ctx.sharedScratchSpace.set(indexInGroup, interpreter.scratch);
+    }
```
ratik21, line 655: Can we do it inside interpreter only (after executing all opcodes)? Could you please check?
Inside interpreter we have access to ctx (`this.runtime.ctx`). We can also get `indexInGroup` (using `ctx.gtxs`).

If possible then we won't need to pass index in each `run`.

packages/runtime/test/src/interpreter/opcode-list.ts
```diff
@@ -4074,4 +4074,93 @@ describe("Teal Opcodes", function () {
       );
     });
   });
+
+  describe("shared data between contracts", () => {
```
ratik21, line 4078:
```suggestion
  describe("TEALv4: shared data between contracts", () => {
```

packages/runtime/src/errors/errors-list.ts
amityadav0: message says scratch space doesn't exist, description states that either that tx is not application call or tx is not even in group, how is it unrelated ?
I will add there a reason may be tag

packages/runtime/src/errors/errors-list.ts
amityadav0: What do you propose ?

packages/runtime/src/interpreter/opcode-list.ts
amityadav0: it will be with integration tests. I have been waiting for this one's review.

packages/runtime/src/interpreter/opcode-list.ts
amityadav0: in this we pass only one variable, so constructor checks are different, plus execute function loads from stack instead of reading from value stored using constructor. So I don't think this good idea

packages/runtime/src/runtime.ts
amityadav0: Good idea, I will try this.

packages/runtime/src/errors/errors-list.ts
ratik21: By different I meant the wording. During catch I think error message is shown, so I suggested to add a reason in message.

packages/runtime/src/errors/errors-list.ts
ratik21: Simple, `scratch space not found`/`scratch space does not exist`

packages/runtime/src/interpreter/opcode-list.ts
ratik21: Ok let's repeat code

packages/runtime/src/errors/errors-list.ts
amityadav0: ok

packages/runtime/src/errors/errors-list.ts
amityadav0: same as message, OK

packages/runtime/src/runtime.ts
amityadav0: i checked it, to get indexInGroup from ctx.gtxs there are two options 
1. traverse gtxs and match tx(this seems not good  and slow) 
2. have a tx variable set in runtime but this will be similar to what we have
3. I like current implementation because it is easy to understand .

packages/runtime/src/ctx.ts
amityadav0: ok

packages/runtime/src/ctx.ts
amityadav0: when user directly calls updateApp or maybe we can handle it in runtime.ts, it will be more clear

amityadav0: moved

packages/runtime/src/runtime.ts
ratik21: yes, i was thinking the first one (`indexInGroup = gtxs.indexOf(tx);`). It introduces an extra loop. 
Ok, let's keep what we have.

ratik21 (approved):
lgtm

PR #444, opened 2021-08-13
TEALv4: Transaction Array Changes

ratik21:
## Proposed Changes

Transaction array changes:
+ Array length validation for `appArgs`, `appAccounts`, `appForeignAssets`, `appForeginApps`.
+ From `tealv4`, user can pass "address" directly instead of an index to `tx.Accounts[]` array.
+ From `tealv4`, user can pass offset to `foreginApps`/`foreignAssets` in TEAL, OR pass assetID/appID directly.

### DIFF
--- a/packages/runtime/src/interpreter/interpreter.ts
+++ b/packages/runtime/src/interpreter/interpreter.ts
@@ -136,9 +136,13 @@ export class Interpreter {
   }
 
   /**
-   * Queries appIndex from Txn.ForeignApps or `ctx.tx.apid` (current app if `appIndex==0`).
-   * If appIndex is passed directly, then verifies that appID is present in foreignApps array.
-   * @param appRef index to foreign apps array OR applicationID
+   * Queries appIndex by app reference (offset to foreignApps array OR index directly)
+   * + Since TEALv4, any reference is supported (but it must be present in foreignApps array)
+   * + For older versions, if foreign === true, reference is treated as offset to foreignApps array,
+   * otherwise it is treated as a direct reference.
+   * @param appRef an offset to foreign app array OR appID
+   * @param foreign for older teal versions(<= 3), foreign bool represent if ref is
+   * treated as an offset/appIndex
    * @param line line number
    * https://developer.algorand.org/articles/introducing-algorand-virtual-machine-avm-09-release/
    */
@@ -178,9 +182,13 @@ export class Interpreter {
   }
 
   /**
-   * Queries assetIndex from Txn.ForeignAssets (note: index 0 means first asset in foreign asset array).
-   * If assetIndex is passed directly, then verifies that assetIndex is present in ForeignAssets array.
+   * Queries assetIndex by asset reference (offset to foreignAssets array OR index directly)
+   * + Since TEALv4, any reference is supported (but it must be present in foreign assets array)
+   * + For older versions, if foreign === true, reference is treated as offset to foreignAssets array,
+   * otherwise it is treated as a direct reference.
    * @param assetRef an offset to foreign assets array OR assetID
+   * @param foreign for older teal versions(<= 3), foreign bool represent if
+   * ref is treated as an offset/assetIndex
    * @param line line number
    * https://developer.algorand.org/articles/introducing-algorand-virtual-machine-avm-09-release/
    */
@@ -284,8 +292,9 @@ export class Interpreter {
 
   /**
    * Assets transaction references (apps, assets, accounts) lengths are valid:
-   * 1. The AVM limits the accounts array to no more than 4
-   * 2. Assets and application arrays combined and totaled with the accounts array can not exceed 8
+   * 1. Application args are limited to max. size of 16.
+   * 2. The AVM limits the accounts array to no more than 4
+   * 3. Assets and application arrays combined and totaled with the accounts array can not exceed 8
    * https://developer.algorand.org/articles/introducing-algorand-virtual-machine-avm-09-release/
    */
   assertValidTxArray (): void {
--- a/packages/runtime/src/interpreter/interpreter.ts
+++ b/packages/runtime/src/interpreter/interpreter.ts
@@ -3,15 +3,16 @@ import { encodeAddress, modelsv2 } from "algosdk";
 import { RUNTIME_ERRORS } from "../errors/errors-list";
 import { RuntimeError } from "../errors/runtime-errors";
 import { Runtime } from "../index";
-import { checkIndexBound } from "../lib/compare";
-import { DEFAULT_STACK_ELEM } from "../lib/constants";
+import { checkIndexBound, compareArray } from "../lib/compare";
+import { ALGORAND_MAX_APP_ARGS_LEN, ALGORAND_MAX_TX_ACCOUNTS_LEN, ALGORAND_MAX_TX_ARRAY_LEN, DEFAULT_STACK_ELEM, MaxTEALVersion } from "../lib/constants";
 import { keyToBytes } from "../lib/parsing";
 import { Stack } from "../lib/stack";
 import { parser } from "../parser/parser";
-import type {
+import {
   AccountStoreI, ExecutionMode, Operator, SSCAttributesM,
   StackElem, TEALStack
 } from "../types";
+import { Op } from "./opcode";
 import { Label } from "./opcode-list";
 
 export class Interpreter {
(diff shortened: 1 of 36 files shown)


packages/runtime/src/interpreter/interpreter.ts
```diff
@@ -65,34 +66,160 @@ export class Interpreter {
...
+   * + Since TEALv4, any reference is supported (but it must be present in foreignApps array)
+   * + For older versions, if foreign === true, reference is treated as offset to foreignApps array,
+   * otherwise it is treated as a direct reference.
+   * @param appRef an offset to foreign app array OR appID
+   * @param foreign for older teal versions(<= 3), foreign bool represent if ref is
+   * treated as an offset/appIndex
+   * @param line line number
+   * https://developer.algorand.org/articles/introducing-algorand-virtual-machine-avm-09-release/
+   */
+  getAppID (appRef: number, foreign: boolean, line: number, op: Op): number {
```
amityadav0, line 149: maybe `getAppIDByReference`

packages/runtime/test/integration/consensus.ts
```diff
@@ -0,0 +1,125 @@
+import { aliceAcc, bobAcc } from "@algo-builder/algob/test/mocks/account";
```
amityadav0, line 1: Please update the filename

packages/runtime/src/interpreter/interpreter.ts
```diff
@@ -65,34 +66,160 @@ export class Interpreter {
...
-      if (this.runtime.ctx.tx.apat) {
-        pkBuffer = this.runtime.ctx.tx.apat[Number(accIndex)];
+    if (typeof accountRef === 'bigint') {
+      if (accountRef === 0n) {
+        address = encodeAddress(this.runtime.ctx.tx.snd);
+        account = this.runtime.ctx.state.accounts.get(address);
       } else {
-        throw new Error("pk Buffer not found");
+        const accIndex = accountRef - 1n;
+        checkIndexBound(Number(accIndex), this.runtime.ctx.tx.apat as Buffer[], line);
```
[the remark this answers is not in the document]
ratik21: from algosdk types `txAccounts` can be undefined, but these are executed by stateful ops. So it is only a type hack.

packages/runtime/src/interpreter/interpreter.ts
ratik21: good idea

packages/runtime/src/interpreter/interpreter.ts
```diff
@@ -65,34 +66,160 @@ export class Interpreter {
...
-      let pkBuffer;
-      if (this.runtime.ctx.tx.apat) {
-        pkBuffer = this.runtime.ctx.tx.apat[Number(accIndex)];
+    if (typeof accountRef === 'bigint') {
+      if (accountRef === 0n) {
+        address = encodeAddress(this.runtime.ctx.tx.snd);
+        account = this.runtime.ctx.state.accounts.get(address);
       } else {
-        throw new Error("pk Buffer not found");
+        const accIndex = accountRef - 1n;
```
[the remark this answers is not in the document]
ratik21: Yes it is verified. For `tx.Accounts` index 0 means `txn.sender` & it is 1 indexed (so we decrement accountRef by 1).

packages/runtime/test/integration/consensus.ts
ratik21: what do you suggest?

amityadav0: maybe transaction-array-tests or something related to feature added

ratik21: ok

packages/runtime/src/interpreter/interpreter.ts
```diff
@@ -176,6 +339,7 @@ export class Interpreter {
   execute (program: string, mode: ExecutionMode, runtime: Runtime, debugStack?: number): void {
     this.runtime = runtime;
     this.instructions = parser(program, mode, this);
+    if (mode === ExecutionMode.APPLICATION) { this.assertValidTxArray(); }
```
amityadav0, line 342: we were not doing it before v4. shouldn't we consider that here ? i think this check should only be called after AVM is introduced

packages/runtime/src/interpreter/opcode-list.ts
```diff
@@ -1793,9 +1793,9 @@ export class AppLocalPut extends Op {
     this.assertMinStackLen(stack, 3, this.line);
     const value = stack.pop();
     const key = this.assertBytes(stack.pop(), this.line);
-    const accountIndex = this.assertBigInt(stack.pop(), this.line);
+    const accountRef: StackElem = stack.pop();
```
amityadav0, line 1796: Here we should follow different approach. what if version here is 3 ? the same opcode will behave differently.
so maybe we can use opcodev2, opcodev4 as suggested by robert.

packages/runtime/src/interpreter/opcode-list.ts
```diff
@@ -2023,18 +2024,9 @@ export class GetAssetDef extends Op {
...
-      Number(foreignAssetsIdx),
-      this.interpreter.runtime.ctx.tx.apas as number[], this.line);
-
-    let assetId;
-    if (this.interpreter.runtime.ctx.tx.apas) {
-      assetId = this.interpreter.runtime.ctx.tx.apas[Number(foreignAssetsIdx)];
-    } else {
-      throw new Error("foreign asset id not found");
-    }
-    const AssetDefinition = this.interpreter.getAssetDef(assetId);
```
amityadav0, line 2037: same for all other opcodes

packages/runtime/test/src/interpreter/opcode-list.ts
```diff
@@ -3243,6 +3337,42 @@ describe("Teal Opcodes", function () {
         RUNTIME_ERRORS.TEAL.INDEX_OUT_OF_BOUND
       );
     });
+
+    it("tealv4: should push correct value accepting offset to foreignAssets", () => {
+      interpreter.tealVersion = 4;
```
amityadav0, line 3342: could you please add tests where tealVersion = 3 or less and value accepted is correct (according to v3)?

packages/runtime/test/integration/consensus.ts
```diff
@@ -1,46 +0,0 @@
-import { assert } from "chai";
```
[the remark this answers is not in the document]
ratik21: renamed consensus.ts to tx-array-len.ts

packages/runtime/src/lib/constants.ts
```diff
@@ -25,6 +25,10 @@ export const LogicSigMaxCost = 20000;
 export const MaxAppProgramCost = 700;
 export const LogicSigMaxSize = 1000;
 export const MaxAppProgramLen = 1024;
+export const ALGORAND_MAX_APP_ARGS_LEN = 16;
+export const ALGORAND_MAX_TX_ACCOUNTS_LEN = 4;
+// the assets and application arrays combined and totaled with the accounts array can not exceed 8
+export const ALGORAND_MAX_TX_ARRAY_LEN = 8;
```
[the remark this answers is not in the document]
ratik21: we have already added link to consensus params in go-algorand.

packages/runtime/src/interpreter/interpreter.ts
ratik21: It was before tealv4 as well. We were not doing it. AVM just introduced dynamic size (total len should be 8), and is not version specific (just like pooled transaction fees). Verified on node as well

packages/runtime/test/src/interpreter/opcode-list.ts
ratik21: Previous test follow version <= 3, so that case is already tested. Added new tests for v4.

packages/runtime/src/interpreter/opcode-list.ts
ratik21: I think it's better to resolve an assetId/appId by reference in this case (opcode & function being teal version agnostic). Previously it was confusing resolving these id's in separate ops (we were handling lot of cases). So i prefer to keep this in a single function [https://github.com/algorand/go-algorand/blob/master/data/transactions/logic/eval.go#L3041](https://github.com/algorand/go-algorand/blob/master/data/transactions/logic/eval.go#L3041)

packages/runtime/src/interpreter/opcode-list.ts
ratik21: same

packages/runtime/src/interpreter/opcode-list.ts
amityadav0: Is it possible that this opcode will behave differently in different versions ?
Or am I misunderstanding this here ?

packages/runtime/src/lib/constants.ts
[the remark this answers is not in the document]
amityadav0: could you please be specific ?

PR #445, opened 2021-08-17
TEALv4: Dynamic Opcode Cost Evaluation

ratik21:
## Proposed Changes

If using version 4 of TEAL or higher, opcode costs are now evaluated dynamically instead of statically.


packages/runtime/src/interpreter/interpreter.ts
```diff
@@ -175,10 +177,15 @@ export class Interpreter {
...
 
+    let dynamicCost = 0;
     while (this.instructionIndex < this.instructions.length) {
       const instruction = this.instructions[this.instructionIndex];
       instruction.execute(this.stack);
 
+      // for teal version >= 4, cost is calculated dynamically at the time of execution
+      dynamicCost += this.lineToCost[instruction.line];
+      if (this.tealVersion >= 4) { assertMaxCost(dynamicCost, mode); }
+
```
amityadav0, line 188: By doing this,  it is possible that cost is very low because if some branches of code are never executed. right ?
or this cost is addition to static cost ?

packages/runtime/src/interpreter/interpreter.ts
```diff
@@ -21,6 +21,7 @@ export class Interpreter {
    */
   readonly stack: TEALStack;
   tealVersion: number; // LogicSigVersion
+  lineToCost: { [key: number]: number }; // cost of each instruction by line
```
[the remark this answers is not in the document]
ratik21: yes, name is `lineToCost`. Will update comment as well

packages/runtime/src/interpreter/interpreter.ts
amityadav0: great

amityadav0 (approved):

PR #446, opened 2021-08-18
TEALv4: subroutines and loops

amityadav0:
## Proposed Changes

+ Add subroutines
+ Add loops


packages/runtime/src/interpreter/interpreter.ts
```diff
@@ -29,6 +29,10 @@ export class Interpreter {
   instructions: Operator[];
   instructionIndex: number;
   runtime: Runtime;
+  // The call stack is separate from the data stack. Only callsub and retsub manipulate it.
+  // It is used to provide sub routine functionality
+  callStack: Stack<number>;
+  labelMap: Map<string, number>; // label string mapped to their respective indexes in instructions array
```
ratik21, line 35: is "respective indexes" same as `line no.` in TEAL file?

packages/runtime/src/interpreter/interpreter.ts
```diff
@@ -146,6 +153,30 @@ export class Interpreter {
...
+  jumpToLabel (label: string, line: number): void {
+    const toInstructionIndex = this.labelMap.get(label);
+    if (toInstructionIndex === undefined) {
+      throw new RuntimeError(RUNTIME_ERRORS.TEAL.LABEL_NOT_FOUND, {
+        label: label,
+        line: line
+      });
+    }
+    let currentIndex = toInstructionIndex;
+    // if next immediate op is also label, then keep continuing, otherwise return
```
ratik21, line 170: So if we have something like:
```teal
loop:
b1:
b2:
int 2
int 3
+ 
..
// jump to loop (10 times)
```
Then it will skip b1 & b2 and set `instructionIndex` to `int 2` ? Is this verified on node?

packages/runtime/src/interpreter/opcode-list.ts
```diff
@@ -1477,7 +1485,11 @@ export class BranchIfNotZero extends Op {
     const last = this.assertBigInt(stack.pop(), this.line);
 
     if (last !== 0n) {
-      this.interpreter.jumpForward(this.label, this.line);
+      if (this.interpreter.tealVersion <= 3) {
+        this.interpreter.jumpForward(this.label, this.line);
+      } else {
+        this.interpreter.jumpToLabel(this.label, this.line);
+      }
```
ratik21, line 1492: We have this if else 3 times. Shall we move this check to `jumpForward`/ `jumpToLabel` ? Eg.
```js
function jumpForward(..) {
 if (this.tealVersion > 3) { this.jumpToLabel(label, line); }
 ..
}
```
OR, add move this to say `interpreter.jump(label, line)` ?

packages/runtime/src/interpreter/interpreter.ts
amityadav0: If that's the case in instructions array, This array is used for internal mapping(used to jump to specific label instruction)

packages/runtime/src/interpreter/interpreter.ts
amityadav0: let's add this case in tests

packages/runtime/src/interpreter/opcode-list.ts
amityadav0: this was discussed in friday call, now we will be adding different class for these type of opcodes, for ex: b3, b4. b4 will inherit b3 if needed.

robert-zaremba: How about having the name convention: `Bv3`, `Bv4` ...

PR #447, opened 2021-08-18
Tealv4 byteslice arithmatic: implement b+ opcode

ratik21:
Closes #x

## Proposed Changes

+ add parsing functions (`bigintToBigEndianBytes`, `bigEndianBytesToBigInt`) with tests
+ implement `b+` op

### DIFF
--- a/packages/runtime/src/lib/constants.ts
+++ b/packages/runtime/src/lib/constants.ts
@@ -29,6 +29,11 @@ export const MaxAppProgramLen = 1024;
 export const MAX_ALGORAND_ACCOUNT_ASSETS = 1000;
 export const MAX_ALGORAND_ACCOUNT_APPS = 10;
 
+// for byteslice arithmetic ops, inputs are limited to 64 bytes,
+// but ouput can be upto 128 bytes (eg. when using b+ OR b*)
+export const MAX_INPUT_BYTE_LEN = 64;
+export const MAX_OUTPUT_BYTE_LEN = 128;
+
 const zeroAddress = new Uint8Array(32);
 const zeroUint64 = 0n;
 const zeroByte = new Uint8Array(0);
--- a/packages/runtime/src/lib/parsing.ts
+++ b/packages/runtime/src/lib/parsing.ts
@@ -171,3 +171,54 @@ export function parseBinaryStrToBigInt (binary: string[] | string): bigint {
   }
   return res;
 }
+
+/**
+ * Parses bigint to big endian bytes (represeted as Uint8array)
+ * NOTE: This is different from decodeUint64, encodeUint64 as it is capable of
+ * handling bigint > 64 bit (8 bytes).
+ * @param b value in bigint to parse
+ */
+export function bigintToBigEndianBytes (b: bigint): Uint8Array {
+  // `toString(base)` works
+  let hex = BigInt(b).toString(16);
+
+  // But it still follows the old behavior of giving
+  // invalid hex strings (due to missing padding),
+  // but we can easily add that back
+  if (hex.length % 2) { hex = '0' + hex; }
+
+  // The byteLength will be half of the hex string length
+  const len = hex.length / 2;
+  const u8 = new Uint8Array(len);
+
+  // And then we can iterate each element by one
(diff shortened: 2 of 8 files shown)


packages/runtime/src/lib/constants.ts
```diff
@@ -29,6 +29,11 @@ export const MaxAppProgramLen = 1024;
 export const MAX_ALGORAND_ACCOUNT_ASSETS = 1000;
 export const MAX_ALGORAND_ACCOUNT_APPS = 10;
 
+// for byteslice arithmetic ops, inputs are limited to 64 bytes,
+// but ouput can be upto 128 bytes (eg. when using b+ OR b*)
+export const MAX_INPUT_BYTE_LEN = 64;
+export const MAX_OUTPUT_BYTE_LEN = 128;
```
robert-zaremba, line 35: I think we should rename BYTE -> BIT. Shall we have `MAX_BIT_LEN` and `MAX_BIT_PLUS_LEN`? It seams that some operations for output will also return 64 bits.

packages/runtime/src/lib/parsing.ts
```diff
@@ -171,3 +171,54 @@ export function parseBinaryStrToBigInt (binary: string[] | string): bigint {
...
+  const u8 = Uint8Array.from(bytes);
+
+  u8.forEach(function (i) {
+    let h = i.toString(16);
+    if (h.length % 2) { h = '0' + h; }
+    hex.push(h);
+  });
+
+  return BigInt('0x' + hex.join(''));
+}
```
robert-zaremba, line 224: We can remove the convertion functions. `BigInt` already has that:
```
> BigInt('0x1')
1n
> BigInt('0xa')
10n
> BigInt('0x12')
18n
```
and for going from big int to hex:
```
> b = BigInt('0x12')
> b.toString(16)
'12'
```

If we need the `0x` prefix then we can have this helper functions to add `0x` prefix.

robert-zaremba (changes requested):

packages/runtime/src/lib/constants.ts
ratik21: `64` & `128` represent bytes here (also each value in a `uint8array` represents 1byte/8 bits). So changing to bit would look like:
```js
export const MAX_BIT_LEN = 64 * 8;
export const MAX_BIT_PLUS_LEN = 128 * 8;
```
Let me know if i should still change it

packages/runtime/src/lib/parsing.ts
ratik21: I think we need to convert bigint -> hex & vice verca, for conversion from/to `bigEndian` bytes. source: [https://coolaj86.com/articles/convert-js-bigints-to-typedarrays/](https://coolaj86.com/articles/convert-js-bigints-to-typedarrays/).
I will add separate helper functions for `hex`.

packages/runtime/src/lib/constants.ts
robert-zaremba: but this is used for numbers, right? I don't think Algorand support `64*8` bit number. This is huge and more then anyone could imagine.

ratik21: This is used for byte slice arithmatic. This is not natively on numbers. 
For eg.
```
byte "A"
byte "B"
b+ // "A" + "B" 
```
So we convert "A", "B" => uint (using `bigEndianBytesToBigInt`), then add them, and convert result back to byte. 
For bytes, the max limit is 64 (for input) & 128 (for output). But it can be handled by bigint (yes these would be veryy large numbers). I have added tests for it.

robert-zaremba: let's put this link next to the limits above: https://github.com/algorand/go-algorand/blob/bd5a00092c8a63dba8314b97851e46ff247cf7c1/data/transactions/logic/eval.go#L1302

robert-zaremba (approved):

packages/runtime/src/lib/constants.ts
ratik21: ok

PR #453, opened 2021-08-31
Update buyTx function/helper functions

amityadav0:
## Proposed Changes

+ https://www.pivotaltracker.com/story/show/179336759
+ Add bond example to workflow


.github/workflows/examples.yaml
```diff
@@ -63,6 +63,12 @@ jobs:
             pipenv run yarn algob run scripts/transfer/master-fund-john.js
             pipenv run yarn algob run scripts/transfer/tesla-to-john.js
 
+      - name: Example bond token
+      working-directory: ./examples/bond
+      run: |
+        pipenv run yarn algob deploy
+        pipenv run yarn algob run scripts/run/run.js
```
robert-zaremba, line 70: shouldn't we run tests as well?

examples/bond/scripts/run/common/common.js
```diff
@@ -119,6 +120,42 @@ function buyTx (buyer, issuerLsig, amount, algoAmount, appID, bondID) {
...
+ * Returns buy transaction for runtime
+ * @param runtime runtime object
+ * @param buyer buyer account
+ * @param issuerLsig Bond issuer logic signature
+ * @param algoAmount amount of Algo to pay for the bonds
+ * @param appID Bond application index
+ * @param bondID Bond ASA index
+ */
+function buyTxRuntime (runtime, buyer, issuerLsig, algoAmount, appID, bondID) {
+  const bondPrice = Number(runtime.getGlobalState(appID, 'issue_price'));
```
robert-zaremba, line 155: it would be nice to be able to use same code to read a global record.

examples/bond/scripts/run/common/common.js
```diff
@@ -119,6 +120,42 @@ function buyTx (buyer, issuerLsig, amount, algoAmount, appID, bondID) {
...
+    const key = Buffer.from(g.key, 'base64').toString();
+    if (key === 'issue_price') {
+      bondPrice = g.value.uint;
+    }
+  }
+  return buyTx(buyer, issuerLsig, algoAmount / bondPrice, algoAmount, appID, bondID);
+}
+
+/**
+ * Returns buy transaction for runtime
```
robert-zaremba, line 146:
```suggestion
 * Returns buy transaction for test/runtime
```

examples/bond/scripts/run/common/common.js
```diff
@@ -119,6 +120,42 @@ function buyTx (buyer, issuerLsig, amount, algoAmount, appID, bondID) {
   ];
 }
 
+/**
+ * Returns buy transaction for algorand node
```
robert-zaremba, line 124:
```suggestion
 * Returns buy transaction using algorand node
```

robert-zaremba (approved):
left few 2 suggestions.

.github/workflows/examples.yaml
amityadav0: yes, we are already running tests from `main.yaml`.

examples/bond/scripts/run/common/common.js
amityadav0: okay, i think we can traverse results of `readGlobalStateSSC` at the time when function is called, and instead of returning same response as algorand, we will change it to map, so user will be able to access node global state like this.

amityadav0: what do you think ? should i update `readGlobalStateSSC` as above ?

robert-zaremba: Yes, let's update `readGlobalStateSSC`

PR #454, opened 2021-08-31
Upgrade error in catch

amityadav0:
https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/#use-unknown-catch-variables

- Previoulsy default error was any, in new update it defaults to unknown

### DIFF
--- a/packages/algob/src/internal/core/params/argument-types.ts
+++ b/packages/algob/src/internal/core/params/argument-types.ts
@@ -212,7 +212,7 @@ export const inputFile: ArgumentType<string> = {
         // tslint:disable-next-line only-algob-error
         throw new Error(`${strValue} is a directory, not a file`);
       }
-    } catch (error) {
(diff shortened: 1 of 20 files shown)


packages/algob/src/internal/core/params/argument-types.ts
```diff
@@ -239,15 +241,17 @@ export const inputFile: ArgumentType<string> = {
...
+        throw new BuilderError(
+          ERRORS.ARGUMENTS.INVALID_VALUE_FOR_TYPE,
+          {
+            value,
+            name: argName,
+            type: inputFile.name
+          },
+          error
+        );
+      }
```
robert-zaremba, line 254: we should have `else` and thrown an `UnknownError(error)`

robert-zaremba (approved):
in all catch places - we should not ignore an error if by mistake it's not a BuilderError

PR #455, opened 2021-09-02
Add fund account function in runtime

amityadav0:
## Proposed Changes

+ Add fundAccount function in runtime
+ Update all examples

### DIFF
--- a/examples/bond/test/bond-token-flow.js
+++ b/examples/bond/test/bond-token-flow.js
@@ -115,15 +115,7 @@ describe('Bond token Tests', function () {
     console.log('Issuer Address: ', issuerLsigAddress);
 
     // fund escrow with some minimum balance first
-    const fundEscrowParam = {
-      type: types.TransactionType.TransferAlgo,
-      sign: types.SignType.SecretKey,
-      fromAccount: master.account,
-      toAccountAddr: issuerLsigAddress,
-      amountMicroAlgos: minBalance + 10000,
-      payFlags: {}
-    };
-    runtime.executeTx(fundEscrowParam);
+    runtime.fundAccount(master.account, issuerLsigAddress, minBalance + 10000);
 
     // verify global state
     assert.isDefined(applicationId);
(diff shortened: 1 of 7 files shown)


examples/bond/test/bond-token-flow.js
```diff
@@ -115,15 +115,7 @@ describe('Bond token Tests', function () {
...
-    const fundEscrowParam = {
-      type: types.TransactionType.TransferAlgo,
-      sign: types.SignType.SecretKey,
-      fromAccount: master.account,
-      toAccountAddr: issuerLsigAddress,
-      amountMicroAlgos: minBalance + 10000,
-      payFlags: {}
-    };
-    runtime.executeTx(fundEscrowParam);
+    runtime.fundAccount(master.account, issuerLsigAddress, minBalance + 10000);
```
ratik21, line 118: How abouot renaming `fundAccount` to `fundLsig`? It will be same as deployer API, and we update all instances in  examples  where we're funding an escrow contract. what do you think?

packages/runtime/src/runtime.ts
```diff
@@ -572,6 +573,24 @@ export class Runtime {
     return lsig;
   }
 
+  /**
+   * Transfers `amount` of microAlgos from `from` address to `to` address
+   * @param from From account
+   * @param to to address
+   * @param amount amount of algo in microalgos
+   */
+  fundAccount (from: RuntimeAccount, to: string, amount: number): void {
```
ratik21, line 582:
```suggestion
  fundAccount (from: RuntimeAccount, to: AccountAddress, amount: number): void {
```
I think we have a type `AccountAddress = string`

examples/bond/test/bond-token-flow.js
amityadav0: i thought so, but fundAccount is more general, user can use it to fund accounts/lsig accounts. I don't have preference in both cases. what do you think?

ratik21: I would prefer to keep the name as `fundLsig`. In algob scripts user uses `executeTransaction.TransferAlgo` to fund other accounts, and will use `deployer.fundLsig` for escrow contracts

amityadav0: ok will remove more generalised name.

ratik21 (approved):

PR #458, opened 2021-09-08
DAO

ratik21:
## Proposed Changes

Implement DAO example.

## Potential followups

Tests


examples/dao/assets/dao-fund-lsig.py
```diff
@@ -0,0 +1,42 @@
...
+    """
+
+    # check no rekeying, close remainder to, asset close to for a txn
+    def basic_checks(txn: Txn): return And(
+        txn.rekey_to() == Global.zero_address(),
+        txn.close_remainder_to() == Global.zero_address(),
+        txn.asset_close_to() == Global.zero_address()
+    )
+
+    # ideally we should retrict the sender here(so that anyone cannot withdraw funds)
```
robert-zaremba, line 15:
```suggestion
    # ideally we should restrict the sender here (no one can withdraw funds)
```

examples/dao/assets/dao-fund-lsig.py
```diff
@@ -0,0 +1,42 @@
...
+    receive_or_send = basic_checks(Txn)
+
+    # verify funds are transfered only when paired with DAO app (during execute call)
+    payment = And(
+        # verify first transaction
+        basic_checks(Gtxn[0]),
+        Gtxn[0].type_enum() == TxnType.ApplicationCall,
+        Gtxn[0].application_id() == Tmpl.Int("TMPL_DAO_APP_ID"),
```
robert-zaremba, line 23: how about using function arguments rather than PyTEAL tmpl?

examples/dao/assets/dao-fund-lsig.py
```diff
@@ -0,0 +1,42 @@
...
+        Or(
+            Gtxn[1].type_enum() == TxnType.AssetTransfer,
+            Gtxn[1].type_enum() == TxnType.Payment,
+        )
+    )
+
+    program = program = Cond(
+        [Global.group_size() == Int(1), receive_or_send],
+        [Global.group_size() == Int(2), payment]
+    )
```
robert-zaremba, line 37: we need to restrict the access to the account:
* anyone can send to the dao-fund
* anyone can opt-in 
* withdraw (payment) - as below
* other transactions must be forbidden

examples/dao/assets/proposal-lsig.py
```diff
@@ -0,0 +1,11 @@
+from pyteal import *
+
+def proposal_lsig():
+    """
+    Represents Proposal Lsig (a contract account owned by the proposer)
+    """
```
robert-zaremba, line 6: we should have conditions here. What about:
1. algo and asset withdraw must be done by "owner" (template value)
2. opt-in only by the owner
3. only owner can close reminder to
4. ... maybe more?

This way we assure only owner can manage the assets of the lsig.

examples/dao/scripts/deploy.js
```diff
@@ -0,0 +1,100 @@
...
+  console.log('* ASA distribution (Gov tokens) *');
+  await Promise.all([
+    deployer.optInLsigToASA(asaInfo.assetIndex, depositLsig, { totalFee: 1000 }),
+    deployer.optInLsigToASA(asaInfo.assetIndex, daoFundLsig, { totalFee: 1000 }),
+    deployer.optInAcountToASA(asaInfo.assetIndex, proposer.name, {}),
+    deployer.optInAcountToASA(asaInfo.assetIndex, voterA.name, {}),
+    deployer.optInAcountToASA(asaInfo.assetIndex, voterB.name, {})
+  ]);
+
+  const fundASAParams = {
```
robert-zaremba, line 83:
```suggestion
  const distributeGovTokenParams = {
```

examples/dao/scripts/run/add_proposal.js
```diff
@@ -0,0 +1,68 @@
...
+  const daoAppInfo = deployer.getApp('dao-app-approval.py', 'dao-app-clear.py');
+  const proposalLsig = await deployer.loadLogic('proposal-lsig.py');
+  try {
+    await deployer.optInLsigToApp(daoAppInfo.appID, proposalLsig, {}, {});
+  } catch (e) {
+    console.log(e.message);
+  }
+
+  const daoFundLsig = await getDAOFundLsig(deployer);
+  const appArgs = [
```
robert-zaremba, line 22:
```suggestion
  const daoParams = [
```

examples/dao/scripts/run/add_proposal.js
```diff
@@ -0,0 +1,68 @@
...
+    `int:${now + (5 * 60)}`, // execute_before (now + 5min)
+    `int:${ProposalType.ALGO_TRANSFER}`, // type
+    `addr:${daoFundLsig.address()}`, // from (DAO treasury)
+    `addr:${proposer.addr}`, // recepient
+    `int:${2e6}` // amount (in microalgos)
+  ];
+
+  const depositLsig = await getDepositLsig(deployer);
+  const govToken = deployer.asa.get('gov-token');
+  const groupTx = [
```
robert-zaremba, line 39:
```suggestion
  const addProposalTx = [
```

examples/dao/scripts/run/clear_vote_record.js
```diff
@@ -0,0 +1,31 @@
+const { executeTx } = require('./common/common.js');
+const { types } = require('@algo-builder/web');
+const { accounts } = require('./common/accounts.js');
+
+async function clearVoteRecord (deployer, voterAcc, proposalLsig) {
+  const daoAppInfo = deployer.getApp('dao-app-approval.py', 'dao-app-clear.py');
+
+  console.log(`* Clearing vote record of ${voterAcc.addr} from proposal ${proposalLsig.address()} *`);
+  const withdrawVoteParam = {
```
robert-zaremba, line 9:
```suggestion
  const clearVoteParams = {
```

examples/dao/scripts/run/clear_vote_record.js
```diff
@@ -0,0 +1,31 @@
...
+    type: types.TransactionType.CallApp,
+    sign: types.SignType.SecretKey,
+    fromAccount: voterAcc,
+    appID: daoAppInfo.appID,
+    payFlags: { totalFee: 1000 },
+    appArgs: ['str:clear_vote_record'],
+    accounts: [proposalLsig.address()]
+  };
+
+  await executeTx(deployer, withdrawVoteParam);
```
robert-zaremba, line 19:
```suggestion
  await executeTx(deployer, clearVoteParams);
```

examples/dao/scripts/run/clear_vote_record.js
```diff
@@ -0,0 +1,31 @@
+const { executeTx } = require('./common/common.js');
+const { types } = require('@algo-builder/web');
+const { accounts } = require('./common/accounts.js');
+
+async function clearVoteRecord (deployer, voterAcc, proposalLsig) {
```
robert-zaremba, line 5: how about renaming `proposalLsig` to `proposalAddr`? In other places, only in scripts, as well (but only when we use it as an address, not the TEAL).

examples/dao/scripts/run/common/common.js
```diff
@@ -0,0 +1,61 @@
...
+    console.error('Transaction Failed', e.response ? e.response.error : e);
+  }
+};
+
+/**
+ * Fund accounts from master with 20 Algos
+ * @param {*} deployer algobDeployer
+ * @param {*} account or list of accounts to fund
+ */
+async function fundAccount (deployer, account) {
```
robert-zaremba, line 19: Please rename `account` to `accounts`

examples/dao/scripts/run/execute.js
```diff
@@ -0,0 +1,43 @@
+const { executeTx } = require('./common/common.js');
+const { types } = require('@algo-builder/web');
+const { accounts, getDAOFundLsig } = require('./common/accounts.js');
+
+async function execute (deployer, account, proposalLsig) {
```
robert-zaremba, line 5:
```suggestion
// Executes a proposal
async function execute (deployer, account, proposalLsig) {
```

robert-zaremba (changes requested):
Fore each "use-case" in the TEAL code:
* please list transactions (eg tx0:... tx1: ...) - no need to explain them, we have it in spec. Just list.
* let's recheck all lsig contracts.

examples/asa/assets/teal/4-gold-asa.teal
```diff
@@ -1,5 +1,5 @@
 #pragma version 2
-// Accepts only if (transaction type is OPT-IN OR (transaction type is asset transfer, 
+// Accepts only if (transaction type is OPT-IN OR (transaction type is asset transfer,
```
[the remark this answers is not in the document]
ratik21: OK

examples/dao/assets/dao-app-approval.py
```diff
@@ -0,0 +1,449 @@
...
+    asa_id = App.localGet(Int(1), Bytes("asa_id"))
+    recipient = App.localGet(Int(1), Bytes("recipient"))
+    amount = App.localGet(Int(1), Bytes("amount"))
+
+    # Computes result of proposal. Saves result in scratch. Args:
+    # * idx - index of account where proposal_lsig.address() is passed
+    # NOTE: idx == Int(0) means proposalLsig is Txn.sender()
+    def compute_result(idx: Int):
+        return Seq([
+            Cond(
```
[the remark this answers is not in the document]
ratik21: i thought so, but here we're computing a result against latest time (`Global.latest_timestamp()`). So result could be different for different block timestamps.

examples/dao/assets/dao-app-approval.py
```diff
@@ -0,0 +1,449 @@
...
+                            App.localGet(idx, Bytes("yes")) <= App.localGet(idx, Bytes("no"))
+                        )
+                    ) == Int(1),
+                    scratchvar_result.store(Int(2))
+                ],
+                # 3 if voting is still in progress (now <= voting_end)
+                [Global.latest_timestamp() <= App.localGet(idx, Bytes("voting_end")), scratchvar_result.store(Int(3))],
+                # 4 if proposal expired (now > proposal.execute_before())
+                [Global.latest_timestamp() > App.localGet(idx, Bytes("execute_before")), scratchvar_result.store(Int(4))],
+            ),
```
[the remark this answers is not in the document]
ratik21: OK, thanks

examples/dao/assets/dao-app-approval.py
```diff
@@ -0,0 +1,449 @@
...
+            Gtxn[tx_index].xfer_asset() == Tmpl.Int("TMPL_GOV_TOKEN"),
+            Gtxn[tx_index].asset_receiver() == receiver,
+            Gtxn[tx_index].asset_amount() >= Int(0)
+        ))
+
+    # scratch vars to save result() & is_proposal_active()
+    scratchvar_result = ScratchVar(TealType.uint64)
+    scratchvar_is_proposal_active = ScratchVar(TealType.uint64)
+
+    # Fetch key(s) from Txn.Accounts[1] (passed as proposalLsig), for current app
```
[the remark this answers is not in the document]
ratik21: sender is the proposer account (owner of `proposal_lsig`), and passes the lsig as `Txn.Accounts[1]`.

examples/dao/assets/dao-app-approval.py
```diff
@@ -0,0 +1,449 @@
...
+        App.localPut(Int(0), Bytes("voting_end"), Btoi(Txn.application_args[6])),
+        # execute_before must be after voting_end
+        Assert(Btoi(Txn.application_args[7]) > Btoi(Txn.application_args[6])),
+        App.localPut(Int(0), Bytes("execute_before"), Btoi(Txn.application_args[7])),
+        # type must be 1, 2 OR 3
+        Assert(
+            Or(
+                Btoi(Txn.application_args[8]) == Int(1),
+                Btoi(Txn.application_args[8]) == Int(2),
+                Btoi(Txn.application_args[8]) == Int(3)
```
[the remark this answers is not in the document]
ratik21: good idea

examples/dao/assets/dao-app-approval.py
```diff
@@ -0,0 +1,449 @@
...
+        Return(Int(1))
+    ])
+
+    # sender.deposit
+    deposit = App.localGet(Int(0), Bytes("deposit"))
+
+    # Records votes deposited by user by updating sender.deposit
+    deposit_vote = Seq([
+        # Verify deposit of votes to deposit_lsig account (with atleast 1 vote)
+        # Ques: do we care if sender is voter here?
```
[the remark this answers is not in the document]
ratik21: yes, clarified on spec. forgot to remove from here.

examples/dao/assets/dao-app-approval.py
```diff
@@ -0,0 +1,449 @@
...
+        is_proposal_active(Int(1)),
+        Assert(Global.group_size() == Int(1)),
+        # fail if proposal is active (can’t remove proposal record of an active proposal)
+        If(
+            p_proposal.hasValue() == Int(1),
+            If(
+                And(
+                    p_proposal.value() == proposal_id,
+                    scratchvar_is_proposal_active.load() == Int(1)
+                ) == Int(1),
```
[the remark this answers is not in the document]
ratik21: so this is tricky, here we're only moving forward if "p_propsal is set" (i.e `p_proposal.hasValue() == Int(1)`), otherwise we don't. 
Because if we don't do it this way, and value does not exist, then `p_proposal.value()` will return Int(0) and `proposal_id` will return `Bytes`. This will throw a pyTEAL error.

examples/dao/scripts/run/clear_vote_record.js
ratik21: > but only when we use it as an address

yes, let's rename to proposalAddr in these cases.

examples/dao/assets/dao-app-approval.py
```diff
@@ -0,0 +1,449 @@
...
+    """
+
+    # Verfies deposit of gov_token to an address for a transaction. Args:
+    # * tx_index (index of deposit transaction)
+    # * receiver (receiver of gov_token)
+    def verify_deposit(tx_index: Int, receiver: Addr):
+        return Assert(And(
+            Global.group_size() >= tx_index,
+            Gtxn[tx_index].type_enum() == TxnType.AssetTransfer,
+            Gtxn[tx_index].xfer_asset() == Tmpl.Int("TMPL_GOV_TOKEN"),
```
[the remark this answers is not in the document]
ratik21: yes, i prefer that as well

examples/dao/assets/dao-fund-lsig.py
ratik21: done

examples/dao/assets/dao-fund-lsig.py
ratik21: We're following all these points (except that anyone can withdraw funds). We can add a hardcoded address for withdrawal.

examples/dao/assets/proposal-lsig.py
ratik21: Ok, i didn't do it before because we don't control this account (it belongs to the proposer). We can provide a template with these checks, and recommend to use it.

robert-zaremba: yes, that's the idea. We also need to document / explain it.

examples/dao/assets/dao-fund-lsig.py
robert-zaremba: For withdraw:
* we don't restrict recipient (transfer can go anywhere)
* but the transaction must be in a group transaction with the `dao_app execute` call.

examples/dao/assets/dao-app-approval.py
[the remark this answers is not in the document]
robert-zaremba: could you write in a comment that sender = proposer?

examples/dao/assets/dao-fund-lsig.py
```diff
@@ -0,0 +1,54 @@
...
+    """
+
+    # check no rekeying, close remainder to, asset close to for a txn
+    def basic_checks(txn: Txn): return And(
+        txn.rekey_to() == Global.zero_address(),
+        txn.close_remainder_to() == Global.zero_address(),
+        txn.asset_close_to() == Global.zero_address()
+    )
+
+    # ideally we should restrict the sender here (no one can withdraw funds)
```
amityadav0, line 19: currently anyone with this DAO address can use it's funds, shouldn't we block it ?

examples/dao/assets/dao-fund-lsig.py
```diff
@@ -0,0 +1,54 @@
...
+        basic_checks(Gtxn[1]),
+        Or(
+            Gtxn[1].type_enum() == TxnType.AssetTransfer,
+            Gtxn[1].type_enum() == TxnType.Payment,
+        )
+    )
+
+    program = program = Cond(
+        [Global.group_size() == Int(1), receive_or_send],
+        [Global.group_size() == Int(2), payment]
```
amityadav0, line 40: could you please add a check here for opt-in ? so that malicious user cannot spam proposal lsig with ASA opt-in transactions ?

examples/dao/assets/dao-fund-lsig.py
```diff
@@ -0,0 +1,54 @@
+import sys
+sys.path.insert(0,'..')
+
+from algobpy.parse import parse_params
+from pyteal import *
+
+def proposal_lsig(ARG_DAO_APP_ID):
```
amityadav0, line 7: can we rename this function ?

amityadav0: maybe `treasury_lsig`

examples/dao/assets/proposal-lsig.py
```diff
@@ -0,0 +1,56 @@
...
+
+from algobpy.parse import parse_params
+from pyteal import *
+
+def proposal_lsig(ARG_OWNER):
+    """
+    Represents Proposal Lsig (a contract account owned by the proposer)
+    """
+
+    # check no rekeying, close remainder to, asset close to for a txn
```
amityadav0, line 12: below we are checking for `rekey_to` only.

examples/dao/assets/proposal-lsig.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def basic_checks(txn: Txn): return And(
+        txn.rekey_to() == Global.zero_address(),
+        txn.close_remainder_to() == Global.zero_address(),
+        txn.asset_close_to() == Global.zero_address()
+    )
+
+    # Opt-in transaction.
+    opt_in = And(
+        basic_owner_checks(Txn),
```
amityadav0, line 28: i think while opting in `txn.sender` will be proposal lsig address, not owner address. so this check may fail here. could yo u please check ?

examples/dao/assets/dao-app-approval.py
```diff
@@ -0,0 +1,470 @@
...
+from pyteal import *
+
+def approval_program(ARG_GOV_TOKEN):
+    """
+    A stateful app with governance rules. Stores
+    deposit, min_support, min_duration, max_duration, url.
+
+    Commands:
+        add_proposal            Save proposal record in lsig
+        deposit_vote_token            records deposited votes in voter.account
```
amityadav0, line 14:
```suggestion
        deposit_vote_token       records deposited votes in voter.account
```

examples/dao/assets/dao-fund-lsig.py
robert-zaremba: yes, we are restricting it. The comment should be updated.

examples/dao/assets/dao-fund-lsig.py
robert-zaremba: we were thinking about it. There can be many DAO funds. So in essence it is not a problem. So anyone can create a DAO.
Maybe we can restrict it and add a random owner and explain that the owner is only there to control what assets to opt-in, but we can use any fund lsig with DAO.

examples/dao/assets/dao-fund-lsig.py
robert-zaremba: or `dao_fund_lsig` ;)

examples/dao/assets/dao-app-approval.py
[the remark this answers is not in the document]
ratik21: sure

examples/dao/README.md
```diff
@@ -0,0 +1,160 @@
...
+
+
+	yarn run algob run scripts/run/clear_vote_record.js
+
+
+
+To clear proposal record (from proposal_lsig as a sender), fails if the proposal is still in progress:
+
+
+
```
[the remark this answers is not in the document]
ratik21: removed

examples/dao/assets/dao-fund-lsig.py
ratik21: Missed this one. Renamed to `dao_fund_lsig` (it's the file name as well)

examples/dao/assets/dao-fund-lsig.py
ratik21: got it, so we're doing this as well (in the `payment` branch), where we ensure funds are only transferred when paired with the `execute` call.

examples/dao/assets/dao-fund-lsig.py
ratik21: done. [https://github.com/scale-it/algo-builder/pull/458#discussion_r710951355](https://github.com/scale-it/algo-builder/pull/458#discussion_r710951355)

examples/dao/assets/proposal-lsig.py
ratik21: updated

examples/dao/assets/proposal-lsig.py
ratik21: yes, removed

examples/dao/assets/dao-app-approval.py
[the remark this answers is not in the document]
robert-zaremba: As far as I know scratch space is reset after the execution. Could you validate?

examples/dao/assets/dao-app-approval.py
```diff
@@ -0,0 +1,471 @@
...
+            ),
+            # if proposal is not expired, not in progess, and not passed, then reject (set result == Int(2))
+            If(
+                And(
+                    scratchvar_result.load() != Int(4), # expired
+                    scratchvar_result.load() != Int(3), # in progess
+                    scratchvar_result.load() != Int(1), # passed
+                ),
+                scratchvar_result.store(Int(2))
+            )
```
robert-zaremba, line 88: instead of having the `If`, maybe we can add put into conditions (at the end):
```
[true, scratchvar_result.store(Int(1))]
```

examples/dao/assets/dao-app-approval.py
```diff
@@ -0,0 +1,471 @@
...
+        App.localPut(Int(0), Bytes("voting_end"), scratchvar_voting_end.load()),
+        # execute_before must be after voting_end
+        Assert(scratchvar_execute_before.load() > scratchvar_voting_end.load()),
+        App.localPut(Int(0), Bytes("execute_before"), scratchvar_execute_before.load()),
+        # type must be 1, 2 OR 3
+        Assert(
+            Or(
+                scratchvar_proposal_type.load() == Int(1),
+                scratchvar_proposal_type.load() == Int(2),
+                scratchvar_proposal_type.load() == Int(3)
```
robert-zaremba, line 203: do you know if this is an optimal way? loading 3 times doesn't create lot of TEAL code? Ideally we should load once and check top of the stack against 3 values.

examples/dao/assets/dao-app-approval.py
```diff
@@ -0,0 +1,471 @@
...
+            )
+        ),
+        If(
+            p_proposal.hasValue() == Int(0),
+            # If Sender.p_<proposal> is not set then set p_<proposal> := proposal.id
+            App.localPut(Int(0), byte_p_proposal, proposal_id),
+            # if Sender.p_<proposal> != proposal.id then overwrite by setting the new proposal.id, fail otherwise
+            If(p_proposal.value() != proposal_id,
+                App.localPut(Int(0), byte_p_proposal, proposal_id),
+                Err()),
```
robert-zaremba, line 288:
```suggestion
                Err()),  # double vote
```

examples/dao/assets/dao-app-approval.py
```diff
@@ -0,0 +1,471 @@
...
+                App.localGet(Int(0), Bytes("deposit")) > Int(0)
+            )
+        ),
+        If(
+            p_proposal.hasValue() == Int(0),
+            # If Sender.p_<proposal> is not set then set p_<proposal> := proposal.id
+            App.localPut(Int(0), byte_p_proposal, proposal_id),
+            # if Sender.p_<proposal> != proposal.id then overwrite by setting the new proposal.id, fail otherwise
+            If(p_proposal.value() != proposal_id,
+                App.localPut(Int(0), byte_p_proposal, proposal_id),
```
robert-zaremba, line 287: this should be only possible if the existing proposal is over. Otherwise we can double  vote by creating two proposals and alternate voting between these proposals.

examples/dao/assets/dao-app-approval.py
```diff
@@ -0,0 +1,471 @@
...
+    # p_<proposal> is a concatenation of p_ with the proposal address to avoid some weird attacks.
+    byte_p_proposal = Concat(Bytes("p_"), Txn.accounts[1])
+    p_proposal = App.localGetEx(Int(0), Int(0), byte_p_proposal)  # value = proposal.id when a user voted or 0
+    yes = Bytes("yes")
+    no = Bytes("no")
+    abstain = Bytes("abstain")
+
+    # Register user votes in proposal_lsig by saving Sender.p_<proposal>. Arguments:
+    # * proposal : lsig account address with the proposal record (provided as the first external account).
+    # * vote (bytes): abstain, yes, no
```
robert-zaremba, line 268:
```suggestion
    # Register user votes in proposal_lsig by saving Sender.p_<proposal>.
    # * External Accounts: `proposal` : lsig account address with the proposal record (provided as the first external account).
    # * Call arguments: `vote` (bytes): { abstain, yes, no} 
```

examples/dao/assets/dao-app-approval.py
```diff
@@ -0,0 +1,471 @@
...
+    register_vote = Seq([
+        p_proposal,
+        Assert(
+            And(
+                Global.group_size() == Int(1),
+                # voting_start <= now <= voting_end
+                # voting_start <= Global.latest_timestamp(),
+                # Global.latest_timestamp() <= voting_end,
```
robert-zaremba, line 276: where do we have this checks? They are essential. Voting is only possible when `voting_start <= now <= voting_end`

examples/dao/assets/dao-app-approval.py
```diff
@@ -0,0 +1,471 @@
...
+        p_proposal,
+        scratch_proposal_active(Int(1)),
+        Assert(Global.group_size() == Int(1)),
+        # fail if proposal is active (can’t remove proposal record of an active proposal)
+        If(
+            p_proposal.hasValue() == Int(1),
+            If(
+                And(
+                    p_proposal.value() == proposal_id,
+                    scratchvar_proposal_active.load() == Int(1)
```
robert-zaremba, line 316: clearing a vote record should be only possible when the proposal status `== 3` (is active).
`1` means that voting is over and it passed.

examples/dao/assets/dao-app-approval.py
```diff
@@ -0,0 +1,471 @@
...
+                ) == Int(1),
```
robert-zaremba, line 317: Why not merging 2 `If` in one:
```
If ( And( p_proposal.value() == proposal_id
              status == 3)
     Err()
)
```
             ))
```

**Review by robert-zaremba — CHANGES_REQUESTED** [review]

File: examples/dao/assets/dao-app-approval.py (hunk 14 above)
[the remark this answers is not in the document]
Comment (ratik21) [reply examples/dao/assets/dao-app-approval.py:65]: sorry, i don't follow.
+ we're only using this function once per branch (for each execution), so it will only be filled once (you're right about scratch space is reset after the execution).
+ but even if this function was called multiple times (at a single branch), then we should compute result again (yes, scratch space will be filled, but we have moven forward in time, so result could be different).

Am i saying something wrong here?

File: examples/dao/assets/dao-app-approval.py (hunk 28 above)
Comment (ratik21) [reply examples/dao/assets/dao-app-approval.py:88]: I was putting it into conditions first, but an error is thrown that "we cannot load from scratch before store". Hence i put it after the conditions.

File: examples/dao/assets/dao-app-approval.py (hunk 29 above)
Comment (ratik21) [reply examples/dao/assets/dao-app-approval.py:203]: `scratch.load` loads a [copy](https://developer.algorand.org/docs/reference/teal/opcodes/#load-i) from scratch slot X to top of stack (cost of both `load X` & `store X` is 1). The other way would be to load once, create 2 duplicates (using `dup`), and then compare.
Current:
```teal
load X // loads a copy from scratch slot X
int 1
==
load X
int 2
==
||
load X
int 3
==
||
```
Using `dup`:
```
load X
dup
dup
int 1
==
int 2 
== 
||
int 3
==
||
```
both have same cost & length. Will update if you prefer the latter. Let me know

examples/dao/assets/dao-app-approval.py
ratik21: yes, i was having some issue with block timestamp in a development network. Will test & uncomment

examples/dao/assets/dao-app-approval.py
ratik21: Could you please check the spec: [https://paper.dropbox.com/doc/Algo-Builder-DAO--BSmq7oxrGfMChl9VhEhjwJO4Ag-ncLdytuFa7EJrRerIASSl](https://paper.dropbox.com/doc/Algo-Builder-DAO--BSmq7oxrGfMChl9VhEhjwJO4Ag-ncLdytuFa7EJrRerIASSl) ?

`scratch_proposal_active()` is implemented as per spec, where `1` means proposal is still in voting.

examples/dao/assets/dao-app-approval.py
ratik21: could you please [this](https://github.com/scale-it/algo-builder/pull/458#discussion_r709100082) comment? I explain why we do it this way.

examples/dao/assets/dao-app-approval.py
ratik21: from the spec (`vote(…)` section):
```txt
if Sender.p_<proposal> != proposal.id then overwrite by setting the new proposal.id (this means that the proposer did a new proposal and the previous one was executed or expired);
```
we reject `add_proposal` if proposal is already set. And an existing proposal can only be cleared by `clear_proposal` command (which asserts proposal is not active).

examples/dao/assets/dao-app-approval.py
robert-zaremba: ah, ok. Please add a comment to the code.

examples/dao/assets/dao-app-approval.py
robert-zaremba: OK

examples/dao/assets/dao-app-approval.py
robert-zaremba: OK, the first one is simpler to read I think. So you can leave.

examples/dao/assets/proposal-lsig.py
amityadav0: but we need to ensure opt-in can only be done by owner, so we can ensure that by checking if transaction is signed by owner.

ratik21: ok, adding check for sender

examples/dao/assets/dao-app-approval.py
robert-zaremba: ping - we need to add the voting_start and voting_end checks.

examples/dao/assets/dao-fund-lsig.py
```diff
@@ -0,0 +1,58 @@
...
+    # verify funds are transfered only when paired with DAO app (during execute call)
+    payment = And(
+        # verify first transaction
+        basic_checks(Gtxn[0]),
+        Gtxn[0].type_enum() == TxnType.ApplicationCall,
+        Gtxn[0].application_id() == Int(ARG_DAO_APP_ID),
+        Gtxn[0].application_args[0] == Bytes("execute"),
+
+        # verify second transaction (either payment in asa or ALGO)
+        basic_checks(Gtxn[1]),
```
robert-zaremba, line 28: shall we verify that _this_ transaction has index == 1?

robert-zaremba: ah we check it later that the group length == 2

examples/dao/assets/dao-app-approval.py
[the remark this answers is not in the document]
robert-zaremba: How possible that the result may change within the same transaction?

examples/dao/assets/dao-app-approval.py
robert-zaremba: strange, in `scratchvar_result.store` we shouldn't load a value - it only saves the value.

examples/dao/assets/dao-app-approval.py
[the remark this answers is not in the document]
ratik21: ah, yes. OK, i was trying to do it, i keep getting a python error (with slot assignment):
image
Screenshot from 2021-09-23 17-13-27

examples/dao/assets/dao-app-approval.py
ratik21: yeah, it's a pyTEAL error (when compiling pyTEAL -> TEAL).

examples/dao/assets/dao-app-approval.py
ratik21: done, i increased the buffer by 1min `{ voting start = now + 1min, voting start = now + 3min }`.

examples/dao/assets/proposal-lsig.py
```diff
@@ -0,0 +1,57 @@
...
+        txn.close_remainder_to() == Global.zero_address(),
+        txn.asset_close_to() == Global.zero_address()
+    )
+
+    # Opt-in transaction.
+    opt_in = And(
+        basic_checks(Txn),
+        Txn.type_enum() == TxnType.AssetTransfer,
+        Txn.asset_amount() == Int(0),
+        Txn.sender() == Addr(ARG_OWNER)
```
amityadav0, line 31: why are we checking `Txn.sender` to be `ARG_OWNER` here ?

ratik21: updating to `asset_sender`

examples/dao/README.md
```diff
@@ -0,0 +1,103 @@
...
+## Spec document
+
+Algo Builder DAO [specification](https://paper.dropbox.com/doc/Algo-Builder-DAO--BRlh~FwufNzIzk4wNUuAjLTuAg-ncLdytuFa7EJrRerIASSl).
+
+## Deploy script
+
+We created a deploy script in `scripts/deploy`, This script deploys initial Gov token, deploys DAO app, fund lsig's, saves deposit_lsig address to DAO app, and does initial distribution of ASA (Gov token).
+
+## Run scripts flow
```
robert-zaremba, line 73:
```suggestion
## Scripts
```

examples/dao/assets/deposit-lsig.py
```diff
@@ -0,0 +1,72 @@
...
+        Gtxn[0].application_id() == Int(ARG_DAO_APP_ID),
+        Or(
+            Gtxn[0].application_args[0] == app_arg_1,
+            Gtxn[0].application_args[0] == app_arg_2
+        ),
+
+        # verify second transaction
+        basic_checks(Gtxn[1]),
+        Gtxn[1].type_enum() == TxnType.AssetTransfer,
+        Gtxn[1].xfer_asset() == Int(ARG_GOV_TOKEN),
```
robert-zaremba, line 45: shall we check that `this_tx_index == 1`?

examples/dao/assets/deposit-lsig.py
```diff
@@ -0,0 +1,72 @@
...
+    def deposit_or_withdraw(app_arg_1: Bytes, app_arg_2: Bytes): return And(
+        Global.group_size() == Int(2),
+
+        # verify first transaction
+        basic_checks(Gtxn[0]),
+        Gtxn[0].type_enum() == TxnType.ApplicationCall,
+        Gtxn[0].application_id() == Int(ARG_DAO_APP_ID),
+        Or(
+            Gtxn[0].application_args[0] == app_arg_1,
+            Gtxn[0].application_args[0] == app_arg_2
```
robert-zaremba, line 39: Can we instead of `payment = Or(....`, can we have a bigger `Or` statement here, where we check if : 
```
Gtxn[0].application_args[0] is in {"add_proposal", "deposit_vote_token", "withdraw_vote_deposit", "clear_proposal")
```

examples/dao/assets/deposit-lsig.py
```diff
@@ -0,0 +1,72 @@
...
+        ),
+
```
robert-zaremba, line 41: let's add a comment that the transfer attributes (sender, recipient, amount) are checked in the app.

examples/dao/assets/proposal-lsig.py
```diff
@@ -0,0 +1,89 @@
...
+        Gtxn[1].type_enum() == TxnType.AssetTransfer,
+        Gtxn[1].asset_amount() == Int(0)
+    )
+
+    # Allow app call to DAO (eg. OptInToApp, add_proposal, etc)
+    allow_app_call = And(
+        Or(
+            Global.group_size() == Int(1),
+            Global.group_size() == Int(2)
+        ),
```
robert-zaremba, line 45:
```suggestion
        Global.group_size() <= Int(2),
```

examples/dao/scripts/deploy.js
```diff
@@ -0,0 +1,101 @@
...
+      globalInts: 4,
+      globalBytes: 2,
+      appArgs: appArgs
+    }, {}, templateParam);
+  console.log(daoAppInfo);
+
+  // fund lsig's
+  await Promise.all([
+    deployer.fundLsig('deposit-lsig.py',
+      { funder: creator, fundingMicroAlgo: 2e6 }, {},
```
robert-zaremba, line 47:
```suggestion
      { funder: creator, fundingMicroAlgo: 2e6 }, {},  // 2 algo
```

robert-zaremba: Why this needs 2 Algo? Can we decrease it?

examples/dao/scripts/run/add_proposal.js
```diff
@@ -0,0 +1,68 @@
...
+  const daoAppInfo = deployer.getApp('dao-app-approval.py', 'dao-app-clear.py');
+  const proposalLsig = await getProposalLsig(deployer);
+  try {
+    await deployer.optInLsigToApp(daoAppInfo.appID, proposalLsig, {}, {});
+  } catch (e) {
+    console.log(e.message);
+  }
+
+  const daoFundLsig = await getDAOFundLsig(deployer);
+  const daoParams = [
```
robert-zaremba, line 22:
```suggestion
  const proposalParams = [
```

examples/dao/scripts/run/add_proposal.js
```diff
@@ -0,0 +1,68 @@
...
+      payFlags: { totalFee: 1000 }
+    }
+  ];
+
+  // Transaction PASS
+  await executeTx(deployer, addProposalTx);
+
+  // Transaction FAIL (asset_transfer amount is less than min_deposit)
+  addProposalTx[1].amount = 10;
+  await executeTx(deployer, addProposalTx);
```
robert-zaremba, line 65: let's change the order: firstly try to call a failing scenario, and then a positive one.

examples/dao/scripts/run/clear_proposal.js
```diff
@@ -0,0 +1,72 @@
...
+      sign: types.SignType.LogicSignature,
+      fromAccountAddr: proposalLsig.address(),
+      lsig: proposalLsig,
+      assetID: govToken.assetIndex,
+      payFlags: {}
+    }
+  ];
+  await executeTx(deployer, optInTx);
+
+  // Transaction FAIL: deposit amount is not same as app.global("deposit")
```
robert-zaremba, line 65:
```suggestion
  // Transaction FAIL: deposit amount is not the same as app.global("deposit")
```

examples/dao/scripts/run/common/common.js
```diff
@@ -0,0 +1,61 @@
+const {
+  executeTransaction
+} = require('@algo-builder/algob');
+const { types } = require('@algo-builder/web');
+
+async function executeTx (deployer, txnParams) {
```
robert-zaremba, line 6:
```suggestion
async function tryExecuteTx (deployer, txnParams) {
```

robert-zaremba (approved):
Preapproing. Added few comments. there are also other comments to resolve.

examples/dao/assets/dao-app-approval.py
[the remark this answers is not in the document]
robert-zaremba: OK, please add a comment in the code

examples/dao/assets/dao-app-approval.py
robert-zaremba: let's add a comment in the code as well with more explanation

examples/dao/assets/deposit-lsig.py
ratik21: yes

examples/dao/assets/deposit-lsig.py
ratik21: ok

examples/dao/assets/deposit-lsig.py
ratik21: yes, looks simpler.

examples/dao/scripts/run/add_proposal.js
ratik21: ok

examples/dao/scripts/deploy.js
ratik21: decreased to 1.

PR #464, opened 2021-09-13
[Bug] Random address for logicsig(runtime)

amityadav0:


CHANGELOG.md
```diff
@@ -20,6 +20,7 @@
 * Fixed `yarn add @algo-builder/web` (was failing because of missing dependency `zod` in packages/web).
 * Fix metadatahash type
 * Fix init project-name bug(`algob init <project-name>` command was not working properly)
+* Fix random address for logicsig(runtime)
```
ratik21, line 23: Could you please describe the bug?

robert-zaremba: The bug is that we need a predictable address of an lsig. Creating two times an lsig account from the same TEAL code should return the same address.

packages/runtime/src/logicsig.ts
```diff
@@ -26,10 +27,22 @@ export class LogicSig {
...
+    programArgs?: Array<Uint8Array | Buffer> | null
+  ) {
     this.tag = Buffer.from("Program");
     this.logic = parsing.stringToBytes(program);
-    this.lsigAddress = generateAccount().addr;
+
+    const hash = murmurhash.v3(program);
+    const result = logicSigAddresses.get(hash);
+    if (result === undefined) {
+      const addr = generateAccount().addr;
```
robert-zaremba, line 40: how about the following:
* `generateAccount` will take an optional argument: `code: string`.
* if `code` is provided it will use it to generate an address by hashing the code.
* otherwise it will use a random number.

This way:
* we don't need `logicSigAddresses` map
* we assure the addresses will be deterministic for lsig.

robert-zaremba (commented):
I suggest to have _always_ deterministic method for generating lsig accounts.

packages/runtime/src/logicsig.ts
amityadav0: i thought this, i wanted to keep same address as algorand but to do so, we need a logic of program in bytes which is returned by compiler and then uint8array is used to create a hash like this:
```
const toBeSigned = utils.concatArrays(this.tag, this.logic);
const hash = nacl.genericHash(toBeSigned);
return address.encodeAddress(new Uint8Array(hash));
```
But we cannot get uint8array that's why i didn't follow this.
will that be ok if address is not `ALGORAND ADDRESS` ?
looking more into encoding algorand addresses

amityadav0: Screenshot from 2021-09-16 11-50-39

I tried to use hash to make `encodeAddress(parsing.uint64ToBigEndian(hash));`

also i tried using `encodeAddress(parsing.stringToBytes(program));`
as

amityadav0: verification is failing because address is not `ALGORAND_ADDRESS`, we have added test for verification so that only logic sig address can send transactions from logic sig.

amityadav0: found js-sdk is sending request to `.compile(),do()` to go-algorand and then there we have compiler which is used to compile string to bytes

robert-zaremba: We just need to find how to construct a valid algorand address based on a seed (here the hash of the code).

Here is how the address is constructed: https://developer.algorand.org/docs/features/accounts/#transformation-public-key-to-algorand-address

So what we need:
1. `key = hash(code).slice(0, 32)`  -- take 32 bytes from the code hash
2. `checksum = sha256(key)`
3. `address_bytes = key.concat(checksum.slice(-4))`
4. `address = base32(address_bytes)`

amityadav0: done added key_pair generation using random seed, random seed used is hash of program

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@
 * Fix init project-name bug(`algob init <project-name>` command was not working properly)
 * Fix zod package was missing from runtime(but zod was being used in runtime)
 * Added support for passing http token as an `object` as well. User can now use `{ "X-Algo-API-Token": <token> }` notation for passing token in `algob.config.js`.
+* Fix random address for logic sig, creating two times an lsig account from the same TEAL code should return the same address.
```
robert-zaremba, line 30: Please create a subsection for bugs in the `testing framework` -- to indicate that these bug fixes are not impacting users in production.

packages/runtime/src/logicsig.ts
```diff
@@ -26,10 +27,19 @@ export class LogicSig {
...
+    programArgs?: Array<Uint8Array | Buffer> | null
+  ) {
     this.tag = Buffer.from("Program");
     this.logic = parsing.stringToBytes(program);
-    this.lsigAddress = generateAccount().addr;
+
+    const seedBytes = parsing.uint64ToBigEndian(murmurhash.v3(program));
+    const append = new Uint8Array(Number(seedBytes.length) + Number(24));
+    append.set(seedBytes);
+    const pair = tweet.sign_keyPair_fromSeed(append);
```
robert-zaremba, line 40:
```suggestion
    const extendedSeed = new Uint8Array(Number(seedBytes.length) + Number(24));
    extendedSeed.set(seedBytes);
    const pair = tweet.sign_keyPair_fromSeed(extendedSeed);
```

robert-zaremba (approved):
Pre approving. left 2  more comments.

CHANGELOG.md
amityadav0: ok

PR #472, opened 2021-09-20
Add guide for using purestake-api (algodv2)

ratik21:


packages/algob/src/internal/deployer_cfg.ts
```diff
@@ -35,7 +35,7 @@ export class DeployerConfig {
     this.runtimeEnv = runtimeEnv;
     this.cpData = loadCheckpointsRecursive();
     this.algoOp = algoOp;
-    this.accounts = mkAccountIndex(runtimeEnv.network.config.accounts);
+    this.accounts = mkAccountIndex(runtimeEnv.network.config.accounts ?? []);
```
amityadav0, line 38: do we not need accounts in purestake api ?

docs/guide/purestake-api.md
```diff
@@ -0,0 +1,40 @@
...
+};
+
+module.exports = {
+  networks: {
+    default: defaultCfg,
+    purestake: purestakeTestNetCfg
+  }
+};
+```
+
```
amityadav0, line 39: could you please add an example call using api ?

packages/algob/src/internal/deployer_cfg.ts
ratik21: This is to handle a case if no accounts are passed in config.

docs/guide/purestake-api.md
ratik21: The usage is same as we use `algodClient` in scripts. We just demonstrate how to add/switch to this network (originally users were not aware how to use algob with purestake)

amityadav0 (approved):

docs/guide/purestake-api.md
amityadav0: could you please add a `accounts` variable it could be blank `accounts: []` here ? It would be good for users

ratik21: yes

PR #476, opened 2021-09-23
Add support for algosdk.Transaction object in executeTranasction

amityadav0:


packages/algob/src/lib/tx.ts
```diff
@@ -266,16 +269,64 @@ export async function makeAndSignTx (
...
+    const signedTxn = signTransactionObject(execParams, signer);
+    const confirmedTx = await deployer.sendAndWait(signedTxn);
+    console.log(confirmedTx);
+    return confirmedTx;
+  }
+
+  // Update type here because we are sure this is not transaction object type
+  execParams = Array.isArray(execParams)
+    ? execParams as wtypes.ExecParams[]
+    : execParams;
```
ratik21, line 328: It seems to me we can also consider keeping this as a separate function. Maybe `executeSDKTransaction`? Is there much of an overlap?

packages/algob/src/lib/tx.ts
```diff
@@ -266,16 +269,64 @@ export async function makeAndSignTx (
   return [txns, signedTxn];
 }
 
+/**
+ * Signs transaction object/s and returns raw signed transaction
+ * @param txn Transaction object/s
+ * @param signer signer
+ */
+export function signTransactionObject (
```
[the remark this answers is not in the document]
amityadav0: added a doc string above explaining what is `signTransactionObject`

packages/algob/src/lib/tx.ts
amityadav0: the idea was to extend `executeTransaction`. So that it is easier for user

packages/web/src/types.ts
```diff
@@ -263,3 +265,18 @@ export function isRequestError (object: unknown): object is RequestError {
    Object.prototype.hasOwnProperty.call(object, "response.error");
   return res && Object.prototype.hasOwnProperty.call(object, "error");
 }
+
+// This function checks if given object implements `Transaction` class
+export function isTransaction (object: unknown): object is Transaction {
```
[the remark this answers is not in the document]
amityadav0: yes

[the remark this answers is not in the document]
amityadav0: because it is `type` checking a object

packages/algob/src/lib/tx.ts
```diff
@@ -266,16 +269,64 @@ export async function makeAndSignTx (
...
+      signedTxn = txns.map((txn: Transaction, index: number) => {
+        return signTransaction(txn, signer);
+      });
+    }
+    return signedTxn;
+  } else {
+    if (Array.isArray(signer)) {
+      throw new Error("cannot sign single transaction with multiple signers");
+    }
+    return signTransaction(txns, signer);
```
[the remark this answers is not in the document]
amityadav0: I prefer this way, simpler for user

packages/web/src/types.ts
```diff
@@ -263,3 +265,18 @@ export function isRequestError (object: unknown): object is RequestError {
...
+    "genesisID", "genesisHash", "voteKey", "selectionKey", "voteFirst", "voteLast",
+    "voteKeyDilution", "assetIndex"
+  ];
+  let res = Object.prototype.hasOwnProperty.call(object, "name");
+  for (const prop of props) {
+    res = res && Object.prototype.hasOwnProperty.call(object, prop);
+  }
+
+  return res;
+}
```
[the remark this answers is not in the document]
amityadav0: of course

packages/algob/src/lib/tx.ts
```diff
@@ -123,7 +124,9 @@ export function makeASAOptInTx (
  * @param txn unsigned transaction
  * @param execParams transaction execution parametrs
  */
-function signTransaction (txn: Transaction, execParams: wtypes.ExecParams): Uint8Array {
+function signTransaction (
+  txn: Transaction, execParams: wtypes.ExecParams | wtypes.Signer
```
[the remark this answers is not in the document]
amityadav0: This is done to remove complexity from `signTransaction(txn, execParams[index]);` execute transaction, i will add a comment explaining difference, If we choose to remove Execparams, we will have to create new objects each time we call this execute transaction

packages/algob/src/lib/tx.ts
```diff
@@ -1,6 +1,7 @@
 import { parseASADef, types as rtypes } from "@algo-builder/runtime";
 import { tx as webTx, types as wtypes } from "@algo-builder/web";
 import algosdk, { Algodv2, decodeSignedTransaction, SuggestedParams, Transaction } from "algosdk";
+import { sign } from "crypto";
```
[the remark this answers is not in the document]
amityadav0: ok

packages/algob/src/lib/tx.ts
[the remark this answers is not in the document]
ratik21: I thought user would use `executeTransaction`. If we expose this to user, why not use an object?. 
Eg. `[{ txn: <Txn1>, signer: <sig1> }, { .. } ]`. 
Rather than using two arrays `([ txn1, txn2 ..], [sig1, sig2 ..])`.

The idea is signer is associated with a transaction, so it makes sense to pass them together. What do you think?

packages/algob/src/lib/tx.ts
ratik21: `executeTransaction(depoloyer, execParams)`

`executeTransaction(depoloyer, sdkTxn, signer)`

They seem different, Ok.

amityadav0: What do you mean ?

packages/algob/src/lib/tx.ts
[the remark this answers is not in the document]
amityadav0: ok

packages/algob/src/lib/tx.ts
```diff
@@ -266,16 +268,68 @@ export async function makeAndSignTx (
   return [txns, signedTxn];
 }
 
+/**
+ * Signs transaction object(s) and returns raw signed transaction
+ * Note: Sign transaction is used to sign single transaction and `signTransactionObject` takes
+ * SDK transaction object, signer and signs it.
+ * @param txn Transaction object(s)
+ * @param signer signer
```
robert-zaremba, line 276: please add more documentation about the behavior of signer when it's a single object or an array. Add that if it's an array, then the length must be the same as `txns` length.

packages/algob/src/lib/tx.ts
```diff
@@ -266,16 +268,68 @@ export async function makeAndSignTx (
...
+ */
+export function signTransactionObject (
+  txns: Transaction | Transaction[], signer: wtypes.Signer | wtypes.Signer[]
```
robert-zaremba, line 279: This function is too much overloaded and it's not good to return a union type. It doesn't make sense to use this function with single transaction and single signer (because in that case we should use `signTransaction` directly). So I propose to update this function to:
```
export function signTransactions(txns:  Transaction[], signer: wtypes.Signer | wtypes.Signer[]): Uint8Array[]
```

packages/algob/src/lib/tx.ts
```diff
@@ -266,16 +268,68 @@ export async function makeAndSignTx (
   return [txns, signedTxn];
 }
 
+/**
+ * Signs transaction object(s) and returns raw signed transaction
+ * Note: Sign transaction is used to sign single transaction and `signTransactionObject` takes
+ * SDK transaction object, signer and signs it.
+ * @param txn Transaction object(s)
```
robert-zaremba, line 275:
```suggestion
 * @param txns Transaction objects
```

robert-zaremba (commented):
I think we over-engineer here. Let's discuss more in Discord if needed.

packages/web/src/types.ts
```diff
@@ -1,10 +1,12 @@
-import { Account as AccountSDK, LogicSigAccount } from 'algosdk';
+import { Account as AccountSDK, LogicSigAccount, Transaction } from 'algosdk';
 import * as z from 'zod';
 
 import type { ASADefSchema, ASADefsSchema } from "./types-input";
 
 export type AccountAddress = string;
 
+export type Signer = SignWithSk | SignWithLsig;
```
robert-zaremba, line 8: this duplicates `type Sign = SignWithSk | SignWithLsig;`. Let's remove the type above.

robert-zaremba (changes requested):

packages/algob/src/lib/tx.ts
```diff
@@ -266,16 +268,68 @@ export async function makeAndSignTx (
...
  * https://github.com/scale-it/algo-builder/blob/docs/docs/guide/execute-transaction.md
  */
 export async function executeTransaction (
   deployer: Deployer,
-  execParams: wtypes.ExecParams | wtypes.ExecParams[]):
-  Promise<ConfirmedTxInfo> {
+  execParams: (wtypes.ExecParams | Transaction) | (wtypes.ExecParams[] | Transaction[]),
+  signer?: wtypes.Signer | wtypes.Signer[]
+): Promise<ConfirmedTxInfo> {
+  if (
```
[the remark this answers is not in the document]
amityadav0: ok

packages/web/src/types.ts
amityadav0: removed

packages/algob/src/lib/tx.ts
amityadav0: ok

packages/algob/src/lib/tx.ts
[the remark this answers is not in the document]
amityadav0: updated

packages/algob/src/lib/tx.ts
```diff
@@ -266,16 +268,68 @@ export async function makeAndSignTx (
...
+    (Array.isArray(execParams) && wtypes.isSDKTransaction(execParams)) || wtypes.isSDKTransaction(execParams)
```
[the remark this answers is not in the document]
amityadav0: updated

packages/algob/src/lib/tx.ts
amityadav0: updated merged both the types. `TransactionAndSign` in one

packages/algob/src/lib/tx.ts
```diff
@@ -266,20 +268,70 @@ export async function makeAndSignTx (
...
 
+/**
+ * Signs transaction object(s) and returns raw signed transaction
+ * Note: Sign transaction is used to sign single transaction and `signTransactions` takes
+ * SDK transaction object, signer and signs it.
+ * @param txnAndSign Transaction object(s) with signers
+ */
+export function signTransactions (
+  txnAndSign: wtypes.TransactionAndSign[]
+): Uint8Array | Uint8Array[] {
```
robert-zaremba, line 279: After the update, the return type is always array
```suggestion
): Uint8Array[] {
```

packages/algob/src/lib/tx.ts
```diff
@@ -266,20 +268,70 @@ export async function makeAndSignTx (
...
+  let txns: Transaction[] = [];
+  const signers: wtypes.Sign[] = [];
+  txnAndSign.forEach(function (value, idx) {
+    txns[idx] = value.transaction;
+    signers[idx] = value.sign;
+  });
```
robert-zaremba, line 285: let's use for loop instead of `foreach`.

packages/web/src/errors/errors-list.ts
```diff
@@ -177,6 +177,12 @@ Please check algob output for more details.`
     description: `Invalid ASA definition: '%source%'.
     Reason: %reason%
     Please check your ASA file`
+  },
+  EXECPARAMS_LENGTH_ERROR: {
+    number: 18,
+    message: `Execparams length is zero'.`,
+    title: "Invalid parameters length",
+    description: `Length should be greater than 0'.`
```
robert-zaremba, line 185:
```suggestion
    message: `Exec params length is zero'.`,
    title: "Invalid parameter",
    description: `Empty list is not supported.`
```

packages/algob/src/lib/tx.ts
```diff
@@ -266,20 +268,70 @@ export async function makeAndSignTx (
...
+ * @param transactionParam transaction parameters or atomic transaction parameters
  * https://github.com/scale-it/algo-builder/blob/docs/docs/guide/execute-transaction.md
+ * or TransactionAndSign object(SDK transaction object and signer parameters)
  */
 export async function executeTransaction (
   deployer: Deployer,
-  execParams: wtypes.ExecParams | wtypes.ExecParams[]):
-  Promise<ConfirmedTxInfo> {
-  deployer.assertCPNotDeleted(execParams);
+  transactionParam: (wtypes.ExecParams | wtypes.TransactionAndSign)
```
robert-zaremba, line 304: let's rename `transactionParam` to `transactions`

packages/algob/src/lib/tx.ts
```diff
@@ -266,20 +268,70 @@ export async function makeAndSignTx (
...
+      const confirmedTx = await deployer.sendAndWait(signedTxn);
+      console.log(confirmedTx);
+      return confirmedTx;
+    }
+  } else {
+    if (wtypes.isSDKTransactionAndSign(transactionParam)) {
+      const signedTxn = signTransaction(transactionParam.transaction, transactionParam.sign);
+      const confirmedTx = await deployer.sendAndWait(signedTxn);
+      console.log(confirmedTx);
+      return confirmedTx;
```
robert-zaremba, line 322: we duplicate code in the conditions above. let's try to make a condition which will allow us to not duplicate the code, eg:
```
let isArray = ...

if isArray and length == 0 { throw exception }
if (isArray and isSDKTransactionAndSign(transactions[0])) || (!isArray and isSDKTransactionAndSign(transactions[0])) { ... }

File: packages/algob/src/lib/tx.ts [hunk 19]
```diff
@@ -266,20 +268,70 @@ export async function makeAndSignTx (
...
 /**
  * Execute single transaction or group of transactions (atomic transaction)
+ * executes `ExecParams` or `Transaction` Object, SDK Transaction object passed to this function
+ * will be signed and sent to network. User can use SDK functions to create transactions.
+ * Note: If passing transaction object a signer/s must be provided.
  * @param deployer Deployer
- * @param execParams transaction parameters or atomic transaction parameters
+ * @param transactionParam transaction parameters or atomic transaction parameters
  * https://github.com/scale-it/algo-builder/blob/docs/docs/guide/execute-transaction.md
+ * or TransactionAndSign object(SDK transaction object and signer parameters)
```
robert-zaremba, line 300: let's update the documentation. The function may be confused. Please add few examples to the documentation.

robert-zaremba (commented):
Pre-approving
* Please add a changelog (both about the new feature and new `signTransactions` function).
* update the guide (docs/guide)
* let's update the `executeTransaction` tests

packages/algob/src/lib/tx.ts
amityadav0: ok will update the first condition, but there are different functions used in both the conditions `signTransaction` `signTransactions`.

packages/algob/src/lib/tx.ts
amityadav0: ok

packages/algob/src/lib/tx.ts
amityadav0: ok

packages/algob/src/lib/tx.ts
amityadav0: is there any specific reason ?

packages/algob/src/lib/tx.ts
amityadav0: Moved common code after condition checks

packages/algob/src/lib/tx.ts
robert-zaremba: This comment is not blocking (you can merge). Having a function to modify outside variables is not very clear.

packages/algob/src/lib/tx.ts
robert-zaremba: ok

robert-zaremba (approved):
ACK.

packages/algob/src/lib/tx.ts
amityadav0: ok, updating to `for`

PR #490, opened 2021-10-17
A change to support better automated testing in the Aqua compiler

ashleydavis:
Factored out a new function "executeWithResult" that returns the result of evaluating the TEAL program so the caller can decide what to do with the result.

## Motivation:

This change will be used in the Aqua compiler to facilitate better automated testing, it allows tests of the following nature:

```typescript
import { compile } from "aqua-compiler";

export function execute(teal: string): any {
    const runtime = new Runtime([]);
    return interpreter.executeWithResult(teal, types.ExecutionMode.APPLICATION, runtime, undefined);
}

describe("runtime tests", () => {

    it("1 + 2", () => {
        const result = execute(compile("1 + 2;"));
        expect(Number(result)).toBe(3);
    });
});
```

This is a better method of testing for Aqua because it means we don't have to make complex assertions about the generated TEAL assembly code, instead we can simply run the assembly code through algo-builder/runtime and then verify that the result is as expected.

### DIFF
--- a/.github/workflows/examples.yaml
+++ b/.github/workflows/examples.yaml
@@ -118,3 +118,9 @@ jobs:
         run: |
           pipenv run yarn algob deploy
           pipenv run yarn algob run scripts/interaction/call_application.js
+
+      - name: Example DAO
+        working-directory: ./examples/dao
+        run: |
+          pipenv run yarn algob deploy
+          pipenv run yarn algob run scripts/run/add_proposal.js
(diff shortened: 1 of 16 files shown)


ashleydavis:
Thanks @robert-zaremba, I don't have write access. Someone will have to merge for me.

robert-zaremba:
The linter is failing. I will merge it and fix directly.

PR #506, opened 2021-10-29
Rename get lsig

robert-zaremba:
Following discussion on Discord, the `getLogicSig` is not very clear. So updating the comment and renaming it. It's a breaking change unfortunately.

### DIFF
--- a/examples/bond/test/bond-token-flow.js
+++ b/examples/bond/test/bond-token-flow.js
@@ -107,7 +107,7 @@ describe('Bond token Tests', function () {
       TMPL_APP_MANAGER: appManager.address
     };
     const issuerLsigProg = getProgram('issuer-lsig.py', scInitParam);
-    lsig = runtime.createLSigAccount(issuerLsigProg, []);
+    lsig = runtime.createLsigAccount(issuerLsigProg, []);
     issuerLsigAddress = lsig.address();
 
     // sync escrow account
@@ -246,7 +246,7 @@ describe('Bond token Tests', function () {
       TMPL_BOND: bond2
     };
     const buyLsigProgram = getProgram('buyback-lsig.py', scParam);
-    const buybackLsig = runtime.createLSigAccount(buyLsigProgram, []);
+    const buybackLsig = runtime.createLsigAccount(buyLsigProgram, []);
 
     // fund dex with some minimum balance first
     runtime.fundLsig(master.account, buybackLsig.address(), minBalance + 10000);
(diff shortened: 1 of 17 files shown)


examples/bond/test/bond-token-flow.js
```diff
@@ -107,7 +107,7 @@ describe('Bond token Tests', function () {
       TMPL_APP_MANAGER: appManager.address
     };
     const issuerLsigProg = getProgram('issuer-lsig.py', scInitParam);
-    lsig = runtime.getLogicSig(issuerLsigProg, []);
+    lsig = runtime.createLSigAccount(issuerLsigProg, []);
```
ratik21, line 110: just to confirm, `LSig` <--> `Lsig`.

packages/algob/src/lib/lsig.ts
```diff
@@ -14,8 +14,8 @@ import { CompileOp } from "./compile";
 export async function getLsig (
   name: string,
   algodClient: Algodv2,
-  scTmplParams?: SCParams):
-  Promise<LogicSigAccount> {
+  scTmplParams?: SCParams): Promise<LogicSigAccount> {
+
```
ratik21, line 18: could you please fix the failing github action?

packages/runtime/src/runtime.ts
```diff
@@ -558,11 +558,14 @@ export class Runtime {
   }
 
   /**
-   * Returns logic signature
+   * Creates a new account with logic signature and smart contract arguments
+   * in the runtime store. The arguments are use when we send a transaction with this
+   * account and verify it.
```
ratik21, line 563:
```suggestion
   * Creates a new account with logic signature and smart contract arguments
   * in the runtime store. The arguments are used when we send a transaction with this
   * account and verify it.
```

packages/runtime/src/runtime.ts
```diff
@@ -558,11 +558,14 @@ export class Runtime {
...
    * @param program TEAL code
    * @param args arguments passed
+   * @returns logic signature account with arguments.
```
ratik21, line 566:
```suggestion
   * @returns logic signature with arguments.
```

ratik21 (approved):

examples/bond/test/bond-token-flow.js
robert-zaremba: Good catch, in other places we are using `Lsig`, so I will rename it here as well.

robert-zaremba: renamed.

PR #507, opened 2021-10-30
Pass test filename to arguments of algob test command

vuvth:
## Proposed Changes

+ Move loadFilenames function(and test) from builtin-tasks/deploy.ts to lib/files.ts
+ Add addOptionalVariadicPositionalParam testFiles. Now we can run test file we want to run. 
Ex:  `algob test ./test/test.js`.

Please, review it! Thanks for your support! ❤️

### DIFF
--- a/packages/algob/src/builtin-tasks/test.ts
+++ b/packages/algob/src/builtin-tasks/test.ts
@@ -4,11 +4,11 @@ import Mocha from "mocha";
 import { task } from "../internal/core/config/config-env";
 import { loadFilenames } from "../lib/files";
 import { testsDirectory } from "../lib/script-checkpoints";
-import type { Config } from "../types";
+import type { TaskTestConfig } from "../types";
 import { TASK_TEST } from "./task-names";
 
 const TEST_DIR = 'test';
-async function runTests (config: Config): Promise<void> {
+async function runTests (config: TaskTestConfig): Promise<void> {
   try {
     const tsPath = findupSync("tsconfig.json", { cwd: process.cwd() });
     if (tsPath) {
@@ -16,8 +16,12 @@ async function runTests (config: Config): Promise<void> {
       process.env.TS_NODE_PROJECT = tsPath;
       require('ts-mocha');
     }
+    // User pass testFiles to arguments so just run those files.
+    // else we run test files in default test dir.
+    const testFiles = config.testFiles.length !== 0
+      ? config.testFiles
+      : loadFilenames(testsDirectory, TEST_DIR);
 
-    const testFiles = loadFilenames(testsDirectory, TEST_DIR);
     const mocha = new Mocha(config.mocha);
     // Adding test files to mocha object
     testFiles.forEach((file) => mocha.addFile(file));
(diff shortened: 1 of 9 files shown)


packages/algob/src/builtin-tasks/test.ts
```diff
@@ -2,22 +2,26 @@ import findupSync from "findup-sync";
...
+async function runTests (config: TaskTestConfig): Promise<void> {
   try {
     const tsPath = findupSync("tsconfig.json", { cwd: process.cwd() });
     if (tsPath) {
       // run tests via ts-mocha, if project is in typescript
       process.env.TS_NODE_PROJECT = tsPath;
       require('ts-mocha');
     }
+    // User pass testFiles to arguments so just run those files.
+    // else we run test files in default test dir.
```
ratik21, line 20:
```suggestion
    // else we run test files in default(root) test dir.
```

packages/algob/src/builtin-tasks/test.ts
```diff
@@ -32,5 +36,10 @@ async function runTests (config: Config): Promise<void> {
 
 export default function (): void {
   task(TASK_TEST, "Run tests using mocha in project root")
+    .addOptionalVariadicPositionalParam(
+      "testFiles",
+      "An optional list of files to test",
```
ratik21, line 41:
```suggestion
      "An optional list of file path(s) to test",
```

packages/algob/src/lib/files.ts
```diff
@@ -48,3 +50,27 @@ export function loadEncodedTxFromFile (fileName: string): Uint8Array | undefined
...
   }
 }
+
+/**
+ * Load .js, .ts files from /scripts (default) directory
+ * @param directory directory to load files from
+ * @param taskType task type (eg. test)
+ * @returns array of paths as string eg. ['scripts/file1.js', 'scripts/file2.js', ..]
+ */
+export function loadFilenames (directory: string, taskType?: string): string[] {
```
ratik21, line 60: how about moving this somewhere in `/internal`? atm, i don't think we need it in `/src/lib`.

ratik21 (commented):
let one comment, thanks for contribution!

packages/algob/src/lib/files.ts
vuvth: Hi @ratik21! Nice to have conversations with you.

I see in the built-in tasks we use functions inside folder src/lib. This is the reason did I move the function to src/lib.

vuvth: I think it will keep the loadFilenames function in this file. It more sense if I do refactor code in another pull request.

robert-zaremba: The idea is that we use `src/internal` for things we don't want to expose to the user. 
Is a user going to import `loadFilenames` in his local project? If the function is going to be used only in the built-in tasks or other package functions, then it makes more sense to move it to `src/internal`.

vuvth: Thanks! I got it.

vuvth: I only refactor loadFilenames function. I see some other files want to refactor too. I think I will do it in another pull request.

robert-zaremba (approved):

robert-zaremba:
FYI, @vuvth , I'm using `yarn test path/to/file` for running tests. You can also use a watch mode:
```
yarn test -w path/to/file
```

PR #510, opened 2021-11-01
Support application account in runtime, update checkpoint structure

ratik21:
## Proposed Changes

+ Support `applicationAccount` in runtime
+ Add key `applicationAccount` to app's checkpoints

### DIFF
--- a/packages/algob/src/lib/algo-operator.ts
+++ b/packages/algob/src/lib/algo-operator.ts
@@ -328,6 +328,7 @@ export class AlgoOperatorImpl implements AlgoOperator {
       localBytes: flags.localBytes,
       globalInts: flags.globalInts,
       globalBytes: flags.globalBytes,
+      extraPages: flags.extraPages,
       accounts: flags.accounts,
       foreignApps: flags.foreignApps,
       foreignAssets: flags.foreignAssets,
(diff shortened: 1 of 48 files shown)


packages/algob/src/lib/algo-operator.ts
```diff
@@ -1,6 +1,6 @@
 import { types as rtypes } from "@algo-builder/runtime";
 import { BuilderError, ERRORS, tx as webTx, types as wtypes } from "@algo-builder/web";
-import algosdk, { LogicSigAccount, modelsv2 } from "algosdk";
+import algosdk, { getApplicationAddress, LogicSigAccount, modelsv2 } from "algosdk";
```
robert-zaremba, line 3: Can we rename to `getAppAddress`?

robert-zaremba: Where is this defined? Probably some file is missing in this PR?

ratik21: it is defined in the js-sdk: [https://github.com/algorand/js-algorand-sdk/blob/develop/src/encoding/address.ts#L187](https://github.com/algorand/js-algorand-sdk/blob/develop/src/encoding/address.ts#L187)

robert-zaremba: ah, sorry, didn't look at the end of the statement ;)

robert-zaremba (approved):

PR #515, opened 2021-11-03
Inner Transactions: more tests, ixtn op, added missing fields

ratik21:
## Proposed Changes

+ more tests for inner transactions (asset freeze, delete, modify, deploy)
+ opcodes: `itxn f`, `itxna`
+ Added missing fields


pbennett:
Trying to use this PR to create an ASA via inner transactions I'm getting:
```
     Error: RUNTIME_ERR1500: Invalid ASA definition: 'undefined'.
    Reason:           [decimals]: Invalid input
            [decimals]: Required
            [decimals]: Required

          [decimals]: Decimals must be between 0 (non divisible) and 19
```

For whatever reason, the decimals value I'm setting isn't coming across [it's arriving undefined] - yet the TEAL code:
int 0
itxn_field ConfigAssetDecimals
is there correctly.

Hacking ctx.ts addASADet to explicitly set asaDef.decimals = 0 lets my code continue.

I'm also seeing that setting of the ASA URL to ipfs://ABCDEF is being change to just 'ipfs' (I suspect because of str:ipfs://ABCDEF call in smart-contract appCall transaction - and the processing for str: is doing something here?)

pbennett:
The SC code I have creating the ASA which is failing is like:
```
    return Seq(
        InnerTxnBuilder.Begin(),
        InnerTxnBuilder.SetFields(
            {
                TxnField.type_enum: TxnType.AssetConfig,
                TxnField.config_asset_unit_name: Bytes("TEST"),
                TxnField.config_asset_name: Txn.application_args[1],
                TxnField.config_asset_total: Int(1),
                TxnField.config_asset_decimals: Int(0),
                TxnField.config_asset_default_frozen: Int(1),
                TxnField.config_asset_url: Txn.application_args[2],
                TxnField.config_asset_manager: Global.current_application_address(),
                TxnField.config_asset_reserve: Global.current_application_address(),
                TxnField.config_asset_freeze: Global.current_application_address(),
                TxnField.config_asset_clawback: Global.current_application_address(),
            }
        ),
        InnerTxnBuilder.Submit(),
        ```
        
From this, name is undefined, decimals is undefined, and the URL is being stripped.

**Comment by ratik21** [comment]
@pbennett thanks, i have pushed the fixes with tests. yes, they're related to parsing (decimals, `str:` issue)

**Comment by pbennett** [comment]
Thanks(!) this fixed a number of issues but it looks like the 'name' field isn't being set at all.  UnitName is, but not the Asset Name.

**Comment by pbennett** [comment]
I've also noticed that after creating an ASA via an itxn, the contract address of the SC doesn't contain the asset.
ie, this fails:
        const NFDASAHolding = runtime.getAssetHolding(createdAssetID, nfdContractAddress);

**Comment by pbennett** [comment]
I've also noticed that after creating an ASA via an itxn - the contract address does not contain the asset like it should.
Ie, this fails:
        const asaHolding = runtime.getAssetHolding(createdAssetID, contractAddress);
yet I can confirm that the contract address is correct and matches the manager/reserve/freeze/clawback addresses - all set via the itxn and the Global.current_application_address()

**Comment by ratik21** [comment]
> but it looks like the 'name' field isn't being set at all. UnitName is, but not the Asset Name.

oh - atm we don't store the asset name in our [schema](https://github.com/scale-it/algo-builder/blob/master/packages/web/src/types-input.ts#L8), it's used as a key to get asset info. eg.
```js
const asaDefFromName = runtime.getAssetInfoFromName("asa_name")?.assetDef;
```
[here](https://github.com/scale-it/algo-builder/pull/515/files#diff-0464562c698a01be913681450dc6011b3bdd9355c9fbe936aa32769aa1d9715aR421) is the reference in test. Let me know if you think we should store asaName in schema as well.

> I've also noticed that after creating an ASA via an itxn - the contract address does not contain the asset like it should.

weird, i am able to see `contractAssetHolding` after asa deployment. Maybe you didn't run `syncAccounts()`? [here](https://github.com/scale-it/algo-builder/pull/515/commits/03a862789e562de7d9f54e0fc07e3b290cef112f) is the commit

pbennett:
> > but it looks like the 'name' field isn't being set at all. UnitName is, but not the Asset Name.
> oh - atm we don't store the asset name in our [schema](https://github.com/scale-it/algo-builder/blob/master/packages/web/src/types-input.ts#L8), it's used as a key to get asset info. eg.

This obviously needs corrected.  Not even storing the name of an asset in an asset definition seems like a pretty significant oversight for something meant to model, as closely as possible, another system.  Creating an asset and verifying the name is what it is supposed to be is pretty fundamental in a simple unit test.

> 
> ```js
> const asaDefFromName = runtime.getAssetInfoFromName("asa_name")?.assetDef;
> ```

This isn't correct either - assets aren't defined by name - they're defined by ID.  Asset names aren't unique in Algorand.

> [here](https://github.com/scale-it/algo-builder/pull/515/files#diff-0464562c698a01be913681450dc6011b3bdd9355c9fbe936aa32769aa1d9715aR421) is the reference in test. Let me know if you think we should store asaName in schema as well.

Yes, the asset schema needs to have everything a real Algorand asset would have when created/queried.

> weird, i am able to see `contractAssetHolding` after asa deployment. Maybe you didn't run `syncAccounts()`? [here](https://github.com/scale-it/algo-builder/pull/515/commits/03a862789e562de7d9f54e0fc07e3b290cef112f) is the commit

I found the issue - the problem is the runtime.getAssetDef and runtime.getAssetHolding don't properly handle receiving a bigint which is what 'real' asset ids will normally be.  During unit tests it's probably not an issue but once I converted to a Number it worked.

pbennett:
Ah, found another bug.  If the itxn creating an asset sets defaultFrozen to true - the created asset isn't frozen and I think should be.

ratik21:
> Yes, the asset schema needs to have everything a real Algorand asset would have when created/queried.

Understood, will add a task for it.

> Ah, found another bug. If the itxn creating an asset sets defaultFrozen to true - the created asset isn't frozen and I think should be.

Sure? In `asaDef` `defaultFrozen` is being set as `true`. 
If you're talking "ASA Holding", then for the creator account `"is-frozen"` will be `false`, even if defaultFrozen is `true`.
For any account other than creator, `"is-frozen"` in ASA Holding will be `true`.

pbennett:
> Sure? In `asaDef` `defaultFrozen` is being set as `true`. If you're talking "ASA Holding", then for the creator account `"is-frozen"` will be `false`, even if defaultFrozen is `true`. For any account other than creator, `"is-frozen"` in ASA Holding will be `true`.

Got it - I'd verify that it should show as not frozen but I've just commented it out from my test assertions for now.

robert-zaremba (approved):

robert-zaremba:
@pbennett thanks for the review and testing.  
We use ASA ID internally (eg: checkpoints), it seams there was an oversight in runtime. 
We promote ASA name usage because user can declare it in an asa definition file (in algob) which is indexed by name and then translated to a proper transaction with asa name set in it. ASA ID is a number globally unique.

PR #519, opened 2021-11-14
Support rekeying

### LINKED ISSUE #512: Add rekying support to runtime
## Is your feature request related to a problem? Please describe.

Algo Builder test environment doesn't support account rekeying.

## Describe the solution you'd like

Update the [RuntimeAccount](https://algobuilder.dev/api/runtime/interfaces/types.RuntimeAccount.html)  type and functionality (eg by adding class methods) to support account rekeying and verification.

A complete solution will require tests.

vuvth:
Closes #512

## Proposed Changes

+ Create and modify interface(RuntimeAccountI) and class for RuntimeAccount.
+ Add rekey logic to class AccountStore. (rekeyTo method).
+ Add spend accoun query method (AccountStore.getSpendAddr())
+ Check rekey account logic in runtime (validateAccountSignature, validateLsigAndRun).
+ Test cases (account -> account, account -> lsig, lsig -> acccount, lsig -> lsig) [detail here](https://github.com/vuvth/algorand-rekey-experiment)
+ Better message error for rekey transaction. 
+ Fixbug miss rekeyTo field when create algorand transaction.
Please review it! Thanks ❤️


ratik21 (changes requested):
Thanks for the PR. Left an initial review. Will review & test the rekeying logic after initial comments are addressed.

vuvth:
> Thanks for the PR. Left an initial review. Will review & test the rekeying logic after initial comments are addressed.

Thanks for the review and comment, @ratik21 . I will fix it. Sorry because of some mistake I made. I will do it carefully in the time.

packages/runtime/src/account.ts
```diff
@@ -23,8 +24,70 @@ const globalState = "global-state";
...
+  }
+
+  rekey (authAccount: RuntimeAccountI): void {
+    this.authAccount = authAccount;
+  }
+}
+
+export class RuntimeAccountBuilder {
+  sk: Uint8Array
+  addr: string
```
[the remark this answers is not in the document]
vuvth: I implemented it from `Account` interface from JS SDK. So i will keep it `public` like interface.

packages/runtime/src/account.ts
```diff
@@ -23,8 +24,70 @@ const globalState = "global-state";
...
+
+  public setSk (sk: Uint8Array): RuntimeAccountBuilder {
+    this.sk = sk;
+    return this;
+  }
+
+  public setAddr (addr: string): RuntimeAccountBuilder {
+    this.addr = addr;
+    return this;
+  }
```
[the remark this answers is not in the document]
vuvth: Yes, i will remove it. With class RuntimeAccountBuilder,  I think`from` and `genKey()`  method is enough for build account.

packages/runtime/src/account.ts
```diff
@@ -24,8 +24,71 @@ const globalState = "global-state";
...
+    this.name = builder.name;
+  }
+
+  rekeyTo (authAccount: RuntimeAccountI): void {
+    this.authAccount = authAccount;
+  }
+}
+
+export class RuntimeAccountBuilder {
+  sk: Uint8Array
```
robert-zaremba, line 45:
```suggestion
  sk: Uint8Array  // signing key (private key or lsig)
```

packages/runtime/src/account.ts
```diff
@@ -24,8 +24,71 @@ const globalState = "global-state";
 const localStateSchema = "local-state-schema";
 const globalStateSchema = "global-state-schema";
 
+class RuntimeAccount implements RuntimeAccountI {
+  sk: Uint8Array
```
robert-zaremba, line 28:
```suggestion
  sk: Uint8Array // signing key (private key or lsig)
```

packages/runtime/src/account.ts
```diff
@@ -24,8 +24,71 @@ const globalState = "global-state";
...
+    this.addr = builder.addr;
+    this.name = builder.name;
+  }
+
+  rekeyTo (authAccount: RuntimeAccountI): void {
+    this.authAccount = authAccount;
+  }
+}
+
+export class RuntimeAccountBuilder {
```
robert-zaremba, line 44: I'm not sure why do we need a builder for account? The constructor and logic for the `RuntimeAccount`. The account can be only rekeyed. I don't see how adding a builder here will help. If it's only for setting up key generation, then maybe it's overengineering?

packages/runtime/src/account.ts
```diff
@@ -34,21 +97,23 @@ export class AccountStore implements AccountStoreI {
...
+  constructor (balance: number | bigint, account?: RuntimeAccountI | AccountSDK | string) {
     if (typeof account === 'string') {
-      this.account = generateAccount();
-      this.account.name = account;
-      this.address = this.account.addr;
-    } else if (account) {
-      // set config if account is passed by user
+      // create new account with name
+      this.account = new RuntimeAccountBuilder().genKey().setName(account).build();
+    } else if (account instanceof RuntimeAccount) {
```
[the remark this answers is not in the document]
robert-zaremba: re. a) JavaScript doesn't understand interfaces. so `instanceof RuntimeAccountI` will be always false. We would need to use a base class.

packages/runtime/src/account.ts
vuvth: I think you right. We just need move `from` and `genKey` method to RunttimeAccount, code more clear and sense.

packages/runtime/src/account.ts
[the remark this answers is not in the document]
vuvth: I think we should create the issues for question b.

packages/runtime/src/ctx.ts
```diff
@@ -547,6 +559,21 @@ export class Ctx implements Context {
...
+          if (authAccount.addr !== signAccount.account.addr) {
+            // TODO: should use right message error
+            throw new RuntimeError(RUNTIME_ERRORS.GENERAL.INVALID_AUTH_ACCOUNT);
+          }
+          // throw new RUNTIME_ERRORS
+        } else {
+          if (fromAccountAddr !== signAccount.account.addr) {
+            throw new RuntimeError(RUNTIME_ERRORS.GENERAL.INVALID_SECRET_KEY);
+          }
+        }
```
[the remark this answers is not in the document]
vuvth: I created a new function `validateAccountSignature` in `runtime/src/runtime.ts`

packages/runtime/src/ctx.ts
```diff
@@ -154,6 +155,11 @@ export class Ctx implements Context {
     toAccount.amount += BigInt(txParam.amountMicroAlgos); // add 'x' algo to receiver
     this.assertAccBalAboveMin(fromAccount.address);
 
+    if (txParam.payFlags.rekeyTo) {
+      const spend = txParam.payFlags.rekeyTo;
+      fromAccount.rekeyTo(spend);
+    }
```
ratik21, line 161: so we can rekey only for `TransferAlgo` transaction? I think we can rekey account with more txtypes (asset transfer, call app, optin ..etc) as well. Could you please check/confirm?

packages/runtime/src/ctx.ts
```diff
@@ -598,6 +605,8 @@ export class Ctx implements Context {
         this.tx = this.gtxs[idx]; // update current tx to index of stateless
         r = this.runtime.validateLsigAndRun(txParam, this.debugStack);
         this.tx = this.gtxs[0]; // after executing stateless tx updating current tx to default (index 0)
+      } else if (txParam.sign === types.SignType.SecretKey) {
+        this.runtime.validateAccountSignature(txParam);
```
ratik21, line 609: :+1:

packages/runtime/src/types.ts
```diff
@@ -196,8 +196,12 @@ export interface CreatedAppM {
   attributes: SSCAttributesM
 }
 
-export interface RuntimeAccount extends AccountSDK {
+export interface RuntimeAccountI extends AccountSDK {
   name?: string
+  spend: types.AccountAddress
+
+  rekeyTo: (authAccountAddress: types.AccountAddress) => void
+  getSpend: () => types.AccountAddress
```
ratik21, line 204: i think you use `getSpendAddress`

packages/runtime/test/integration/rekey-transaction.ts
```diff
@@ -0,0 +1,348 @@
...
+      syncAccounts();
+      const aliceBalanceAfter = alice.balance();
+      const bobBalanceAfter = bob.balance();
+
+      // transaction fee will paid by `from account` not `signer account`
+      assert.equal(aliceBalanceBefore, aliceBalanceAfter + amount + BigInt(fee));
+      assert.equal(bobBalanceBefore + amount, bobBalanceAfter);
+    });
+
+    it("Should failed because cloneLsig is invalid spend address of alice account", () => {
```
ratik21, line 192: could you also add a failing test, when trying to send ALGO's from "original" account as signer, when account has been rekeyed to lsig ?

packages/runtime/src/account.ts
```diff
@@ -24,8 +25,32 @@ const globalState = "global-state";
 const localStateSchema = "local-state-schema";
 const globalStateSchema = "global-state-schema";
 
+export class RuntimeAccount implements RuntimeAccountI {
+  sk: Uint8Array // signing key (private key or lsig)
+  addr: string
+  name?: string;
+  spend: types.AccountAddress;
```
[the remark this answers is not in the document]
vuvth: I think only sk and addr, because we may change when we rekey account.

packages/runtime/src/ctx.ts
vuvth: In Rekey document they wrote, "A rekey-to transaction is a `payment` type transaction which includes the rekey-to parameter set to a well-formed Algorand address." and in this [docs](https://developer.algorand.org/docs/get-details/transactions/transactions/#payment-transaction) we can see only ALGO transfer is a payment transaction.

packages/runtime/src/account.ts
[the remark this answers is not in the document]
ratik21: ah, right. So only `sk` & `addr`

packages/runtime/src/ctx.ts
ratik21: Could you please test it? I think it's for other types as well (otherwise we would need to update `algob` code as well). Maybe test with `axfer`?

packages/runtime/src/ctx.ts
```diff
@@ -582,6 +577,17 @@ export class Ctx implements Context {
...
+  // apply rekey config on from account
+  rekeyTo (txParam: types.ExecParams): void {
+    // get from account
+    const fromAccount = this.getAccount(webTx.getFromAddress(txParam));
+    // apply rekey
+    if (txParam.payFlags.rekeyTo) {
+      const spend = txParam.payFlags.rekeyTo;
+      fromAccount.rekeyTo(spend);
+    }
+  }
```
ratik21, line 589:
```suggestion
  rekeyTo (txParam: types.ExecParams): void {
    if (!txParam.payFlags.rekeyTo) { return; }
    
    const fromAccount = this.getAccount(webTx.getFromAddress(txParam));
    fromAccount.rekeyTo(txParam.payFlags.rekeyTo);
  }
```

packages/runtime/src/ctx.ts
```diff
@@ -747,6 +753,9 @@ export class Ctx implements Context {
         }
       }
 
+      // apply rekey after pass all logic
+      this.rekeyTo(txParam);
```
ratik21, line 757: cool, should work. Could you please also check few direct API's which we have (which doesn't use processTransactions function)? If not, we're good

ratik21 (approved):

packages/runtime/src/ctx.ts
vuvth: We have some APIs like runtime.{deployApp, updateApp..., } which parameters didn't provide the `rekey` field. So I think we will do it in other PR and ensure Runtime and Algob.Deployer are compatible.

vuvth: > cool, should work. Could you please also check few direct API's which we have (which doesn't use processTransactions function)? If not, we're good

We have some APIs like runtime.{deployApp, updateApp..., } which parameters didn't provide the rekey field. So I think we will do it in other PR and ensure Runtime and Algob.Deployer are compatible.

PR #520, opened 2021-11-16
Tealv5 transaction ops, TxReceipt in runtime

ratik21:
## Proposed Changes

+ added opcodes: `Txnas, Gtxnas, Gtxnsas, Args, Log`. Added tests.
+ Update all transaction functions (eg. `executeTx`, `addAsset`, `addApp` ..etc) to return a transaction receipt. 

NOTE: this introduces a breaking change for `.addAsset` & `.addApp`

## Potential followups
[DONE] integration test for `log` op

### DIFF
--- a/packages/runtime/src/ctx.ts
+++ b/packages/runtime/src/ctx.ts
@@ -11,16 +11,16 @@ import {
   AccountAddress, AccountStoreI,
   AppDeploymentFlags,
   ASADeploymentFlags, AssetHoldingM,
-  Context, DeployedAppTxReceipt, DeployedAssetTxReceipt, ExecutionMode,
-  ID, SSCAttributesM, StackElem, State, Txn, TxReceipt
+  Context, DeployedAppTxReceipt, DeployedAssetTxReceipt, EncTx, ExecutionMode,
+  ID, SSCAttributesM, StackElem, State, TxReceipt
 } from "./types";
 
 const APPROVAL_PROGRAM = "approval-program";
 
 export class Ctx implements Context {
   state: State;
-  tx: Txn;
-  gtxs: Txn[];
+  tx: EncTx;
+  gtxs: EncTx[];
   args: Uint8Array[];
   runtime: Runtime;
   debugStack?: number; //  max number of top elements from the stack to print after each opcode execution.
@@ -31,7 +31,7 @@ export class Ctx implements Context {
   isInnerTx: boolean; // true if "ctx" is switched to an inner transaction
   createdAssetID: number; // Asset ID allocated by the creation of an ASA (for an inner-tx)
 
-  constructor (state: State, tx: Txn, gtxs: Txn[], args: Uint8Array[],
+  constructor (state: State, tx: EncTx, gtxs: EncTx[], args: Uint8Array[],
     runtime: Runtime, debugStack?: number) {
     this.state = state;
     this.tx = tx;
--- a/packages/runtime/src/ctx.ts
+++ b/packages/runtime/src/ctx.ts
@@ -12,7 +12,7 @@ import {
   AppDeploymentFlags,
   ASADeploymentFlags, AssetHoldingM,
   Context, ExecutionMode,
-  ID, SSCAttributesM, StackElem, State, Txn
+  ID, SSCAttributesM, StackElem, State, Txn, TxReceipt
 } from "./types";
 
 const APPROVAL_PROGRAM = "approval-program";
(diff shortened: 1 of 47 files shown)


packages/runtime/src/runtime.ts
```diff
@@ -60,6 +63,14 @@ export class Runtime {
     this.timestamp = 1;
   }
 
+  /**
+   * Returns transaction receipt for a particular transaction
+   * @param txID transaction ID
+   */
+  getTransactionInfo (txID: string): TxReceipt | undefined {
```
robert-zaremba, line 70:
```suggestion
  getTxReceipt (txID: string): TxReceipt | undefined {
```

packages/runtime/src/runtime.ts
```diff
@@ -336,32 +347,38 @@ export class Runtime {
         : asaDef.metadataHash,
       mockSuggestedParams(flags, this.round)
     );
+
+    if (this.ctx.tx === undefined || this.ctx.tx.type !== "acfg") { // could already be defined (if used as a txGroup in this.executeTx())
```
robert-zaremba, line 351: what's "acfg"?

robert-zaremba (commented):
+ in some places we are using `tx` in other `txn`. Let's stick to one - for functions, variables, infixes... If Algorand uses `txn` then maybe we should follow that? On the other hand, in our code we have `tx`. So let's use what makes most of the sense (I think `tx`)

packages/runtime/src/ctx.ts
```diff
@@ -49,6 +49,12 @@ export class Ctx implements Context {
     this.createdAssetID = 0;
   }
 
+  private setAndGetTxnInfo (): TxReceipt {
```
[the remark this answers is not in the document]
ratik21: ok

packages/runtime/src/ctx.ts
```diff
@@ -49,6 +49,12 @@ export class Ctx implements Context {
...
+    const info = { txn: this.tx, txID: this.tx.txID };
+    this.state.txnInfo.set(this.tx.txID, info);
```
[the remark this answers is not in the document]
ratik21: done

packages/runtime/src/runtime.ts
ratik21: It's a type of transaction (asset deploy/modify/delete ..): [https://developer.algorand.org/docs/get-details/dapps/avm/teal/specification/#typeenum-constants](https://developer.algorand.org/docs/get-details/dapps/avm/teal/specification/#typeenum-constants)

I think using an Enum here would be more intuitive. Updating

packages/runtime/src/interpreter/opcode-list.ts
```diff
@@ -4056,3 +4058,177 @@ export class ITxna extends Op {
...
+export class Args extends Arg {
+  /**
+   * Gets the argument value from interpreter.args array.
+   * store the value in _arg variable
+   * @param args Expected arguments: none
+   * @param line line number in TEAL file
+   * @param interpreter interpreter object
+   */
+  constructor (args: string[], line: number, interpreter: Interpreter) {
+    super([...args, "100"], line, interpreter);
```
robert-zaremba, line 4171: why adding "100"?

packages/runtime/src/lib/constants.ts
```diff
@@ -32,6 +32,8 @@ export const ALGORAND_MAX_TX_ACCOUNTS_LEN = 4;
 // the assets and application arrays combined and totaled with the accounts array can not exceed 8
 export const ALGORAND_MAX_TX_ARRAY_LEN = 8;
 export const MAX_INNER_TRANSACTIONS = 16;
+export const ALGORAND_MAX_LOGS_COUNT = 32;
+export const ALGORAND_MAX_LOGS_LENGTH = 1024;
```
robert-zaremba, line 36: let's check if this is defined in JS-SDK as well, if so, we should use SDK directly.

robert-zaremba (approved):

packages/runtime/src/interpreter/interpreter.ts
```diff
@@ -10,7 +10,7 @@ import { Stack } from "../lib/stack";
 import { assertMaxCost, parser } from "../parser/parser";
 import {
   AccountStoreI, ExecutionMode, Operator, SSCAttributesM,
-  StackElem, TEALStack, Txn
+  StackElem, TEALStack, Txn, TxReceipt
```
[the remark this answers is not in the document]
ratik21: alright

packages/runtime/src/interpreter/opcode-list.ts
ratik21: just a mock value, to inherit from super class. In this op the value corresponding to "100" will be fetched from stack

packages/runtime/src/lib/constants.ts
ratik21: unfortunately, this is only exported from `go-algorand`

PR #532, opened 2021-12-08
Update zod version to 3.11.9

vuvth:
## Proposed Changes

+ Updated zob version to 3.11.9.
+ remove deprecated  method (`nonstrict`) and replace by `passthrough`.

### DIFF
--- a/packages/algob/src/internal/core/config/config-validation.ts
+++ b/packages/algob/src/internal/core/config/config-validation.ts
@@ -35,7 +35,7 @@ const ChainType = z.object({
   throwOnCallFailures: z.boolean().optional(),
   loggingEnabled: z.boolean().optional(),
   initialDate: z.string().optional()
-}).nonstrict();
+}).passthrough();
 
 const HttpHeaders = z.record(z.string());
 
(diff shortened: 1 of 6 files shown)


packages/algob/src/internal/core/config/config-validation.ts
```diff
@@ -35,7 +35,7 @@ const ChainType = z.object({
   throwOnCallFailures: z.boolean().optional(),
   loggingEnabled: z.boolean().optional(),
   initialDate: z.string().optional()
-}).nonstrict();
+}).passthrough();
```
ratik21, line 38: what's the difference b/w `nonstrict` & `passthrough`?

packages/runtime/src/errors/validation-errors.ts
```diff
@@ -1,8 +1,7 @@
-import * as z from 'zod';
-import { ZodErrorCode, ZodSuberror } from "zod/lib/src/ZodError";
+import { z } from 'zod';
 
-function parseZodSubError (e: ZodSuberror, indent: number): string {
-  if (e.code === ZodErrorCode.invalid_union) {
+function parseZodSubError (e: z.ZodIssueOptionalMessage, indent: number): string {
+  if (e.code === z.ZodIssueCode.invalid_union) {
```
ratik21, line 4: Are these changes due to version upgradation or was it incorrect before?

vuvth: Because the version update.

packages/algob/src/internal/core/config/config-validation.ts
vuvth: I think it same, but nonstrict is deprecated and zod recommend use passthrough

packages/runtime/src/errors/validation-errors.ts
ratik21: ok

ratik21 (approved):

PR #533, opened 2021-12-09
Update @algo-builder/web package

ratik21:
## Proposed Changes

+ remove nullish coalaesing operator
+ add `forceCreate()`

### DIFF
--- a/packages/web/src/lib/wallectconnect-mode.ts
+++ b/packages/web/src/lib/wallectconnect-mode.ts
@@ -40,16 +40,6 @@ export class WallectConnectSession {
   }
 
   /**
-   * Kills a session if already exists, create new session
-   */
-  async forceCreate (): Promise<void> {
-    if (this.connector.connected) {
-      try { await this.close(); } catch (e) {};
-    }
-    return await this.create();
-  }
-
-  /**
    * Create new session
    * @param force if true, kills an existing session and creates new one.
    * By default force is false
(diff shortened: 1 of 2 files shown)


packages/web/src/lib/wallectconnect-mode.ts
```diff
@@ -39,13 +39,22 @@ export class WallectConnectSession {
...
+    return await this.create();
+  }
+
   /**
    * Create new session
    */
   async create (): Promise<void> {
     if (this.connector.connected) {
       console.warn(`A session is already active`);
-      return;
```
vuvth, line 48: Why you delete this line?

robert-zaremba: good question, probably we should  return a current session if possible.

ratik21: right, this shouldn't be removed. thanks

packages/web/src/lib/wallectconnect-mode.ts
```diff
@@ -39,6 +39,16 @@ export class WallectConnectSession {
     this.wcAccounts = this.connector.accounts;
   }
 
+  /**
+   * Kills a session if already exists, create new session
+   */
+  async forceCreate (): Promise<void> {
```
robert-zaremba, line 45: how about adding a `force` parameter to the function below? By default it will be `false`

ratik21: yes

robert-zaremba: the idea was to remove this function.

robert-zaremba (approved):
pre-approving, let's remove `forceCreate` and use the argument in `create`

packages/web/src/lib/wallectconnect-mode.ts
ratik21: ah yes, i forgot to remove it.

PR #535, opened 2021-12-09
add devmode check for suggested params

jzjones:
Closes #x

## Proposed Changes

Check for devmode when doing firstRound check on suggestedParams, otherwise fails against a fresh dev mode sandbox network

### DIFF
--- a/packages/algob/test/lib/tx.ts
+++ b/packages/algob/test/lib/tx.ts
@@ -14,7 +14,7 @@ import { expectBuilderError, expectBuilderErrorAsync } from "../helpers/errors";
 import { mkEnv } from "../helpers/params";
 import { useFixtureProject, useFixtureProjectCopy } from "../helpers/project";
 import { aliceAcc, bobAcc } from "../mocks/account";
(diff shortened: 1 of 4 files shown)


packages/algob/test/lib/tx.ts
```diff
@@ -70,6 +70,8 @@ describe("Opt-In to ASA", () => {
     };
     sinon.stub(algod.algodClient, "getTransactionParams")
       .returns({ do: async () => mockSuggestedParam } as ReturnType<algosdk.Algodv2['getTransactionParams']>);
+    sinon.stub(algod.algodClient, "genesis")
+      .returns({ do: async () => mockGenesisInfo } as ReturnType<algosdk.Algodv2['genesis']>);
```
robert-zaremba, line 74: this is repeated, shall we put it into a function? Eg: `stubAlgodGenesis(algod)`?

ratik21: yes

ratik21 (approved):

PR #539, opened 2021-12-15
Update parameters in runtime add asset and add app

vuvth:
## Proposed Changes

+ Renamed Runtime.addApp to Runtime.deployApp, Runtime.addAsset to Runtime.deployAsset, Runtime.addAssetDef to Runtime.deployAssetDef.
+ Make order of Runtime.deloyApp like Deloyer.deployApp.
+ Update unit tests.

Please review. Thanks!

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -17,7 +17,6 @@ module.exports = {
   },
   rules:  {
     // "ter-indent": [2, {"FunctionDeclaration": {"parameters": "first"}}],
-
     "import/no-extraneous-dependencies": 0,
     "import/prefer-default-export": "off",
     "max-classes-per-file": 0,
@@ -35,6 +34,7 @@ module.exports = {
     "semi": "off",  // required for a proper work
     "@typescript-eslint/semi": ["error", "always"],
     "@typescript-eslint/strict-boolean-expressions": "off",
-    "sonarjs/cognitive-complexity": ["error", 16]
+    "sonarjs/cognitive-complexity": ["error", 16],
+    "sonarjs/no-identical-functions": 0
   }
 }
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -34,6 +34,6 @@ module.exports = {
     "semi": "off",  // required for a proper work
     "@typescript-eslint/semi": ["error", "always"],
     "@typescript-eslint/strict-boolean-expressions": "off",
-    "sonarjs/cognitive-complexity": ["error", 16],
+    "sonarjs/cognitive-complexity": ["error", 16]
   }
 }
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -35,6 +35,5 @@ module.exports = {
     "@typescript-eslint/semi": ["error", "always"],
     "@typescript-eslint/strict-boolean-expressions": "off",
     "sonarjs/cognitive-complexity": ["error", 16],
-    "sonarjs/no-identical-functions": 0
   }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,20 +50,20 @@
     const appID = receipt.appID;
     ```
 + `getProgram` is moved to `@algo-builder/runtime` from `@algo-builder/algob`.
-+ Renamed `Runtime.addApp` to `Runtime.deployApp`, `Runtime.addAsset` to `Runtime.deployAsset`, `Runtime.addAssetDef` to `Runtime.deployAssetDef`. 
-+ Make order of `Runtime.deloyApp` like `Deloyer.deployApp`.
++ `runtime.addAsset`, `runtime.addAssetDef` and `runtime.addApp` are deprecated.
++ Update `runtime.deloyApp` to be compatible with `deployer.deployApp`. Example:
     * For instance:
     ```javascript
-    //from 
-    Runtime.deployApp(
+    // from 
+    runtime.deployApp(
         flags: AppDeploymentFlags, 
         payFlags: types.TxParams,
         approvalProgram: string, 
         clearProgram: string,
         debugStack?: number
     )
-    //to
-    Runtime.deployApp(
+    // to
+    runtime.deployApp(
         approvalProgram: string, 
         clearProgram: string,
         flags: AppDeploymentFlags, 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,7 +50,27 @@
     const appID = receipt.appID;
     ```
 + `getProgram` is moved to `@algo-builder/runtime` from `@algo-builder/algob`.
-
++ Renamed `Runtime.addApp` to `Runtime.deployApp`, `Runtime.addAsset` to `Runtime.deployAsset`, `Runtime.addAssetDef` to `Runtime.deployAssetDef`. 
++ Make order of `Runtime.deloyApp` like `Deloyer.deployApp`.
+    * For instance:
+    ```javascript
+    //from 
+    Runtime.deployApp(
+        flags: AppDeploymentFlags, 
+        payFlags: types.TxParams,
+        approvalProgram: string, 
+        clearProgram: string,
+        debugStack?: number
+    )
+    //to
+    Runtime.deployApp(
+        approvalProgram: string, 
+        clearProgram: string,
+        flags: AppDeploymentFlags, 
+        payFlags: types.TxParams,
+        debugStack?: number
+    )
+    ```
 ### Bug Fixes
 +  Fix bug substring3 opcode pop wrong order [/#505](https://github.com/scale-it/algo-builder/pull/505), contribution: @vuvth.
 + Fix bug: `runtime.optinToApp` updating state even after opt-in fails. Reported by @patrick
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,7 +50,8 @@
     const appID = receipt.appID;
     ```
 + `getProgram` is moved to `@algo-builder/runtime` from `@algo-builder/algob`.
-+ `runtime.addAsset`, `runtime.addAssetDef` and `runtime.addApp` are deprecated.
++ `runtime.addAsset`, `runtime.addAssetDef` and `runtime.addApp` are deprecated. 
+Please use `runtime.deployASA`, `runtime.deployASADef` and `runtime.deployAdd` instead of the above functions.
 + Update `runtime.deloyApp` to be compatible with `deployer.deployApp`. Example:
     * For instance:
     ```javascript
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,26 +52,7 @@
 + `getProgram` is moved to `@algo-builder/runtime` from `@algo-builder/algob`.
 + `runtime.addAsset`, `runtime.addAssetDef` and `runtime.addApp` are deprecated. 
 Please use `runtime.deployASA`, `runtime.deployASADef` and `runtime.deployAdd` instead of the above functions.
-+ Update `runtime.deloyApp` to be compatible with `deployer.deployApp`. Example:
-    * For instance:
-    ```javascript
-    // from 
-    runtime.deployApp(
-        flags: AppDeploymentFlags, 
-        payFlags: types.TxParams,
-        approvalProgram: string, 
-        clearProgram: string,
-        debugStack?: number
-    )
-    // to
-    runtime.deployApp(
-        approvalProgram: string, 
-        clearProgram: string,
-        flags: AppDeploymentFlags, 
-        payFlags: types.TxParams,
-        debugStack?: number
-    )
-    ```
++ Update `runtime.deloyApp` to be compatible with `deployer.deployApp`.
 ### Bug Fixes
 +  Fix bug substring3 opcode pop wrong order [/#505](https://github.com/scale-it/algo-builder/pull/505), contribution: @vuvth.
 + Fix bug: `runtime.optinToApp` updating state even after opt-in fails. Reported by @patrick
--- a/examples/crowdfunding/test/crowdfundingTest.js
+++ b/examples/crowdfunding/test/crowdfundingTest.js
@@ -93,7 +93,7 @@ describe('Crowdfunding Tests', function () {
     const creationFlags = Object.assign({}, flags);
 
     // create application
-    applicationId = runtime.addApp(
+    applicationId = runtime.deployApp(
       { ...creationFlags, appArgs: creationArgs }, {}, approvalProgram, clearProgram).appID;
     const creatorPk = convert.addressToPk(creator.address);
 
@@ -296,7 +296,7 @@ describe('Crowdfunding Tests', function () {
   it('should be rejected by logic when claiming funds if goal is not met', () => {
     // create application
     const creationFlags = Object.assign({}, flags);
-    const applicationId = runtime.addApp(
+    const applicationId = runtime.deployApp(
       { ...creationFlags, appArgs: creationArgs }, {}, approvalProgram, clearProgram).appID;
 
     // setup escrow account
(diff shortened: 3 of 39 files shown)


CHANGELOG.md
```diff
@@ -50,7 +50,27 @@
     const appID = receipt.appID;
     ```
 + `getProgram` is moved to `@algo-builder/runtime` from `@algo-builder/algob`.
-
++ Renamed `Runtime.addApp` to `Runtime.deployApp`, `Runtime.addAsset` to `Runtime.deployAsset`, `Runtime.addAssetDef` to `Runtime.deployAssetDef`. 
++ Make order of `Runtime.deloyApp` like `Deloyer.deployApp`.
+    * For instance:
```
robert-zaremba, line 55:
```suggestion
+ Update `Runtime.deloyApp` to be compatible with `Deloyer.deployApp`. Example:
```

CHANGELOG.md
```diff
@@ -50,7 +50,27 @@
...
+    )
+    //to
+    Runtime.deployApp(
+        approvalProgram: string, 
+        clearProgram: string,
+        flags: AppDeploymentFlags, 
+        payFlags: types.TxParams,
+        debugStack?: number
+    )
+    ```
```
robert-zaremba, line 73: please add a changelog entry that `runtime.addAsset` ... are deprecated

packages/runtime/src/account.ts
```diff
@@ -269,7 +269,7 @@ export class AccountStore implements AccountStoreI {
    * @param clearProgram application clear program
    * NOTE - approval and clear program must be the TEAL code as string
    */
-  addApp (appID: number, params: AppDeploymentFlags,
+  deployApp (appID: number, params: AppDeploymentFlags,
```
robert-zaremba, line 272: We should do: 
1. rename
2. Create a function with old names and parameters, calling the renamed function and mark it [deprecated](https://stackoverflow.com/a/62991642/297094) , eg:
```
/// @deprecated `deployASA` should be used instead.
addAsset (assetId: number, name: string, asaDef: types.ASADef): modelsv2.AssetParams {
    return this.deployASA(assetId, name, asaDef)
```

robert-zaremba (commented):
We need to deprecate the methods and update in Context as  well. Moreover, please double check that in all our code we don't use deprecated methods.

.eslintrc.js
```diff
@@ -35,6 +34,7 @@ module.exports = {
     "semi": "off",  // required for a proper work
     "@typescript-eslint/semi": ["error", "always"],
     "@typescript-eslint/strict-boolean-expressions": "off",
-    "sonarjs/cognitive-complexity": ["error", 16]
+    "sonarjs/cognitive-complexity": ["error", 16],
+    "sonarjs/no-identical-functions": 0
```
ratik21, line 38: Why are we turning this rule off?

CHANGELOG.md
```diff
@@ -50,7 +50,28 @@
     const appID = receipt.appID;
     ```
 + `getProgram` is moved to `@algo-builder/runtime` from `@algo-builder/algob`.
-
++ `runtime.addAsset`, `runtime.addAssetDef` and `runtime.addApp` are deprecated. 
+Please use `runtime.deployASA`, `runtime.deployASADef` and `runtime.deployAdd` instead of the above functions.
```
ratik21, line 54:
```suggestion
Please use `runtime.deployASA`, `runtime.deployASADef` and `runtime.deployApp` instead of the above functions.
```
```suggestion
Please use `runtime.deployASA`, `runtime.deployASADef` and `runtime.deployAdd` instead of the above functions.
```

CHANGELOG.md
```diff
@@ -50,7 +50,28 @@
...
+        debugStack?: number
+    )
+    // to
+    runtime.deployApp(
+        approvalProgram: string, 
+        clearProgram: string,
+        flags: AppDeploymentFlags, 
+        payFlags: types.TxParams,
+        debugStack?: number
+    )
```
ratik21, line 73: both functions are identical here

examples/crowdfunding/test/happypaths.js
```diff
@@ -101,9 +101,14 @@ describe('Crowdfunding Tests - Happy Paths', function () {
...
 
-    // create application
-    applicationId = runtime.addApp(
-      { ...creationFlags, appArgs: creationArgs }, {}, approvalProgram, clearProgram).appID;
+    // deploy application
+    applicationId = runtime.deployApp(
+      approvalProgram,
+      clearProgram,
+      { ...creationFlags, appArgs: creationArgs },
+      {}
```
ratik21, line 109: i am not sure if this is how we pass params in `deployer.deployApp`. Could you please confirm?

packages/runtime/src/runtime.ts
```diff
@@ -459,23 +487,47 @@ export class Runtime {
...
+    approvalProgram: string, clearProgram: string,
     flags: AppDeploymentFlags, payFlags: types.TxParams,
+    debugStack?: number
+  ): DeployedAppTxReceipt {
+    this.addCtxAppCreateTxn(flags, payFlags);
+    this.ctx.debugStack = debugStack;
+    const txReceipt = this.ctx.deployApp(flags.sender.addr, flags, approvalProgram, clearProgram, 0);
+
+    this.store = this.ctx.state;
+    return txReceipt;
```
ratik21, line 510: why not just use `this.deployApp`?

packages/runtime/src/types.ts
```diff
@@ -215,10 +221,11 @@ export interface AccountStoreI {
   balance: () => bigint
   getApp: (appID: number) => SSCAttributesM | undefined
   getAppFromLocal: (appID: number) => AppLocalStateM | undefined
-  addApp: (appID: number, params: AppDeploymentFlags,
```
ratik21, line 218: why are we removing `addApp` from definition here?

.eslintrc.js
vuvth: Because the deprecated method is same code with new method, just name is different. Should i do it inline instead of config this rule in .eslintrc, in this case?

packages/runtime/src/types.ts
vuvth: Uhm, I remember i added it again. I will fix it.

examples/crowdfunding/test/happypaths.js
vuvth: When i check the deploy.deployApp, this is list of parammeters 
```
async deployApp (
    approvalProgram: string,
    clearProgram: string,
    flags: rtypes.AppDeploymentFlags,
    payFlags: wtypes.TxParams,
    scTmplParams?: SCParams,
    appName?: string)
```
and I just update runtime.deployApp() follow this.

packages/runtime/src/runtime.ts
vuvth: Oh, That is great. Thanks for your suggestion (y).

.eslintrc.js
ratik21: yes, we should do it inline (we don't need to turn this off for whole package). but i don't follow why we need to turn this off, as i don't think we're repeating code here.

examples/crowdfunding/test/happypaths.js
ratik21: difference: `approvalProgram` & `clearProgram` are file names here, and in `runtime.deployApp` it's the TEAL code

ratik21: so we probably need to add `scTmplParams?` as well

.eslintrc.js
robert-zaremba: The warning makes sense - you should not copy the whole function. Instead, you should call the new function from the old one, as I wrote in the review comment: https://github.com/scale-it/algo-builder/pull/539/files#r770378696

examples/crowdfunding/test/happypaths.js
vuvth: Can you give me a short description about scTmplParams?

robert-zaremba: @vuvth you have it in the parameters list: `scTmplParams?: SCParams,`

vuvth: Yes, i added it to runtime.deployApp.

examples/crowdfunding/test/crowdfundingTest.js
```diff
@@ -20,8 +20,11 @@ describe('Crowdfunding Tests', function () {
   let runtime;
   let flags;
   let applicationId;
-  const approvalProgram = getProgram('crowdFundApproval.teal');
-  const clearProgram = getProgram('crowdFundClear.teal');
+  const crowdFundApprovalFileName = 'crowdFundApproval.teal';
+  const crowdFundClearFileName = 'crowdFundClear.teal';
+
+  const crowdFundApprovalProgram = getProgram(crowdFundApprovalFileName);
+  const crowdFundClearProgram = getProgram(crowdFundClearFileName);
```
ratik21, line 27: same as above

packages/runtime/src/account.ts
```diff
@@ -172,11 +172,11 @@ export class AccountStore implements AccountStoreI {
   }
 
   /**
-   * Creates Asset in account's state
+   * Deploy Asset in account's state
    * @param name Asset Name
    * @param asaDef Asset Definitions
    */
-  addAsset (assetId: number, name: string, asaDef: types.ASADef): modelsv2.AssetParams {
+  deployASA (assetId: number, name: string, asaDef: types.ASADef): modelsv2.AssetParams {
```
ratik21, line 179: could you please update the params in description? assetid is missing

packages/runtime/src/ctx.ts
```diff
@@ -163,36 +163,52 @@ export class Ctx implements Context {
...
 
   /**
-   * Add Asset without using asa.yaml file
+   * create asset using asa.yaml file
+   * @deprecated `deployASA` should be used instead.
+   * @param name asset name
+   * @param fromAccountAddr account address
+   * @param flags asa deployment flags
+   */
+  addAsset (
```
ratik21, line 187: i don't think we need to add this here (in `ctx.ts`). User is not exposed to this function, end user will only use the `runtime` API (eg. `runtime.deployApp`).

packages/runtime/src/ctx.ts
```diff
@@ -220,6 +236,21 @@ export class Ctx implements Context {
...
 
+  /**
+   * Deploy Asset without using asa.yaml file
+   * @deprecated `deployASADef` should be used instead.
+   * @param name ASA name
+   * @param asaDef asset defitions
+   * @param fromAccountAddr account address of creator
+   * @param flags ASA Deployment Flags
+   */
+  addASADef (
```
ratik21, line 247: same as above, we don't need to add & deprecate this here

packages/runtime/src/ctx.ts
```diff
@@ -313,6 +344,26 @@ export class Ctx implements Context {
...
+   * NOTE:
+   * - approval and clear program must be the TEAL code as string (not compiled code)
+   * - When creating or opting into an app, the minimum balance grows before the app code runs
+   */
+  addApp (
+    fromAccountAddr: AccountAddress, flags: AppDeploymentFlags,
+    approvalProgram: string, clearProgram: string, idx: number
+  ): DeployedAppTxReceipt {
+    return this.deployApp(fromAccountAddr, flags, approvalProgram, clearProgram, idx);
+  }
```
ratik21, line 365: same

packages/runtime/src/runtime.ts
```diff
@@ -459,23 +479,56 @@ export class Runtime {
...
+    scTmplParams?: SCParams,
+    debugStack?: number
+  ): DeployedAppTxReceipt {
+    this.addCtxAppCreateTxn(flags, payFlags);
+    this.ctx.debugStack = debugStack;
+    const approvalProgram = getProgram(approvalProgramFileName, scTmplParams);
+
+    // if clearProgramFileName is empty => clearProgram = ""
+    // else load it by getProgram
+    const clearProgram = clearProgramFileName ? getProgram(clearProgramFileName, scTmplParams) : "";
```
ratik21, line 529: a) logic should be handled for both - approval & clear programs
b) i think we simply need to check if it's a fileName first, if not then proceed with `getProgram`. something like:
 ```js
  const clearProgram = clearProgramFileName.endsWith(".teal" OR ".py") ? getProgram(clearProgramFileName, scTmplParams) : clearProgramFileName;
   ```

ratik21: BDW, i think we should the compilation in `ctx.deployApp` (so move this logic to `ctx.ts`), this way we will also handle `runtime.getAppInfoFromName`. Otherwise it will always store TEAL code as key even if user passes file name.

And probably which is why you take both filenames & tealcode in examples
```js
  const approvalProgramFileName = 'crowdFundApproval.teal';
  const clearProgramFileName = 'crowdFundClear.teal';
  const approvalProgram = getProgram(approvalProgramFileName);
  const clearProgram = getProgram(clearProgramFileName);
```

packages/runtime/src/runtime.ts
```diff
@@ -459,23 +479,56 @@ export class Runtime {
...
+    const txReceipt = this.ctx.deployApp(flags.sender.addr, flags, approvalProgram, clearProgram, 0);
+
+    this.store = this.ctx.state;
+    return txReceipt;
+  }
+
+  /**
+   * deploy a new application and returns application id
+   * @param approvalProgramFileName application approval program filename
+   * @param clearProgramFileName application clear program filename
```
ratik21, line 508: Rename *approvalProgramFileName* => *approvalProgram*. Update comment (program can be TEAL code OR filename as well)

Same for clear

packages/runtime/src/runtime.ts
```diff
@@ -459,23 +479,56 @@ export class Runtime {
...
+   * @param flags SSCDeployment flags
+   * @param payFlags Transaction parameters
+   * @param debugStack: if passed then TEAL Stack is logged to console after
+   * each opcode execution (upto depth = debugStack)
+   * NOTE - approval and clear program must be the TEAL code as string (not compiled code)
```
ratik21, line 513: We should update this comment now, please check in other places as well

packages/runtime/src/runtime.ts
```diff
@@ -459,23 +479,56 @@ export class Runtime {
...
+   */
+  deployApp (
+    approvalProgramFileName: string,
+    clearProgramFileName: string,
+    flags: AppDeploymentFlags,
+    payFlags: types.TxParams,
+    scTmplParams?: SCParams,
```
ratik21, line 520: please add this param in description as well

packages/runtime/src/types.ts
```diff
@@ -156,6 +162,10 @@ export interface Context {
   ) => DeployedAssetTxReceipt
   optIntoASA: (
     assetIndex: number, address: AccountAddress, flags: types.TxParams) => TxReceipt
+  deployApp: (
+    fromAccountAddr: string, flags: AppDeploymentFlags,
+    approvalProgram: string, clearProgram: string, idx: number
+  ) => DeployedAppTxReceipt
   addApp: (
```
ratik21, line 169: explained in above comment as well, we can simply replace addApp with `deployApp` in ctx. Sorry i should have seen in  the initial review. could you please update?

ratik21 (commented):
Few updates needs to be done.

BDW, could you also look at user guide (under `docs/guide`) as well?

packages/runtime/src/runtime.ts
vuvth: In runtime.updateApp also pass program instead of name file, so why i wrote code like above.

ratik21: i updated it

ratik21 (approved):

PR #540, opened 2021-12-16
Update balance of function

vuvth:
## Proposed Changes

+ make balanceOf return amount(number|bigInt)
+ Unit test for it
+ Update code use new `balanceOf` method.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,7 +50,7 @@
     const appID = receipt.appID;
     ```
 + `getProgram` is moved to `@algo-builder/runtime` from `@algo-builder/algob`.
-
++ Modify `balanceOf` in package `@algo-builder/algob` will return amount(number) of asset account holding. If the account does not hold an asset it will return `undefined`.
 ### Bug Fixes
 +  Fix bug substring3 opcode pop wrong order [/#505](https://github.com/scale-it/algo-builder/pull/505), contribution: @vuvth.
 + Fix bug: `runtime.optinToApp` updating state even after opt-in fails. Reported by @patrick
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,7 +50,7 @@
     const appID = receipt.appID;
     ```
 + `getProgram` is moved to `@algo-builder/runtime` from `@algo-builder/algob`.
-+ Modify `balanceOf` in package `@algo-builder/algob` will return amount(number) of asset account holding. If the account does not hold an asset it will return `undefined`.
++ `balanceOf` in `@algo-builder/algob` package now return amount (number) of asset account holding and won't print them. If the account does not hold an asset it will return 0.
 ### Bug Fixes
 +  Fix bug substring3 opcode pop wrong order [/#505](https://github.com/scale-it/algo-builder/pull/505), contribution: @vuvth.
 + Fix bug: `runtime.optinToApp` updating state even after opt-in fails. Reported by @patrick
--- a/packages/algob/src/lib/status.ts
+++ b/packages/algob/src/lib/status.ts
@@ -13,14 +13,14 @@ export async function balanceOf (
   deployer: Deployer,
   accountAddress: AccountAddress,
   assetID: number
-): Promise<number|bigint|undefined> {
+): Promise<number|bigint> {
   const accountInfo = await deployer.algodClient.accountInformation(accountAddress).do();
   for (const asset of accountInfo.assets) {
     if (asset['asset-id'] === assetID) {
       return asset.amount;
     }
   }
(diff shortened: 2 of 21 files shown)


packages/algob/src/lib/status.ts
```diff
@@ -13,15 +13,14 @@ export async function balanceOf (
...
   const accountInfo = await deployer.algodClient.accountInformation(accountAddress).do();
   for (const asset of accountInfo.assets) {
     if (asset['asset-id'] === assetID) {
-      console.log("Asset Holding Info:", asset);
-      return asset;
+      return asset.amount;
     }
   }
-  return undefined;
+  return 0;
```
ratik21, line 23: this could be misleading, as the asset holding is not found but we return `0`. Maybe it makes sense to return `undefined` here?

packages/algob/test/lib/status.ts
```diff
@@ -0,0 +1,42 @@
...
+  afterEach(async () => {
+    (algod.algodClient.accountInformation as sinon.SinonStub).restore();
+  });
+
+  it("balanceOf should return corrent amount when account hold asset", async () => {
+    const assetID = mockAccountInformation.assets[0]["asset-id"];
+    const amount = mockAccountInformation.assets[0].amount;
+    expect(await balanceOf(deployer, mockAccountInformation.address, assetID)).to.equal(amount);
+  });
+  it("balanceOf should return 0 when account not hold asset", async () => {
```
ratik21, line 38:
```suggestion
  });
  
  it("balanceOf should return 0 when account not hold asset", async () => {
```

ratik21 (commented):
We need a changelog update

CHANGELOG.md
```diff
@@ -50,7 +50,7 @@
     const appID = receipt.appID;
     ```
 + `getProgram` is moved to `@algo-builder/runtime` from `@algo-builder/algob`.
-
++ Modify `balanceOf` in package `@algo-builder/algob` will return amount(number) of asset account holding. If the account does not hold an asset it will return `undefined`.
```
robert-zaremba, line 53:
```suggestion
+ `balanceOf` in `@algo-builder/algob` package now return amount (number) of asset account holding and won't print them. If the account does not hold an asset it will return 0.
```

packages/algob/test/lib/status.ts
```diff
@@ -0,0 +1,43 @@
...
+import { DeployerRunMode } from "../../src/internal/deployer";
+import { DeployerConfig } from "../../src/internal/deployer_cfg";
+import { balanceOf } from "../../src/lib/status";
+import { Deployer } from "../../src/types";
+import { mkEnv } from "../helpers/params";
+import { useFixtureProject } from "../helpers/project";
+import { mockAccountInformation } from "../mocks/tx";
+import { AlgoOperatorDryRunImpl } from "../stubs/algo-operator";
+
+describe("Status library", () => {
```
robert-zaremba, line 14:
```suggestion
describe("Status package", () => {
```

packages/algob/src/lib/status.ts
```diff
@@ -13,12 +13,11 @@ export async function balanceOf (
...
+): Promise<number|bigint|undefined> {
   const accountInfo = await deployer.algodClient.accountInformation(accountAddress).do();
   for (const asset of accountInfo.assets) {
     if (asset['asset-id'] === assetID) {
-      console.log("Asset Holding Info:", asset);
-      return asset;
+      return asset.amount;
     }
   }
   return undefined;
```
robert-zaremba, line 23: let's change it and return `0n` (bigint) here.

packages/algob/src/lib/status.ts
vuvth: @robert-zaremba How do you think about @ratik21 suggestion?

packages/algob/src/lib/status.ts
```diff
@@ -13,12 +13,11 @@ export async function balanceOf (
   deployer: Deployer,
   accountAddress: AccountAddress,
   assetID: number
-): Promise<modelsv2.AssetHolding | undefined> {
+): Promise<number|bigint|undefined> {
```
robert-zaremba, line 16:
```suggestion
): Promise<number|bigint> {
```

packages/algob/src/lib/status.ts
robert-zaremba: @ratik21 , balance should be a number (eg, if you have zero rubels and don't have rubel bank account then you still have 0, not undefined). I will add another function `getAssetHolding` which will follow the old behavior and return undefined if account didn't opt-in.

robert-zaremba (approved):
looks good

PR #541, opened 2021-12-16
Asset holding

robert-zaremba:
## Proposed Changes

+ adding new function to return asset holding and use it in `balanceOf`

### DIFF
--- a/packages/web/src/lib/status.ts
+++ b/packages/web/src/lib/status.ts
@@ -1,9 +1,9 @@
 import type { Algodv2, modelsv2 } from "algosdk";
 
 /**
- * Returns `account` balance of `assetID`. Returns 0 if the account has not
- * opt-in to the given asset id.
- * @param deployer algob deployer
+ * Returns `account` balance of `assetID`. Returns 0 if the account is not
+ * opted-in to the given asset id.
+ * @param algodClient Algodv2 instance
  * @param accountAddress account to return assetholding info
  * @param assetID asset index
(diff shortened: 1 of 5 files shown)


packages/web/src/lib/status.ts
```diff
@@ -0,0 +1,22 @@
+import type { Algodv2, modelsv2 } from "algosdk";
+
+/**
+ * Returns `account` balance of `assetID`. Returns 0 if the account has not
+ * opt-in to the given asset id.
+ * @param deployer algob deployer
```
ratik21, line 6: this param should be `algodClient`

packages/web/src/lib/status.ts
```diff
@@ -0,0 +1,22 @@
+import type { Algodv2, modelsv2 } from "algosdk";
+
+/**
+ * Returns `account` balance of `assetID`. Returns 0 if the account has not
+ * opt-in to the given asset id.
```
ratik21, line 5:
```suggestion
 * opted-in to the given asset id.
```

ratik21 (approved):

packages/web/src/lib/status.ts
robert-zaremba: updated

packages/web/src/lib/status.ts
robert-zaremba: I've updated to "did not opt-in"

PR #550, opened 2022-01-13
Fixbug on ItxnField opcode

vuvth:
## Proposed Changes

+ Allow random address when creating asset and re-config when using the Inner transaction.
+ Teal ItxnField opcode understands address 32 bytes so add tests for this case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Bug Fixes
 
-- Fixbug ItxnField not allow random address. Reported by @patrick
+- `Runtime` env is enforcing address use in ItxnField already `exist`. However, those addresses used in `acfg` or create asset transactions the addresses use for config may not exist. We fixed it in this [PR](https://github.com/scale-it/algo-builder/pull/550). Reported by @patrick
 
 ## v3.0.0 2021-12-22
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Bug Fixes
+
+- Fixbug ItxnField not allow random address. Reported by @patrick
+
 ## v3.0.0 2021-12-22
 
 ### Improvements
--- a/packages/runtime/src/interpreter/opcode.ts
+++ b/packages/runtime/src/interpreter/opcode.ts
@@ -1,4 +1,6 @@
 /* eslint sonarjs/no-identical-functions: 0 */
+import { encodeAddress, isValidAddress } from "algosdk";
+
 import { RUNTIME_ERRORS } from "../errors/errors-list";
 import { RuntimeError } from "../errors/runtime-errors";
 import { GlobalFields, ITxArrFields, ITxnFields, MAX_UINT6, MAX_UINT8, MAX_UINT64, MAX_UINT128, MIN_UINT8, MIN_UINT64, TxArrFields, TxnFields } from "../lib/constants";
@@ -131,6 +133,20 @@ export class Op {
       });
     }
     return b;
+  }
+
+  /**
+   * asserts if given variable type is address
+   * @param a - value to assert (should in bytes)
+   * @param line - line number in TEAL file
+   */
+  assertAddress (a: unknown, line: number): Uint8Array {
+    const bytes = this.assertBytes(a, line);
+    const addr = encodeAddress(bytes);
+    if (!isValidAddress(addr)) {
+      throw new RuntimeError(RUNTIME_ERRORS.TEAL.INVALID_ADDR, { addr: a, line: line });
+    }
+    return bytes;
   }
 
   /**
--- a/packages/runtime/src/interpreter/opcode.ts
+++ b/packages/runtime/src/interpreter/opcode.ts
@@ -140,7 +140,7 @@ export class Op {
    * @param a - value to assert (should in bytes)
    * @param line - line number in TEAL file
    */
-  assertAddress (a: unknown, line: number): Uint8Array {
+  assertAlgorandAddress (a: unknown, line: number): Uint8Array {
     const bytes = this.assertBytes(a, line);
     const addr = encodeAddress(bytes);
     if (!isValidAddress(addr)) {
--- a/packages/runtime/src/lib/itxn.ts
+++ b/packages/runtime/src/lib/itxn.ts
@@ -65,9 +65,8 @@ export function setInnerTxField (
   }
 
   if (addrTxnFields.has(field)) {
-    const assertedVal = op.assertBytes(val, line);
-    const accountState = interpreter.getAccount(assertedVal, line);
-    txValue = Buffer.from(decodeAddress(accountState.address).publicKey);
+    const assertedAddr = op.assertAddress(val, line);
+    txValue = assertedAddr;
   }
 
   const encodedField = TxnFields[interpreter.tealVersion][field]; // eg 'rcv'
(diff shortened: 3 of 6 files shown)


packages/runtime/src/interpreter/opcode.ts
```diff
@@ -133,6 +135,20 @@ export class Op {
     return b;
   }
 
+  /**
+   * asserts if given variable type is address
+   * @param a - value to assert (should in bytes)
+   * @param line - line number in TEAL file
+   */
+  assertAddress (a: unknown, line: number): Uint8Array {
```
robert-zaremba, line 143:
```suggestion
  assertAddress (a: Uint8Array, line: number): Uint8Array {
```

packages/runtime/src/lib/itxn.ts
```diff
@@ -65,9 +65,8 @@ export function setInnerTxField (
   }
 
   if (addrTxnFields.has(field)) {
-    const assertedVal = op.assertBytes(val, line);
-    const accountState = interpreter.getAccount(assertedVal, line);
-    txValue = Buffer.from(decodeAddress(accountState.address).publicKey);
+    const assertedAddr = op.assertAddress(val, line);
+    txValue = assertedAddr;
```
robert-zaremba, line 69:
```suggestion
    txValue = op.assertAddress(val, line);
```

packages/runtime/src/lib/itxn.ts
```diff
@@ -183,6 +182,14 @@ const _getRuntimeAccountAddr = (publickey: Buffer | undefined,
   return _getRuntimeAccount(publickey, interpreter, line)?.addr;
 };
 
+const _getASAConfigAddr = (addr?: Uint8Array): string => {
+  if (addr) {
+    return encodeAddress(addr);
+  }
+  // if addr not declare return ""
```
robert-zaremba, line 189: I think this comment is not needed, the code is short and clear.

packages/runtime/test/src/interpreter/inner-transaction.ts
```diff
@@ -404,33 +404,35 @@ describe("TEALv5: Inner Transactions", function () {
...
+    it(`should fail: invalid ref, not an account`, function () {
       // but a b32 string rep is not an account
       tealCode = `
         itxn_begin
         byte "GAYTEMZUGU3DOOBZGAYTEMZUGU3DOOBZGAYTEMZUGU3DOOBZGAYZIZD42E"
         itxn_field AssetCloseTo
       `;
-      expectRuntimeError(() => executeTEAL(tealCode), RUNTIME_ERRORS.TEAL.ADDR_NOT_VALID);
+      expectRuntimeError(() => executeTEAL(tealCode), RUNTIME_ERRORS.TEAL.INVALID_ADDR);
     });
```
robert-zaremba, line 437: please add a test:
* inner transaction of type asa config, where we put address which is not registered in the runtime (it should pass without failure).

robert-zaremba (commented):
Let's add more tests.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Bug Fixes
+
+- Fixbug ItxnField not allow random address. Reported by @patrick
```
ratik21, line 7: please  desribe a little bit more about the issue. maybe a link to PR?

packages/runtime/src/interpreter/opcode.ts
ratik21, line 143: Can  we rename to `assertValidAlgorandAddr` (OR `assertAddressValid`)?

packages/runtime/src/lib/itxn.ts
ratik21: This is not entirely correct. We only want to remove `interpreter.getAccount` check only in the case of "acfg" transaction. So for example if i do an asset transfer transaction from smart contract's account to `<addr>`, then that "<addr>" must be present in the accounts array (passed in first txn account).

ratik21: Also, we can modify `interpreter.getAccount` function here. The address must be present in the array (if txtype is not "acfg"), but it doesn't have to be an already initialized runtime account

packages/runtime/src/interpreter/opcode.ts
```diff
@@ -133,6 +135,20 @@ export class Op {
...
+    const bytes = this.assertBytes(a, line);
+    const addr = encodeAddress(bytes);
+    if (!isValidAddress(addr)) {
+      throw new RuntimeError(RUNTIME_ERRORS.TEAL.INVALID_ADDR, { addr: a, line: line });
+    }
```
ratik21, line 148: i don't follow this logic -- i think `encodeAddress` will fail if `bytes` is not a valid algorand address. You try to encode first, and then validate.

packages/runtime/test/src/interpreter/inner-transaction.ts
```diff
@@ -404,33 +404,35 @@ describe("TEALv5: Inner Transactions", function () {
...
         itxn_field AssetReceiver
+        int 1
       `;
-      expectRuntimeError(() => executeTEAL(tealCode),
-        RUNTIME_ERRORS.TEAL.ADDR_NOT_FOUND_IN_TXN_ACCOUNT);
 
+      assert.doesNotThrow(() => executeTEAL(tealCode));
+    });
+
+    it(`should fail: invalid ref, not an account`, function () {
```
ratik21, line 443:
```suggestion
    it(`should fail: invalid ref, not a valid account address`, function () {
```

ratik21 (commented):
Need to add test(s) before merging

packages/runtime/src/interpreter/opcode.ts
vuvth: In my view, `unknow` is better because `a` can be `int` or `bytes`(stack element). By accident, they can write contracts like 
`
int 9 
itxn_field AssetCloseTo
` 
we also can't compile if we Uint8Array.

packages/runtime/src/lib/itxn.ts
vuvth: > This is not entirely correct. We only want to remove `interpreter.getAccount` check only in the case of "acfg" transaction. So for example if i do an asset transfer transaction from smart contract's account to `<addr>`, then that "" must be present in the accounts array (passed in first txn account).

Teal allow we declare `itxn_field TypeEnum` before and after `itxn_field <address>` and Teal only wants to know this is the correct address or not.

packages/runtime/src/interpreter/opcode.ts
vuvth: > i don't follow this logic -- i think `encodeAddress` will fail if `bytes` is not a valid algorand address. You try to encode first, and then validate.

If `bytes` is present for an invalid address `encodeAddress` still return answer but when we check with `isValidAddress` It's will return false.

ratik21: > If bytes is present for an invalid address encodeAddress still return answer

is this tested? So something like `encodeAddress(strToBytes("Hello World"))` doesn't break & return an answer/address?

vuvth: > > If bytes is present for an invalid address encodeAddress still return answer
> 
> is this tested? So something like `encodeAddress(strToBytes("Hello World"))` doesn't break & return an answer/address?
isValidAddress address will return false

vuvth: > > If bytes is present for an invalid address encodeAddress still return answer
> 
> is this tested? So something like `encodeAddress(strToBytes("Hello World"))` doesn't break & return an answer/address?

yes and `isValidAddress` address will return false

packages/runtime/src/lib/itxn.ts
```diff
@@ -31,10 +31,13 @@ const byteTxnFields = new Set([
   'ConfigAssetMetadataHash', 'ConfigAssetURL'
 ]);
 
-const addrTxnFields = new Set([
+const acfgAddrTxnFields = new Set([
+  'ConfigAssetManager', 'ConfigAssetReserve', 'ConfigAssetFreeze', 'ConfigAssetClawback'
+]);
+
+const otherTxnFields = new Set([
```
ratik21, line 38: *otherTxnFields* => *otherAddrTxnFields*

ratik21 (approved):

PR #552, opened 2022-01-18
Fix bug in deployer.getDelegatedLsig 

vuvth:
Fix bug when we convert logicSigAccount data in checkpoint to LogicSigAccount object.


ratik21 (approved):
*ACK*

PR #554, opened 2022-01-19
Add `deployer.getCompiledASCInfo`

ratik21:


packages/algob/src/internal/deployer.ts
```diff
@@ -194,6 +195,17 @@ class DeployerBasicMode {
     return await getLsigFromCache(name);
   }
 
+  /**
+   * Returns program compiled info: bytecode (compiled program) and it's hash.
+   * @param name ASC name
+   * @param scTmplParams: Smart contract template parameters (used only when compiling PyTEAL to TEAL)
+   * @returns loaded logic signature from artifacts/cache/<file_name>.teal.yaml
+   */
+  async getCompiledASCInfo (name: string,
```
robert-zaremba, line 204: Maybe we can remove `Info` from the name?

packages/algob/src/internal/deployer.ts
```diff
@@ -194,6 +195,17 @@ class DeployerBasicMode {
...
+    scTmplParams?: SCParams): Promise<ASCCache> {
+    return await getBytecodeAndHash(name, this.algoOp.algodClient, scTmplParams);
+  }
```
robert-zaremba, line 207: Do we support here a program name (not the file name) similarly to `loadLogic`? This is to avoid providing the template params.

packages/algob/src/internal/deployer.ts
ratik21: ok

packages/algob/src/internal/deployer.ts
robert-zaremba: Yeah, I was chatting about it in Discord: https://discord.com/channels/754286650540032030/754286650540032033/934040373628657675
I would propose the following names:
* `getCompiledASC(filename)`
* `getDeployedASC(name)`

with appropriate docs.

ratik21: ok

robert-zaremba (approved):

PR #556, opened 2022-01-24
Fixbug unconver opcode

vuvth:
## Proposed Changes

+ Fix bug in uncover opcode

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,9 +10,10 @@
 
 - `Runtime` env is enforcing address use in ItxnField already `exist`. However, those addresses used in `acfg` or create asset transactions the addresses use for config may not exist. We fixed it in this [PR](https://github.com/scale-it/algo-builder/pull/550). Reported by @patrick
 
-- Fix bug can't get LogicSigAccount `deployer.getDelegatedLsig`
+- Can't get LogicSigAccount from `deployer.getDelegatedLsig`.
 
-- Fix bug `uncover` opcode push/pop wrong order.
+- `uncover` opcode push/pop wrong order.
+
 ## v3.0.0 2021-12-22
 
 ### Improvements
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,9 @@
 
 - `Runtime` env is enforcing address use in ItxnField already `exist`. However, those addresses used in `acfg` or create asset transactions the addresses use for config may not exist. We fixed it in this [PR](https://github.com/scale-it/algo-builder/pull/550). Reported by @patrick
 
-- Fixbug can't get LogicSigAccount `deployer.getDelegatedLsig`
+- Fix bug can't get LogicSigAccount `deployer.getDelegatedLsig`
 
+- Fix bug `uncover` opcode push/pop wrong order.
 ## v3.0.0 2021-12-22
 
 ### Improvements
--- a/packages/runtime/test/src/interpreter/opcode-list.ts
+++ b/packages/runtime/test/src/interpreter/opcode-list.ts
@@ -5560,17 +5560,69 @@ describe("Teal Opcodes", function () {
       );
     });
 
+    it("uncover: n = 1", () => {
+      push(stack, 4); // stack state = [1, 2, 3, 4]
+      const op = new Uncover(['1'], 1);
+
+      // move top to below 1 element
+      op.execute(stack);
+
+      // stack state = [1, 2, 4, 3]
+      assert.equal(stack.length(), 4);
+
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -10,8 +10,9 @@
 
 - `Runtime` env is enforcing address use in ItxnField already `exist`. However, those addresses used in `acfg` or create asset transactions the addresses use for config may not exist. We fixed it in this [PR](https://github.com/scale-it/algo-builder/pull/550). Reported by @patrick
 
-- Fixbug can't get LogicSigAccount `deployer.getDelegatedLsig`
+- Fix bug can't get LogicSigAccount `deployer.getDelegatedLsig`
 
+- Fix bug `uncover` opcode push/pop wrong order.
```
robert-zaremba, line 15:
```suggestion
- Can't get LogicSigAccount from `deployer.getDelegatedLsig`
- `uncover` opcode push/pop wrong order.
```

packages/runtime/test/src/interpreter/opcode-list.ts
```diff
@@ -5560,17 +5560,69 @@ describe("Teal Opcodes", function () {
       );
     });
 
+    it("uncover: n = 1", () => {
+      push(stack, 4); // stack state = [1, 2, 3, 4]
+      const op = new Uncover(['1'], 1);
```
robert-zaremba, line 5565:
```suggestion
      const op = new Uncover([1n], 1);
```
I think we can use `1n` (big number notation) instead of string. Could you check please?

packages/runtime/test/src/interpreter/opcode-list.ts
```diff
@@ -5560,17 +5560,69 @@ describe("Teal Opcodes", function () {
...
+      assert.equal(stack.pop(), 3n);
+      assert.equal(stack.pop(), 4n);
+      assert.equal(stack.pop(), 2n);
+      assert.equal(stack.pop(), 1n);
+    });
+
+    it("uncover: n = 0 - stack state should not change", () => {
+      push(stack, 4); // stack state = [1, 2, 3, 4]
+
+      const op = new Uncover(['0'], 1);
```
robert-zaremba, line 5582: ditto

packages/runtime/test/src/interpreter/opcode-list.ts
```diff
@@ -5560,17 +5560,69 @@ describe("Teal Opcodes", function () {
...
+      op.execute(stack);
+
+      // stack state = [1, 2, 3, 4]
+      assert.equal(stack.length(), 4);
+
+      assert.equal(stack.pop(), 4n);
+      assert.equal(stack.pop(), 3n);
+      assert.equal(stack.pop(), 2n);
+      assert.equal(stack.pop(), 1n);
```
robert-zaremba, line 5591: why not using (same in the other test below:
```
assert.deepEqual(stack, [1n, 2n, 3n, 4n])
```

https://www.chaijs.com/api/assert/#method_deepequal

robert-zaremba (approved):
Preapproving. Left few comments. Let's try to use big numbers rather than string

packages/runtime/test/src/interpreter/opcode-list.ts
vuvth: No, We can't. Constructor of opcode is Opcode(args: string[], line: number) so args is string[] we can't use BigInt.

packages/runtime/test/src/interpreter/opcode-list.ts
vuvth: We can't because stack and array [1n, 2n, 3n, 4n] is not same type. Shell we create utils function to compare it?

packages/runtime/test/src/interpreter/opcode-list.ts
vuvth: No, we can't.

packages/runtime/test/src/interpreter/opcode-list.ts
robert-zaremba: ah, it will be part of parsing. Thanks.

packages/runtime/test/src/interpreter/opcode-list.ts
robert-zaremba: I thought we are using array here.

vuvth: Stack is [class](https://github.com/scale-it/algo-builder/blob/master/packages/runtime/src/lib/stack.ts#L9)

PR #561, opened 2022-01-25
fix github workflow

robert-zaremba:


.github/workflows/05-examples.yaml
```diff
@@ -0,0 +1,125 @@
...
+      - develop
+
+jobs:
+  run-examples:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      # - name: Setup Node
+      #   uses: actions/setup-node@v2-beta
```
ratik21, line 20: is this still in progress?

PR #562, opened 2022-01-26
Fixbug int opcode can't start with 0x

vuvth:
## Proposed Changes

+ Support prefix 0x(hex) and 0(oct) for `Int` opcode
+ Add assertNumber for assert data is number with 3 format dec, oct, hex.
+ Unit tests.


packages/runtime/src/lib/parsing.ts
```diff
@@ -21,6 +21,26 @@ export function assertOnlyDigits (val: string, line: number): void {
...
+ * return val if format is correct
+ * @param val : string
+ */
+export function assertNumber (val: string, line: number): string {
+  if (reOct.test(val)) {
+    // typescript use 0o postfix instade of 0 postfix for oct format.
+    return "0o".concat(val.substring(1));
+  }
+
+  if (reDec.test(val) || reHex.test(val)) return val;
```
robert-zaremba, line 35: before returning, should we try to parse the number and check if it's smaller than `1<<64` (2**64)?

packages/runtime/test/src/parser/parsing.ts
```diff
@@ -84,6 +84,39 @@ describe("Get Encoding for Byte Data", () => {
   });
 });
 
+describe("assertNumber", function () {
+  it("should pass with hex and dec", () => {
+    const hexValue = "0xaa3C3";
+    assert.deepEqual(assertNumber(hexValue, 1), hexValue);
+    const decValue = "343434";
+    assert.deepEqual(assertNumber(decValue, 1), decValue);
```
robert-zaremba, line 92: don't need to use `deepEqual`. Simple equal is perfect here

robert-zaremba (approved):
pre-approving

packages/runtime/src/lib/parsing.ts
vuvth: We check it in Int opcode [here](https://github.com/scale-it/algo-builder/blob/dc544299516315c33ce0269d01fcbe18decbc68f/packages/runtime/src/interpreter/opcode-list.ts#L2154)

packages/runtime/test/src/parser/parsing.ts
vuvth: Thanks!

packages/runtime/src/lib/parsing.ts
```diff
@@ -21,6 +21,26 @@ export function assertOnlyDigits (val: string, line: number): void {
   }
 }
 
+/**
+ * assert if string is valid number respesent.
```
ratik21, line 25:
```suggestion
 * assert if string is valid algorand number respesentation (octal / hex / unsigned int).
```

ratik21 (commented):
We need changelog update

PR #567, opened 2022-01-29
Algobpy

robert-zaremba:
Use strict parsing rules when decoding PyTEAL teamplate parameters using `algobpy`. Previously, on decode failure, the script was continuing with partially update template params, now we fail with an exception.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
   * `getCompiledASC`: compiles a contract in real time, returns info after compilation (eg. bytecode, bytecode hash, timestamp etc).
   * `getDeployedASC`: returns cached program (in artifacts/cache) compiled info(bytecode, hash, filename etc).
 - Added `sandbox-up-dev` and `sandbox-reset` commands into Makefile in `infrastructure/`.
-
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -8,7 +8,7 @@
   * `getCompiledASC`: compiles a contract in real time, returns info after compilation (eg. bytecode, bytecode hash, timestamp etc).
   * `getDeployedASC`: returns cached program (in artifacts/cache) compiled info(bytecode, hash, filename etc).
 - Added `sandbox-up-dev` and `sandbox-reset` commands into Makefile in `infrastructure/`.
-
+- Use strict parsing rules when decoding PyTEAL teamplate parameters using `algobpy`. Previously, on decode failure, the script was continuing with partially update template params, now we fail with an exception.
```
ratik21, line 11:
```suggestion
- Use strict parsing rules when decoding PyTEAL teamplate parameters using `algobpy`. Previously, on decode failure, the script was continuing with partially updated template params, now we fail with an exception.
```

ratik21 (approved):

PR #568, opened 2022-02-01
Get compiled asc

robert-zaremba:
Closes #x

## Proposed Changes

+ renamed `getCompiledASC` to `compileASC` - it's shorter an more clear. Also reuse the `ensureCompiled`.
+ deprecate `ensureCompiled` in favor of `compileASC`
+ remove useless async/ await - if no need to wait in the middle of a function we can return promise directly.

### DIFF
--- a/docs/guide/deployer.md
+++ b/docs/guide/deployer.md
@@ -170,11 +170,11 @@ You can learn more about Stateless Smart Contracts [here](https://developer.algo
 
 #### Compile contracts
 
-You can use the deployer API to compile smart contracts and get the contract's bytecode, hash, compilation timestamp etc:
-  * `getCompiledASC`: compiles a contract in real time, returns info after compilation (eg. bytecode, bytecode hash, timestamp etc).Eg.
+You can use the deployer API to compile smart contracts (ASC) and get the contract's bytecode, hash, compilation timestamp etc:
+  * `compiledASC`: compiles a contract in real time, returns `ASCCache` object. Example:
   ```js
-  const info = await deployer.getCompiledASC('buyback-lsig.py', { ARG_DAO_APP_ID: 23 });
+  const info = await deployer.compileASC('buyback-lsig.py', { ARG_DAO_APP_ID: 23 }, false);
   const bytecode = info.compiled;
   ```
 
-  * `getDeployedASC`: Similar to above, but instead of compiling in real time, it returns cached program (in artifacts/cache) compiled info(bytecode, hash, filename etc).
+  * `getDeployedASC`: Similar to above, but instead of compiling, it returns cached program (from artifacts/cache) by deployment name. 
--- a/packages/algob/src/internal/deployer.ts
+++ b/packages/algob/src/internal/deployer.ts
@@ -10,7 +10,7 @@ import * as algosdk from "algosdk";
 
 import { txWriter } from "../internal/tx-log-writer";
 import { AlgoOperator } from "../lib/algo-operator";
-import { CompileOp, getBytecodeAndHash } from "../lib/compile";
+import { CompileOp } from "../lib/compile";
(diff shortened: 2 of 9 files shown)


packages/algob/src/internal/deployer.ts
```diff
@@ -242,8 +256,8 @@ class DeployerBasicMode {
    * Send signed transaction to network and wait for confirmation
    * @param rawTxns Signed Transaction(s)
    */
-  async sendAndWait (rawTxns: Uint8Array | Uint8Array[]): Promise<ConfirmedTxInfo> {
-    return await this.algoOp.sendAndWait(rawTxns);
+  sendAndWait (rawTxns: Uint8Array | Uint8Array[]): Promise<ConfirmedTxInfo> {
+    return this.algoOp.sendAndWait(rawTxns);
```
ratik21, line 260: Why do we remove `async/await` from here? I thought sending and waiting tx confirmation from network should be promise based? And note: you return `Promise<ConfirmedTxInfo>`

packages/algob/src/types.ts
```diff
@@ -653,24 +653,23 @@ export interface Deployer {
...
+   * @param force: if force is true file will be compiled for sure, even if it's checkpoint exist
    */
-  getCompiledASC: (name: string, scTmplParams?: SCParams) => Promise<ASCCache>
+  compileASC: (name: string, scTmplParams?: SCParams, force?: boolean) => Promise<ASCCache>
 
   /**
-   * Returns cached program (in artifacts/cache) compiled info(bytecode, hash, filename etc).
-   * @param name ASC file name
+   * Returns cached program (from artifacts/cache) `ASCCache` object by deployment name.
+   * @param name ASC name used during deployment
```
ratik21, line 672: currently we use name only for App (since we only "deploy" an application). Maybe we change to `getDeployedApp`?

docs/guide/deployer.md
```diff
@@ -170,11 +170,11 @@ You can learn more about Stateless Smart Contracts [here](https://developer.algo
...
+  * `compiledASC`: compiles a contract in real time, returns `ASCCache` object. Example:
   ```js
-  const info = await deployer.getCompiledASC('buyback-lsig.py', { ARG_DAO_APP_ID: 23 });
+  const info = await deployer.compileASC('buyback-lsig.py', { ARG_DAO_APP_ID: 23 }, false);
   const bytecode = info.compiled;
   ```
 
-  * `getDeployedASC`: Similar to above, but instead of compiling in real time, it returns cached program (in artifacts/cache) compiled info(bytecode, hash, filename etc).
\ No newline at end of file
+  * `getDeployedASC`: Similar to above, but instead of compiling, it returns cached program (from artifacts/cache) by deployment name.
```
ratik21, line 180: `it returns cached program (from artifacts/cache) by deployment name.`

This is not true. 
a) we only use names for app. 
b) even for app as well, appName is used in checkpoint only, in the `/cache` we still store by respective file names ( this is because we have both approval and clear program's compiled metadata to store)

packages/algob/src/internal/deployer.ts
robert-zaremba: Yes, this is what the return type is. 
we don't need async / await here because we don't need result of the function here, it can stay in prromise

docs/guide/deployer.md
robert-zaremba: I know, I wanted to already update it, because we really need to support app names.
In the meantime, I can change from:
> by deployment name.
to
> by filename

packages/algob/src/internal/deployer.ts
robert-zaremba: Or as you suggested below, we can change to `getDeployedApp`

ratik21 (approved):

PR #569, opened 2022-02-01
Fixbug from address is alway application address when use inner Tx.

vuvth:
## Proposed Changes

+ Fix bug sender is always application address in inner tx. 
+ Test for this case(rekey-inner-tx.ts)


packages/runtime/src/lib/itxn.ts
```diff
@@ -202,10 +202,15 @@ const _getASAConfigAddr = (addr?: Uint8Array): string => {
 /* eslint-disable sonarjs/cognitive-complexity */
 export function parseEncodedTxnToExecParams (tx: EncTx,
   interpreter: Interpreter, line: number): types.ExecParams {
+  // signer is the contract
+  const appID = interpreter.runtime.ctx.tx.apid ?? 0;
+  const applicationAccount = getApplicationAddress(appID);
```
ratik21, line 207:
```suggestion
  const appAddress = getApplicationAddress(appID);
```

packages/runtime/src/lib/itxn.ts
```diff
@@ -202,10 +202,15 @@ const _getASAConfigAddr = (addr?: Uint8Array): string => {
...
+
   // initial common fields
   const execParams: any = {
     sign: types.SignType.SecretKey,
-    fromAccount: _getRuntimeAccount(tx.snd, interpreter, line), // sender is the contract
+    fromAccount: { addr: applicationAccount, sk: Buffer.from([]) }, // signer is the contract
```
ratik21, line 212:
```suggestion
    fromAccount: { addr: appAddress, sk: Buffer.from([]) }, // signer is the contract
```

packages/runtime/test/integration/rekey-inner-tx.ts
```diff
@@ -0,0 +1,139 @@
...
+  let master: AccountStore;
+  let alice: AccountStore;
+  let bob: AccountStore;
+  let john: AccountStore;
+
+  let runtime: Runtime;
+
+  let appInfo: DeployedAppTxReceipt;
+
+  let txnParams: types.ExecParams;
```
ratik21, line 35:
```suggestion
  let runtime: Runtime;
  let appInfo: DeployedAppTxReceipt;
  let txnParams: types.ExecParams;
```

packages/runtime/src/lib/itxn.ts
ratik21: What if i rekey the contract to another account? In that case i would need to change the signer, ryt? Can i set `authAddr` in an inner transaction? Could you please check?

vuvth: > What if i rekey the contract to another account? In that case i would need to change the signer, ryt? Can i set `authAddr` in an inner transaction? Could you please check?

I tried to do it before, but I stuck with rekeyTo on the inner transaction. I think it is fine now on teal v6. I will check it.

ratik21 (approved):
Please add TODO, and update changelog before merging

PR #572, opened 2022-02-07
Use Algo DAO app account instead of lsig

ratik21:
Use app account instead of `deposit_lsig`


robert-zaremba (approved):
lgtm, please add changelog entry

PR #573, opened 2022-02-07
Enable implicit account

vuvth:
## Proposed Changes

+ Enable transfer ALGO to implicit account. 
+ Tests


packages/runtime/src/ctx.ts
```diff
@@ -615,6 +615,15 @@ export class Ctx implements Context {
       // https://developer.algorand.org/docs/features/asc1/stateful/#the-lifecycle-of-a-stateful-smart-contract
       switch (txParam.type) {
         case types.TransactionType.TransferAlgo: {
+          // if to account not exist in runtime env
```
robert-zaremba, line 618:
```suggestion
          // if to account doesn't exist in runtime env
```

packages/runtime/src/runtime.ts
```diff
@@ -202,6 +202,17 @@ export class Runtime {
...
+  /**
+   * Update sk for account.
+   * @param address account address
+   * @param sk secret key of account
+   */
+  updateAccountSecretKey (address: string, sk: Uint8Array): void {
+    const account = this.getAccount(address);
+    account.account.sk = sk;
+    this.store.accounts.set(address, account);
+  }
```
robert-zaremba, line 214: why do we need this function? I don't think we should expose that in API.

vuvth: > why do we need this function? I don't think we should expose that in API.

If we don't provide this method. A user will be modify `runtime.store.accounts` Map directly. Is it OK?

robert-zaremba: But where and why the user should modify the account's store key?

vuvth: Oh, I recheck In runtime we don't need sk to sign transaction. My bad. I think we don't want this method.

robert-zaremba: yes, let's remove. I will pre-approve

robert-zaremba (approved):
lgtm

PR #575, opened 2022-02-08
Fix: close remainer to not remove auth account for rekeyed account

vuvth:
## Proposed Changes

+ Fixbug not throw error when closeRemainderTo equal sender
+ Fixbug not remove auth/spend address when close account 
+ Add test cases for both case above.


packages/runtime/src/ctx.ts
```diff
@@ -374,6 +375,17 @@ export class Ctx implements Context {
     }
   }
 
+  /**
+  * Verify closeRemainderTo field is different with fromAccountAddr
```
robert-zaremba, line 379:
```suggestion
  * Verify closeRemainderTo field is different than fromAccountAddr
```

packages/runtime/test/integration/rekey/rekey-transaction.ts
```diff
@@ -409,6 +481,33 @@ describe("Re-keying transactions", function () {
...
+        }
+      };
+
+      runtime.executeTx(txParams);
+
+      // check account state after close
+      syncAccounts();
+      assert.equal(lsigAccount.balance(), 0n);
+      assert.equal(alice.balance(), lsigBalanceBefore + aliceBalanceBefore - BigInt(fee));
+      assert.equal(lsigAccount.getSpendAddress(), lsigAccount.address);
```
robert-zaremba, line 509: this is almost copy  paste in few palces. Let's create a function for that:

```
function verifyCloseAccount(runtime, from, toAddress) {
  txParams = { ... }
  runtime.executeTx(txParams);
  syncAccounts()
  assert...
}
```

robert-zaremba (commented):
Let's reduce copy-paste

packages/runtime/test/integration/rekey/rekey-transaction.ts
```diff
@@ -0,0 +1,430 @@
...
+    // balance before rekey
+    const fromAccountBalanceBefore = from.balance();
+    const toAccountBalanceBefore = to.balance();
+
+    // transfer ALGO by spend account
+    mkTxAlgoTransferFromAccount(
+      runtime, signer, from, to, algoAmount,
+      {
+        totalFee: FEE
+      }
```
robert-zaremba, line 161:
```suggestion
      { totalFee: FEE }
```

robert-zaremba (approved):
Look great, thanks!

packages/runtime/test/integration/rekey/rekey-transaction.ts
robert-zaremba: we can do something similar below as well

PR #581, opened 2022-02-10
feat: KeyRegistration transaction

ratik21:


packages/runtime/src/ctx.ts
```diff
@@ -645,6 +645,11 @@ export class Ctx implements Context {
           r = this.transferAsset(txParam);
           break;
         }
+        case types.TransactionType.KeyRegistration: {
```
vuvth, line 648: https://developer.algorand.org/docs/run-a-node/participate/online/ 
I think key registration will happened after 320 rounds. what happened if will call one more key registration transaction before 320 rounds? This transaction will failed or successful.

ratik21: in runtime this is a `noop`, so transaction is always accepted. Is there a use case that a user will use key registration transaction in runtime?

PR #587, opened 2022-02-16
Runtime default accounts

sczembor:
Closes #x

## Proposed Changes

+ `runtime.defaultAccounts` - a list of pre-generated 16 accounts with pre-defined addresses and keys, each with 1e9 microAlgos (1e3 Algos)
+ `runtime.resetDefaultAccounts()` - will recreate the default accounts (reset their state).
+ unit tests that cover new scenarios when  `runtime.defaultAccounts` and `runtime.resetDefaultAccounts()` are used.
+ changed `bond-token-flow.js` test to use runtime.defaultAccounts.
+ updated the `guide/testing-teal.md` section with the defaultAccounts explanation


packages/runtime/test/src/runtime.ts
```diff
@@ -959,3 +959,91 @@ describe("Stateful Smart Contracts", function () {
...
+
+  function syncAccounts (): void {
+    [alice, bob] = runtime.defaultAccounts;
+    charlie = runtime.getAccount(charlie.address);
+  }
+  this.beforeEach(() => {
+    runtime = new Runtime([charlie]);
+    [alice, bob] = runtime.defaultAccounts;
+  })
+  it("Should have an address", () => {
```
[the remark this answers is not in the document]
robert-zaremba: let's add some room

packages/runtime/src/runtime.ts
```diff
@@ -66,6 +69,38 @@ export class Runtime {
     this.timestamp = 1;
   }
 
+  get defaultBalance (): number {
+    return 1e9; // 1000 Algos
```
ratik21, line 73: Let's use 100 ALGO's (`100 * 1e6`) as the default balance. (same as the test accounts on a ganache mainnet fork)

packages/runtime/src/account.ts
```diff
@@ -492,3 +492,185 @@ export class BaseModelI implements BaseModel {
     throw new Error("get_obj_for_encoding Not Implemented");
   }
 }
+
+//list of 16 predefined accounts 
+export const defaultSDKAccounts = {
```
ratik21, line 497: let's put this somewhere in `constants` (or maybe `default_accounts.ts`)

packages/runtime/src/runtime.ts
```diff
@@ -66,6 +69,38 @@ export class Runtime {
...
+  }
+
+  /**
+   * Returns a list of initialized default accounts created using static accountSDK from account.ts
+   *  and funded with default balance
```
ratik21, line 78:
```suggestion
   *  and funded with default balance (100 ALGO)
```

packages/runtime/src/runtime.ts
```diff
@@ -66,6 +69,38 @@ export class Runtime {
...
+   * @returns list of AccountStore
+   */
+  _setupDefaultAccounts (): AccountStore[] {
+    const balance = this.defaultBalance;
+    const accounts = Object.values(defaultSDKAccounts)
+      .map(accountInfo => new AccountStore(balance, accountInfo));
+    this.initializeAccounts(accounts);
```
ratik21, line 85: Do we need to do this here? At line `60` can we intialize all accounts at a single place? Eg. 
```js
this.initializeAccounts([...accounts, ...this._defaultAccounts]);
```

packages/runtime/src/runtime.ts
```diff
@@ -66,6 +69,38 @@ export class Runtime {
...
+  resetDefaultAccounts (): void {
+    this._defaultAccounts = this._setupDefaultAccounts();
+  }
+
+  /**
+   * Getter for _defaultAccounts, returns a synced version of the accounts list
+   * @returns list of AccountStore
+   */
+  defaultAccounts (): AccountStore[] {
+    return this._defaultAccounts.map((account) => this.getAccount(account.address));
```
ratik21, line 101: hmm, could you please explain this? We're already inside runtime, do we need to use "this.getAccount" (sync) here? Sync is required when you declare variables outside the runtime's context (eg. in a test), but need to use the updated state after each transaction.

packages/runtime/test/src/runtime.ts
```diff
@@ -959,3 +959,91 @@ describe("Stateful Smart Contracts", function () {
...
+    runtime.resetDefaultAccounts();
+    syncAccounts();
+
+    assert.equal(initialAliceBalance, alice.balance());
+    assert.equal(initialBobBalance, bob.balance());
+  });
+
+  it("Should not reset the state of the other accounts stored in runtime", () => {
+    const initialCharlieBalance = charlie.balance();
+    const ALGOTransferTxParam: types.AlgoTransferParam = {
```
ratik21, line 1033: wrong naming convention for javascript variables. use `algoTransferTxParam`

packages/runtime/test/src/runtime.ts
```diff
@@ -959,3 +959,91 @@ describe("Stateful Smart Contracts", function () {
...
+describe("Deafult Accounts", function () {
+  let alice: AccountStore;
+  let bob: AccountStore;
+  let charlie = new AccountStore(minBalance);
+  let runtime: Runtime;
+  const amount = 1e6;
+  const fee = 1000;
+
+  function syncAccounts (): void {
+    [alice, bob] = runtime.defaultAccounts();
```
ratik21, line 972: just for note: one idea for using default Accounts was to avoid running sync

packages/runtime/test/src/runtime.ts
```diff
@@ -959,3 +959,91 @@ describe("Stateful Smart Contracts", function () {
...
+
+    runtime.executeTx(ALGOTransferTxParam);
+
+    syncAccounts();
+
+    assert.equal(initialAliceBalance, alice.balance() + BigInt(amount) + BigInt(fee));
+    assert.equal(initialBobBalance + BigInt(amount), bob.balance());
+  });
+
+  it("Should reset the state of the deafult accounts", () => {
```
ratik21, line 1008:
```suggestion
  it("Should reset the state of the default accounts", () => {
```

packages/runtime/test/src/runtime.ts
```diff
@@ -959,3 +959,91 @@ describe("Stateful Smart Contracts", function () {
...
+    runtime = new Runtime([charlie]);
+    [alice, bob] = runtime.defaultAccounts();
+  });
+
+  it("Should be properly initialized", () => {
+    assert.exists(alice.address);
+    assert.equal(alice.balance(), BigInt(runtime.defaultBalance), "Alice balance must be correct");
+  });
+
+  it("Should update the state of the accounts", () => {
```
ratik21, line 985:
```suggestion
  it("Should update the state of the default accounts", () => {
```

examples/bond/test/bond-token-flow.js
```diff
@@ -16,28 +16,24 @@ const { buyTxRuntime, issueTx } = require('../scripts/run/common/common');
...
+  const runtime = new Runtime([master, issuerAddress]);
+  [appManager, bondTokenCreator, elon, bob, dex1, dex2] = runtime.defaultAccounts();
 
-  let runtime;
   let flags;
   let applicationId;
   let issuerLsigAddress;
   let lsig;
 
   this.beforeAll(async function () {
```
vuvth, line 36: We don't use `await` in this.beforeAll. can we remove async at here?

robert-zaremba: the async here is not needed at all.

packages/runtime/src/account.ts
robert-zaremba: I don't think we should put all possible constants to `contstants`. I think here is OK, it's `accounts` module, so easy to find. `default_accounts.ts` is also OK.

packages/runtime/src/runtime.ts
robert-zaremba: I'm pretty sure, that after transaction, runtime will create a new instance (copy) of account, so the `this._defaultAccounts` won't be updated.

packages/runtime/src/runtime.ts
sczembor: I agree, but then every time a `resetDefaultAccount()` method is called the `this.initializeAccounts()` will have to be called inside of it anyways. I don't know which is better

sczembor: The change would look something like that: eg. 
from this:
` resetDefaultAccounts (): void {
    this._defaultAccounts = this._setupDefaultAccounts();
  }`
to this:
` resetDefaultAccounts (): void {
    this._defaultAccounts = this._setupDefaultAccounts();
    this.initializeAccounts(this._defaultAccounts);
  }`

ratik21: ah, ok. Then let's keep what we have currently.

robert-zaremba (approved):

PR #589, opened 2022-02-16
Add keyreg to inner tx

vuvth:
## Proposed Changes

+ support `keyreg` transaction to inner transaction.
+ Test cases.
+ updated changelog

### DIFF
--- a/packages/algob/test/lib/tx.ts
+++ b/packages/algob/test/lib/tx.ts
@@ -774,7 +774,7 @@ describe("Update transaction test in run mode", () => {
       type: wtypes.TransactionType.UpdateApp,
       sign: wtypes.SignType.SecretKey,
       fromAccount: bobAcc,
-      appID: appInfo?.appID,
+      appID: appInfo?.appID as number,
       newApprovalProgram: "approval.teal",
       newClearProgram: "clear.teal",
       payFlags: {}
--- a/packages/runtime/test/integration/atomic-transfer.ts
+++ b/packages/runtime/test/integration/atomic-transfer.ts
@@ -75,8 +75,8 @@ describe("Algorand Smart Contracts - Atomic Transfers", function () {
         payFlags: { totalFee: 1000 }
       }
     ];
-    const initialJohnAssets = john.getAssetHolding(assetId)?.amount;
-    const initialAliceAssets = alice.getAssetHolding(assetId)?.amount;
+    const initialJohnAssets = john.getAssetHolding(assetId)?.amount as bigint;
+    const initialAliceAssets = alice.getAssetHolding(assetId)?.amount as bigint;
     assert.isDefined(initialJohnAssets);
     assert.isDefined(initialAliceAssets);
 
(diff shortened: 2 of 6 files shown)


packages/algob/test/lib/tx.ts
```diff
@@ -774,7 +774,7 @@ describe("Update transaction test in run mode", () => {
       type: wtypes.TransactionType.UpdateApp,
       sign: wtypes.SignType.SecretKey,
       fromAccount: bobAcc,
-      appID: appInfo?.appID,
+      appID: appInfo?.appID as number,
```
ratik21, line 777: I think linter will wipe this out. you're pushing with `--no-verify`. You should use `if (appInfo !== undefined) { // do stuff }`

packages/runtime/test/integration/atomic-transfer.ts
```diff
@@ -75,8 +75,8 @@ describe("Algorand Smart Contracts - Atomic Transfers", function () {
         payFlags: { totalFee: 1000 }
       }
     ];
-    const initialJohnAssets = john.getAssetHolding(assetId)?.amount;
-    const initialAliceAssets = alice.getAssetHolding(assetId)?.amount;
+    const initialJohnAssets = john.getAssetHolding(assetId)?.amount as bigint;
+    const initialAliceAssets = alice.getAssetHolding(assetId)?.amount as bigint;
```
ratik21, line 79: same as above. Linter. I think i fixed this before, why was it reverted? :(

packages/runtime/test/src/interpreter/inner-transaction.ts
```diff
@@ -1194,19 +1200,64 @@ describe("Inner Transactions", function () {
...
+      int 5
+      itxn_field VoteKeyDilution
+      int 1
+      itxn_field Nonparticipation
+      itxn_submit
+      int 1
+
+      `;
+      executeTEAL(program);
+      //  assert.doesNotThrow(() => executeTEAL(program));
```
ratik21, line 1244: why not use 1244?

packages/runtime/test/src/interpreter/inner-transaction.ts
```diff
@@ -1194,19 +1200,64 @@ describe("Inner Transactions", function () {
...
+    });
+
+    it("Invalid field keyreg transaction", () => {
```
ratik21, line 1247:
```suggestion
    it("should fail on invalid field keyreg transaction", () => {
```

packages/runtime/test/src/interpreter/inner-transaction.ts
vuvth: > why not use 1244?

Thanks!

packages/runtime/test/integration/atomic-transfer.ts
vuvth: It's also happened a lot of time with me. :(

ratik21 (approved):

PR #598, opened 2022-02-23
more dependencies are needed in quickstart

shaih:
## Proposed Changes

+ Added tslib and @walletconnect/encoding as dependencies. The process in quickstart was failing without them


robert-zaremba:
Why `tslib` is needed? It should be only needed in a typescript project, but then, it's no relevant to a general setup. Algo Builder should work with JS and TS.

I think we should add them to the dependencies of `packages/web/package.json`. 
Could you update? The PR?

PR #600, opened 2022-02-24
solve `modelsv2.DryRunRequest` type issue from algob

ratik21:
Still had to do a type hack :/

### DIFF
--- a/packages/algob/src/lib/dryrun.ts
+++ b/packages/algob/src/lib/dryrun.ts
@@ -118,8 +118,10 @@ export class Tealdbg {
         // load pyCache from "artifacts/cache"
         if (fs.existsSync(path.join(CACHE_DIR, file + ".yaml"))) {
           const pathToPyCache = getPathFromDirRecursive(CACHE_DIR, file + ".yaml");
-          const pyCache = loadFromYamlFileSilent(pathToPyCache) as PyASCCache;
-          tealFromPyTEAL = pyCache.tealCode;
+          if (pathToPyCache) {
+            const pyCache = loadFromYamlFileSilent(pathToPyCache) as PyASCCache;
+            tealFromPyTEAL = pyCache.tealCode;
+          }
         }
 
         /* Use cached TEAL code if:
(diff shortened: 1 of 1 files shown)


packages/algob/src/lib/dryrun.ts
```diff
@@ -266,6 +176,18 @@ export class Tealdbg {
...
+
+    /*
+      Encoding fails on taking empty arrays ([]), so we need to convert
+      to undefined first.
+    */
+    if (dryRunRequest.accounts.length === 0) {
+      (dryRunRequest as any).accounts = undefined;
+    }
+    if (dryRunRequest.apps.length === 0) {
+      (dryRunRequest as any).apps = undefined;
```
robert-zaremba, line 188: can you try to not use `as`? We want to limit it, as it already created few problems.

robert-zaremba (commented):
Left one comment about the `as` usage.

robert-zaremba:
test is also not passing

packages/algob/src/lib/dryrun.ts
ratik21: [type hack] Need to do it here, as `dryRunReq.apps` has type `modelsv2.Application[]` (but it should be `Application[] | undefined`) here. If you don't set it as `undefined`, `encodeObj` will fail.

I can create an issue on their repo though

ratik21:
> test is also not passing

yeah that's a separate issue with workflow. Tests are failing in all branches
image

robert-zaremba:
I think i fixed tests in develop. Need to recheck tomorrow

packages/algob/src/lib/dryrun.ts
robert-zaremba: OK, let's add a comment to explain that

robert-zaremba (approved):

robert-zaremba:
Hmm, ok the changes bare there, but I don't know why the test fails. We use pipenv run, so it should work.

packages/algob/src/lib/dryrun.ts
ratik21: ok

PR #605, opened 2022-02-28
feat: support bsqrt and divw opcodes

vuvth:
## Proposed Changes

+ Add bsqrt and divw opcodes (Check packages/runtime/src/interpreter/opcode-list.ts)
+ Uint test for individual opcode(Check packages/runtime/test/src/interpreter/opcode-list.ts)
+ Modify parser to support both opcode. 

## Ref:
+ Link to opcodes docs [here](https://github.com/algorand/go-algorand/blob/1504f5f39ec2893d2b0ac9b60d1b87f6e7d4e2fc/data/transactions/logic/TEAL_opcodes.md#bsqrt)
 
TODO: 
- [ ] update Changelog after review.

Please review it! Thanks!


packages/runtime/src/interpreter/opcode-list.ts
```diff
@@ -1027,6 +1028,41 @@ export class Mulw extends Op {
...
+    this.assertMinStackLen(stack, 3, this.line);
+    const valueC = this.assertBigInt(stack.pop(), this.line);
+    const valueB = this.assertBigInt(stack.pop(), this.line);
+    const valueA = this.assertBigInt(stack.pop(), this.line);
+
+    if (valueC === 0n) {
+      throw new RuntimeError(RUNTIME_ERRORS.TEAL.ZERO_DIV, { line: this.line });
+    }
+
+    const result = ((valueA << BigInt('64')) + valueB) / valueC;
```
robert-zaremba, line 1059:
```suggestion
    const result = ((valueA << 64n) + valueB) / valueC;
```

robert-zaremba (approved):

packages/runtime/src/interpreter/opcode-list.ts
```diff
@@ -1027,6 +1028,41 @@ export class Mulw extends Op {
...
+   * @param args Expected arguments: [] // none
+   * @param line line number in TEAL file
+   */
+  constructor (args: string[], line: number) {
+    super();
+    this.line = line;
+    assertLen(args.length, 0, line);
+  };
+
+  execute (stack: TEALStack): void {
```
thdailong, line 1049: Can you explain for me what `execute` is?(Is it a class, a method, a function???)

vuvth: This is method in class. You can read more details at [here](https://www.typescriptlang.org/docs/handbook/classes.html)

thdailong (approved):

PR #607, opened 2022-03-01
Add doc string for execution modes (app/sig)

ratik21:


robert-zaremba (approved):
Thanks Ratik!
Approving, but let's require @sczembor  review as well. I think originally that was part of his task.

robert-zaremba:
if you merge latest develop into this branch then the tests will pass

robert-zaremba (approved):

PR #612, opened 2022-03-07
Integrated MyAlgo wallet in web

Megha-Dev-19:
- Integrate MyAlgo wallet in web.
- Changed TestNet url as the earlier one was deprecated.
- Updated keys in walletConnect as per their documentation


packages/web/src/lib/api.ts
```diff
@@ -3,16 +3,16 @@ import algosdk, { Algodv2, ALGORAND_MIN_TX_FEE, SuggestedParams } from "algosdk"
 import { ChainType, TxParams } from "../types";
 
 const mainNetClient = new Algodv2("", "https://algoexplorerapi.io", "");
-const testNetClient = new Algodv2("", "https://testnet.algoexplorerapi.io", "");
+const testNetClient = new Algodv2("", "https://node.testnet.algoexplorerapi.io", "");
```
vuvth, line 6: Could we use constant to store this URL? Btw can we import the URL from env?

packages/web/src/lib/myalgowallet-mode.ts
```diff
@@ -0,0 +1,135 @@
...
+      const txn = mkTransaction(
+        execParams,
+        await mkTxParams(this.algodClient, execParams.payFlags)
+      );
+      signedTxn = await this.signTransaction(txn);
+      const Uint8ArraySignedTx = signedTxn.blob;
+      confirmedTx = await this.sendAndWait(Uint8ArraySignedTx);
+    }
+
+    console.log("confirmedTx: ", confirmedTx);
```
vuvth, line 132: Could we use a logger to handle log instead of `console.log`? How about think about it?

vuvth (commented):
Thanks for you PR. I added a few comment.

packages/web/src/lib/api.ts
Megha-Dev-19: just a confirmation, we don't have an .env file right now, so you want me to create one and store it as a const in there and then import it from there. right?

vuvth: In my view, we should make testnetclient flexible. User can config this value(they can update/change it) . Not sure about use .env is best solution.

rmeena840: Storing them in a constant file or env file is a good idea. Here is more info about it: https://medium.com/chingu/an-introduction-to-environment-variables-and-how-to-use-them-f602f66d15fa

rmeena840 (commented):
Reviewed

packages/web/src/lib/api.ts
Megha-Dev-19: @vuvth why will the user change it .. ? I am not much familiar about it but I think it will always be the same host, as @rmeena840 also talked about creating an .env file, so how should I proceed?

packages/web/src/lib/myalgowallet-mode.ts
Megha-Dev-19: I think we can create a new task, where we can update all logs with that logger.

packages/web/src/lib/api.ts
vuvth: > @vuvth why will the user change it .. ? I am not much familiar about it but I think it will always be the same host

I think we shouldn't maintain this network URL. User should do that. I not sure this will change again or they will create multi test net in the next time. @Megha-Dev-19

packages/web/src/lib/myalgowallet-mode.ts
vuvth: Yes, sound goods. Could you do it?

packages/web/src/lib/api.ts
robert-zaremba: We should not have that global variable at all. Also no need to use `.env` - this should be avoided for libraries, this must work easily by just importing the library.

Instead, I propose to:
1. Define const values:
```
const mainnetAlgoExplorer = "https://algoexplorerapi.io";
const testnetAlgoExplorer = "https://node.testnet.algoexplorerapi.io"
```
2. rename `clientForChain` to `algoexplorerAlgod`

packages/web/src/lib/api.ts
```diff
@@ -38,24 +38,24 @@ export async function getSuggestedParams(algocl: Algodv2): Promise<SuggestedPara
...
-		userParams.firstValid === undefined || userParams.validRounds === undefined // eslint-disable-line @typescript-eslint/prefer-nullish-coalescing
-			? s.lastRound
-			: Number(userParams.firstValid) + Number(userParams.validRounds);
-	return s;
+  s.firstRound = userParams.firstValid || s.firstRound; // eslint-disable-line @typescript-eslint/prefer-nullish-coalescing
+  s.lastRound =
+    userParams.firstValid === undefined || userParams.validRounds === undefined // eslint-disable-line @typescript-eslint/prefer-nullish-coalescing
+      ? s.lastRound
+      : Number(userParams.firstValid) + Number(userParams.validRounds);
+  return s;
```
robert-zaremba, line 60: lot of formatting problems. Seams that you don't use `yarn format` or you are not using the changes in develop we pushed.

Megha-Dev-19: I ran `yarn format` and updated the code.

packages/web/src/lib/wallectconnect-mode.ts
```diff
@@ -57,7 +56,7 @@ export class WallectConnectSession {
 				try {
 					await this.close();
 				} catch (e) {
-					console.error("Can't close walletconnect connection", e);
+					console.log(e);
```
robert-zaremba, line 59: why changing this?

Megha-Dev-19: while testing, I must have changed it, will revert back.

Megha-Dev-19:
The PR tests are failing due to `@randlab/communication-bridge` library which is used by myAlgo wallet, since it is using window but we are running it server side.
Here is a link to the issue:
https://github.com/randlabs/myalgo-connect/issues/47 

I don't know how to use dynamic import to fix this.
If anyone can help.?

packages/web/src/lib/myalgowallet-mode.ts
```diff
@@ -0,0 +1,136 @@
...
+
+const CONFIRMED_ROUND = "confirmed-round";
+const LAST_ROUND = "last-round";
+
+export class MyAlgoWalletSession {
+	readonly connector: MyAlgoConnect;
+	private readonly algodClient: algosdk.Algodv2;
+	accounts: Accounts[] = [];
+	addresses: Address[] = [];
+
```
robert-zaremba, line 17: To avoid importing `@randlabs/myalgo-connect` create / copy interface types here and you them in the code instead of the types imported from `@randlabs/myalgo-connect`. Then in constructor only dynamically import `MyAlgoConnect` to initialize the connection object.

````
type Connector interface {
   ...   
}
```

packages/web/src/lib/api.ts
```diff
@@ -2,17 +2,17 @@ import algosdk, { Algodv2, ALGORAND_MIN_TX_FEE, SuggestedParams } from "algosdk"
 
 import { ChainType, TxParams } from "../types";
 
-const mainNetClient = new Algodv2("", "https://algoexplorerapi.io", "");
-const testNetClient = new Algodv2("", "https://testnet.algoexplorerapi.io", "");
+const mainnetAlgoExplorer = new Algodv2("", "https://algoexplorerapi.io", "");
+const testnetAlgoExplorer = new Algodv2("", "https://node.testnet.algoexplorerapi.io", "");
```
robert-zaremba, line 6: we should not declare them as global, please move inside the `algoexplorerAlgod` function

packages/web/src/lib/myalgowallet-mode.ts
```diff
@@ -0,0 +1,142 @@
...
+			const pendingInfo = await this.algodClient.pendingTransactionInformation(txId).do();
+			if (pendingInfo["pool-error"]) {
+				throw new Error(`Transaction Pool Error: ${pendingInfo["pool-error"] as string}`);
+			}
+			if (pendingInfo[CONFIRMED_ROUND] !== null && pendingInfo[CONFIRMED_ROUND] > 0) {
+				return pendingInfo as algosdk.modelsv2.PendingTransactionResponse;
+			}
+			lastround++;
+			await this.algodClient.statusAfterBlock(lastround).do();
+		}
```
robert-zaremba, line 98: please use `algosdk.waitForConfirmation` instead of this loop

packages/web/src/lib/myalgowallet-mode.ts
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	/**
+	 * Execute single transaction or group of transactions (atomic transaction)
+	 * @param execParams transaction parameters or atomic transaction parameters
+	 */
+	async executeTransaction(
```
robert-zaremba, line 105: I have added a followup task (to be done after this pr will be merged): https://www.pivotaltracker.com/story/show/181626750

packages/web/src/lib/myalgowallet-types.ts
```diff
@@ -0,0 +1,214 @@
+// https://github.com/randlabs/myalgo-connect/blob/master/index.d.ts
+/* eslint-disable no-unused-vars */
+export type Address = string;
+export type Base64 = string;
+export type TxHash = string;
+
```
robert-zaremba, line 6: Please only copy the minimum. I think only `MyAlgoConnect` and `Account has to be copied.
Other types are already in AlgoSDK, eg: https://algorand.github.io/js-algorand-sdk/classes/Transaction.html can be used instead of `Txn`.

Also, check other types defined in:
+ https://github.com/scale-it/algo-builder/blob/master/packages/web/src/types.ts

Another approach would be to use:
```
import type {...} from "@randlabs/myalgo-connect"
````
Let's see if this work - maybe we don't need this file at all :) `import type` is done only for compilation, and after compilation it is removed. so when running it should not create the weird error about undefined `window`.


Let's limit the copies to the minimum.

File: packages/web/src/lib/wallectconnect-mode.ts [hunk 10]
```diff
@@ -25,8 +24,8 @@ export class WallectConnectSession {
 	private readonly algodClient: algosdk.Algodv2;
 	wcAccounts: string[];
 
-	constructor(chain: ChainType, connector?: WalletConnect) {
-		this.algodClient = clientForChain(chain);
+	constructor(chain: string, connector?: WalletConnect) {
```
robert-zaremba, line 27: why changing this?

robert-zaremba (changes requested):
let's reduce amount of type copies. Maybe the solution will be to use `import type ...` - it just came to my mind when doing a review.

packages/web/src/lib/wallectconnect-mode.ts
Megha-Dev-19: because when I checked the `algoexplorerAlgod` which accepts ChainType as a parameter, I realised that it should be a string (i.e the network type TestNet/MainNet) not an object, thus I updated it

robert-zaremba: `ChainType` is an enum of strings.
I think the only problem is when a user will like to integrate his other chain in a wallet. I'm not sure if he can do it though.

robert-zaremba: let's leave it as a it was. We will change it in https://www.pivotaltracker.com/story/show/181620989 anyway.

Megha-Dev-19: okay

robert-zaremba (approved):

PR #614, opened 2022-03-07
Prettier

robert-zaremba:
* Migrate to use prettier for formating instead of eslint-standard
* Use eslint for linting and prettier for formatting


sczembor (approved):
There is some problem with the yarn format when run in docs, the test does not pass.
https://github.com/scale-it/algo-builder/runs/5448282372?check_suite_focus=true#step:6:16

robert-zaremba:
> There is some problem with the yarn format when run in docs, the test does not pass. https://github.com/scale-it/algo-builder/runs/5448282372?check_suite_focus=true#step:6:16

@sczembor I've just fixed it.

PR #617, opened 2022-03-08
Remove some printlines when running executeTx, deployApp, deployASA

sczembor:
Closes #x

## Proposed Changes

+ removed console.log from `deployer.ts` and replaced with console.debug
+ only when run with the flag --verbose the errors from deployer will be displayed


vuvth (approved):
LGTM.
 
Btw, Could you add more test? Test suit should follow: 

1. Mock `algoOp.deployApp()`, `algoOp.updateApp`, ... etc throw error when use `deployer.deployApp`, `deployer.updateApp`, ...etc. 
2. Use `assert.true(console.error.calledWith('some string value'))` to check message will log on screen. 

Thanks

PR #618, opened 2022-03-10
feat: add acct_params_get opcode

vuvth:
## Proposed Changes

+ Add AcctParamsGet to handle acct_params_get opcode(check opcode-list.ts)
+ Add acct_params_get to parser (parser.ts).
+ Create `create` for interpreter.getAcccount (check interpreter.ts). When `create` flag is false we will throws exception if account is not found else if `create` is true we will create new account and add this account to context.
+ Unit test for them

## TODO
- [x] Update changelog.md after PR approval.

Please review my PR! Thank you a lot.


packages/runtime/src/errors/errors-list.ts
```diff
@@ -377,6 +377,13 @@ maximun uint128`,
 		title: "Execution mode not valid",
 		description: `Execution mode not valid`,
 	},
+	UNKNOWN_ACCT_FIELD: {
+		number: 1053,
```
robert-zaremba, line 381:
```suggestion
		number: 1055,
```

packages/runtime/src/interpreter/interpreter.ts
```diff
@@ -107,14 +108,28 @@ export class Interpreter {
...
+	
 	/**
 	 * Beginning from TEALv4, user can directly pass address instead of index to Txn.Accounts.
 	 * However, the address must still be present in tx.Accounts OR should be equal to Txn.Sender
 	 * @param accountPk public key of account
 	 * @param line line number in TEAL file
 	 * https://developer.algorand.org/articles/introducing-algorand-virtual-machine-avm-09-release/
 	 */
-	private _getAccountFromAddr(accountPk: Uint8Array, line: number): AccountStoreI {
+	private _getAccountFromAddr(accountPk: Uint8Array, line: number, strict = true): AccountStoreI {
```
robert-zaremba, line 132: need to add docs about `strict`

robert-zaremba: Also, I suggest to rename `strict` to `create`

packages/runtime/src/interpreter/interpreter.ts
```diff
@@ -145,7 +160,13 @@ export class Interpreter {
 			compareArray(accountPk, decodeAddress(getApplicationAddress(appID)).publicKey)
 		) {
 			const address = encodeAddress(pkBuffer);
-			const account = this.runtime.ctx.state.accounts.get(address);
+			let account = this.runtime.ctx.state.accounts.get(address);
+			// strict is turn on we will create account store
+			// we will create account if it not exist
+			if (!strict) {
+				account = this.createAccountIfAbsent(address);
+			}
```
robert-zaremba, line 168:
```suggestion
			let account = create ? this.createAccountIfAbsent(address) : this.runtime.ctx.state.accounts.get(address);
```

packages/runtime/src/interpreter/interpreter.ts
```diff
@@ -181,9 +204,12 @@ export class Interpreter {
 				}
 				address = encodeAddress(pkBuffer);
 				account = this.runtime.ctx.state.accounts.get(address);
+				if (!strict) {
+					account = this.createAccountIfAbsent(address);
+				}
```
robert-zaremba, line 209:
```suggestion
				account = create ? this.createAccountIfAbsent(address) : this.runtime.ctx.state.accounts.get(address);
```

packages/runtime/src/interpreter/interpreter.ts
```diff
@@ -158,12 +179,14 @@ export class Interpreter {
...
-	 * Throws exception if account is not found.
+	 * when `strict` is true we will throws exception if account is not found. 
+	 * when `strict` is false we will create new account and add it to context.
 	 * @param accountRef index of account to fetch from account list
 	 * @param line line number
+	 * @param strict strict flag
 	 * NOTE: index 0 represents txn sender account
 	 */
-	getAccount(accountRef: StackElem, line: number): AccountStoreI {
+	getAccount(accountRef: StackElem, line: number, strict = true): AccountStoreI {
```
robert-zaremba, line 189: let's rename strict as well

packages/runtime/src/interpreter/opcode-list.ts
```diff
@@ -4562,3 +4563,70 @@ export class AppParamsGet extends Op {
...
 	}
 }
+
+export class AcctParamsGet extends Op {
+	readonly interpreter: Interpreter;
+	readonly line: number;
+	readonly field: string;
+	/**
+	 * Asserts 1 arguments are passed.
+	 * @param args Expected arguments: [] // none
```
robert-zaremba, line 4573:
```suggestion
	 * @param args Expected arguments: [account_param]
```

packages/runtime/src/interpreter/opcode-list.ts
```diff
@@ -4562,3 +4563,70 @@ export class AppParamsGet extends Op {
...
+				break;
+			}
+			case "AcctAuthAddr": {
+				if (accountInfo.getSpendAddress() === accountInfo.address) {
+					value = ZERO_ADDRESS;
+				} else {
+					value = Buffer.from(decodeAddress(accountInfo.getSpendAddress()).publicKey);
+				}
+				break;
+			}
```
robert-zaremba, line 4622: add a default case, which will return an error , something like `'unknown account parameter "%s"'`

packages/runtime/test/src/interpreter/opcode-list.ts
```diff
@@ -6252,4 +6254,118 @@ describe("Teal Opcodes", function () {
...
+		let interpreter: Interpreter;
+		let alice: AccountStoreI;
+		let bob: AccountStoreI;
+		let op: AcctParamsGet;
+		const zeroBalanceAddr = "WWYNX3TKQYVEREVSW6QQP3SXSFOCE3SKUSEIVJ7YAGUPEACNI5UGI4DZCE";
+
+		this.beforeEach(() => {
+			interpreter = new Interpreter();
+			interpreter.runtime = new Runtime([]);
+			[alice, bob] = interpreter.runtime.defaultAccounts()
```
robert-zaremba, line 6270: we can use default accounts :) cc: @sczembor

packages/runtime/test/src/interpreter/opcode-list.ts
```diff
@@ -6252,4 +6254,118 @@ describe("Teal Opcodes", function () {
...
+
+			// valid address but not in tx accounts list
+			stack.push(parsing.stringToBytes("01234567890123456789012345678901"));
+			expectRuntimeError(
+				() => op.execute(stack),
+				RUNTIME_ERRORS.TEAL.ADDR_NOT_FOUND_IN_TXN_ACCOUNT
+			)
+		});
+
+		it("Should throw error if top element in stack is not address type", () => {
```
robert-zaremba, line 6361:
```suggestion
		it("Should throw error if top element in stack is not an address", () => {
```

robert-zaremba (approved):
pre approving. Left few suggestions.

packages/runtime/src/interpreter/opcode-list.ts
vuvth: We can't create Opcode class if the parameter invalid(will throw error) so we don't need handle this case.

PR #621, opened 2022-03-16
Add an example run script in AlgoDAO to transfer from proposalLsig

sczembor:
Closes #x

## Proposed Changes

+ Added a script `withdraw_from_proposal` that withdraws from `proposalLsig` back to the owner.
+ Updated the `proposal-lsig.py` smart contract to allow withdrawals only back to the owner account.
+ Added some test to check the withdrawal capability.
+ Updated the `README.md`.


examples/dao/scripts/run/withdraw_from_proposal.js
```diff
@@ -0,0 +1,45 @@
+const { tryExecuteTx } = require("./common/common.js");
+const { types } = require("@algo-builder/web");
+const { accounts } = require("./common/accounts.js");
+const { mkWithdrawFromProposalTx } = require("./common/tx-params.js");
+
+async function withdrawFromProposal(deployer, proposalLsig, proposer, voterA) {
+	const govTokenID = deployer.asa.get("gov-token").assetIndex;
+	console.log("gov token id: ", govTokenID);
+
+	// Needed to add the fund operation for it to work.
```
sczembor, line 10: Not sure about this issue. Maybe the way I run it is incorrect

robert-zaremba: because the proposal must have tokens in the first place. In this proposal, you are basically sending gov tokens to the proposal. This is not what we want. You should remove `fundProposaltx`.
The flow is:
+ owner creates proposal
+ people vote (they firstly need to deposit vote tokens)
+ we execute the proposal (finalize it)
+ owner calls "clear_proposal" <- here the DAO will return deposit back to the proposal lsig.
+ and only then, an owner can call withdraw from the proposal lsig.

So basically this script should be called after voting and after finalizing and after `clear_proposal.js` - after that it  should work.

examples/dao/test/dao-app-flow.js
```diff
@@ -944,5 +948,30 @@ describe("DAO test", function () {
...
+				amount: 5,
+				lsig: proposalALsig,
+				assetID: govTokenID,
+				payFlags: {},
+			},
+		];
+		let proposerABalance = proposerA.getAssetHolding(govTokenID).amount;
+		let proposalLsigABalance = proposalALsigAcc.getAssetHolding(govTokenID).amount;
+
+		runtime.executeTx(withdrawTx);
```
robert-zaremba, line 968: as discussed, w should have two transactions: 
- one withdrawing let's say 2 tokens to a new user (he must opt-in before)
- second: withdrawing  let's say 8 tokens to the owner.
Both tx should succeed and we need to add balance checks after each transaction.

Also, let's add a test with withdraw from not an owner - it should fail ( we can do it in this function, before the "positive" withdraws )

robert-zaremba (commented):
let's update the tests and the script.

examples/dao/README.md
```diff
@@ -65,6 +65,9 @@ We use functional notation to describe use cases we will implement.
   - _tx0_: Call to DAO App with _appArg_ == "clear_proposal".
   - _tx1_: Withdraw _gov_tokens_ from _depositLsig_ back to _proposalLsig_ (ASA transfer).
 
+- `withdraw_from_proposal()` - Withdraw the assets from the proposalLsig back to the owner. Receiver must be the owner of the proposalLsig. Transaction composition:
```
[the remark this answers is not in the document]
robert-zaremba: I'm wrong, the receiver must be owner.

examples/dao/scripts/run/withdraw_from_proposal.js
sczembor: The script was updated and the scenario works in the [dao-app-flow.js](https://github.com/scale-it/algo-builder/pull/621#diff-05aea0ea48d2b68d5f013c73e2277d2af1fe75bb4887582778a113381b68faf6). However when executing all the scripts manually I could not get the `execute` transaction to succeed. Not sure if im doing something incorrectly. Maybe we should ad a story?

examples/dao/test/dao-app-flow.js
```diff
@@ -944,5 +950,44 @@ describe("DAO test", function () {
...
+
+		assert.equal(proposerA.getAssetHolding(govTokenID).amount, proposerABalance + 1n);
+		assert.equal(
+			proposalALsigAcc.getAssetHolding(govTokenID).amount,
+			proposalLsigABalance - 1n
+		);
+
+		//verify that the funds from proposalLsig cannot be withdrawed to an non-owner account
+		runtime.optIntoASA(govTokenID, alice.address, {});
+		[alice] = runtime.defaultAccounts();
```
robert-zaremba, line 982: this line is not needed

robert-zaremba (approved):

examples/dao/test/dao-app-flow.js
sczembor: `[alice] = runtime.defaultAccounts();`
Is needed for the assertion below where we check if the account opted-in to the GOV token
`assert.isDefined(alice.getAssetHolding(govTokenID));`

PR #624, opened 2022-03-17
Upload file .gitignore to sample project

thdailong:
Closes #x

## Proposed Changes

+
+Upload file .gitignore to sample project
+

### DIFF
--- a/packages/algob/sample-project/common/.gitignore
+++ b/packages/algob/sample-project/common/.gitignore
@@ -1,0 +1,2 @@
+/node_data
+.dockerignore
--- a/packages/algob/sample-project/common/.gitignore
+++ b/packages/algob/sample-project/common/.gitignore
@@ -1,2 +1,6 @@
 /node_data
 .dockerignore
+artifacts/
+/algob.config-local.js
+node_modules
+.DS_Store


rmeena840 (approved):

packages/algob/sample-project/common/.gitignore
```diff
@@ -0,0 +1,2 @@
+/node_data
+.dockerignore
```
ratik21, line 2: Please also add
```
artifacts/
/algob.config-local.js
node_modules
```

vuvth: Please also add

```
.DS_Store
```
Why you add `/node_data`? Its mean `node_data` in `root` folder right? @thdailong

thdailong: I copy the .gitignore file in the same "Sample project" folder and copy to this "common" folder.
I am thinking that the node_data is in the root folder but I am not realy sure

packages/algob/sample-project/common/.gitignore
```diff
@@ -0,0 +1,6 @@
+/node_data
```
vuvth, line 1: Why you add node_date here? and why use add `/`(in root folder)?

packages/algob/sample-project/common/.gitignore
```diff
@@ -0,0 +1,6 @@
...
+.dockerignore
+artifacts/
+/algob.config-local.js
```
vuvth, line 4: Why use add /algob.config-local.js here?

packages/algob/sample-project/common/.gitignore
thdailong: Because I think the node_data reference to data of node in algorand so we do not need to commit node_data

packages/algob/sample-project/common/.gitignore
thdailong: I think it use to ignore the local file in the computer?

packages/algob/sample-project/common/.gitignore
vuvth: Oh / mean from root project not root of OS, right?

thdailong: Yes that's right

vuvth (approved):

PR #625, opened 2022-03-17
Check the max number of schema entries

sczembor:
Closes #x

## Proposed Changes

+ Added two constants with `MaxGlobalSchemaEntries = 64` and `MaxLocalSchemaEntries = 16`.
+ Added check if the parameters does not exceed those values above.
+ Fixed the unit tests that where failing because of the parameters exceeded the limits.


packages/runtime/test/src/lib/txn.ts
```diff
@@ -170,10 +170,10 @@ describe("Convert encoded Txn to ExecParams", function () {
...
 				approvalProgram: "counter-approval.teal",
 				clearProgram: "clear.teal",
-				globalBytes: 10,
-				globalInts: 10,
-				localBytes: 10,
-				localInts: 10,
+				globalBytes: 1,
+				globalInts: 1,
+				localBytes: 1,
+				localInts: 1,
```
robert-zaremba, line 176: why changing all of this?

robert-zaremba (commented):
- [x] we need to add a test
- [ ] we should update the runtime checking that we don't exceed the amount of variables during the execution: that should happen whenever we save something to store. I think we already encapsulate it in a function.

packages/runtime/src/ctx.ts
```diff
@@ -329,6 +333,20 @@ export class Ctx implements Context {
 
 		this.verifyTEALVersionIsMatch(approvalProgTEAL, clearProgTEAL);
 
+		//verify that MaxGlobalSchemaEntries <= 64 and MaxLocalSchemaEntries <= 16
+		const globalSchemaEntries = flags.globalInts + flags.globalBytes;
+		const localSchemaEntries = flags.localInts + flags.localBytes;
```
vuvth, line 338: What happen when localInts(or other value) is negative number? 
I also don't think use will pass negative number, but better we should check it too.

packages/runtime/src/lib/constants.ts
```diff
@@ -44,6 +44,10 @@ export const MAX_ALGORAND_ACCOUNT_CREATED_APPS = 10;
 
 export const MAX_ALGORAND_ACCOUNT_OPTEDIN_APPS = 50;
 
+//smart contract constraints
```
vuvth, line 47: Could you add link to algorand docs we refer?

vuvth (commented):
Thanks! I also add few comments.

packages/runtime/test/src/lib/txn.ts
ratik21: i think this would throw error otherwise now. local state entries total to 20 (max 16 allowed)

packages/runtime/src/ctx.ts
sczembor: It already throws an error when there is a negative number passed

vuvth: Thanks.

packages/runtime/test/src/runtime.ts
```diff
@@ -992,6 +992,26 @@ describe("Stateful Smart Contracts", function () {
 			RUNTIME_ERRORS.TEAL.PROGRAM_VERSION_MISMATCH
 		);
 	});
+
+	it("Should throw and error when local/global schema entries exceeds the limit (AppDeploymentFlags)", () => {
```
vuvth, line 996: Add one more test when global schema bound of limit.

vuvth (approved):

packages/runtime/test/src/runtime.ts
sczembor: Done

sczembor:
> * [x]  we need to add a test
> * [ ]  we should update the runtime checking that we don't exceed the amount of variables during the execution: that should happen whenever we save something to store. I think we already encapsulate it in a function.

I think we already have this functionality: https://github.com/scale-it/algo-builder/blob/0567da28f9b8ac2c83cbd4673280eca7a0ec13ac/packages/runtime/src/account.ts#L201

robert-zaremba (approved):

PR #629, opened 2022-03-23
Update waitForConfirmation in web

Megha-Dev-19:
- Updated `waitForConfirmation` function to use `algosdk.waitForConfirmation` in walletConnect and MyAlgo wallet (done in last PR) rather than a loop.
- In AlgoSigner, it already have a function of `algod` to get the status of transaction, thus not updating it there

### DIFF
--- a/packages/web/src/lib/wallectconnect-mode.ts
+++ b/packages/web/src/lib/wallectconnect-mode.ts
@@ -196,20 +196,10 @@ export class WallectConnectSession {
 	private async waitForConfirmation(
 		txId: string
 	): Promise<algosdk.modelsv2.PendingTransactionResponse> {
-		const response = await this.algodClient.status().do();
-		let lastround = response[LAST_ROUND];
-		/*eslint no-constant-condition: ["error", { "checkLoops": false }]*/
-		while (true) {
-			const pendingInfo = await this.algodClient.pendingTransactionInformation(txId).do();
-			if (pendingInfo["pool-error"]) {
-				throw new Error(`Transaction Pool Error: ${pendingInfo["pool-error"] as string}`);
-			}
-			if (pendingInfo[CONFIRMED_ROUND] !== null && pendingInfo[CONFIRMED_ROUND] > 0) {
-				return pendingInfo as algosdk.modelsv2.PendingTransactionResponse;
-			}
-			lastround++;
-			await this.algodClient.statusAfterBlock(lastround).do();
-		}
+		const pendingInfo = await algosdk.waitForConfirmation(this.algodClient, txId, 10);
+		if (pendingInfo["pool-error"]) {
+			throw new Error(`Transaction Pool Error: ${pendingInfo["pool-error"] as string}`);
+		} else return pendingInfo as algosdk.modelsv2.PendingTransactionResponse;
 	}
 
 	/**
--- a/packages/web/src/lib/wallectconnect-mode.ts
+++ b/packages/web/src/lib/wallectconnect-mode.ts
@@ -199,9 +199,8 @@ export class WallectConnectSession {
 		const pendingInfo = await algosdk.waitForConfirmation(this.algodClient, txId, WAIT_ROUNDS);
 		if (pendingInfo["pool-error"]) {
 			throw new Error(`Transaction Pool Error: ${pendingInfo["pool-error"] as string}`);
-		} else {
-		   return pendingInfo as algosdk.modelsv2.PendingTransactionResponse;
 		}
+		return pendingInfo as algosdk.modelsv2.PendingTransactionResponse;
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


packages/web/src/lib/wallectconnect-mode.ts
```diff
@@ -196,20 +196,10 @@ export class WallectConnectSession {
...
-			if (pendingInfo["pool-error"]) {
-				throw new Error(`Transaction Pool Error: ${pendingInfo["pool-error"] as string}`);
-			}
-			if (pendingInfo[CONFIRMED_ROUND] !== null && pendingInfo[CONFIRMED_ROUND] > 0) {
-				return pendingInfo as algosdk.modelsv2.PendingTransactionResponse;
-			}
-			lastround++;
-			await this.algodClient.statusAfterBlock(lastround).do();
-		}
+		const pendingInfo = await algosdk.waitForConfirmation(this.algodClient, txId, 10);
```
vuvth, line 199: Shall we store `10`(waitRounds) as constant and use it in all web modes(my wallet, wallet connect, etc,...)?

vuvth (changes requested):
Thanks for your PR!
I left a few comments. Please check it! <3

packages/web/src/lib/wallectconnect-mode.ts
```diff
@@ -196,20 +196,10 @@ export class WallectConnectSession {
 	private async waitForConfirmation(
 		txId: string
 	): Promise<algosdk.modelsv2.PendingTransactionResponse> {
```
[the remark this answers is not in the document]
Megha-Dev-19: yeah, I told that in comment, that's `AlgoSigner` wallet and it has its own `algod` function which provides transaction status , so I don't think we can use `algosdk.waitForConfirmation` in that. If it can be done, please let me know.

packages/web/src/lib/wallectconnect-mode.ts
Megha-Dev-19: yeah, that's a nice suggestion. Will do

packages/web/src/lib/wallectconnect-mode.ts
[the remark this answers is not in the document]
vuvth: > yeah, I told that in comment, that's `AlgoSigner` wallet and it has its own `algod` function which provides transaction status , so I don't think we can use `algosdk.waitForConfirmation` in that. If it can be done, please let me know.

Sorry, I didn't check the web-mode.ts using `AlgoSigner`. Thanks for your inform!

packages/web/src/lib/wallectconnect-mode.ts
```diff
@@ -196,19 +196,11 @@ export class WallectConnectSession {
...
-			if (pendingInfo[CONFIRMED_ROUND] !== null && pendingInfo[CONFIRMED_ROUND] > 0) {
-				return pendingInfo as algosdk.modelsv2.PendingTransactionResponse;
-			}
-			lastround++;
-			await this.algodClient.statusAfterBlock(lastround).do();
+		const pendingInfo = await algosdk.waitForConfirmation(this.algodClient, txId, WAIT_ROUNDS);
+		if (pendingInfo["pool-error"]) {
+			throw new Error(`Transaction Pool Error: ${pendingInfo["pool-error"] as string}`);
+		} else {
+		   return pendingInfo as algosdk.modelsv2.PendingTransactionResponse;
```
ratik21, line 203: you don't need an `else` if you're throwing an `error` in `If` condition

packages/web/src/lib/wallectconnect-mode.ts
```diff
@@ -196,19 +196,11 @@ export class WallectConnectSession {
...
-		while (true) {
-			const pendingInfo = await this.algodClient.pendingTransactionInformation(txId).do();
-			if (pendingInfo["pool-error"]) {
-				throw new Error(`Transaction Pool Error: ${pendingInfo["pool-error"] as string}`);
-			}
-			if (pendingInfo[CONFIRMED_ROUND] !== null && pendingInfo[CONFIRMED_ROUND] > 0) {
-				return pendingInfo as algosdk.modelsv2.PendingTransactionResponse;
-			}
-			lastround++;
-			await this.algodClient.statusAfterBlock(lastround).do();
```
ratik21, line 211: to confirm: did we do the same for `waitFromConfirmation` in `Algo Operator` in `algob`?

Megha-Dev-19: I m not sure, because the task I was doing in that only web packages was mentioned
https://www.pivotaltracker.com/story/show/181591414

vuvth (approved):
Approved! Thank for your PR <3

packages/web/src/lib/wallectconnect-mode.ts
ratik21: could you please check and create apt task?

PR #630, opened 2022-03-23
feat:add itxnas opcode

vuvth:
## Proposed Changes

+ Add itxnas opcode 
+ Unit test


packages/runtime/src/interpreter/opcode-list.ts
```diff
@@ -4121,7 +4121,7 @@ export class ITxn extends Op {
 	execute(stack: TEALStack): void {
 		if (this.interpreter.innerTxnGroups.length === 0) {
 			throw new RuntimeError(RUNTIME_ERRORS.TEAL.NO_INNER_TRANSACTION_AVAILABLE, {
-				version: this.interpreter.tealVersion,
```
robert-zaremba, line 4124: why renaming?

robert-zaremba (approved):

packages/runtime/src/interpreter/opcode-list.ts
vuvth: Oh! Its not match with message map in ERROR. So I fix it to same with other.

PR #634, opened 2022-03-28
Responsive layout fix

Megha-Dev-19:
- Made Logo image and audit image responsive
- Made address text wrap


robert-zaremba (approved):
didn't test it, but I think it's OK.

PR #639, opened 2022-03-30
use algosdk.waitForConfirmation rather than a loop as done in web package

rmeena840:


vuvth (approved):
Looks good! 
Could you add test cases when it timeout? Just option, not required.

packages/algob/src/lib/constants.ts
```diff
@@ -3,5 +3,6 @@ import { Algodv2 } from "algosdk";
 export const globalZeroAddress = "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAY5HFKQ";
 export const MIN_UINT64 = 0n;
 export const MAX_UINT64 = 0xffffffffffffffffn;
+export const WAIT_ROUNDS = 10; // for transaction confirmation
```
robert-zaremba, line 6:
```suggestion
export const WAIT_ROUNDS = 6; // for transaction confirmation
```

robert-zaremba (approved):

PR #640, opened 2022-03-30
Bug in javascript/typescript template

rmeena840:
Bug in javascript/typescript template


packages/algob/sample-project/js/package.json
```diff
@@ -4,9 +4,11 @@
 	"main": "index.js",
 	"license": "MIT",
 	"devDependencies": {
-		"mocha": "^9.2.0"
+		"@types/mocha": "^9.1.0",
+		"mocha": "^9.1.3",
+		"ts-mocha": "^9.0.2"
```
robert-zaremba, line 9: ts-mocha and @types/mocha should be definitely only in `ts` project, not `js`

robert-zaremba (changes requested):
please fix he dependencies.

packages/algob/sample-project/js/package.json
rmeena840: I have only now kept it to ts project. Removed from js.

packages/algob/sample-project/js/package.json
```diff
@@ -4,9 +4,9 @@
 	"main": "index.js",
 	"license": "MIT",
 	"devDependencies": {
-		"mocha": "^9.2.0"
+		"mocha": "^9.1.3"
```
rmeena840, line 7:
```suggestion
		"mocha": "^9.2.0"
```

rmeena840:
Fixed

packages/algob/sample-project/ts/.gitignore
```diff
@@ -0,0 +1,6 @@
+nodenode_modules
```
vuvth, line 1:
```suggestion
```

PR #641, opened 2022-03-31
Allow token to be empty in testnet and mainet

thdailong:
Closes #x

## Proposed Changes

+ Allow token to be empty in the testnet and mainet
+ Write test to check 
+


packages/algob/src/internal/core/config/config-validation.ts
```diff
@@ -78,7 +78,10 @@ const HttpNetworkType = z
 		// from: z.string().optional(),
 		host: z.string().optional(),
 		port: z.union([z.number(), z.string()]).optional(),
-		token: z.union([z.string(), AlgodTokenHeaderType, CustomTokenHeaderType]).optional(),
+		token: z
+			.union([z.string(), AlgodTokenHeaderType, CustomTokenHeaderType])
+			.nullable()
```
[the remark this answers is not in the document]
thdailong: nullable in this case is mean that the token can be null

packages/algob/src/internal/core/config/config-validation.ts
```diff
@@ -156,7 +159,12 @@ export function getValidationErrors(config: any): CfgErrors {
 				errors.push(net, "host", host, "hostname string (eg: http://example.com)");
 			}
 			const token = hcfg.token;
-			if (typeof token === "string" && token.length < 10) {
+			if (
+				typeof token === "string" &&
+				token.length < 10 &&
+				host !== "https://node.testnet.algoexplorerapi.io" &&
+				host !== "https://node.algoexplorerapi.io"
```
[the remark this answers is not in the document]
thdailong: Host can not start with HTTP because it would cause an error
image

CHANGELOG.md
```diff
@@ -83,12 +83,14 @@ Added:
...
 
 - The ` PyASCCache` has been merged to `ASCCache` and is not used anymore.
+- Rename the executeTransaction to executeTx
 
 ### Bug fixes
 
 - Return error when closeRemainderTo and fromAccountAddr is the same.
 - When close account should remove auth/spend address. Fixed in [#575](https://github.com/scale-it/algo-builder/pull/575).
 - Approval program and clear propram should throw error if they are mismatch version. Fixed in [620](https://github.com/scale-it/algo-builder/pull/620)
+- Allow token can be empty if the host is mainet or testnet
```
rmeena840, line 93:
```suggestion
- Allow token to be empty if the host is mainet or testnet
```

packages/algob/test/internal/core/config/config-validation.ts
```diff
@@ -276,6 +276,30 @@ describe("Config validation", function () {
 						);
 					});
 
+					it("Token can be empty due to the testnet", function () {
+						const errors = getValidationErrors({
+							networks: {
+								custom: {
+									host: "https://node.testnet.algoexplorerapi.io",
```
rmeena840, line 283: use the const value declared by you above.

packages/algob/test/internal/core/config/config-validation.ts
```diff
@@ -276,6 +276,30 @@ describe("Config validation", function () {
...
+							},
+						});
+						assert.isEmpty(errors.toString());
+					});
+
+					it("Token can be empty due to the mainet", function () {
+						const errors = getValidationErrors({
+							networks: {
+								custom: {
+									host: "https://node.algoexplorerapi.io",
```
rmeena840, line 295: ditto.

store it in const

rmeena840 (changes requested):

packages/web/src/lib/api.ts
```diff
@@ -3,8 +3,10 @@ import algosdk, { Algodv2, ALGORAND_MIN_TX_FEE, SuggestedParams } from "algosdk"
 import { ChainType, TxParams } from "../types";
 
 export function algoexplorerAlgod(chain: string): algosdk.Algodv2 {
-	const mainnetAlgoExplorer = new Algodv2("", "https://algoexplorerapi.io", "");
-	const testnetAlgoExplorer = new Algodv2("", "https://node.testnet.algoexplorerapi.io", "");
+	const mainnetURL = "https://node.algoexplorerapi.io";
```
vuvth, line 6: Move it to constant.ts I think

CHANGELOG.md
thdailong: sure let me fix that

packages/web/src/lib/api.ts
thdailong: Okay I will add to the file constant.ts

rmeena840 (approved):

packages/algob/src/internal/core/config/config-validation.ts
```diff
@@ -6,6 +6,7 @@ import { validateAccount } from "../../../lib/account";
 // import { Account } from "algosdk";
 import type { ChainCfg, HttpNetworkConfig, NetworkConfig } from "../../../types";
 import { ALGOB_CHAIN_NAME } from "../../constants";
+import { mainnetURL, testnetURL } from "./../../../../../web/src/lib/constants";
```
ratik21, line 9: This is wrong. Don't do direct import. Import from `@algo-builder/web` instead.

packages/algob/src/internal/core/config/config-validation.ts
```diff
@@ -156,7 +160,12 @@ export function getValidationErrors(config: any): CfgErrors {
...
 			}
 			const token = hcfg.token;
-			if (typeof token === "string" && token.length < 10) {
+			if (
+				typeof token === "string" &&
+				token.length < 10 &&
+				host !== testnetURL &&
+				host !== mainnetURL
+			) {
 				errors.push(net, "token", token, "string (with length > 10)");
```
ratik21, line 169: This doesn't look right either, You're validating upon host values, but pushing `"token", token ..` in `errors`.

packages/algob/test/internal/core/config/config-validation.ts
```diff
@@ -9,6 +9,7 @@ import {
 } from "../../../../src/internal/core/config/config-validation";
 import { expectBuilderError } from "../../../helpers/errors";
 import { account1 } from "../../../mocks/account";
+import { mainnetURL, testnetURL } from "./../../../../../web/src/lib/constants";
```
ratik21, line 12: same as above

packages/algob/src/internal/core/config/config-validation.ts
```diff
@@ -156,7 +160,12 @@ export function getValidationErrors(config: any): CfgErrors {
 				errors.push(net, "host", host, "hostname string (eg: http://example.com)");
 			}
 			const token = hcfg.token;
-			if (typeof token === "string" && token.length < 10) {
+			if (
+				typeof token === "string" &&
+				token.length < 10 &&
+				host !== testnetURL &&
+				host !== mainnetURL
```
ratik21, line 167: a) I think you should do this validation at line 159 (check above)

b) If you're doing this, you can do something like:
```js
const allowedHosts = [ testnetURL, mainnetURL, ..];

// later
if (!allowedHosts.includes(host)) { // handle }
```

c) But why do we need this at all? Host can be multiple values (local, purestake, algoexplorerapi ..etc). Why do we need this?

ratik21 (changes requested):

CHANGELOG.md
```diff
@@ -83,6 +83,7 @@ Added:
 - `printGlobalStateSCC` renamed to `printGlobalStateApp`.
 
 - The ` PyASCCache` has been merged to `ASCCache` and is not used anymore.
+- Only use list transaction in executeTx.
```
ratik21, line 86: ?

packages/algob/src/internal/core/config/config-validation.ts
```diff
@@ -156,10 +159,13 @@ export function getValidationErrors(config: any): CfgErrors {
 				errors.push(net, "host", host, "hostname string (eg: http://example.com)");
 			}
 			const token = hcfg.token;
-			if (typeof token === "string" && token.length < 10) {
-				errors.push(net, "token", token, "string (with length > 10)");
-			} else if (typeof token !== "object" && typeof token !== "string") {
-				errors.push(net, "token", token, "string or object");
+			const AllowHost = [testnetURL, mainnetURL];
+			if (!AllowHost.includes(host)) {
```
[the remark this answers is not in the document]
thdailong: this is the code to check whether the token are valid or not. But how do we pass though the checking valid token code while the host value is the allowed host? We should check the host before the token and if host is allowed so we run to the final line?

CHANGELOG.md
thdailong: Oh this was my mistake when I see the file changelog.md on the `dev` branch and put the different to the this commit

packages/algob/src/internal/core/config/config-validation.ts
```diff
@@ -156,9 +159,7 @@ export function getValidationErrors(config: any): CfgErrors {
 				errors.push(net, "host", host, "hostname string (eg: http://example.com)");
 			}
 			const token = hcfg.token;
-			if (typeof token === "string" && token.length < 10) {
-				errors.push(net, "token", token, "string (with length > 10)");
-			} else if (typeof token !== "object" && typeof token !== "string") {
+			if (typeof token !== "object" && typeof token !== "string") {
```
ratik21, line 162: Could you check against `null` as well? Or could it only be `null`

ratik21 (approved):
Update the changelog before merging

packages/algob/src/internal/core/config/config-validation.ts
thdailong: Yes, `Null` is also allow as well

PR #642, opened 2022-03-31
Fix/only support list txn in execute tx

vuvth:
## Proposed Changes

+ use list in executeTx

Note: I don't support excuteTx in web, because it don't support TransactionAndSign for now. 
Should I update it in this PR? Please left the comment for this problem.

TODO: 
- [x] Update docs/changelog

### DIFF
--- a/packages/runtime/src/runtime.ts
+++ b/packages/runtime/src/runtime.ts
@@ -819,7 +819,7 @@ export class Runtime {
 			amountMicroAlgos: amount,
 			payFlags: { totalFee: 1000 },
 		};
-		return this.executeTx([fundParam]) as TxReceipt;
+		return this.executeTx([fundParam])[0];
 	}
 
 	/**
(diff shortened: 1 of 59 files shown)


packages/runtime/src/runtime.ts
```diff
@@ -875,28 +875,22 @@ export class Runtime {
 	 * each opcode execution (upto depth = debugStack)
 	 */
 	executeTx(
-		txnParams:
-			| types.ExecParams
-			| types.ExecParams[]
-			| types.TransactionAndSign
-			| types.TransactionAndSign[],
+		txnParams: types.ExecParams[] | types.TransactionAndSign[],
```
robert-zaremba, line 878: need to update the changelog, for all packages: algob, runtime and web

robert-zaremba (approved):
Great job! Please don't forget to update the changelog

PR #643, opened 2022-04-03
Logger added to web package

Megha-Dev-19:
Follow up:
To create logger object in other packages and update the logs.


packages/web/src/lib/logger.ts
```diff
@@ -0,0 +1,52 @@
...
+		namespace: string;
+		destroy: () => boolean;
+		extend: (namespace: string, delimiter?: string) => Debugger;
+	}
+}
+
+const log = debug("algob/web:log");
+
+const error = debug("algob/web:error");
+error.log = console.error.bind(console);
```
robert-zaremba, line 45: why do we need this?

packages/web/src/lib/wallectconnect-mode.ts
```diff
@@ -38,9 +36,7 @@ export class WallectConnectSession {
 
 		// if connection not already established, log message to create one
 		if (!this.connector.connected) {
-			console.warn(
-				`Connection not established, please use "this.create()" to create new session`
-			);
+			log(`Connection not established, please use "this.create()" to create new session`);
```
robert-zaremba, line 39: why not using warn?

robert-zaremba (commented):
looks good, thanks for updating logs. I will add a new task to update the configuration.

Also, 
- [ ] add a changelog 
- [ ] in the web/README file add a note how to enable and disable different log levels.

packages/web/src/lib/logger.ts
Megha-Dev-19: It shows logs in different colors, error is shown in red, log in yellow

packages/web/src/lib/wallectconnect-mode.ts
Megha-Dev-19: will update

Megha-Dev-19:
> * [ ]  add a changelog
 
Can you please explain, what I have to update in changelog?

packages/web/README.md
```diff
@@ -44,4 +44,48 @@ This code will make the transaction, let the user sign it using algosigner and s
 
 You can also use `web.sendTransaction()` or `web.signTransaction()` in a react app.
 
-**Note:** We don't support checkpoints yet. Currently `deployASA`, `deploySSC` functions don't work. User should directly pass assetIndex, appIndex instead of asaName, appName.
+**NOTE**:
+
+1. We don't support checkpoints yet. Currently `deployASA`, `deploySSC` functions don't work. User should directly pass assetIndex, appIndex instead of asaName, appName.
```
ratik21, line 49:
```suggestion
1. We don't support checkpoints yet. Currently `deployASA`, `deployApp` functions don't work. User should directly pass assetIndex, appIndex instead of asaName, appName.
```

packages/web/src/lib/logger.ts
```diff
@@ -0,0 +1,52 @@
+declare const debug: debug.Debug & { debug: debug.Debug; default: debug.Debug };
+// eslint-disable-next-line @typescript-eslint/no-namespace
+declare namespace debug {
```
ratik21, line 3: I am not aware for naming conventions of namespaces. Should it start from small case (debug)?

packages/web/src/lib/logger.ts
Megha-Dev-19: yes for `console.error(msg) => error(msg) and for console.log(msg) => log(msg)` can be used.

robert-zaremba (approved):

PR #644, opened 2022-04-04
Update algob/web README

rmeena840:
Update algob/web README
Task: https://www.pivotaltracker.com/story/show/181762041

### DIFF
--- a/packages/web/README.md
+++ b/packages/web/README.md
@@ -12,6 +12,12 @@ You can use `@algo-builder/web` with [pipeline UI](https://www.pipeline-ui.com/d
 
 In the `@algo-builder/web` package we pass transaction [parameters](https://github.com/scale-it/algo-builder/blob/master/docs/guide/execute-transaction.md) in the same way as we do in `algob`.
 
+## Wallets supported
+
+- AlgoSigner
+- My Algo Connect
+- Wallet Connect
+
 ## Important links
 
 - [Home Page](https://algobuilder.dev/)
@@ -43,5 +49,8 @@ Now you can use it to execute a transaction:
 This code will make the transaction, let the user sign it using algosigner and send it to the network.
 
 You can also use `web.sendTransaction()` or `web.signTransaction()` in a react app.
+
+### deployASA and deploySSC
+`deploySSC` has been renamed to `deployASA`.
 
 **Note:** We don't support checkpoints yet. Currently `deployASA`, `deploySSC` functions don't work. User should directly pass assetIndex, appIndex instead of asaName, appName.


packages/web/README.md
```diff
@@ -44,4 +50,7 @@ This code will make the transaction, let the user sign it using algosigner and s
 
 You can also use `web.sendTransaction()` or `web.signTransaction()` in a react app.
 
+### deployASA and deploySSC
+`deploySSC` has been renamed to `deployASA`.
```
vuvth, line 54:
```suggestion
`deploySSC` has been renamed to `deployApp`.
```

robert-zaremba, line 54: you were supposed to update the examples and and text in this readme, rather than adding this note.

vuvth (approved):
LGTM! Thanks

packages/web/README.md
rmeena840: I have also added an example. Please check.

packages/web/README.md
```diff
@@ -44,4 +49,38 @@ This code will make the transaction, let the user sign it using algosigner and s
...
+	{},
+	{},
+	"DAO App"
+); // app name passed here
+
+// now during querying, you only need this app name
+const appInfo = deployer.getApp("DAO App");
+```
+
 **Note:** We don't support checkpoints yet. Currently `deployASA`, `deploySSC` functions don't work. User should directly pass assetIndex, appIndex instead of asaName, appName.
```
robert-zaremba, line 86: `deploySSC` doesn't exists - this have to be updated. Please see the rest of the README if there is any update needed.

robert-zaremba (changes requested):
not everything has been updated

packages/web/README.md
rmeena840: `deploySSC ` has not been used anywhere. It has not been used in the entire codebase. Should we delete this from this file as well?

robert-zaremba: then why do I see it in the "old" line 89? see the line attached tho this comment.

rmeena840: Changemode says that it has been converted to deployApp: https://github.com/scale-it/algo-builder/blob/2a06f2fba9aa7dc422061b6bdac82b28500bf601/CHANGELOG.md?plain=1#L217

packages/web/README.md
```diff
@@ -44,4 +49,38 @@ This code will make the transaction, let the user sign it using algosigner and s
 
 You can also use `web.sendTransaction()` or `web.signTransaction()` in a react app.
 
-**Note:** We don't support checkpoints yet. Currently `deployASA`, `deploySSC` functions don't work. User should directly pass assetIndex, appIndex instead of asaName, appName.
+### deployApp
+`deployer.deployApp` deploys stateful smart contract.
+### Parameters:
```
ratik21, line 54:
```suggestion
`deployer.deployApp` deploys stateful smart contract.

### Parameters:
```

packages/web/README.md
```diff
@@ -44,4 +49,38 @@ This code will make the transaction, let the user sign it using algosigner and s
...
+- approvalProgram:  approval program filename (must be present in assets folder)
+- clearProgram:  clear program filename (must be present in assets folder)
+- flags:  AppDeploymentFlags
+- payFlags:  Transaction Parameters
+- scTmplParams:  Smart contract template parameters (used only when compiling PyTEAL to TEAL). This is an optional parameter.
+- appName: name of the app to deploy. This name (if passed) will be used as
+the checkpoint "key", and app information will be stored agaisnt this name. This is an optional parameter.
+
+### deployApp example
```
ratik21, line 63:
```suggestion
### Example
```

packages/web/README.md
```diff
@@ -44,4 +49,38 @@ This code will make the transaction, let the user sign it using algosigner and s
 
 You can also use `web.sendTransaction()` or `web.signTransaction()` in a react app.
 
-**Note:** We don't support checkpoints yet. Currently `deployASA`, `deploySSC` functions don't work. User should directly pass assetIndex, appIndex instead of asaName, appName.
+### deployApp
+`deployer.deployApp` deploys stateful smart contract.
+### Parameters:
+- approvalProgram:  approval program filename (must be present in assets folder)
+- clearProgram:  clear program filename (must be present in assets folder)
+- flags:  AppDeploymentFlags
```
ratik21, line 57: Please add a link to API docs here for `AppDeploymentFlags`

packages/web/README.md
```diff
@@ -44,4 +49,38 @@ This code will make the transaction, let the user sign it using algosigner and s
...
+- payFlags:  Transaction Parameters
```
ratik21, line 58: same

packages/web/README.md
```diff
@@ -44,4 +49,38 @@ This code will make the transaction, let the user sign it using algosigner and s
...
+- scTmplParams:  Smart contract template parameters (used only when compiling PyTEAL to TEAL). This is an optional parameter.
+- appName: name of the app to deploy. This name (if passed) will be used as
+the checkpoint "key", and app information will be stored agaisnt this name. This is an optional parameter.
```
ratik21, line 61: Are you sure about "This is an optional parameter"? I think we changed the design to make it required by default. Could you please check? cc: @robert-zaremba

packages/web/README.md
```diff
@@ -44,4 +49,39 @@ This code will make the transaction, let the user sign it using algosigner and s
 
 You can also use `web.sendTransaction()` or `web.signTransaction()` in a react app.
 
-**Note:** We don't support checkpoints yet. Currently `deployASA`, `deploySSC` functions don't work. User should directly pass assetIndex, appIndex instead of asaName, appName.
+### deployApp
+`deployer.deployApp` deploys stateful smart contract.
+
+### Parameters:
```
robert-zaremba, line 55:
```suggestion
Parameters:
```

packages/web/README.md
robert-zaremba: Correct. Both `scTmplParams` and `appName` are optionsl.

                scTmplParams?: SCParams,
                appName?: string

Instead of listing all arguments, I think it's easier and better to link to the API documentation: https://algobuilder.dev/api/algob/interfaces/types.Deployer.html#deployApp

This way it will be always up to date with the released version.

packages/web/README.md
```diff
@@ -44,4 +49,39 @@ This code will make the transaction, let the user sign it using algosigner and s
...
+		localInts: 9,
+		localBytes: 7,
+		globalInts: 4,
+		globalBytes: 2,
+		appArgs: appArgs,
+	},
+	{},
+	{},
+	"DAO App"
+); // app name passed here
```
robert-zaremba, line 81: comment is in wrong place.

robert-zaremba (approved):
pre-approving. Please look at the comments before merging.

packages/web/README.md
rmeena840: Added on line 80

packages/web/README.md
```diff
@@ -44,4 +49,39 @@ This code will make the transaction, let the user sign it using algosigner and s
...
+	{},
+	{},
+	"DAO App" // app name passed here
+);
+
+// now during querying, you only need this app name
+const appInfo = deployer.getApp("DAO App");
+```
+
+**Note:** We don't support checkpoints yet. Currently `deployASA`, `deployApp` functions don't work. User should directly pass assetIndex, appIndex instead of asaName, appName.
```
vuvth, line 87: Are we support deployASA and deployApp on web-mode? 
If we already, should we remove those line above.

packages/web/README.md
rmeena840: Ok, the parameters link has been added.

PR #646, opened 2022-04-06
fix:fee computation on inner tx

vuvth:
## Proposed Changes
+ Fix bug fee computation on inner tx

Details:
https://www.pivotaltracker.com/n/projects/2452320/stories/181642443

### DIFF
--- a/packages/runtime/test/integration/c2c-call/c2c-call.ts
+++ b/packages/runtime/test/integration/c2c-call/c2c-call.ts
@@ -3,7 +3,7 @@ import { assert } from "chai";
 
 import { RUNTIME_ERRORS } from "../../../src/errors/errors-list";
 import { Runtime } from "../../../src/index";
-import { AccountStoreI, AppDeploymentFlags, AppInfo, TxReceipt } from "../../../src/types";
+import { AccountStoreI, AppDeploymentFlags, AppInfo } from "../../../src/types";
 import { useFixture } from "../../helpers/integration";
 import { expectRuntimeError } from "../../helpers/runtime-errors";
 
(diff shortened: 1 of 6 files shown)


packages/runtime/test/integration/c2c-call/c2c-call.ts
```diff
@@ -91,61 +91,59 @@ describe("C2C call", function () {
...
-	// 			fromAccount: alice.account,
-	// 			appID: firstApp.appID,
-	// 			foreignApps: [secondApp.appID],
-	// 			appArgs: ['str:call_method', `int:${1}`],
-	// 			payFlags: {
-	// 			totalFee: 1000
-	// 			}
-	// 		}
+	describe("Inner transaction in group", function () {
+		it("should succeed", () => {
```
sczembor, line 95: Maybe we should add a slightly longer explanation for the test case

packages/runtime/test/integration/c2c-call/c2c-call.ts
```diff
@@ -91,61 +91,59 @@ describe("C2C call", function () {
...
-	// 			sign: types.SignType.SecretKey,
-	// 			fromAccount: alice.account,
-	// 			appID: firstApp.appID,
-	// 			foreignApps: [secondApp.appID],
-	// 			appArgs: ['str:call_method', `int:${1}`],
-	// 			payFlags: {
-	// 			totalFee: 1000
-	// 			}
-	// 		}
+		it("should fail", () => {
```
sczembor, line 119: Same as above. Maybe something like "Should throw an exception when... "

sczembor (approved):
Left few comments. Otherwise looks good

packages/runtime/test/integration/c2c-call/c2c-call.ts
vuvth: yes, I will add more detail about test cases.

PR #648, opened 2022-04-06
Update `executeTransaction` / `executeTx` method in `web` and `algob` to make it compatible. 

sczembor:
Closes #x

## Proposed Changes

+ add `TransactionAndSign[]` as a parameter to `executeTx`
+ rename `executeTx` parameter name `execParams` to `transactions`
+ moved `executeTx` as a `Deployer` method

## Potential followups

+ We do not have any tests for `executeTx` in web mode.

### TODO

- [x] Update changelog
- [ ] Update documentation and tutorials
- [ ] Update the rest of the examples that use common.js 
- [x] Refactor deployer.ts (two very similar methods `executeTx`)


packages/web/src/lib/wallectconnect-mode.ts
```diff
@@ -201,20 +203,23 @@ export class WallectConnectSession {
 
 	/**
 	 * Execute single transaction or group of transactions (atomic transaction)
-	 * @param execParams transaction parameters or atomic transaction parameters
+	 * @param transactions transaction parameters,  atomic transaction parameters
+	 *  or TransactionAndSign object(SDK transaction object and signer parameters)
 	 */
 	async executeTx(
```
robert-zaremba, line 209: can we use `WebMode.executeTx` here?

robert-zaremba (commented):
changelog update is missing.

packages/web/src/lib/wallectconnect-mode.ts
```diff
@@ -201,20 +203,23 @@ export class WallectConnectSession {
...
 	): Promise<algosdk.modelsv2.PendingTransactionResponse> {
 		let signedTxn;
 		let txns: Transaction[] = [];
-		if (execParams.length > 16) {
+		if (transactions.length > 16) {
 			throw new Error("Maximum size of an atomic transfer group is 16");
 		}
-		for (const [_, txn] of execParams.entries()) {
-			txns.push(mkTransaction(txn, await mkTxParams(this.algodClient, txn.payFlags)));
+		if (!isSDKTransactionAndSign(transactions[0])) {
```
vuvth, line 217: What if transactions type is `TransactionAndSign`? 
I think we don't need require `sign` field in this case. I think we need few discuss about this problem.

packages/web/src/lib/wallectconnect-mode.ts
sczembor: I think this will be use to do once all the methods return the same type

packages/algob/src/internal/deployer.ts
```diff
@@ -878,6 +883,52 @@ export class DeployerDeployMode extends DeployerBasicMode implements Deployer {
...
+	 * Execute single transaction or group of transactions (atomic transaction)
+	 * executes `ExecParams` or `Transaction` Object, SDK Transaction object passed to this function
+	 * will be signed and sent to network. User can use SDK functions to create transactions.
+	 * Note: If passing transaction object a signer/s must be provided.
+	 * @param deployer Deployer
+	 * @param transactionParam transaction parameters or atomic transaction parameters
+	 * https://github.com/scale-it/algo-builder/blob/docs/docs/guide/execute-transaction.md
+	 * or TransactionAndSign object(SDK transaction object and signer parameters)
+	 */
+	async executeTx(
```
vuvth, line 896: How about move `executeTx` method to `DeployerBasicMode`? Because `DeployerRunMode` and `DeployerDeployerMode` both use same `executeTx`.

vuvth (commented):
Thank you for PR! I left few comments.

packages/algob/src/internal/deployer.ts
```diff
@@ -1054,4 +1105,44 @@ export class DeployerRunMode extends DeployerBasicMode implements Deployer {
...
+		}
+
+		const execParams = transactions as wtypes.ExecParams[];
+
+		this.assertCPNotDeleted(execParams);
+		const txIdxMap = new Map<number, [string, wtypes.ASADef]>();
+		const [txns, signedTxns] = await makeAndSignTx(this, execParams, txIdxMap);
+		await this.sendAndWait(signedTxns);
+		const confirmedTx = await this.getReceiptTxns(txns);
+		console.debug(confirmedTx);
```
robert-zaremba, line 1145: this lines are repeated, can we move it to a common private function?

robert-zaremba (approved):
lgtm!

packages/algob/src/internal/deployer.ts
sczembor: I am not sure what you mean by a private function. It cannot be moved 'up' because the base class does not implement the Deployer interface. To not repeat the code I moved it back as a function and just invoke from inside of the executeTx method in Deployer class. It will make the old code compatible but also allows users to use it as a deployer method now. LMK what do you think about it

robert-zaremba: I was thinking about some function in the file, which will take deployer as an argument. But if it's too hacky then we can keep as is.

PR #652, opened 2022-04-07
Feat/improve algob init

vuvth:
## Proposed Changes

+ Install algo-builder/web when create algob project 
+ CI test for algob create project


.github/workflows/05-examples.yaml
```diff
@@ -33,8 +33,8 @@ jobs:
           cache: "pipenv"
       - name: Install pipenv
         run: python -m pip install pipenv
-      - run: pipenv sync
-
+      - run: pipenv sync   
+
```
robert-zaremba, line 37: let's remove trailing whitespace

robert-zaremba (approved):

.github/workflows/05-examples.yaml
vuvth: Thanks! I didn't see it

PR #654, opened 2022-04-08
feat:global OpcodeBudget

vuvth:
## Proposed Changes

+ Global OpcodeBudget
+ Test case


rmeena840 (changes requested):

packages/runtime/test/integration/global-opcodebudget.ts
```diff
@@ -0,0 +1,146 @@
+import { parsing, tx, types } from "@algo-builder/web";
+import { getApplicationAddress } from "algosdk";
+import { assert } from "chai";
+
```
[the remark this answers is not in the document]
vuvth: fixed

packages/runtime/src/interpreter/opcode-list.ts
```diff
@@ -4064,6 +4076,13 @@ export class ITxnSubmit extends Op {
 			}
 		}
 
+		// increase Budget when submit application call transaction
+		for (const tx of this.interpreter.currentInnerTxnGroup) {
```
rmeena840, line 4080: I think you can also use the array filter method here as well. Filter by `tx.type === TransactionTypeEnum.APPLICATION_CALL`

get filtered array length and your `this.interpreter.runtime.ctx.budget = MAX_APP_PROGRAM_COST  * filtered array length;`

The same approach we discussed other places.

rmeena840 (changes requested):

packages/runtime/src/interpreter/opcode-list.ts
vuvth: Updated. I will apply this approach when I see a pattern like above.

rmeena840 (approved):
LGTM!. Thanks

packages/runtime/src/interpreter/interpreter.ts
```diff
@@ -67,12 +67,14 @@ export class Interpreter {
 	labelMap: Map<string, number>; // label string mapped to their respective indexes in instructions array
 	currentInnerTxnGroup: EncTx[]; // "current" inner transaction
 	innerTxnGroups: EncTx[][]; // executed inner transactions
-
+	cost: number; // total code
 	constructor() {
 		this.stack = new Stack<StackElem>();
+		this.mode = ExecutionMode.APPLICATION;
```
ratik21, line 73: why setting this? Interpreter can also run TEAL in `signature` mode, no?

packages/runtime/src/runtime.ts
```diff
@@ -727,6 +719,7 @@ export class Runtime {
 	): TxReceipt {
 		this.addCtxAppUpdateTx(senderAddr, appID, payFlags, flags);
 		this.ctx.debugStack = debugStack;
+		this.ctx.budget = MAX_APP_PROGRAM_COST;
```
ratik21, line 722: Same as above. I think instead of doing this in each function, we should have a global function (eg. `computeTxBudget(..)`), which computes the `this.ctx.budget` depending on the execution mode and type of transaction. It will reduce the complexity in each function

packages/runtime/src/runtime.ts
```diff
@@ -922,6 +915,13 @@ export class Runtime {
 			types.isSDKTransactionAndSign(txn) ? transactionAndSignToExecParams(txn, this.ctx) : txn
 		);
 
+		// calculate budget for single/group tx
+		const applCallTxNumber = gtxs.filter(
+			(txn) => txn.type === TransactionTypeEnum.APPLICATION_CALL
+		).length;
+
+		this.ctx.budget = MAX_APP_PROGRAM_COST * applCallTxNumber;
```
ratik21, line 923: same

packages/runtime/test/fixtures/stateful/assets/budget-opcode.teal
```diff
@@ -0,0 +1,35 @@
...
+
+int 1
+return
+
+budget: 
+int 1 
+return
+
+on_creation:
+int 1
```
ratik21, line 34: try to use pyTEAL from next time

packages/runtime/test/integration/global-opcodebudget.ts
```diff
@@ -0,0 +1,141 @@
...
+	let flags: AppDeploymentFlags;
+	let appID: number;
+	let dummyAppID: number;
+	let txnParam: types.ExecParams;
+	let lsig: LogicSigAccount;
+	this.beforeAll(async function () {
+		runtime = new Runtime([john]); // setup test
+
+		flags = {
+			sender: john.account,
```
ratik21, line 24: indentation is wrong. Could you please check your editor? A tab should be set to 2 spaces for javascript.

packages/runtime/test/integration/global-opcodebudget.ts
```diff
@@ -0,0 +1,141 @@
...
+				assert.deepEqual(
+					receipts[0].logs[0],
+					new TextDecoder().decode(parsing.uint64ToBigEndian(1396n))
+				);
+			} else {
+				assert.fail("log is undefined");
+			}
+		});
+	});
+});
```
ratik21, line 141: indentation for this whole file is wrong

packages/runtime/test/integration/global-opcodebudget.ts
```diff
@@ -0,0 +1,141 @@
...
+				payFlags: {
+					totalFee: 1000,
+				},
+			};
+
+			const receipts = runtime.executeTx([txnParam]);
+			if (receipts[0].logs) {
+				assert.deepEqual(
+					receipts[0].logs[0],
+					new TextDecoder().decode(parsing.uint64ToBigEndian(693n))
```
ratik21, line 84: What are we doing here?

packages/runtime/test/integration/global-opcodebudget.ts
```diff
@@ -0,0 +1,141 @@
...
+			};
+			const receipts = runtime.executeTx([txnParam, buffTx]);
+			if (receipts[0].logs) {
+				assert.deepEqual(
+					receipts[0].logs[0],
+					new TextDecoder().decode(parsing.uint64ToBigEndian(1393n))
+				);
+			} else {
+				assert.fail("log is undefined");
+			}
```
ratik21, line 114: I don't think this design is correct. No need to put in `if {} else { .. }`. Assert should fail if logs is undefined.

packages/runtime/test/integration/pooled-op-cost.ts
```diff
@@ -77,4 +77,67 @@ describe("TEALv5: Pooled Opcode Cost calculation", function () {
...
+					totalFee: 1000,
+				},
+			};
+		});
+
+		it("Should succeed because enough budget when group 2 appl call transaction", () => {
+			assert.doesNotThrow(() => runtime.executeTx([increaseBudgetTx, callAppTx]));
+		});
+
+		it("Should failed because not enough budget when group appl call and non appl call", () => {
```
ratik21, line 136: could you make it more verbose?

packages/runtime/src/interpreter/interpreter.ts
vuvth: When we call [interpreter.execute](https://github.com/scale-it/algo-builder/blob/feat/global-OpcodeBudget/packages/runtime/src/interpreter/interpreter.ts#L485), we pass ExecuteMode to parameter. I just initial value at here.

packages/runtime/src/interpreter/interpreter.ts
```diff
@@ -527,21 +536,20 @@ export class Interpreter {
...
+					assertMaxCost(dynamicCost, this.mode);
+					txReceipt.gas = this.cost;
 				} else {
 					this.runtime.ctx.pooledApplCost += this.lineToCost[instruction.line];
-					const maxPooledApplCost = MaxAppProgramCost * this.runtime.ctx.gtxs.length;
 					assertMaxCost(
 						this.runtime.ctx.pooledApplCost,
 						ExecutionMode.APPLICATION,
-						maxPooledApplCost
+						this.getBudget()
```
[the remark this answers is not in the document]
vuvth: No, depend on interpreter `mode`. We set interpreter mode when call interpreter.execute().

packages/runtime/src/interpreter/opcode-list.ts
```diff
@@ -1741,6 +1743,16 @@ export class Global extends Op {
...
 				break;
 			}
+
+			case "OpcodeBudget": {
+				const maxBudget = this.interpreter.getBudget();
+				const currentTotalCost =
+					this.interpreter.mode === ExecutionMode.SIGNATURE
+						? this.interpreter.cost
+						: this.interpreter.runtime.ctx.pooledApplCost;
+				result = maxBudget - (currentTotalCost + 1); // include global OpcodeBudget
```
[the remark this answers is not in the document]
vuvth: We also count `global Budget` as 1 cost, so I plus one to remain budget. Uhm, maybe it's not clean. Will improve it.

packages/runtime/src/runtime.ts
```diff
@@ -581,18 +582,7 @@ export class Runtime {
 		payFlags: types.TxParams,
 		debugStack?: number
 	): AppInfo {
```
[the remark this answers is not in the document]
vuvth: We already deprecated it.

packages/runtime/src/runtime.ts
vuvth: Good idea! 1000 👍🏻

packages/runtime/test/fixtures/global-opcodebudget/assets/app.teal
```diff
@@ -0,0 +1,45 @@
...
+global OpcodeBudget 
+itob
+log
+
+int 1
+return
+
+on_creation:
+int 1 
+return
```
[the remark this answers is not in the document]
vuvth: I will note it! Thanks

packages/runtime/test/integration/global-opcodebudget.ts
vuvth: I remember we are ran prettier in husky hook before commit. I will check it.

packages/runtime/test/integration/global-opcodebudget.ts
vuvth: We log opcodeBudget value, i check it is equal with expect value i want.

packages/runtime/test/integration/global-opcodebudget.ts
vuvth: I think so 👍🏻

packages/runtime/test/fixtures/global-opcodebudget/assets/app.teal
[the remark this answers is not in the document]
vuvth: I think this case we should use TEAL. Because we need compute cost of program. Pyteal make it harder to "count".

packages/runtime/src/parser/parser.ts
```diff
@@ -882,6 +882,50 @@ export function parser(
 	return opCodeList;
 }
 
+// check algorand is auto added intcblock for optimize size contract
+export function isAddIntcblock(ops: Operator[], interpreter: Interpreter): boolean {
```
rmeena840, line 886: Try to use arrow method here.

packages/runtime/src/parser/parser.ts
```diff
@@ -882,6 +882,50 @@ export function parser(
...
+			intCount[op.uint64.toString()] += 1;
+			if (intCount[op.uint64.toString()] >= 2) return true;
+		}
+	}
+
+	return false;
+}
+
+// check algorand is auto added byteblock for optimize size contract
+export function isAddedBytecblock(ops: Operator[], interpreter: Interpreter): boolean {
```
rmeena840, line 908: ditto

rmeena840 (commented):
Left comments

packages/runtime/src/parser/parser.ts
vuvth: Why we need to you arrow method?

vuvth: Okay ~~. We discussed, should use `function` keyword in this case

PR #657, opened 2022-04-16
log should have Uint8Array type

vuvth:
## Proposed Changes

+ Return log as Uint8Array type.


packages/runtime/src/interpreter/opcode-list.ts
```diff
@@ -4234,8 +4234,7 @@ export class ITxna extends Op {
 		if (this.interpreter.tealVersion >= 5 && this.field === "Logs") {
 			// handle Logs
 			const txReceipt = this.interpreter.runtime.ctx.state.txReceipts.get(tx.txID);
-			let logs: Buffer[] | string[] = txReceipt?.logs ?? [];
-			logs = logs.map((log) => convertToBuffer(log));
+			const logs: Uint8Array[] = txReceipt?.logs ?? [];
```
thdailong, line 4237: Why should we change from string to Uint8Array?

rmeena840 (approved):

PR #658, opened 2022-04-18
Added dao name to global state

Megha-Dev-19:
Added dao name to global state

### DIFF
--- a/examples/dao/assets/dao-app-approval.py
+++ b/examples/dao/assets/dao-app-approval.py
@@ -7,7 +7,7 @@ from pyteal import *
 def approval_program(ARG_GOV_TOKEN):
     """
     A stateful app with governance rules. Stores
-    deposit, min_support, min_duration, max_duration, url.
+    deposit, min_support, min_duration, max_duration, url, dao_name.
 
     Commands:
         add_proposal            Save proposal record in lsig
@@ -136,11 +136,13 @@ def approval_program(ARG_GOV_TOKEN):
     max_duration = Bytes("max_duration")
     # a url with more information about the DAO
     url = Bytes("url")
+     # dao name
+    dao_name = Bytes("dao_name")
 
 
     # initialization
     # Expected arguments:
-    #   [deposit, min_support, min_duration, max_duration, url]
+    #   [deposit, min_support, min_duration, max_duration, url, dao_name]
     on_initialize = Seq([
         Assert(
             And(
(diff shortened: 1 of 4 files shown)


vuvth (approved):
Preapproval! 
Is the Dao name used for all proposals or will each proposal have a different name?

examples/dao/assets/dao-app-approval.py
```diff
@@ -136,11 +136,13 @@ def scratch_proposal_active(idx: Int):
     max_duration = Bytes("max_duration")
     # a url with more information about the DAO
     url = Bytes("url")
+     # dao name
+    dao_name = Bytes("dao_name")
```
ratik21, line 140: you're adding more storage to global state in contract. But you're not updating the global state schema during deployment.

examples/dao/scripts/deploy.js
```diff
@@ -20,13 +20,15 @@ async function run(runtimeEnv, deployer) {
 	const minDuration = 1 * 60; // 1min (minimum voting time in number of seconds)
 	const maxDuration = 5 * 60; // 5min (maximum voting time in number of seconds)
 	const url = "www.my-url.com";
+	const daoName = "DAO";
```
ratik21, line 23: Is `daoName` different from `Dao App Name`?

examples/dao/test/dao-app-flow.js
```diff
@@ -118,6 +119,7 @@ describe("DAO test", function () {
 			`int:${minDuration}`,
 			`int:${maxDuration}`,
 			`str:${url}`,
+			`str:${daoName}`,
```
ratik21, line 122: Please add an assertion in test as well

ratik21 (changes requested):

examples/dao/assets/dao-app-approval.py
vuvth: I checked it. Number global state still is 2 slots.

Megha-Dev-19: `globalBytes: 2` yes we have 2 slots

ratik21 (approved):

examples/dao/scripts/deploy.js
Megha-Dev-19: yes, we use dao app name to store data in check point, while daoName is what we need to display on UI in sigma dao

Megha-Dev-19:
> Preapproval! Is the Dao name used for all proposals or will each proposal have a different name?

All proposals will have same dao name

PR #661, opened 2022-04-21
feat(algo-builder/web):early support smart signature in web mode

vuvoth:
## Proposed Changes

+ Early support smart signature transaction in web-mode.ts
+ Simple unit test for web-mode.ts
## Test plan 
On my machine 

Please review it! Thanks


packages/web/src/lib/web-mode.ts
```diff
@@ -142,23 +143,44 @@ export class WebMode {
...
+		const toBeSignedTxns = base64Txs.map((txn: string, txnId: number) => {
+			return execParams[txnId].sign === types.SignType.LogicSignature
+				? { txn: txn, signers: [] }
+				: { txn: txn };
 		});
+
 		const signedTxn = await this.signTransaction(toBeSignedTxns);
+
+		// sign smart signature transaction
+		for (let i = 0; i < txns.length; ++i) {
```
rmeena840, line 164: Just a note. Try to use below method for iterating to array.

`for (let profile of myArray) {
    console.log(profile.name) // profile is object in myArray
}`

This looks more clean. The advantage of this is that we don't have to calculate the length of array  on every iteration which is done here: `i < txns.length`

rmeena840 (commented):
Left comments.

packages/web/src/lib/web-mode.ts
vuvth: done

packages/web/src/lib/web-mode.ts
```diff
@@ -1,5 +1,6 @@
 import algosdk, { SuggestedParams, Transaction } from "algosdk";
 
+import { types } from "..";
```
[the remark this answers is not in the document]
vuvth: fixed

rmeena840 (commented):
It's a new featur that you have added. I think it's a good idea to add an unit test of it.

packages/web/test/mocks/algo-signer-mock.ts
```diff
@@ -0,0 +1,115 @@
...
+	AlgoSigner,
+	Encoding,
+	JsonPayload,
+	MultisigTransaction,
+	RequestErrors,
+	Transaction,
+	WalletTransaction,
+} from "../../src/algo-signer-types";
+
+const suggestedParamsMock = {
```
rmeena840, line 15: Shouldn't we also define the type of this? I think it's suggested params.

packages/web/test/src/lib/web-mode.ts
```diff
@@ -0,0 +1,31 @@
...
+	let sender: Account;
+	let receiver: Account;
+
+	this.beforeEach(() => {
+		sender = algosdk.generateAccount();
+		receiver = algosdk.generateAccount();
+		webMode = new WebMode(new AlgoSignerMock(), "Test");
+	});
+
+	it("#executeTx", async function () {
```
sczembor, line 18:
```suggestion
	it("Should executeTx without throwing an error", async function () {
```

packages/web/test/src/lib/web-mode.ts
```diff
@@ -0,0 +1,31 @@
+import algosdk, { Account } from "algosdk";
+import assert from "assert";
+
+import { types, WebMode } from "../../../src";
+import { AlgoSignerMock } from "../../mocks/algo-signer-mock";
+
+describe("Webmode test", function () {
```
sczembor, line 7: We could be more specific here

packages/web/test/mocks/algo-signer-mock.ts
```diff
@@ -0,0 +1,115 @@
...
+	signTxn(transactions: WalletTransaction[], error?: RequestErrors): Promise<JsonPayload> {
+		return new Promise((resolve, reject) => {
+			const result = Array(WallectConnectSession.length);
+			for (let txnId = 0; txnId < transactions.length; ++txnId) {
+				result[txnId] = {
+					txID: "tx-id",
+					blob: "blod",
+				};
+			}
+			return resolve(result as unknown as JsonPayload);
```
rmeena840, line 106: What are we trying to do here? First to unknown then JsonPayLoad type. Can't we do directly?

rmeena840 (changes requested):
Reviewed. Left comments.

sczembor (approved):
Left few comments

packages/web/test/mocks/algo-signer-mock.ts
vuvoth: We can't do it directly. TS throw error if we do it.

packages/web/test/mocks/algo-signer-mock.ts
vuvoth: Seem likes we can't use suggested params.

PR #663, opened 2022-04-26
Change returning type of executeTx

sczembor:
Closes #x

## Proposed Changes

+  `executeTx` method from `WebMode` class now returns `Promise<algosdk.modelsv2.PendingTransactionResponse>`.

### DIFF
--- a/packages/web/src/lib/web-mode.ts
+++ b/packages/web/src/lib/web-mode.ts
@@ -3,6 +3,7 @@ import algosdk, { SuggestedParams, Transaction } from "algosdk";
 import { AlgoSigner, JsonPayload, WalletTransaction } from "../algo-signer-types";
 import { BuilderError, ERRORS } from "../errors/errors";
 import { ExecParams, isSDKTransactionAndSign, TransactionAndSign, TxParams } from "../types";
+import { WAIT_ROUNDS } from "./constants";
 import { log } from "./logger";
 import { mkTransaction } from "./txn";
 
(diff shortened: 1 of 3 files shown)


packages/web/src/lib/web-mode.ts
```diff
@@ -38,7 +40,7 @@ export class WebMode {
 				pendingInfo[CONFIRMED_ROUND] !== null &&
 				(pendingInfo[CONFIRMED_ROUND] as number) > 0
 			) {
-				return pendingInfo;
+				return pendingInfo as unknown as algosdk.modelsv2.PendingTransactionResponse;
```
vuvth, line 43: Can we only use as `algosdk.modelsv2.PendingTransactionResponse` at here?

sczembor: No we can't, there are some properties missing

packages/web/src/lib/web-mode.ts
```diff
@@ -129,18 +131,22 @@ export class WebMode {
...
+	): Promise<algosdk.modelsv2.PendingTransactionResponse> {
 		let txns: Transaction[] = [];
-		if (execParams.length > 16) {
+		if (transactions.length > 16) {
 			throw new Error("Maximum size of an atomic transfer group is 16");
 		}
-
-		for (const [_, txn] of execParams.entries()) {
-			txns.push(mkTransaction(txn, await this.getSuggestedParams(txn.payFlags)));
+		if (!isSDKTransactionAndSign(transactions[0])) {
```
robert-zaremba, line 144: add a check before the if:
```
if (transactions.length == 0)
   throw new SomeErrorClass("at least one transaction must be provided")
```

Please find the best error for the error class from the errors we have defined.

sczembor: done

packages/web/src/lib/web-mode.ts
```diff
@@ -38,7 +41,7 @@ export class WebMode {
 				pendingInfo[CONFIRMED_ROUND] !== null &&
```
vuvth, line 41: How about only waiting 10 or 6 round?
We defined [WAIT_ROUNDS](https://github.com/scale-it/algo-builder/blob/develop/packages/web/src/lib/constants.ts#L4) in constants.ts

vuvth (approved):
lgtm! Just have on suggestion!

PR #665, opened 2022-05-02
base64_decode opcode cost

sczembor:
Closes #x

## Proposed Changes

+ Update the `execute` method to return the cost 
+ Add unit tests for the method
+ Update the rest of the opcodes to return their cost from `execute` method (Only the opcodes that cost is different from 1 return their cost

## Potential followups
+ Add testing for the interpreter to check if the the cost is correct

### DIFF
--- a/packages/runtime/src/interpreter/interpreter.ts
+++ b/packages/runtime/src/interpreter/interpreter.ts
@@ -554,7 +554,7 @@ export class Interpreter {
 			if (costFromExecute === undefined) {
 				this.cost += this.lineToCost[instruction.line];
 			} else {
-				this.cost = costFromExecute;
+				this.cost += costFromExecute;
 			}
 			if (this.tealVersion < 4) {
 				txReceipt.gas = this.gas;
@@ -564,7 +564,8 @@ export class Interpreter {
 					assertMaxCost(this.cost, this.mode);
 					txReceipt.gas = this.cost;
 				} else {
-					this.runtime.ctx.pooledApplCost += this.lineToCost[instruction.line];
+					//TODO here we need some updating probably the if else statement from above
+					this.runtime.ctx.pooledApplCost += this.cost;
 					assertMaxCost(
 						this.runtime.ctx.pooledApplCost,
 						ExecutionMode.APPLICATION,
--- a/packages/runtime/src/interpreter/interpreter.ts
+++ b/packages/runtime/src/interpreter/interpreter.ts
@@ -536,9 +536,6 @@ export class Interpreter {
 
 		while (this.instructionIndex < this.instructions.length) {
 			const instruction = this.instructions[this.instructionIndex];
-			//TODO this should return cost
-			const costFromExecute = instruction.execute(this.stack);
-
 			if (
 				this.runtime.ctx.isInnerTx &&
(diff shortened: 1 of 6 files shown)


packages/runtime/src/interpreter/interpreter.ts
```diff
@@ -548,14 +545,13 @@ export class Interpreter {
 					tealVersion: this.tealVersion,
 				});
 			}
+			const costFromExecute = instruction.execute(this.stack);
+			const cost =
+				costFromExecute === undefined ? this.lineToCost[instruction.line] : costFromExecute;
```
vuvth, line 550:
```suggestion
			const costFromExecute = instruction.execute(this.stack);
			const cost = costFromExecute ?? this.lineToCost[instruction.line];
```

vuvth: Should we return cost with all `instruction.execute()` ? Maybe we should do it in another PR.

packages/runtime/src/interpreter/opcode-list.ts
```diff
@@ -4794,9 +4794,10 @@ export class Base64Decode extends Op {
 		}
 	}
 
-	execute(stack: TEALStack): void {
+	execute(stack: TEALStack): number {
 		this.assertMinStackLen(stack, 1, this.line);
 		const last = this.assertBytes(stack.pop(), this.line);
+		const cost = 1 + Math.ceil(last.length / 16);
```
vuvth, line 4800: Please add comment to explain compute `cost` formula.

packages/runtime/test/src/interpreter/opcode-list.ts
```diff
@@ -6667,6 +6667,41 @@ describe("Teal Opcodes", function () {
...
+			assert.deepEqual(5, cost); // base64_decode cost = 5 (60 bytes -> 1 + ceil(60/16))
+
+			toPush = Buffer.from(
+				"ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-_AA==",
+				"utf-8"
+			);
+			stack.push(toPush);
+			op = new Base64Decode(["URLEncoding"], 0);
+			cost = op.execute(stack);
+			assert.deepEqual(6, cost); // base64_decode cost = 6 (68 bytes -> 1 + ceil(68/16))
```
vuvth, line 6702: How about the case `base64_opcode StdEncoding`?

vuvth: Ah, this test case for compute cost. Thanks!

vuvth (approved):
Nice work! I felt few comments.

packages/runtime/src/interpreter/interpreter.ts
sczembor: I will do it in next PR

packages/runtime/src/interpreter/opcode-list.ts
sczembor: Done

packages/runtime/src/interpreter/interpreter.ts
sczembor: I added the cost for the rest of the opcodes

PR #666, opened 2022-05-02
Fix/extract opcode

vuvoth:
## Proposed Changes

+ Parameter in extract* opcode should be bigint

## Test plan 
+ Unit test.

### DIFF
--- a/packages/runtime/test/src/interpreter/opcode-list.ts
+++ b/packages/runtime/test/src/interpreter/opcode-list.ts
@@ -5690,7 +5690,6 @@ describe("Teal Opcodes", function () {
 			stack.push(3n);
 			let op = new Extract3([], 1);
 			op.execute(stack);
-
 			assert.deepEqual(stack.pop(), new Uint8Array([23, 3, 2]));
 
 			stack.push(longByte);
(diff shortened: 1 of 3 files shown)


sczembor (approved):
Should we update the changeling in this case?

vuvoth:
> Should we update the changeling in this case?

Yes, I forgot it. Thanks

packages/runtime/test/src/interpreter/opcode-list.ts
```diff
@@ -5700,7 +5716,16 @@ describe("Teal Opcodes", function () {
...
 			op.execute(stack);
 
-			const expected = bigEndianBytesToBigInt(new Uint8Array([3, 4]));
+			let expected = bigEndianBytesToBigInt(new Uint8Array([3, 4]));
+			assert.equal(stack.pop(), expected);
+
+			// long byte
+			stack.push(longByte);
+			stack.push(300n);
+
```
rmeena840, line 5725: There is a lot of empty lines I see. You should remove them wherever not needed.

rmeena840 (changes requested):
Reviewed. Left comments.

packages/runtime/test/src/interpreter/opcode-list.ts
vuvoth: Done

rmeena840 (approved):

PR #667, opened 2022-05-03
fix/Can not issue itxn in clear program

vuvoth:
## Proposed Changes

+ Can not issue itxn in clear program
+ Unit test

### DIFF
--- a/packages/runtime/src/errors/errors-list.ts
+++ b/packages/runtime/src/errors/errors-list.ts
@@ -413,9 +413,9 @@ maximun uint128`,
 	},
 	ISSUE_ITXN_WHEN_CLEAR_PROGRAM: {
 		number: 1060,
-		message: "Clear state programs can not issue inner transactions.",
-		title: "Clear state programs can not issue inner transactions.",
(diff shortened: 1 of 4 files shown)


packages/runtime/src/errors/errors-list.ts
```diff
@@ -411,6 +411,12 @@ maximun uint128`,
 		title: PARSE_ERROR,
 		description: `value %exp% is not base64Url`,
 	},
+	ISSUE_ITXN_WHEN_CLEAR_PROGRAM: {
+		number: 1060,
+		message: "Clear state programs can not issue inner transactions.",
+		title: "Clear state programs can not issue inner transactions.",
+		description: "Clear state programs can not issue inner transactions",
```
ratik21, line 418:
```suggestion
		message: "Clear state programs cannot issue inner transactions.",
		title: "Clear state programs cannot issue inner transactions.",
		description: "Clear state programs cannot issue inner transactions",
```

ratik21 (approved):
We need a changelog update

PR #692, opened 2022-05-26
Add test scenarios for interpreter

thdailong:
Closes #x

## Proposed Changes

- Setup for interpreter to execute
- Add test `gas` for the test cost
- Add test non trivial code with if-else

### DIFF
--- a/packages/runtime/test/src/interpreter/interpreter.ts
+++ b/packages/runtime/test/src/interpreter/interpreter.ts
@@ -59,7 +59,7 @@ describe("Interpreter", function () {
 		// setup 2nd account
 		johnAcc = new AccountStore(0, johnAccount);
 
-		// setup 2nd account
+		// setup 3rd account
 		bobAccount = new AccountStore(1000000, bobAcc);
 
 		// setup application account
--- a/packages/runtime/test/src/interpreter/interpreter.ts
+++ b/packages/runtime/test/src/interpreter/interpreter.ts
@@ -1,0 +1,126 @@
+import { bobAcc } from "@algo-builder/algob/test/mocks/account";
+import { decodeAddress, getApplicationAddress } from "algosdk";
+import { assert } from "chai";
+
+import { AccountStore, getProgram, Interpreter, Runtime } from "../../../src";
+import RUNTIME_ERRORS from "../../../src/errors/errors-list";
+import { ALGORAND_ACCOUNT_MIN_BALANCE } from "../../../src/lib/constants";
+import { Stack } from "../../../src/lib/stack";
+import { AccountAddress, AccountStoreI, ExecutionMode, StackElem } from "../../../src/types";
+import { useFixture } from "../../helpers/integration";
+import { expectRuntimeError } from "../../helpers/runtime-errors";
+import { elonMuskAccount, johnAccount } from "../../mocks/account";
+import { accInfo } from "../../mocks/stateful";
+import { elonAddr, TXN_OBJ } from "../../mocks/txn";
+
+export function setDummyAccInfo(acc: AccountStoreI): void {
+	acc.appsLocalState = accInfo[0].appsLocalState;
+	acc.appsTotalSchema = accInfo[0].appsTotalSchema;
+	acc.createdApps = accInfo[0].createdApps;
+}
+
+describe("Interpreter", function () {
+	let interpreter: Interpreter;
+	let elonAcc: AccountStoreI;
+	let johnAcc: AccountStoreI;
+	let bobAccount: AccountStoreI;
+	let appAccAddr: AccountAddress;
+	let applicationAccount: AccountStoreI;
+	const elonPk = decodeAddress(elonAddr).publicKey;
+
+	const reset = (): void => {
+		while (interpreter.stack.length() !== 0) {
+			interpreter.stack.pop();
+		}
+		interpreter.currentInnerTxnGroup = [];
+		interpreter.runtime.ctx.pooledApplCost = 0;
+		interpreter.instructions = [];
+		interpreter.innerTxnGroups = [];
+		interpreter.instructionIndex = 0;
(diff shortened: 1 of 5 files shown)


packages/runtime/test/src/interpreter/interpreter.ts
```diff
@@ -0,0 +1,126 @@
...
+			const file = "test-label.teal";
+			expectRuntimeError(
+				() =>
+					interpreter.execute(getProgram(file), ExecutionMode.SIGNATURE, interpreter.runtime),
+				RUNTIME_ERRORS.TEAL.REJECTED_BY_LOGIC
+			);
+		});
+	});
+
+	// it("Should return correct cost for a .teal program", () => {
```
sczembor, line 122: Why is it commented out?

thdailong: Oh sorry, this one I test yesterday and I find that this file is logic failed. 
Let me delete this

sczembor: But should it fail? If the .teal file is correct but it fails we should take a closer look to it

packages/runtime/test/src/interpreter/interpreter.ts
```diff
@@ -0,0 +1,126 @@
...
+describe("Interpreter", function () {
+	let interpreter: Interpreter;
+	let elonAcc: AccountStoreI;
+	let johnAcc: AccountStoreI;
+	let bobAccount: AccountStoreI;
+	let appAccAddr: AccountAddress;
+	let applicationAccount: AccountStoreI;
+	const elonPk = decodeAddress(elonAddr).publicKey;
+
+	const reset = (): void => {
```
sczembor, line 31: Maybe we could name it `resetInterpreterState`

packages/runtime/test/src/interpreter/interpreter.ts
```diff
@@ -0,0 +1,126 @@
...
+
+export function setDummyAccInfo(acc: AccountStoreI): void {
+	acc.appsLocalState = accInfo[0].appsLocalState;
+	acc.appsTotalSchema = accInfo[0].appsTotalSchema;
+	acc.createdApps = accInfo[0].createdApps;
+}
+
+describe("Interpreter", function () {
+	let interpreter: Interpreter;
+	let elonAcc: AccountStoreI;
```
sczembor, line 24: take a look at the `defaultAccounts` in runtime. We use them in other tests

packages/runtime/test/src/interpreter/interpreter.ts
thdailong: It is like the file teal "test-label.teal" does not have any thing to return and maybe it is wrong logic.
But I find that code like this::

`#pragma version 4\n
int 5\n
int 3\n
==`

would cause the error TEAL.REJECTED_BY_LOGIC

packages/runtime/test/src/interpreter/interpreter.ts
```diff
@@ -0,0 +1,126 @@
...
+			sk: new Uint8Array(0),
+		});
+
+		interpreter = new Interpreter();
+		interpreter.runtime = new Runtime([elonAcc, johnAcc, bobAccount, applicationAccount]);
+		interpreter.tealVersion = tealVersion;
+		reset();
+	};
+	this.beforeAll(() => {
+		setUpInterpreter(6); //setup interpreter for execute
```
sczembor, line 77: why do we set tealVersion to 6 here? I think it should be based on the #pragma version

thdailong: I put default tealVersion to 6 for the largest version. And in the process interpreter.execute, the tealVersion with change based on the #pragma version like you said. 
However, now I think I should put default version to 1 in case there is no op codes #pragma version

sczembor: Oh right, also I think we should add test scenarios for different teal versions, because there are some opcodes that have different costs in different teal versions.

packages/runtime/test/src/interpreter/interpreter.ts
thdailong: I use `defaultAccounts` which has already set up in the inner-transactions file.
So is there things get wrong here?

packages/runtime/test/src/interpreter/interpreter.ts
```diff
@@ -0,0 +1,134 @@
...
+		appBalance: number = ALGORAND_ACCOUNT_MIN_BALANCE
+	): void => {
+		// setup 1st account (to be used as sender)
+		elonAcc = new AccountStore(0, elonMuskAccount); // setup test account
+		setDummyAccInfo(elonAcc);
+
+		// setup 2nd account
+		johnAcc = new AccountStore(0, johnAccount);
+
+		// setup 2nd account
```
sczembor, line 62: something wrong here

thdailong: Is it? I am just set up the account like the set up in the inner-transactions (https://github.com/scale-it/algo-builder/blob/11b6f9c318e80ad1b4ea866837a6b9777c4897fc/packages/runtime/test/src/interpreter/inner-transaction.ts#L52)

packages/runtime/test/src/interpreter/interpreter.ts
```diff
@@ -0,0 +1,134 @@
...
+			resetInterpreterState(); //after each execute need to resetInterpreterState interpreter
+			interpreter.cost = 0;
+			file = "test-file-3.teal";
+			interpreter.execute(getProgram(file), ExecutionMode.SIGNATURE, interpreter.runtime);
+			assert.equal(interpreter.cost, 3);
+
+			resetInterpreterState();
+			interpreter.cost = 0;
+			file = "test-interpreter.teal";
+			interpreter.execute(getProgram(file), ExecutionMode.SIGNATURE, interpreter.runtime);
```
vuvoth, line 100: I recommend move it to separate `it` block.

PR #702, opened 2022-06-02
Add txn LastLog

vuvoth:
## Proposed Changes

- Txn LastLog
- unit test

### DIFF
--- a/packages/runtime/src/lib/constants.ts
+++ b/packages/runtime/src/lib/constants.ts
@@ -152,6 +152,7 @@ TxnFields[5] = {
 
 TxnFields[6] = {
 	...TxnFields[5],
+	LastLog: null,
 };
 
 TxnFields[7] = {
--- a/packages/runtime/src/lib/txn.ts
+++ b/packages/runtime/src/lib/txn.ts
@@ -57,6 +57,13 @@ export function parseToStackElem(a: unknown, field: TxField): StackElem {
 	if (Buffer.isBuffer(a)) {
 		return new Uint8Array(a);
 	}
+
+	// case Uint8Array but empty, some magic here
+	// I also don't how it work :|
+	if (ArrayBuffer.isView(a)) {
+		return new Uint8Array(a.buffer);
+	}
+
 	if (typeof a === "number" || typeof a === "bigint" || typeof a === "boolean") {
 		return BigInt(a);
 	}
@@ -175,6 +182,7 @@ export function txnSpecByField(
 			break;
 		}
 		case "LastLog": {
+			result = interpreter.runtime.ctx.lastLog;
 			break;
 		}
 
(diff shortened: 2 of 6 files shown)


packages/runtime/src/lib/constants.ts
```diff
@@ -152,6 +152,7 @@ TxnFields[5] = {
 
 TxnFields[6] = {
 	...TxnFields[5],
+	LastLog: null,
```
vuvoth, line 155: I'm not happy with current TxnFields format(another data like this). 
We can do it better like 
```js
TxnFields = Map([
   FieldName: {
      minVersionSupport, 
      minITxnVersionSupport, 
      isArray, 
      mappingEncField: "snd"....
    }
])
```

vuvoth, line 155: I'm not happy with current TxnFields format(another data like this). 
We can do it better like 
```js
TxnFields = Map([
   FieldName: {
      minVersionSupport, 
      minITxnVersionSupport, 
      isArray, 
      mappingEncField: "snd"....
    }
])
```

packages/runtime/src/lib/txn.ts
```diff
@@ -57,6 +57,13 @@ export function parseToStackElem(a: unknown, field: TxField): StackElem {
 	if (Buffer.isBuffer(a)) {
 		return new Uint8Array(a);
 	}
+
+	// case Uint8Array but empty, some magic here
+	// I also don't how it work :|
+	if (ArrayBuffer.isView(a)) {
```
vuvoth, line 63: if a === [] then Buffer.isBuffer will return false.

vuvoth, line 63: if a === [] then Buffer.isBuffer will return false.

packages/runtime/test/src/interpreter/opcode-list.ts
```diff
@@ -13,6 +13,7 @@ import { ec as EC } from "elliptic";
 import { describe } from "mocha";
 
 import { AccountStore } from "../../../src/account";
+import { Ctx } from "../../../src/ctx";
```
vuvoth, line 16:
```suggestion
```

vuvoth, line 16:
```suggestion
```

packages/runtime/src/lib/constants.ts
robert-zaremba: why you are not happy with the current format? I think the format we are using is pretty clear.

packages/runtime/src/lib/txn.ts
robert-zaremba: `ArrayBuffer.isView([]) === false`
So maybe we should check `Array.isArray`? and put it in one if:
```
if (Buffer.isBuffer(a) || Array.isArray(a) || ArrayBuffer.isView(a))
```

robert-zaremba (approved):
left few comments

packages/runtime/src/lib/txn.ts
vuvoth: Yup! I will try update it.

PR #704, opened 2022-06-03
fix typo

vuvoth:
## Proposed Changes

- Docs params

### DIFF
--- a/packages/runtime/src/logicsig.ts
+++ b/packages/runtime/src/logicsig.ts
@@ -123,7 +123,7 @@ export class LogicSig {
 
 	/**
 	 * Performs signature verification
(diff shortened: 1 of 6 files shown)


thdailong (approved):
lgtm looking good!

packages/runtime/src/logicsig.ts
```diff
@@ -123,7 +123,7 @@ export class LogicSig {
 
 	/**
 	 * Performs signature verification
-	 * @param accAddr Sender's account address
+	 * @param accPk Sender's account pk
```
robert-zaremba, line 126:
```suggestion
	 * @param accPk Sender's account public key
```

robert-zaremba (approved):

PR #707, opened 2022-06-06
Algob.balanceOf update

vuvoth:
## Proposed Changes

- For Algob.balanceOf(deployer, accountAddr, assetID) if assetID is undefined then function will return ALGO account balance.


thdailong (approved):
LGTM! You act really fast with my requirement!

rmeena840 (approved):

PR #711, opened 2022-06-09
Add inner tx to execute return types

vuvoth:
## Proposed Changes
Add missing fields to ConfirmedTxInfo

```ts
export interface ConfirmedTxInfo {
	"confirmed-round": number;
	"asset-index": number;
	"application-index": number;
	"global-state-delta"?: algosdk.modelsv2.EvalDeltaKeyValue;
	"local-state-delta"?: algosdk.modelsv2.AccountStateDelta;
	"inner-txns"?: ConfirmedTxInfo;
	txn: algosdk.EncodedSignedTransaction;
}

```

### DIFF
--- a/packages/algob/src/types.ts
+++ b/packages/algob/src/types.ts
@@ -780,6 +780,8 @@ export interface ConfirmedTxInfo {
 	"confirmed-round": number;
 	"asset-index": number;
 	"application-index": number;
-	"global-state-delta": string;
-	"local-state-delta": string;
+	"global-state-delta": algosdk.EncodedGlobalStateSchema;
+	"local-state-delta": algosdk.EncodedLocalStateSchema;
+	"inner-txns": ConfirmedTxInfo;
+	txn: algosdk.EncodedSignedTransaction;
 }
--- a/packages/algob/src/types.ts
+++ b/packages/algob/src/types.ts
@@ -780,8 +780,8 @@ export interface ConfirmedTxInfo {
 	"confirmed-round": number;
 	"asset-index": number;
 	"application-index": number;
-	"global-state-delta": algosdk.EncodedGlobalStateSchema;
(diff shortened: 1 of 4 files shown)


packages/algob/src/types.ts
```diff
@@ -780,6 +780,8 @@ export interface ConfirmedTxInfo {
 	"confirmed-round": number;
 	"asset-index": number;
 	"application-index": number;
-	"global-state-delta": string;
-	"local-state-delta": string;
+	"global-state-delta"?: algosdk.modelsv2.EvalDeltaKeyValue;
+	"local-state-delta"?: algosdk.modelsv2.AccountStateDelta;
+	"inner-txns"?: ConfirmedTxInfo;
+	txn: algosdk.EncodedSignedTransaction;
```
sczembor, line 786: Could you explain why this one is without parenthesis?

packages/algob/test/mocks/tx.ts
```diff
@@ -39,12 +39,77 @@ export const mockSuggestedParam: SuggestedParams = {
 	genesisHash: "SGO1GKSzyE7IEPItTxCByw9x8FmnrCDexi9/cOUJOiI=",
 };
 
+const account = algosdk.generateAccount();
+const addr = algosdk.decodeAddress(account.addr);
+
+export const TXN_OBJ = {
```
sczembor, line 45: maybe we can add a comment here about the type or just link do documentation

sczembor (approved):
lgtm! left few comments

packages/algob/src/types.ts
vuvoth: So { "txn" : any } will equal {txn: any}. JS "sugar syntax" I think

packages/algob/test/mocks/tx.ts
vuvoth: Done

PR #716, opened 2022-06-16
update pr template

vuvoth:
## Proposed Changes

- new pr template


robert-zaremba (approved):
I would us labels for PR type rather than a writing it in the description

PR #724, opened 2022-06-27
Add: Add search by dao name backend support for Sigma Dao with pagination

rmeena840:
Add: Add search by dao name backend support for Sigma Dao with pagination
Task: https://www.pivotaltracker.com/story/show/182496847

### DIFF
--- a/examples/dao/Makefile
+++ b/examples/dao/Makefile
@@ -1,4 +1,4 @@
-export SIGMA_DAO_UNAME = sigma_dao_user
+export SIGMA_DAO_UNAME = test_user
 export SIGMA_DAO_PASSWORD = SigmaDao@1234
 # Below are objects needed for sigma dao app. Extend it if more sigma dao objects needed
 export SIGMA_DAO_OBJECTS = sigma_daos, asset, account_asset, account_app
@@ -13,8 +13,8 @@ drop-create-sigma-dao-user: drop-sigma-dao-user create-sigma-dao-user
 # create sigma_dao_user
 create-sigma-dao-user:
 	@echo "Setting up User=$(SIGMA_DAO_UNAME)"
-	@sudo -u postgres createuser $(SIGMA_DAO_UNAME)
-	@sudo -u postgres psql -c "ALTER USER $(SIGMA_DAO_UNAME) WITH ENCRYPTED PASSWORD '$(SIGMA_DAO_PASSWORD)';"
+	@psql -U postgres -c "CREATE USER $(SIGMA_DAO_UNAME);"
+	@psql -U postgres -c "ALTER USER $(SIGMA_DAO_UNAME) WITH ENCRYPTED PASSWORD '$(SIGMA_DAO_PASSWORD)';"
 # grant read only access.
 	@psql -U postgres -d $(POSTGRES_DBNAME) -c "GRANT SELECT ON TABLE $(SIGMA_DAO_OBJECTS) TO $(SIGMA_DAO_UNAME);"
 
--- a/examples/dao/Makefile
+++ b/examples/dao/Makefile
@@ -13,8 +13,7 @@ drop-create-sigma-dao-user: drop-sigma-dao-user create-sigma-dao-user
 # create sigma_dao_user
 create-sigma-dao-user:
 	@echo "Setting up User=$(SIGMA_DAO_UNAME)"
-	@psql -U postgres -c "CREATE USER $(SIGMA_DAO_UNAME);"
-	@psql -U postgres -c "ALTER USER $(SIGMA_DAO_UNAME) WITH ENCRYPTED PASSWORD '$(SIGMA_DAO_PASSWORD)';"
+	@psql -U postgres -c "CREATE USER $(SIGMA_DAO_UNAME); ALTER USER $(SIGMA_DAO_UNAME) WITH ENCRYPTED PASSWORD '$(SIGMA_DAO_PASSWORD)'"
 
 # drop sigma_dao_user
 drop-sigma-dao-user:
--- a/examples/dao/Makefile
+++ b/examples/dao/Makefile
@@ -1,7 +1,5 @@
 export SIGMA_DAO_UNAME = sigma_dao_user
 export SIGMA_DAO_PASSWORD = SigmaDao@1234
-# Below are objects needed for sigma dao app. Extend it if more sigma dao objects needed
-export SIGMA_DAO_OBJECTS = sigma_daos, asset, account_asset, account_app
 export SIGMA_DAO_SETUP_FILE = setup_sigma_dao.sql
 # https://github.com/scale-it/algo-builder/blob/326d6d34a7279dc0efc32cbcab0fc9ad4161c293/infrastructure/Makefile#L17
 export POSTGRES_UNAME = algorand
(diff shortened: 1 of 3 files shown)


examples/dao/Makefile
```diff
@@ -13,10 +13,8 @@ drop-create-sigma-dao-user: drop-sigma-dao-user create-sigma-dao-user
 # create sigma_dao_user
 create-sigma-dao-user:
 	@echo "Setting up User=$(SIGMA_DAO_UNAME)"
-	@sudo -u postgres createuser $(SIGMA_DAO_UNAME)
-	@sudo -u postgres psql -c "ALTER USER $(SIGMA_DAO_UNAME) WITH ENCRYPTED PASSWORD '$(SIGMA_DAO_PASSWORD)';"
-# grant read only access.
-	@psql -U postgres -d $(POSTGRES_DBNAME) -c "GRANT SELECT ON TABLE $(SIGMA_DAO_OBJECTS) TO $(SIGMA_DAO_UNAME);"
```
Megha-Dev-19, line 19: why we are moving it to setup-sigma-dao?

examples/dao/setup_sigma_dao.sql
```diff
@@ -5,9 +5,10 @@
 -- create a sigma_daos table if does not exit already.
 CREATE TABLE IF NOT EXISTS sigma_daos (
   id SERIAL PRIMARY KEY, -- auto increment id
-  app_id bigint, -- application id
-  app_params jsonb, -- application params with approval etc.
-  asset_id bigint -- token id
+  app_id BIGINT, -- application id
+  dao_name CHAR(255), -- dao name
+  app_params JSONB, -- application params with approval etc.
+  asset_id BIGINT -- token id
```
Megha-Dev-19, line 11: why are we moving to Caps here?

rmeena840: I prefer to keep sql keywords in caps. This is also recommended by postgres as well.

examples/dao/Makefile
rmeena840: sigma dao relation should be present and indexer should be running to grant access to sigma dao objects. When I was working on it last time, it was available for me. I noticed this deleting the database.

rmeena840:
@robert-zaremba Please review the PR.

examples/dao/Makefile
```diff
@@ -13,10 +13,8 @@ drop-create-sigma-dao-user: drop-sigma-dao-user create-sigma-dao-user
...
+	@psql -U postgres -c "CREATE USER $(SIGMA_DAO_UNAME);"
+	@psql -U postgres -c "ALTER USER $(SIGMA_DAO_UNAME) WITH ENCRYPTED PASSWORD '$(SIGMA_DAO_PASSWORD)';"
```
robert-zaremba, line 17:
```suggestion
	@psql -U postgres -c "CREATE USER $(SIGMA_DAO_UNAME); ALTER USER $(SIGMA_DAO_UNAME) WITH ENCRYPTED PASSWORD '$(SIGMA_DAO_PASSWORD)'"
```

examples/dao/Makefile
```diff
@@ -25,6 +23,8 @@ drop-sigma-dao-user:
 	@psql -d $(POSTGRES_DBNAME) -c "DROP OWNED BY $(SIGMA_DAO_UNAME);"
 	@psql -d $(POSTGRES_DBNAME) -c "DROP USER $(SIGMA_DAO_UNAME);"
 
-# execute sigma dao setup file
+# execute sigma dao setup file. Indexer should be running before executing this script
 setup-sigma-dao:
 	@psql -U $(POSTGRES_UNAME) -d $(POSTGRES_DBNAME) -a -f $(SIGMA_DAO_SETUP_FILE)
+# grant read only access.
+	@psql -U postgres -d $(POSTGRES_DBNAME) -c "GRANT SELECT ON TABLE $(SIGMA_DAO_OBJECTS) TO $(SIGMA_DAO_UNAME);"
```
robert-zaremba, line 30: can we add this statement to the SIGMA_DAO_SETUP_FILE ?

examples/dao/setup_sigma_dao.sql
```diff
@@ -18,12 +19,14 @@ CREATE OR REPLACE FUNCTION sigmadao_trigger_fn()
...
+	app_bytecode TEXT; -- app hash
+	dao_name_key_bytecode CHAR(255) := 'ZGFvX25hbWU='; -- Byte code of 'dao_name' key
+	dao_name CHAR(255); -- dao_name
+	i RECORD;
 BEGIN
 	/* 	Check to avoid duplicate entries in table. id attribute increments despite pass/fail insertion  */
 	IF EXISTS (SELECT FROM sigma_daos p WHERE p.app_id = NEW.asset) THEN
```
robert-zaremba, line 32: how and when this is possible?

examples/dao/setup_sigma_dao.sql
```diff
@@ -61,3 +68,9 @@ EXECUTE FUNCTION sigmadao_trigger_fn();
 
 -- grant privileges to relation sigma_daos
 GRANT ALL PRIVILEGES ON TABLE sigma_daos TO algorand;
+
+-- Function to search dao name in sigma_daos relation
+CREATE OR REPLACE FUNCTION search_sigma_daos(daoToBeSearched TEXT)
+RETURNS SETOF sigma_daos AS $$
+	SELECT * FROM sigma_daos WHERE dao_name ilike ('%' || daoToBeSearched || '%');
+$$ LANGUAGE SQL STABLE;
```
robert-zaremba, line 76: why do we need this function? We can use SELECT directly?

rmeena840: This is needed because postgraphile does only the exact search. E.g if there a dao named 'DAO' then if you search for 'dao' or 'DA', then no result will be returned.

examples/dao/setup_sigma_dao.sql
rmeena840: txn relation will have two entries for same app id and data, that's why our trigger will be called two times. This is just to avoid duplicate entries in table. This is safe check to avoid duplicates. Also, `id` attributes will increase even if insertion failed on 2nd time,

examples/dao/Makefile
rmeena840: We can move but I would suggest to keep it here because in future if variables value like `POSTGRES_DBNAME `, `SIGMA_DAO_OBJECTS ` changes then we have to do in file as well. Is it ok?

examples/dao/setup_sigma_dao.sql
robert-zaremba: My question was about introducing the function rather than using `select` directly. I thought we can add any handler to postgraphile, including raw SQL.

examples/dao/setup_sigma_dao.sql
robert-zaremba: Could you this to the comment? What are the two entries?

examples/dao/setup_sigma_dao.sql
rmeena840: We have to create function in postgresql so that postgraphile can recognize it as query. For more info on custom query: https://www.graphile.org/postgraphile/custom-queries/

examples/dao/setup_sigma_dao.sql
rmeena840: ok. I have added the comment.

examples/dao/setup_sigma_dao.sql
robert-zaremba: OK, thanks!

examples/dao/Makefile
robert-zaremba: I think it's better to make all setup in one file. It's also easier to test.

robert-zaremba (approved):
lgtm

examples/dao/Makefile
rmeena840: Ok. I have moved it to sql file.

PR #725, opened 2022-06-28
Fix algosigner authorization txn

vuvoth:
## Proposed Changes

- Support authorization transaction(rekey). We can interact with rekeyed account now.


sczembor (approved):
lgtm!

PR #726, opened 2022-06-30
create function to save checkpoint from confirmed txn(inner txn/normal txn)

vuvoth:
## Proposed Changes

- create `checkpointApp` and `checkpointASA`. Those function take confirmTx and save (appInfo and AsaInfo) to checkpoint. Developers will use this function to inner transaction asset if they want. 

## Notes: 
- I don't check type of transaction now. 
- Only create new checkpoint, not update for now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,8 @@ Features, Bug Fixes, API Breaking, Deprecated, Infrastructure, Template Updates
 - Add Runtime.getAppByName(appName): get app by name declared in appDefinition.
 - For Algob.balanceOf(deployer, accountAddr, assetID) if assetID is undefined then function will return ALGO account balance.
 - Add new example [Trampoline](https://github.com/algorand-devrel/demo-avm1.1/tree/master/demos/trampoline)
-
+- Add `checkpointASA(asaName, txConfirmed)` and `checkpointApp(appName, txConfirmed)`. The methods will extract information about application and asa from `txConfirmed` and save it to checkpoint with name custom by developers.
+ 
 ### Bug Fixes
 
 - Fix number transaction in one call should be 256(include inner and atomic transaction).
--- a/packages/algob/src/internal/deployer.ts
+++ b/packages/algob/src/internal/deployer.ts
@@ -575,7 +575,7 @@ export class DeployerDeployMode extends DeployerBasicMode implements Deployer {
 	}
 
 	/**
-	 * Save new ASA information(get from txConfirmation) to checkpoint
+	 * Save new ASA information(loaded from txConfirmation) to checkpoint
 	 * @param asaName name of asa
 	 * @param txConfirmation txn confirmation
 	 */
--- a/packages/algob/src/internal/deployer.ts
+++ b/packages/algob/src/internal/deployer.ts
@@ -579,7 +579,7 @@ export class DeployerDeployMode extends DeployerBasicMode implements Deployer {
 	 * @param asaName name of asa
 	 * @param txConfirmation txn confirmation
 	 */
-	checkpointASA(asaName: string, txConfirmation: ConfirmedTxInfo): void {
+	addCheckpointASA(asaName: string, txConfirmation: ConfirmedTxInfo): void {
 		const txn = txConfirmation.txn.txn;
 
 		this.cpData.registerASA(this.networkName, asaName, {
@@ -600,7 +600,7 @@ export class DeployerDeployMode extends DeployerBasicMode implements Deployer {
 	 * @param appName app name
 	 * @param txConfirmation txn confirmation
 	 */
-	checkpointApp(appName: string, txConfirmation: ConfirmedTxInfo): void {
+	addCheckpointApp(appName: string, txConfirmation: ConfirmedTxInfo): void {
 		const { txn } = txConfirmation.txn;
 		this.cpData.registerSSC(this.networkName, appName, {
 			appID: txConfirmation["application-index"],
@@ -984,6 +984,28 @@ export class DeployerRunMode extends DeployerBasicMode implements Deployer {
 	registerSSCInfo(name: string, sscInfo: rtypes.AppInfo): void {
 		throw new BuilderError(ERRORS.BUILTIN_TASKS.DEPLOYER_EDIT_OUTSIDE_DEPLOY, {
 			methodName: "registerSSCInfo",
+		});
+	}
+
+	/**
+	 * Checkpoint feature disable on run mode
+	 * @param asaName name of asa
+	 * @param txConfirmation txn confirmation
+	 */
+	addCheckpointASA(asaName: string, txConfirmation: ConfirmedTxInfo): void {
+		throw new BuilderError(ERRORS.BUILTIN_TASKS.DEPLOYER_EDIT_OUTSIDE_DEPLOY, {
+			methodName: "addCheckpointASA",
+		});
+	}
+
+	/**
+	 * Checkpoint feature disable on run mode
+	 * @param appName app name
+	 * @param txConfirmation txn confirmation
+	 */
+	addCheckpointApp(appName: string, txConfirmation: ConfirmedTxInfo): void {
+		throw new BuilderError(ERRORS.BUILTIN_TASKS.DEPLOYER_EDIT_OUTSIDE_DEPLOY, {
+			methodName: "addCheckpointApp",
 		});
 	}
 
--- a/packages/algob/src/internal/deployer.ts
+++ b/packages/algob/src/internal/deployer.ts
@@ -574,6 +574,11 @@ export class DeployerDeployMode extends DeployerBasicMode implements Deployer {
 		persistCheckpoint(this.txWriter.scriptName, this.cpData.strippedCP);
 	}
 
+	/**
+	 * Save new ASA information(get from txConfirmation) to checkpoint
+	 * @param asaName name of asa
+	 * @param txConfirmation txn confirmation
+	 */
 	registerASAInfoFromInnerTxn(asaName: string, txConfirmation: ConfirmedTxInfo): void {
 		const txn = txConfirmation.txn.txn;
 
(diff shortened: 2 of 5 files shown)


packages/algob/src/internal/deployer.ts
```diff
@@ -573,6 +574,47 @@ export class DeployerDeployMode extends DeployerBasicMode implements Deployer {
 		persistCheckpoint(this.txWriter.scriptName, this.cpData.strippedCP);
 	}
 
+	/**
+	 * Save new ASA information(get from txConfirmation) to checkpoint
+	 * @param asaName name of asa
+	 * @param txConfirmation txn confirmation
+	 */
+	registerASAInfoFromConfirmTx(asaName: string, txConfirmation: ConfirmedTxInfo): void {
```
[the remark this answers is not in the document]
robert-zaremba: we can simplify the name

packages/algob/src/internal/deployer.ts
```diff
@@ -573,6 +574,47 @@ export class DeployerDeployMode extends DeployerBasicMode implements Deployer {
...
+			applicationAccount: algosdk.getApplicationAddress(txConfirmation["application-index"]),
+			confirmedRound: txConfirmation["confirmed-round"],
+			approvalFile: txn.apap?.toString("base64") as string,
+			clearFile: txn.apsu?.toString("base64") as string,
+			creator: algosdk.encodeAddress(txn.snd),
+			timestamp: Math.round(+new Date() / 1000),
+			deleted: false,
+			txID: algosdk.Transaction.from_obj_for_encoding(txn).txID(),
+		});
+	}
```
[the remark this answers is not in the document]
vuvoth: When transaction confirm, we will return [txn receipt](https://github.com/scale-it/algo-builder/blob/develop/packages/algob/src/types.ts#L785 ), include inner transaction receipt.
 We can use methods below to save new checkpoint from inner txn receipt. This is main use case here.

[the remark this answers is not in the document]
robert-zaremba: `checkpointApp` is never called. Is it visible to a user? If yes, then we should add it to an interface.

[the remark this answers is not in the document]
vuvoth: > Is it visible to a user? 
Yes, user will call it when they want to create checkpoint from txConfirmInfo. 

You mean we should move it to `Deployer` inferface? 

.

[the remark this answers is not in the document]
robert-zaremba: Yes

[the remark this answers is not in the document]
vuvoth: Done!

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ Features, Bug Fixes, API Breaking, Deprecated, Infrastructure, Template Updates
 - Add Runtime.getAppByName(appName): get app by name declared in appDefinition.
 - For Algob.balanceOf(deployer, accountAddr, assetID) if assetID is undefined then function will return ALGO account balance.
 - Add new example [Trampoline](https://github.com/algorand-devrel/demo-avm1.1/tree/master/demos/trampoline)
+- Add `checkpointASA(asaName, txConfirmed)` and `checkpointApp(appName, txConfirmed)`. The methods will extract information about application and asa from `txConfirmed` and save it to checkpoint with name custom by developers.
```
ratik21, line 22: are these for deployer or runtime? We should mention that here

packages/algob/src/internal/deployer.ts
```diff
@@ -573,6 +574,47 @@ export class DeployerDeployMode extends DeployerBasicMode implements Deployer {
 		persistCheckpoint(this.txWriter.scriptName, this.cpData.strippedCP);
 	}
 
+	/**
+	 * Save new ASA information(get from txConfirmation) to checkpoint
```
ratik21, line 578:
```suggestion
	 * Save new ASA information(loaded from txConfirmation) to checkpoint
```

packages/algob/src/internal/deployer.ts
```diff
@@ -573,6 +574,47 @@ export class DeployerDeployMode extends DeployerBasicMode implements Deployer {
...
+	 * @param asaName name of asa
+	 * @param txConfirmation txn confirmation
+	 */
+	addCheckpointASA(asaName: string, txConfirmation: ConfirmedTxInfo): void {
```
ratik21, line 582: rename *txConfirmation* => *confirmedTxObj*

packages/algob/src/internal/deployer.ts
```diff
@@ -573,6 +574,47 @@ export class DeployerDeployMode extends DeployerBasicMode implements Deployer {
...
+		const txn = txConfirmation.txn.txn;
+
+		this.cpData.registerASA(this.networkName, asaName, {
+			assetIndex: txConfirmation["asset-index"],
+			creator: algosdk.encodeAddress(txn.apar?.m as Uint8Array),
+			confirmedRound: txConfirmation["confirmed-round"],
+			deleted: false,
+			assetDef: {
+				total: txn.apar?.t as number,
+				decimals: txn.apar?.dc as number,
```
ratik21, line 592: Do we need `as number`? If we're using optional chaining then perhaps the result could be `undefined`. Shouldn't we handle that? Otherwise please add a comment

packages/algob/src/internal/deployer.ts
```diff
@@ -573,6 +574,47 @@ export class DeployerDeployMode extends DeployerBasicMode implements Deployer {
...
+			txID: algosdk.Transaction.from_obj_for_encoding(txn).txID(),
+		});
+	}
+
+	/**
+	 * Save new application information(get from txConfirmation) to checkpoint
+	 * @param appName app name
+	 * @param txConfirmation txn confirmation
+	 */
+	addCheckpointApp(appName: string, txConfirmation: ConfirmedTxInfo): void {
```
ratik21, line 603: same as above

packages/algob/src/internal/deployer.ts
```diff
@@ -573,6 +574,47 @@ export class DeployerDeployMode extends DeployerBasicMode implements Deployer {
...
+		const { txn } = txConfirmation.txn;
+		this.cpData.registerSSC(this.networkName, appName, {
+			appID: txConfirmation["application-index"],
+			applicationAccount: algosdk.getApplicationAddress(txConfirmation["application-index"]),
```
ratik21, line 607: + Confused here. How do we ensure here that the confirmed transaction is an app call/app deploy? Or if we're calling this specifically "after" that app deploy call -- we should rename this function to `addInnerAppDeployCP`/`addInnerAppDeployCheckpoint`

+ Is it possible to deploy multiple apps during inner transaction? In that case is this function called multiple times?

packages/algob/src/internal/deployer.ts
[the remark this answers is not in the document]
ratik21: > We can use methods below to save new checkpoint from inner txn receipt. This is main use case here.
> Yes, user will call it when they want to create checkpoint from txConfirmInfo.

If we're exposing these functions in the `deployer`, then does the end user calls this function? I don't follow a user using a `registerSSC` (checkpoint specific) function.

I think these should be used internally by the `deployer` and add a checkpoint after tx (and inner transactions) are confirmed. Not sure about an end user using these functions

packages/algob/src/internal/deployer.ts
```diff
@@ -945,6 +987,28 @@ export class DeployerRunMode extends DeployerBasicMode implements Deployer {
...
+	/**
+	 * Checkpoint feature disable on run mode
+	 * @param asaName name of asa
+	 * @param txConfirmation txn confirmation
+	 */
+	addCheckpointASA(asaName: string, txConfirmation: ConfirmedTxInfo): void {
+		throw new BuilderError(ERRORS.BUILTIN_TASKS.DEPLOYER_EDIT_OUTSIDE_DEPLOY, {
+			methodName: "addCheckpointASA",
+		});
+	}
```
ratik21, line 999: What if i am doing an `app call` in *RUN* mode (script), which is actually deploying a new app internally? In that case application deployment (via inner tx), would be *RUN* mode, not *DEPLOY* mode.

packages/algob/test/internal/deployer.ts
```diff
@@ -676,4 +677,74 @@ describe("DeployerDeployMode", () => {
...
+
+			// get checkpoint data
+			const checkpointData = deployerCfg.cpData.precedingCP["network 123"].asa.get("asaName");
+
+			assert.deepEqual(checkpointData, asaInfo);
+
+			assert.deepEqual(asaInfo, deployer.getASAInfo("asaName"));
+		});
+
+		it("should fails when use addCheckpointApp and addCheckpointASA in run mode", () => {
```
ratik21, line 736:
```suggestion
		it("should fail when we use addCheckpointApp and addCheckpointASA in run mode", () => {
```

packages/algob/test/internal/deployer.ts
```diff
@@ -676,4 +677,74 @@ describe("DeployerDeployMode", () => {
...
+
+			deployer.addCheckpointASA("asaName", mockConfirmedTx);
+
+			// get checkpoint data
+			const checkpointData = deployerCfg.cpData.precedingCP["network 123"].asa.get("asaName");
+
+			assert.deepEqual(checkpointData, asaInfo);
+
+			assert.deepEqual(asaInfo, deployer.getASAInfo("asaName"));
+		});
```
ratik21, line 734: Can we also add tests for deploying multiple apps/asa's ?

packages/algob/src/internal/deployer.ts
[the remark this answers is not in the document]
vuvoth: The main challenge is checkpoint need `name` to save data. For inner txn the struct is unpredictable. I think better way is give the work for users who know what they want.

packages/algob/src/internal/deployer.ts
vuvoth: I will check it.

packages/algob/src/internal/deployer.ts
vuvoth: I also think about it! I think we can run something like `yarn algob deploy ./script/run.js`. How you think about it?

packages/algob/test/internal/deployer.ts
vuvoth: Yeah, sure! I am missing the test for dup checkpoint. 😅

PR #727, opened 2022-06-30
SigmaDAO index

robert-zaremba:
## Proposed Changes

- add fast index for DAO name queries

### DIFF
--- a/examples/dao/setup_sigma_dao.sql
+++ b/examples/dao/setup_sigma_dao.sql
@@ -11,8 +11,12 @@ CREATE TABLE IF NOT EXISTS sigma_daos (
   asset_id BIGINT -- token id
 );
 
--- Create unique index on app_id
+-- Create indexes
 CREATE UNIQUE INDEX sigma_daos_app_id_idx ON sigma_daos (app_id);
+-- create an index to do efficient text search using `%<text>%`, it requires
+-- pg_trgm extension.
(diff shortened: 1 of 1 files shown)


examples/dao/setup_sigma_dao.sql
```diff
@@ -11,8 +11,12 @@ CREATE TABLE IF NOT EXISTS sigma_daos (
   asset_id BIGINT -- token id
 );
 
--- Create unique index on app_id
+-- Create indexes
 CREATE UNIQUE INDEX sigma_daos_app_id_idx ON sigma_daos (app_id);
+-- create an index to do efficient text search using `%<text>%`, it requires
+-- pg_trgm extension.
+CREATE EXTENSION pg_trgm;
+CREATE INDEX sigma_daos_dao_name_idx ON users USING gin (dao_name gin_trgm_ops);
```
ratik21, line 19: Are there any tests/performance benchmarks for the querying we're doing using postgres?

robert-zaremba: no, we don't have a test data. Would be good to generate some. But probably later once we are closer to production.

rmeena840, line 19: `users`. What table is this? We don't have such a table.

rmeena840 (commented):
Reviewed

PR #728, opened 2022-06-30
SendAndWait fn added to web

Megha-Dev-19:
Added a send transaction to network and wait for confirmation to all three wallet config.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,9 @@ Features, Bug Fixes, API Breaking, Deprecated, Infrastructure, Template Updates
 
 ### Breaking Changes
 
-- Renamed `optIntoAsa` to `optInToAsa` to remain naming convention consistency across the project. 
+- Renamed `optIntoAsa` to `optInToAsa` to remain naming convention consistency across the project.
+- Renamed `sendTransaction` to `sendAndWait` in WebMode and parameter is updated to accept `string` to bring consistency with other wallets class.
+
 ### Deprecated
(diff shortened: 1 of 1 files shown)


packages/web/src/lib/web-mode.ts
```diff
@@ -63,14 +63,18 @@ export class WebMode {
...
+	 * Send signed transaction to network and wait for confirmation
+	 * @param signedTxn Signed Transaction blob encoded in base64
 	 */
-	async sendTransaction(signedTxn: any): Promise<JsonPayload> {
-		return await this.algoSigner.send({
+	async sendAndWait(signedTxn: string): Promise<algosdk.modelsv2.PendingTransactionResponse> {
+		const txInfo = await this.algoSigner.send({
 			ledger: this.chainName,
-			tx: signedTxn.blob,
+			tx: signedTxn,
```
vuvoth, line 72: Algosigner API is change?

Megha-Dev-19: No, it is same but to keep method consistent with other wallets method in which blob is used, here also the user has to send the blob as parameter rather than the Transaction object

vuvoth (approved):
LGTM! Don't forget update Changelog.md. We are making breaking change

robert-zaremba (approved):

Megha-Dev-19:
> LGTM! Don't forget update Changelog.md. We are making breaking change

yeah sure, will do that

PR #732, opened 2022-07-06
Feature: Filter (all, active, past, future) DB support for sigma dao

rmeena840:


examples/dao/setup_sigma_dao.sql
```diff
@@ -11,6 +11,16 @@ CREATE TABLE IF NOT EXISTS sigma_daos (
   asset_id BIGINT -- token id
 );
 
+-- create a sigma_dao_proposal table if does not exit already.
+CREATE TABLE IF NOT EXISTS sigma_dao_proposals (
+  id SERIAL PRIMARY KEY, -- auto increment id
+  addr BYTEA,
+  app BIGINT,
+  localstate JSONB,
```
robert-zaremba, line 19: how about renaming to "details". Please add a link in a comment to a resource describing what are we going to store here (a link to a code or spec)

examples/dao/setup_sigma_dao.sql
```diff
@@ -60,31 +70,74 @@ BEGIN
...
+CREATE OR REPLACE FUNCTION sigmadao_proposals_trigger_fn()
+RETURNS TRIGGER
+AS $$
+DECLARE
+	voting_start_key CHAR(255) := 'dm90aW5nX3N0YXJ0'; -- Byte code of 'voting_end'
+	voting_end_key CHAR(255) := 'dm90aW5nX2VuZA=='; -- Byte code of 'voting_end'
+	voting_start_value BIGINT;
+	voting_end_value BIGINT;
+BEGIN
+	IF (SELECT NEW.localstate::jsonb -> 'tkv' -> voting_start_key) IS NOT NULL THEN
```
robert-zaremba, line 83: what is `tkv`? Could you add a comment?

examples/dao/setup_sigma_dao.sql
```diff
@@ -60,31 +70,74 @@ BEGIN
...
 CREATE OR REPLACE FUNCTION search_sigma_daos(daoToBeSearched TEXT)
 RETURNS SETOF sigma_daos AS $$
 	SELECT * FROM sigma_daos WHERE dao_name ilike ('%' || daoToBeSearched || '%');
 $$ LANGUAGE SQL STABLE;
 
--- Function to search proposal in account_app relation by app id
-CREATE OR REPLACE FUNCTION search_sigma_daos_proposals(appId BIGINT)
-RETURNS SETOF account_app AS $$
+-- Function to search proposal in sigma_dao_proposals relation by app id
+CREATE OR REPLACE FUNCTION sigma_daos_proposal_filter(appId BIGINT, filterType INT)
```
robert-zaremba, line 123: do we need to add a pagination info? Maybe we can add it later.

examples/dao/setup_sigma_dao.sql
```diff
@@ -60,31 +70,74 @@ BEGIN
...
+RETURNS SETOF sigma_dao_proposals AS $$
 DECLARE
 	voting_end CHAR(255) := 'dm90aW5nX2VuZA=='; -- Byte code of 'voting_end'
+	current_time bigint := extract(epoch FROM NOW()); -- epoch in second
+	filter_all INT := 1; -- filter type -> all
+	filter_active INT := 2; -- filter type -> active
+	filter_future INT := 3; -- filter type -> future
```
robert-zaremba, line 130:
```suggestion
	filter_active INT := 3;  -- ongoing + future
```

examples/dao/setup_sigma_dao.sql
```diff
@@ -60,31 +70,74 @@ BEGIN
...
-CREATE OR REPLACE FUNCTION search_sigma_daos_proposals(appId BIGINT)
-RETURNS SETOF account_app AS $$
+-- Function to search proposal in sigma_dao_proposals relation by app id
+CREATE OR REPLACE FUNCTION sigma_daos_proposal_filter(appId BIGINT, filterType INT)
+RETURNS SETOF sigma_dao_proposals AS $$
 DECLARE
 	voting_end CHAR(255) := 'dm90aW5nX2VuZA=='; -- Byte code of 'voting_end'
+	current_time bigint := extract(epoch FROM NOW()); -- epoch in second
+	filter_all INT := 1; -- filter type -> all
+	filter_active INT := 2; -- filter type -> active
```
robert-zaremba, line 129:
```suggestion
	filter_ongoing INT := 2;
```

examples/dao/setup_sigma_dao.sql
```diff
@@ -60,31 +70,74 @@ BEGIN
...
+	filter_future INT := 3; -- filter type -> future
+	filter_past INT := 4; -- filter type -> past
 BEGIN
-	RETURN QUERY SELECT * FROM account_app WHERE app=appId AND localstate -> 'tkv' -> voting_end IS NOT NULL;
+	IF (filterType = filter_active) THEN
```
robert-zaremba, line 133:
```suggestion
	IF (filterType = filter_ongoing) THEN
```

examples/dao/setup_sigma_dao.sql
```diff
@@ -60,31 +70,74 @@ BEGIN
...
+		RETURN QUERY SELECT * FROM sigma_dao_proposals WHERE app=appId AND current_time BETWEEN sigma_dao_proposals.voting_start AND sigma_dao_proposals.voting_end ORDER BY voting_start;
+	ELSIF (filterType = filter_future) THEN
+		RETURN QUERY SELECT * FROM sigma_dao_proposals WHERE app=appId AND current_time < voting_start ORDER BY voting_start;
```
robert-zaremba, line 136:
```suggestion
	ELSIF (filterType = filter_active) THEN
		RETURN QUERY SELECT * FROM sigma_dao_proposals WHERE app=appId AND current_time < voting_end ORDER BY voting_start;
```

examples/dao/setup_sigma_dao.sql
rmeena840: The other table `account_app` also used the same column name. We are using trigger on that table update. I think keeping the same column name would be good idea.

examples/dao/setup_sigma_dao.sql
rmeena840: done

examples/dao/setup_sigma_dao.sql
rmeena840: No, we don't need that. Postgraphile handles that for us.

rmeena840:
@robert-zaremba This PR is ready for 2nd review.

examples/dao/setup_sigma_dao.sql
robert-zaremba: but this functions returns all records. We don't want to extract all records from DB if it's not necessary.

rmeena840: Yes, we need all the details.

robert-zaremba: record = row. This function doesn't know if it should return only first 100 proposals or the 100-199 proposals.

examples/dao/setup_sigma_dao.sql
```diff
@@ -11,6 +11,16 @@ CREATE TABLE IF NOT EXISTS sigma_daos (
   asset_id BIGINT -- token id
 );
 
+-- create a sigma_dao_proposal table if does not exit already.
+CREATE TABLE IF NOT EXISTS sigma_dao_proposals (
+  id SERIAL PRIMARY KEY, -- auto increment id
+  addr BYTEA,  -- account address
+  app BIGINT, -- app id
```
robert-zaremba, line 18: this should be foreign key. Please add a constraint

robert-zaremba, line 18:
```suggestion
  app BIGINT, -- SigmaDAO app id
```

robert-zaremba (approved):
pre-approving, before merging make sure to add foreign key as noted in the comment

examples/dao/setup_sigma_dao.sql
rmeena840: Done.

PR #737, opened 2022-07-12
Fix asset sender

### LINKED ISSUE #729: Txn.asset_sender() gives different values in runtime and TestNet
## Describe the bug

I have a smart contract in which I verify the amount, ID and sender of the asset received in a group transaction. I am sending the NFT in the second transaction and the function call is the first transaction.

While unit testing I am using the `runtime` and I am able to verify the value of `Gtxn[1].asset_sender()` to the stored address in the smart contract. However I was getting an assertion error for this when I deployed it on the `TestNet` and called the same function.

**Note that in both cases `clawback` address was not set.**

After some debugging I found that in the `TestNet`, the same smart contract was comparing the stored address of the sender to `Global.zero_address()` whose value is `AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAY5HFKQ`.
This address was getting returned from `Gtxn[1].asset_sender()` however, while using AlgoBuilder runtime, I was getting the actual address of the sender instead.

I created the NFT like this:
```js
nft_id = runtime.deployASADef(
                    "ABC",
                    {
                        total: 1,
                        decimals: 0,
                    },
                    {
                        creator: user.account,
                    }
                ).assetIndex;
```
If I print the asset details using `runtime.getAssetDef(nft_id)` I get the following:
```js
 {
  attribute_map: {},
  name: 'ABC',
  creator: 'NWAMR6YPAGXXMFYEGIJAK3L2PQORKHF2RZDCGYIVGXRNUP35EO64TT7CZI',
  total: 1n,
  decimals: 0,
  defaultFrozen: false,
  unitName: undefined,
  url: undefined,
  metadataHash: undefined,
  manager: undefined,
  reserve: undefined,
  freeze: undefined,
  clawback: undefined
}
```
#### Expected Behavior

As mentioned here [https://developer.algorand.org/docs/get-details/transactions/transactions/#assetsender](https://developer.algorand.org/docs/get-details/transactions/transactions/#assetsender) it is supposed to return `ZeroAddress` 

#### Current Behavior
`Gtxn[1].asset_sender()` returning `Global.zero_address()` in `TestNet`
`Gtxn[1].asset_sender()` returning Sender's address in Algobuilder Runtime

## Steps to Reproduce

1. Create a smart contract with a function that can take a group transaction and has a sender's address stored in a global variable
2. In the function assert the value of Gtxn[1].asset_sender() with stored_sender_address
3. Try calling the function with algobuilder runtim - this should pass
4. Try calling the function after deploying on TestNet (AlgoSigner) - this will fail

### Environment

- Node.js version:
- algob version:

vuvoth:
Closes #729 

## Proposed Changes

- Return zero for asset_sender. 
- Fix dao testcases. 

## TODO

- [ ] update change log

### DIFF
--- a/examples/dao/assets/dao-app-approval.py
+++ b/examples/dao/assets/dao-app-approval.py
@@ -438,7 +438,8 @@ def approval_program():
                     And(
                         Global.group_size() == Int(2),
                         Gtxn[1].type_enum() == TxnType.AssetTransfer,
-                        Gtxn[1].asset_sender() == proposal_from,
+                        Gtxn[1].asset_sender() == Global.zero_address(),
+                        Gtxn[1].sender() == proposal_from,
                         Gtxn[1].asset_receiver() == recipient,
                         Gtxn[1].asset_amount() == amount,
                         Gtxn[1].xfer_asset() == asa_id,
--- a/packages/runtime/src/lib/txn.ts
+++ b/packages/runtime/src/lib/txn.ts
@@ -16,6 +16,7 @@ import {
 	TransactionTypeEnum,
 	TxFieldDefaults,
 	TxnFields,
+	ZERO_ADDRESS,
 	ZERO_ADDRESS_STR,
 } from "../lib/constants";
 import {
@@ -159,10 +160,9 @@ export function txnSpecByField(
 			break;
 		}
 		case "AssetSender": {
-			/// + for asset_transfer transactions, we use "snd"
-			/// + for revoke asset tx (also an asset_transfer) tx, we use "asnd"
+			// if tx.asns is undefined we return zero address;
 			if (tx.type === "axfer") {
-				result = tx.asnd ?? tx.snd;
+				result = tx.asnd ?? Buffer.from(ZERO_ADDRESS);
 			}
 			break;
 		}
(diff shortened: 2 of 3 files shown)


kashishkhullar:
Thank you for fixing this issue. I need this change to complete my unit testing. Will this be available in the next release?

vuvoth:
> Thank you for fixing this issue. I need this change to complete my unit testing. Will this be available in the next release?

Yes! and if you want to try it right now. You can build it from source!

robert-zaremba:
@kashishkhullar there are few ways how you can test the build from `develop` branch. I think the easiest one is to clone the repo and install algob in your system (eg, `npm install -g`) or run algob from the algob repo directly (this is what I'm using):

```
git clone https://github.com/scale-it/algo-builder.git
cd algo-builder
yarn build
cd <path to my project>
# you can add the line below as a script in your $PATH directory
<path to algo-builder>/packages/algob/node_modules/.bin/algob init mynewproject
```

Or use a local "dev" project in algob package, this is what we often use for testing.
```
cd algo-builder/packages/algob
./project-dev-script.sh create
```

We will prepare a guide for that.

PR #739, opened 2022-07-13
Docs executetx

robert-zaremba:
- updated deployer and web mode docs related to tx creator.

### DIFF
--- a/packages/algob/src/types.ts
+++ b/packages/algob/src/types.ts
@@ -443,7 +443,8 @@ export interface Deployer {
 	/**
 	 * Creates and deploys ASA defined in asa.yaml.
 	 * @name  ASA name - deployer will search for the ASA in the /assets/asa.yaml file
-	 * @flags  deployment flags */
+	 * @flags  deployment flags. Required.
+	 *   `flags.creator` must be defined - it's an account which will sign the constructed transaction. */
 	deployASA: (
 		name: string,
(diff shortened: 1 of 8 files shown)


packages/algob/src/types.ts
```diff
@@ -443,7 +443,8 @@ export interface Deployer {
 	/**
 	 * Creates and deploys ASA defined in asa.yaml.
 	 * @name  ASA name - deployer will search for the ASA in the /assets/asa.yaml file
-	 * @flags  deployment flags */
+	 * @flags  deployment flags. Required.
+	 *   `flags.creator` must be defined - it's an account which will sign the constructed transaction. */
```
ratik21, line 447: I think we should also add a note for rekeyed accounts. In that case `ExecParams` should be used, or the direct api's should be updated to handle txns via rekeyed accounts

ratik21 (approved):

packages/algob/src/types.ts
robert-zaremba: hmm, maybe I'm missing something. Creator is a account with secret key. So it will have a proper secret key, right?

ratik21: replied on discord

robert-zaremba:
I've added the following note:

> support for rekeyed accounts is limited (creator must have updated sk to properly sign transaction)

PR #752, opened 2022-07-27
Fix gitxna opcode

sczembor:
Closes #x

## Proposed Changes
- `gitxna 1 Logs 0` now works properly
- Added unit tests


packages/runtime/src/lib/txn.ts
```diff
@@ -221,6 +221,14 @@ export function txAppArg(
 ): StackElem {
 	const tealVersion: number = interpreter.tealVersion;
 
+	if (txField === "Logs") {
```
vuvoth, line 224: I think we should move it to [txnSpecByField](https://github.com/scale-it/algo-builder/blob/29657c46caaefa4b8869677e33a69623ca55b11c/packages/runtime/src/lib/txn.ts#L99). get i
Because we don't get it from specific field from Txn object.

packages/runtime/src/lib/constants.ts
```diff
@@ -189,7 +189,7 @@ export const TxArrFields: { [key: number]: Set<string> } = {
 TxArrFields[3] = new Set([...TxArrFields[2], "Assets", "Applications"]);
 TxArrFields[4] = cloneDeep(TxArrFields[3]);
 TxArrFields[5] = cloneDeep(TxArrFields[4]);
-TxArrFields[6] = cloneDeep(TxArrFields[5]);
+TxArrFields[6] = new Set([...TxArrFields[5], "Logs"]);
```
vuvoth, line 192: In the opcode document, I see we have Logs for itxn in version 5. However you add it in teal 6. Is this still work?
https://github.com/algorand/go-algorand/blob/master/data/transactions/logic/TEAL_opcodes.md#txna-f-i

vuvoth (commented):
Thank for your PR! I left few comments.

packages/runtime/src/lib/txn.ts
robert-zaremba: yes

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ Features, Bug Fixes, API Breaking, Deprecated, Infrastructure, Template Updates
 
 - Fix `txn AssetSender` should return zero address by default.
 - Fix `KMDCredentialsFromEnv` loading using KMD_DATA. Algob was trying to use `env.$KMD_DATA` instead of `env.KMD_DATA`
+- Fix `gitxna 1 Logs 0` opcode.
```
ratik21, line 29: Please describe the bug/error, what we fixed here.

packages/runtime/src/lib/constants.ts
ratik21: It should be 5 i think

packages/runtime/src/lib/txn.ts
```diff
@@ -221,6 +221,14 @@ export function txAppArg(
...
+		const txReceipt = interpreter.runtime.ctx.state.txReceipts.get(tx.txID);
+		const logs: Uint8Array[] = txReceipt?.logs ?? [];
+		op.checkIndexBound(idx, logs, op.line);
+		const log = logs[idx];
+		return parseToStackElem(log, txField);
```
ratik21, line 229:
```suggestion
		return parseToStackElem(logs[idx], txField);
```

packages/runtime/src/lib/txn.ts
sczembor: I think we need it in both cases. We can do `gitxna 0 Logs 1` and `gitxnas 0 Logs`

vuvoth: After discuss, we will keep this line at here. However we will refactor it in future

vuvoth (approved):

Issue #759, opened 2022-08-10
TEAL 7 makes runtime fail


MetaB0y:
I am doing `teal = compileTeal(pyteal, mode=Mode.Application, version=MAX_PROGRAM_VERSION)` in my PyTEAL, probably as many other people do.

PyTEAL recently updated `MAX_PROGRAM_VERSION` to 7. It makes Alogbuilder runtime fail with errors like: 
```
TypeError: Cannot read properties of undefined (reading 'has')
      at setInnerTxField (node_modules/.store/@algo-builder-runtime-npm-5.0.1-79534551d4/package/src/lib/itxn.ts:152:33)
      at ITxnField.execute (node_modules/.store/@algo-builder-runtime-npm-5.0.1-79534551d4/package/src/interpreter/opcode-list.ts:4191:38)
```
This particular error occurs because `uintTxtFields` doesn't have a key 7.

 I understand that TEAL 7 is not yet implemented neither in Algobuilder nor Algorand mainnet, but I think it should be run the same as TEAL 6 for now instead of throwing slightly mysterious errors.

robert-zaremba:
we are slowly adding Teal v7 support.
Could you share your PyTeal code (or part of it)? This way we can prioritize the parts which are failing related to your code?

robert-zaremba:
Should be solved here: https://github.com/scale-it/algo-builder/pull/787/files#diff-fb6d679ed8d29771c0ca17e37361f75ff22872dc78ff51f35e147ea868d90a6aR41

robert-zaremba:
It's in our develop release, will be part of the next v6.0 release.

PR #761, opened 2022-08-10
Add "Assets" field to assetIDFields in runtime

MetaB0y:
It works the same way as others, I don't see any reason why it was not included previously; I assume it was forgotten? 

Possible follow-ups:
* Maybe it is not the only field omitted in such a way


robert-zaremba (approved):
utACK, 
Yes, it was forgotten. We will check if other fields are missing.

robert-zaremba:
We will backport it to v5.x release

MetaB0y:
Thanks! If you are working on it, maybe it would make sense to provide stub support for TEAL 7? Similarly to what Algobuilder has for TEAL 6, i.e. `assetIDFields[7] = cloneDeep(assetIDFields[6]);` etc.

See https://github.com/scale-it/algo-builder/issues/759 for some context.

MetaB0y:
Hmm maybe it actually breaks something...

I found myself in a situation where `interpreter.getAssetIDByReference`  fails on line ` if (appForeignAssets.includes(assetRef)) {` because `appForeginAssets` (which equals `this.runtime.ctx.tx.apas`) is `number` while it should be `number[]`.

MetaB0y:
Yes, it is wrong, but I know how to do it properly, I will update this MR later

MetaB0y:
I implemented a proper fix and added it to tests (modified tests were failing before the fix).

I think it also needs the support of `Accounts` which is not implemented at the moment.

I will probably work on it soon.

MetaB0y:
Accounts are done. I don't see anything [important] missing now. I feel like it can be merged and backported

@robert-zaremba PTAL

MetaB0y:
@robert-zaremba friendly ping

MetaB0y:
Don't mean to be intrusive but could somebody take a look? It's blocking me and I have to change compiled js to make algobuilder work for my project which is quite inconvenient.

PabloLION:
Same problem here. I think I'll use your branch. TY.

PabloLION:
Hey @MetaB0y . I git-cloned your forked repo and switched to branch `origin/runtime-assets-field` and seems you added all the fields for pyteal version 5,6,7 instead of only version 7.

Do you want to fix it or should I make a new PR?

PabloLION:
Hey @MetaB0y . I git-cloned your forked repo and switched to branch `origin/runtime-assets-field` and seems you added all the fields for pyteal version `5` and `6` instead of version `7`.

Do you want to fix it or should I make a new PR?

PabloLION:
Hey @MetaB0y . I git-cloned your forked repo and switched to branch `origin/runtime-assets-field` and it didn't work.
It seems that you added all the fields for pyteal version `5` and `6` instead of version `7`.
Maybe you've tested and haven't pushed some commits?

Do you want to fix it or should I make a new PR?

MetaB0y:
@PabloLION hi, I am not sure if it was changed recently but I don't think there was version 7 in algobuilder runtime at all. You can see it in the commit history.

And even if there is version 7, isn't it supposed to copy values from `6` and extend it when needed, like version `6` does now for `5` (`numberTxnFields[6] = cloneDeep(numberTxnFields[5]);`).

Regarding making a new PR - I don't mind, I just want this change to be merged, it doesn't matter how :)

PabloLION:
Yea sorry. I didn't follow up this thread. I realised that algo-builder supports up to pyteal 0.15 or 0.13, which supports up to teal v6.
So I stopped using functionalities from teal v7, and moved back to v6.

I tested your PR by adding building it and using it with my own test cases, after copying all v6 to v7 like you said, "copy values from 6 and extend it". And it worked for everything I needed. Although I didn't run the tests of this repo, I think this is a good PR.

PR #767, opened 2022-08-17
Add replace2 and replace3 opcode

thdailong:
Closes #x

## Proposed Changes

- Add replace2 and replace3 opcode
- Test all testcase from go-algorand

## TODO

- [x] update change log

### DIFF
--- a/packages/runtime/src/errors/errors-list.ts
+++ b/packages/runtime/src/errors/errors-list.ts
@@ -417,6 +417,12 @@ maximun uint128`,
 		title: "Clear state programs cannot issue inner transactions.",
 		description: "Clear state programs cannot issue inner transactions",
 	},
+	BYTES_REPLACE_ERROR: {
+		number: 1061,
+		message: "Can not replace bytes due to invalid start or end replace index",
+		title: "replace got error",
+		description: "Can not replace bytes due to invalid start or end replace index",
+	},
 };
 
 const runtimeGeneralErrors = {
--- a/packages/runtime/src/errors/errors-list.ts
+++ b/packages/runtime/src/errors/errors-list.ts
@@ -419,7 +419,7 @@ maximun uint128`,
 	},
 	BYTES_REPLACE_ERROR: {
 		number: 1061,
-		message: "Can not replace bytes due to invalid start or end replace index",
+		message: "Can not replace bytes due to length of replacing bytes(%lenReplace%) + index replace(%index%) > length of ogirinal bytes(%lenOriginal%) at line %line%",
 		title: "replace got error",
 		description: "Can not replace bytes due to invalid start or end replace index",
 	},
--- a/packages/runtime/src/errors/errors-list.ts
+++ b/packages/runtime/src/errors/errors-list.ts
@@ -419,9 +419,9 @@ maximun uint128`,
 	},
 	BYTES_REPLACE_ERROR: {
 		number: 1061,
-		message: "Can not replace bytes due to length of replacing bytes(%lenReplace%) + index replace(%index%) > length of ogirinal bytes(%lenOriginal%) at line %line%",
-		title: "replace got error",
-		description: "Can not replace bytes due to invalid start or end replace index",
+		message: "Can not replace bytes due to length of replacing bytes(%lenReplace%) + index replace(%index%) > length of original bytes(%lenOriginal%) at line %line%",
+		title: "replace opcode error",
+		description: "Can not replace bytes due to invalid start or end index",
 	},
 };
 
--- a/packages/runtime/src/interpreter/opcode-list.ts
+++ b/packages/runtime/src/interpreter/opcode-list.ts
@@ -4962,45 +4962,64 @@ export class Base64Decode extends Op {
 	}
 }
 
+export class Replace extends Op {
+	readonly line: number;
+	start: number;
+	original: Uint8Array;
+	replace: Uint8Array;
+
+	constructor(start: number, line: number) {
+		super();
+		this.line = line;
+		this.start = start;
+		this.original = new Uint8Array;
+		this.replace = new Uint8Array;
+	}
+
+	execute(stack: TEALStack): number {
+		if (this.start + this.replace.length > this.original.length) {
+			throw new RuntimeError(RUNTIME_ERRORS.TEAL.BYTES_REPLACE_ERROR, {
+				lenReplace: this.replace.length,
+				index: this.start,
+				lenOriginal: this.original.length,
+				line: this.line,
+			});
+		}
+		const result = new Uint8Array(this.original.length);
+		for (let i = 0, j = 0; i < this.original.length; ++i) {
+			if (i >= this.start && i < this.start + this.replace.length) {
+				result[i] = this.replace[j++];
+			}
+			else {
+				result[i] = this.original[i];
+			}
+		}
+		stack.push(result);
+		return this.computeCost();
+	}
+}
+
 /**
  * Opcode: replace2 s
  * Stack: ..., A: []byte, B: []byte → ..., []byte
  * Copy of A with the bytes starting at S replaced by the bytes of B. Fails if S+len(B) exceeds len(A)
  */
-export class replace2 extends Op {
-	readonly line: number;
-	readonly start: number;
+export class Replace2 extends Replace {
 	/**
 	 * Asserts 0 arguments are passed.
 	 * @param args Expected arguments: [start_index]
 	 * @param line line number in TEAL file
 	 */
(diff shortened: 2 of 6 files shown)


packages/runtime/src/errors/errors-list.ts
```diff
@@ -417,6 +417,12 @@ maximun uint128`,
 		title: "Clear state programs cannot issue inner transactions.",
 		description: "Clear state programs cannot issue inner transactions",
 	},
+	BYTES_REPLACE_ERROR: {
```
sczembor, line 420: we can add more details here and inform the user exactly what happened. You can pass arguments to the errors

packages/runtime/src/interpreter/opcode-list.ts
```diff
@@ -4961,3 +4961,86 @@ export class Base64Decode extends Op {
...
+		const original = this.assertBytes(stack.pop(), this.line);
+		if (start + BigInt(replace.length) > original.length) {
+			throw new RuntimeError(RUNTIME_ERRORS.TEAL.BYTES_REPLACE_ERROR, {});
+		}
+		const result = new Uint8Array(original.length);
+		for (let i = 0, j = 0; i < original.length; ++i) {
+			if (i >= start && i < start + BigInt(replace.length)) {
+				result[i] = replace[j++];
+			}
+			else {
```
sczembor, line 5037: we can move it one line up. `} else {`

sczembor (commented):
Left few comments. I will provide my full review tomorrow morning. Good job!

packages/runtime/src/interpreter/opcode-list.ts
```diff
@@ -4961,3 +4961,86 @@ export class Base64Decode extends Op {
...
+		return this.computeCost();
+	}
+}
+
+/**
+ * Opcode: replace3 
+ * Stack: ..., A: []byte, B: uint64, C: []byte → ..., []byte
+ * Copy of A with the bytes starting at B replaced by the bytes of C. Fails if B+len(C) exceeds len(A)
+ */
+ export class replace3 extends Op {
```
vuvoth, line 5012: Because we use very familiar logic(replease2, replease3) so I think you can `extends` `Repleace2` opcode and reuse logic

packages/runtime/src/interpreter/opcode-list.ts
```diff
@@ -4961,3 +4961,86 @@ export class Base64Decode extends Op {
 		return this.computeCost();
 	}
 }
+
+/**
+ * Opcode: replace2 s
+ * Stack: ..., A: []byte, B: []byte → ..., []byte
+ * Copy of A with the bytes starting at S replaced by the bytes of B. Fails if S+len(B) exceeds len(A)
+ */
+export class replace2 extends Op {
```
vuvoth, line 4970:
```suggestion
export class Replace2 extends Op {
```

packages/runtime/src/interpreter/opcode-list.ts
vuvoth, line 5012:
```suggestion
 export class Replace3 extends Op {
```

packages/runtime/test/src/interpreter/opcode-list.ts
```diff
@@ -7022,4 +7024,105 @@ describe("Teal Opcodes", function () {
...
 	});
+
+	describe("Tealv7: replace2 opcode", function () {
+		let stack: Stack<StackElem>;
+		this.beforeEach(() => {
+			stack = new Stack<StackElem>();
+		});
+
+		it("Should replace bytes correctly", () => {
+			const original = "0x11111111";
```
vuvoth, line 7035: You can use Uint8Array directly I think.

packages/runtime/test/src/interpreter/opcode-list.ts
```diff
@@ -7022,4 +7024,105 @@ describe("Teal Opcodes", function () {
...
+			const replace = "0x2222";
+			stack.push(new Uint8Array(Buffer.from(original.slice(2), "hex")));
+			stack.push(new Uint8Array(Buffer.from(replace.slice(2), "hex")));
+			let op = new replace2(["3"], 0);
+			expectRuntimeError(() => op.execute(stack), RUNTIME_ERRORS.TEAL.BYTES_REPLACE_ERROR);
+
+		});
+	});
+
+	describe("Tealv7: replace3 opcode", function () {
```
vuvoth, line 7078: You can group replace2 and repleace3 test cases to one describe, this is optional

vuvoth (commented):
Thanks for your PR! I left few comments.

packages/runtime/test/src/interpreter/opcode-list.ts
thdailong: yeah but I want to set up test like this to be look exactly from this github link
https://github.com/algorand/go-algorand/pull/4129/files#

packages/runtime/test/src/interpreter/opcode-list.ts
thdailong: I want to split this into two replace opcode to test.
It would make confuse if we group them together

packages/runtime/src/interpreter/opcode-list.ts
robert-zaremba: :+1:  class names should be upper case.

packages/runtime/src/interpreter/opcode-list.ts
robert-zaremba: :+1: - during the planning we mentioned to either extend class, or use a common function to delegate the oprations there. I think extending class will be easier.

packages/runtime/test/src/interpreter/opcode-list.ts
```diff
@@ -7022,4 +7024,105 @@ describe("Teal Opcodes", function () {
...
+			const original = "0x11111111";
+			const replace = "0x2222";
+			let hexStr = "0x22221111";
+			let expectedBytes = new Uint8Array(Buffer.from(hexStr.slice(2), "hex"));
+
+			stack.push(new Uint8Array(Buffer.from(original.slice(2), "hex")));
+			stack.push(new Uint8Array(Buffer.from(replace.slice(2), "hex")));
+			let op = new replace2(["0"], 0);
+			op.execute(stack);
+			assert.deepEqual(stack.pop(), expectedBytes);
```
robert-zaremba, line 7044: you should check the whole stack, not only the top element.

robert-zaremba (commented):
1. tests should be updated - there is lot of repeated code, create a helper function (especially for converting string -> bytes array).
2. Create a helper function for setting up the test

packages/runtime/test/src/interpreter/opcode-list.ts
thdailong: shouldn't we check if the opcode replace2 and replace3 run corrrectly right?
so why we should check the whole stack?

robert-zaremba: because this will  be the proper test - you want to make sure that only 2 top elements changed, and rest is the same. Now you only test that the top element changed.

thdailong: oh I see.
let me add a bytes to the stack and check if that bytes remain unchanged!

packages/runtime/src/errors/errors-list.ts
```diff
@@ -417,6 +417,12 @@ maximun uint128`,
...
+		number: 1061,
+		message: "Can not replace bytes due to length of replacing bytes(%lenReplace%) + index replace(%index%) > length of ogirinal bytes(%lenOriginal%) at line %line%",
```
sczembor, line 422:
```suggestion
		message: "Can not replace bytes due to length of replacing bytes(%lenReplace%) + index replace(%index%) > length of original bytes(%lenOriginal%) at line %line%",
```

packages/runtime/src/errors/errors-list.ts
```diff
@@ -417,6 +417,12 @@ maximun uint128`,
...
+		title: "replace got error",
```
sczembor, line 423:
```suggestion
		title: "Replace opcode error",
```

packages/runtime/src/errors/errors-list.ts
```diff
@@ -417,6 +417,12 @@ maximun uint128`,
...
+		description: "Can not replace bytes due to invalid start or end replace index",
```
sczembor, line 424:
```suggestion
		description: "Can not replace bytes due to invalid start or end index",
```

sczembor (approved):
Pre-approval. Left few comments. When you correct them feel free to merge

PR #770, opened 2022-08-19
update myalgo packages

thdailong:
Closes #x

## Proposed Changes

- Update myalgo wallet for support signing multisig transaction

## TODO
- [x] update myalgo packages

### DIFF
--- a/packages/web/src/lib/myalgowallet-mode.ts
+++ b/packages/web/src/lib/myalgowallet-mode.ts
@@ -106,11 +106,12 @@ export class MyAlgoWalletSession {
 	 * @returns array of raw signed txns | null. null representes that the txn in array is NOT signed
 	 * by wallet user (i.e signable by someone else).
 	 */
-	async signTransactionGroup(txns: TransactionInGroup[]): Promise<SignedTx[]> {
+	async signTransactionGroup(txns: TransactionInGroup[], 
+		signOptions?: SignTransactionOptions): Promise<SignedTx[]> {
 		const txnsGroup = txns.map((v) => v.txn);
 		const groupID = algosdk.computeGroupID(txnsGroup);
 		for (let i = 0; i < txns.length; i++) txnsGroup[i].group = groupID;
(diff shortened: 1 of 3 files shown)


packages/web/src/lib/myalgowallet-mode.ts
```diff
@@ -29,17 +30,21 @@ interface MyAlgoConnect {
...
 	/**
 	 * @async
 	 * @description Sign an Algorand Transaction.
 	 * @param transaction Expect a valid Algorand transaction array.
+	 * @param signOptions Sign transactions options object.
 	 * @returns Returns signed an array of signed transactions.
 	 */
-	signTransaction(transaction: (AlgorandTxn | EncodedTransaction)[]): Promise<SignedTx[]>;
+	signTransaction(transaction: (AlgorandTxn | EncodedTransaction)[], 
+		signOptions?: SignTransactionOptions): Promise<SignedTx[]>;
```
robert-zaremba, line 47: why do we have two times the same function? (line 36 and line 47)?

yarn.lock
```diff
@@ -99,7 +99,7 @@ __metadata:
   resolution: "@algo-builder/web@workspace:packages/web"
   dependencies:
     "@json-rpc-tools/utils": ^1.7.6
-    "@randlabs/myalgo-connect": ^1.1.2
+    "@randlabs/myalgo-connect": ^1.1.3
```
robert-zaremba, line 102:
```suggestion
    "@randlabs/myalgo-connect": ^1.2.0
```

robert-zaremba (approved):
pre-approving. Please upgrade myalgo-connect to v1.2.0 and see if we can remove duplicated function.

packages/web/src/lib/myalgowallet-mode.ts
thdailong: Line 36 is the `signTransaction` for signing a transaction but in Line 47,  `signTransaction` for signing an array of transaction

PR #776, opened 2022-08-31
Add opcode ed25519verify_bare

thdailong:
Closes #x

## Proposed Changes

- Add opcode `ed25519verify_bare`
- opcode `ed25519verify_bare` use the same test as opcode `ed25519verrify`

## TODO

- [x] update change log

### DIFF
--- a/packages/runtime/test/src/interpreter/opcode-list.ts
+++ b/packages/runtime/test/src/interpreter/opcode-list.ts
@@ -78,6 +78,7 @@ import {
 	EcdsaPkRecover,
 	EcdsaVerify,
 	Ed25519verify,
+	Ed25519verify_bare,
 	EqualTo,
 	Err,
 	Exp,
@@ -136,8 +137,8 @@ import {
 	Select,
 	SetBit,
 	SetByte,
-	Sha256,
 	Sha3_256,
+	Sha256,
 	Sha512_256,
 	Shl,
 	Shr,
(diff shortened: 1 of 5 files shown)


packages/runtime/test/src/interpreter/opcode-list.ts
```diff
@@ -7184,4 +7185,72 @@ describe("Teal Opcodes", function () {
...
+
+			stack.push(toSign); // data
+			stack.push(signed); // signature
+			stack.push(decodeAddress(account.addr).publicKey); // pk
+
+			const op = new Ed25519verify_bare([], 1);
+			assert.equal(1900, op.execute(stack));
+		});
+	});
+
```
sczembor, line 7255:
```suggestion
```

packages/runtime/test/src/interpreter/opcode-list.ts
```diff
@@ -7184,4 +7185,72 @@ describe("Teal Opcodes", function () {
...
+				stack,
+				[],
+				new Ed25519verify_bare([], 1),
+				RUNTIME_ERRORS.TEAL.ASSERT_STACK_LENGTH
+			)
+		);
+
+		it("Should return correct cost", () => {
+			const account = generateAccount();
+			const toSign = new Uint8Array(Buffer.from([1, 9, 25, 49]));
```
sczembor, line 7244: maybe we can call it `data`, `signature` and get rid of those comments below?

packages/runtime/test/src/interpreter/opcode-list.ts
```diff
@@ -7184,4 +7185,72 @@ describe("Teal Opcodes", function () {
...
+			stack.push(decodeAddress(account.addr).publicKey); // pk
+
+			const op = new Ed25519verify_bare([], 1);
+			op.execute(stack);
+			const top = stack.pop();
+			assert.equal(top, 0n);
+		});
+
+		it(
+			"should throw invalid type error Ed25519verify",
```
sczembor, line 7223:
```suggestion
			"should throw an invalid type error",
```

sczembor: There is no need for repetition of the `Ed25519verify` if we already have the `describe` section

sczembor (commented):
Thanks for the PR. Could you please make sure that the tests scenarios that we are using are exactly the same as in go-algorand implementation. This is the best way to ensure that our implementation is correct.

packages/runtime/test/src/interpreter/opcode-list.ts
```diff
@@ -7184,4 +7185,83 @@ describe("Teal Opcodes", function () {
...
+			"should throw an error if stack is below min length",
+			execExpectError(
+				stack,
+				[],
+				new Ed25519verify_bare([], 1),
+				RUNTIME_ERRORS.TEAL.ASSERT_STACK_LENGTH
+			)
+		);
+
+		it("Should return correct cost", () => {
```
sczembor, line 7254: Could we follow one convention in all the tests? Lets use a capital letter for all of them eg. `Should1`

sczembor (approved):

PR #778, opened 2022-09-06
Find transaction in transaction group by id

MetaB0y:
Without this change `txn GroupIndex` can inappropriately return `-1` despite all fields of `tx` being the same as all fields of `gtxns[x]`. Apparently, the reason is that either `tx` or `gtxns` can be overwritten at some point by the same content but stored in a different memory location (`indexOf` uses `===` which compares objects *by reference*).

I had this problem when providing liquidity twice in a row in [Pact smart contract](https://github.com/pactfi/algorand-testbed/blob/master/constant_product.teal)

This change looks pretty safe, but I don't know if the root cause (change of memory locations) should be found and eliminated in addition or instead of this commit.

### DIFF
--- a/packages/runtime/src/lib/txn.ts
+++ b/packages/runtime/src/lib/txn.ts
@@ -132,7 +132,8 @@ export function txnSpecByField(
 			return parsing.stringToBytes(tx.txID);
 		}
 		case "GroupIndex": {
-			result = gtxns.indexOf(tx);
+      const gtxnIDs = gtxns.map((gtx) => gtx.txID);
+      result = gtxnIDs.indexOf(tx.txID);
 			break;
 		}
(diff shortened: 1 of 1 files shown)


MetaB0y:
Oh, actually it doesn't work for contract-to-contract calls (or inner txs in general) because they don't have ids. Not sure if they should

robert-zaremba:
We should definitely look based on tx id. However yes - we probably need more inspection into that. Thanks for the update!

robert-zaremba:
it seams this breaks our tests.

MetaB0y:
> I've updated the PR by using `gtxns.find` instead of `map + indexOf`

But `find` returns an element, not `index`. So it doesn't work after this change.

PR #781, opened 2022-09-08
Add opcode `json ref`

thdailong:
Closes #x

## Proposed Changes

- Add opcode `json ref`

## TODO

- [x] update change log

### DIFF
--- a/packages/runtime/src/interpreter/opcode-list.ts
+++ b/packages/runtime/src/interpreter/opcode-list.ts
@@ -1,5 +1,6 @@
 /* eslint sonarjs/no-identical-functions: 0 */
 import { parsing, tx as webTx, types } from "@algo-builder/web";
+import { stringToBytes } from "@algo-builder/web/build/lib/parsing";
 import algosdk, {
 	decodeAddress,
 	decodeUint64,
@@ -16,6 +17,7 @@ import chalk from "chalk";
 import { ec as EC } from "elliptic";
 import { Hasher, Message, sha256 } from "js-sha256";
 import { sha512_256 } from "js-sha512";
+import JSONbig from "json-bigint";
 import cloneDeep from "lodash.clonedeep";
 import { Keccak, SHA3 } from "sha3";
 
@@ -30,6 +32,7 @@ import {
 	AssetParamMap,
 	GlobalFields,
 	ITxArrFields,
+	json_refTypes,
 	MathOp,
 	MAX_APP_PROGRAM_COST,
 	MAX_CONCAT_SIZE,
@@ -49,6 +52,7 @@ import { bigintSqrt } from "../lib/math";
 import {
 	assertBase64,
 	assertBase64Url,
+	assertJSON,
 	assertLen,
 	assertNumber,
 	assertOnlyDigits,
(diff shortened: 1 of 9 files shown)


packages/runtime/src/interpreter/opcode-list.ts
```diff
@@ -5087,3 +5091,112 @@ export class Replace3 extends Replace {
...
+	computeCost(): number {
+		return 25 + 2 * Math.ceil(this.length / 7); // cost = 25 + ceil(bytes / 7)
+	}
+	
+	execute(stack: TEALStack): number {
+		this.assertMinStackLen(stack, 2, this.line);
+		const key = this.assertBytes(stack.pop(), this.line);
+		const object = this.assertBytes(stack.pop(), this.line);
+		this.length = object.length;
+		const enc = new TextDecoder("utf-8");
```
sczembor, line 5146: maybe lets name it 'decoder' or 'utf8decoder'

packages/runtime/src/interpreter/opcode-list.ts
```diff
@@ -5087,3 +5091,112 @@ export class Replace3 extends Replace {
...
+		const decoded = enc.decode(object);
```
sczembor, line 5147: lets follow the same convention both for object and the key. Maybe 'decodedObj' and 'decodedKey'?

packages/runtime/test/src/interpreter/opcode-list.ts
```diff
@@ -7283,4 +7284,51 @@ describe("Teal Opcodes", function () {
 			assert.equal(1900, op.execute(stack));
 		});
 	});
+
+	describe("json_ref", function () {
+		const stack = new Stack<StackElem>();
+		const jsonByte = "{\"key0\": 0,\"key1\": \"algo\",\"key2\":{\"key3\": \"teal\", \"key4\": {\"key40\": 10}}, \"key5\": 18446744073709551615 }";
+
+		it("should return correct JSONUint64", function () {
```
sczembor, line 7292:
```suggestion
		it("Should return correct JSONUint64", function () {
```
and lets change it in all the cases

packages/runtime/test/src/interpreter/opcode-list.ts
```diff
@@ -7283,4 +7284,51 @@ describe("Teal Opcodes", function () {
...
+			stack.push(parsing.stringToBytes("{\"maxUint64\": 18446744073709551615}"));
+			stack.push(parsing.stringToBytes("maxUint64"));
+			const op = new Json_ref(["JSONUint64"], 1);
+			op.execute(stack);
+
+			const top = stack.pop();
+			assert.deepEqual(18446744073709551615n, top);
+		});
+
+		it("should throw when get wrong type JSON(expect byte but got uint64)", function () {
```
sczembor, line 7302:
```suggestion
		it("should throw when get wrong JSON type(expect byte but got uint64)", function () {
```

packages/runtime/src/lib/parsing.ts
```diff
@@ -282,4 +283,19 @@ export function bigEndianBytesToBigInt(bytes: Uint8Array | Buffer): bigint {
  */
 export function strHexToBytes(str: string): Uint8Array {
 	return new Uint8Array(Buffer.from(str.slice(2), "hex"));
+}
+
+/**
+ * assert if string given is a valid JSON object
```
sczembor, line 289:
```suggestion
 * assert if given string is a valid JSON object
```

packages/runtime/src/interpreter/opcode-list.ts
sczembor, line 5147: `decodedObj`

sczembor (commented):
Looks good. Let m know if we cover all the important test cases from go-algorand

packages/runtime/test/src/interpreter/opcode-list.ts
```diff
@@ -7283,4 +7284,51 @@ describe("Teal Opcodes", function () {
 			assert.equal(1900, op.execute(stack));
 		});
 	});
+
+	describe("json_ref", function () {
```
sczembor, line 7288:
```suggestion
	describe("json_ref", ()=> {
```

sczembor: maybe lets change this in all the cases. Its shorter

packages/runtime/src/lib/parsing.ts
```diff
@@ -282,4 +283,19 @@ export function bigEndianBytesToBigInt(bytes: Uint8Array | Buffer): bigint {
...
+ * @param jsonString
+ */
+ export function assertJSON(jsonString: string, line: number): void {
+	const strictBigJSON = JSONbig({ strict: true });
+	try {
+		strictBigJSON.parse(jsonString);
+	} catch (e) {
```
sczembor, line 296: do we catch all the error here?

packages/runtime/src/interpreter/opcode-list.ts
```diff
@@ -5087,3 +5091,112 @@ export class Replace3 extends Replace {
...
+					throw new RuntimeError(RUNTIME_ERRORS.TEAL.INVALID_TYPE, {
+						expected: 'object',
+						actual: typeof jsonObject[keyString],
+						line: this.line,
+					});
+				} 
+				break;
+			}
+		}
+
```
sczembor, line 5199:
```suggestion
```

packages/runtime/src/lib/parsing.ts
thdailong: Yes we catch all the errors from JSON.parse

sczembor: Should we handle them? To give a better feedback to the user when it throws an error

thdailong: Well this one function `assertJSON` just to make sure the bytes we parse is correct JSON format. Incase the string was wrong, we throw error invalid json parsing

packages/runtime/test/src/interpreter/opcode-list.ts
```diff
@@ -7283,4 +7284,51 @@ describe("Teal Opcodes", function () {
...
+			op1.execute(stack);
+			stack.push(parsing.stringToBytes("key4"));
+			const op2 = new Json_ref(["JSONObject"], 1);
+			op2.execute(stack);
+
+			const top = stack.pop();
+			const expected = parsing.stringToBytes("{\"key40\":10}");
+			assert.deepEqual(top, expected);
+		});
+	});
```
robert-zaremba, line 7333: please make sure you added few tests from go-algorand.

sczembor (approved):
LGTM

PR #785, opened 2022-09-13
Feature: MultiSig support in executeTx of webmode

rmeena840:
+ Able to get msig txn response from AlgoSigner.
+ Msig support added.
+ ExecParam for msig also working

Respone after signing from AlgoSigner:
Screenshot 2022-09-15 at 12 54 18 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,7 +39,7 @@ from `algosdk` and sends it to the network.
 
 #### @algo-builder/web
 - Added `appendSignMultisigTransaction` function to `WebMode` for appending signature to multisig transaction in the algosigner.
-- Added `MultiSignature` support in `executeTx` method.
+- Added `MultiSignature` support in `executeTx` method for `AlgoSigner`.
 
 ### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@ from `algosdk` and sends it to the network.
 
 #### @algo-builder/web
 - Added `appendSignMultisigTransaction` function to `WebMode` for appending signature to multisig transaction in the algosigner.
+- Added `MultiSignature` support in `executeTx` method.
 
 ### Bug Fixes
 
--- a/packages/web/src/lib/web-mode.ts
+++ b/packages/web/src/lib/web-mode.ts
@@ -9,7 +9,7 @@ import {
 	SignType,
 	TransactionAndSign,
 	TxParams,
-	MultiSig
+	SignWithMultisig
 } from "../types";
 import { WAIT_ROUNDS } from "./constants";
 import { log } from "./logger";
@@ -247,7 +247,7 @@ export class WebMode {
 				case SignType.LogicSignature:
 					return { txn: txn, signers: [] } // logic signature
 				case SignType.MultiSignature: {
-					const msig: MultiSig = execParams[txnId] as MultiSig;
+					const msig: SignWithMultisig = execParams[txnId] as SignWithMultisig;
 					return { txn: txn, msig: msig.mparams }; // multi singature
 				}
 				default:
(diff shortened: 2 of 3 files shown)


robert-zaremba:
@thdailong , @sczembor could you handle the review please?

packages/web/src/types.ts
```diff
@@ -158,9 +159,18 @@ interface SignWithLsig {
...
+	sign: SignType.MultiSignature;
+	mparams: WalletMultisigMetadata;
+	fromAccount?: AccountSDK;
+	fromAccountAddr: AccountAddress;
+}
+
 export type Lsig = SignWithLsig;
 
-export type Sign = SignWithSk | SignWithLsig;
+export type MultiSig = SignWithMultisig;
```
robert-zaremba, line 171: let's don't create a type alias. We can export `SignWithMultisig` directly

packages/web/src/types.ts
```diff
@@ -158,9 +159,18 @@ interface SignWithLsig {
 	args?: Uint8Array[];
 }
 
+interface SignWithMultisig {
+	sign: SignType.MultiSignature;
+	mparams: WalletMultisigMetadata;
+	fromAccount?: AccountSDK;
```
sczembor, line 165: I think the `fromAccount` should be a list of `AccountSDK` since we will be signing the transaction with multiple accounts

packages/web/src/lib/web-mode.ts
```diff
@@ -242,19 +243,26 @@ export class WebMode {
...
 		const toBeSignedTxns = base64Txs.map((txn: string, txnId: number) => {
-			return execParams[txnId].sign === SignType.LogicSignature
-				? { txn: txn, signers: [] } // logic signature
-				: { txn: txn, authAddr: execParams[txnId].fromAccount?.addr }; // set signer
+			switch (execParams[txnId].sign) {
+				case SignType.LogicSignature:
+					return { txn: txn, signers: [] } // logic signature
+				case SignType.MultiSignature: {
+					const msig: MultiSig = execParams[txnId] as MultiSig;
+					return { txn: txn, msig: msig.mparams }; // multi singature
```
sczembor, line 251: I think we also need to pass signers. Also why dont we use `AppendSigToMultiSig` method?

packages/web/src/types.ts
sczembor: And I think it should be named `signers`

rmeena840: No, we are using multi sig param for AccountSDK. One of the param of mparams contains the address array. So , we dont' need it here.

packages/web/src/lib/web-mode.ts
rmeena840: I have got the above structure from here: https://developer.algorand.org/docs/get-details/transactions/signatures/?from_query=multi%20sig#multisignatures.

I used the same structure and I am getting the expected response after signing from AlgoSigner.

rmeena840: Please check the PR description image. It contains the response after signing from AlgoSigner. This is expected response.

packages/web/src/types.ts
rmeena840: Done

CHANGELOG.md
```diff
@@ -39,6 +39,7 @@ from `algosdk` and sends it to the network.
 
 #### @algo-builder/web
 - Added `appendSignMultisigTransaction` function to `WebMode` for appending signature to multisig transaction in the algosigner.
+- Added `MultiSignature` support in `executeTx` method.
```
sczembor, line 42: Lets add a bit more details here. Its only for the algosinger right?

sczembor (approved):
utACK. Before merging please add more details in the change log. Thanks for the PR

CHANGELOG.md
rmeena840: yes. I will add that.

PR #792, opened 2022-09-26
Fix htlc example

Megha-Dev-19:
It used to fail because of failed txn, commented it, added error throw, doesn't need update in changelog

### DIFF
--- a/examples/htlc-pyteal-ts/scripts/withdraw/htlc-withdraw.ts
+++ b/examples/htlc-pyteal-ts/scripts/withdraw/htlc-withdraw.ts
@@ -35,14 +35,21 @@ async function run(
 		payFlags: { totalFee: 1000 },
 	};
 	// Transaction Fails : as wrong secret value is used 
-	// await withdrawExecuteTx(deployer, txnParams); (uncomment to test)
+	try {
+		await withdrawExecuteTx(deployer, txnParams);
+	} catch (error) {
+		console.log(error)
+	}
+
 
 	// Transaction Passes: as right secret value is used
 	txnParams.args = [algob.convert.stringToBytes(secret)];
 
-	await withdrawExecuteTx(deployer, txnParams).catch((error) => {
+	try {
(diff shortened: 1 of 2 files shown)


sczembor (commented):
Can we now run the examples without issue? What was the cause of the problem?

examples/htlc-pyteal-ts/scripts/withdraw/htlc-withdraw.ts
```diff
@@ -34,12 +34,16 @@ async function run(
 		args: [algob.convert.stringToBytes(wrongSecret)],
 		payFlags: { totalFee: 1000 },
 	};
-	// Transaction Fails : as wrong secret value is used
-	await WithdrawExecuteTx(deployer, txnParams);
+	// Transaction Fails : as wrong secret value is used 
+	// await withdrawExecuteTx(deployer, txnParams); (uncomment to test)
```
sczembor, line 38: lets leave both cases and just to try catch and handle them properly

Megha-Dev-19: but it will throw error, and therefore the second txn will not run, so we need to comment it

Megha-Dev-19:
> 

We need to comment the failing txn, it was causing the error, after commenting it we need to build it else the same code runs as it is written in typescript

examples/htlc-pyteal-ts/scripts/withdraw/htlc-withdraw.ts
sczembor: if we use try catch it will

Megha-Dev-19: I did with try catch but once we use throw error it stops the process, i guess instead of throwing the error in incorrect tx we can log it but in correct one we can throw it

Megha-Dev-19:
> Pre-approving. Please update the changelog before merging

But there isn't any change which needs to be added in changeLog

Issue #662, opened 2022-04-25
Add support `loadLogic` for *.teal program and SCParams


huybuidac:
## Is your feature request related to a problem? Please describe.

I'm writing a *.teal lsig program with SCParams. But algo-builder does not support to `loadLogic` for *.teal & SCParams

https://github.com/scale-it/algo-builder/blob/master/packages/runtime/src/runtime.ts#L678
- @param scTmplParams: Smart contract template parameters (used only when compiling PyTEAL to TEAL)

## Describe the solution you'd like

- Can you support this?

## Additional context

-

vuvth:
Thank you for your proposal! I think we shouldn't not support ScParams for teal because it make teal program unportable.

huybuidac:
I found a workaround, but it is better if there is official support.

```js
    const pyOp = new PyCompileOp()

    let lsigProg = getProgram('lsigProg.teal')
    const [replaceParams, param] = pyOp.parseScTmplParam({
      TMPL_a: a,
      TMPL_b b,
      TMPL_c: c,
    }, true);
    lsigProg = pyOp.replaceTempValues(lsigProg, replaceParams)

    const lsigAcc = runtime.createLsigAccount(lsigProg, [])
```

rmeena840:
@huybuidac This has been taken care of and will be available in the next release. Thanks

PR #795, opened 2022-09-28
Feature: Add support loadLogic for *.teal program and SCParam

rmeena840:
Currently, only teal in python can be compiled with smart contract params but was not possible to compile teal with smart contract params. I have now added support for this to compile teal file along with smart contract params.

This resolves: https://github.com/scale-it/algo-builder/issues/662

### DIFF
--- a/packages/runtime/test/fixtures/escrow-account/assets/crowdFundApproval.teal
+++ b/packages/runtime/test/fixtures/escrow-account/assets/crowdFundApproval.teal
@@ -1,0 +1,470 @@
+#pragma version 4
+// Fund Start Date unixtimestamp
+// Fund End Date unixtimestamp
+// Fund Goal
+// Fund Amount - total
+// Escrow Address
+// Creator Address
+
+// check if the app is being created
+// if so save creator
+int 0
+txn ApplicationID
+==
+
+// if not creation skip this section
+bz not_creation
+
+// save the creator address to the global state
+byte "Creator"
+txn Sender
+app_global_put
+
+// verify that 5 arguments are passed
+txn NumAppArgs
+int 5
+==
+bz failed
+
+//store the start date
+byte "StartDate"
+txna ApplicationArgs 0
+btoi
+app_global_put
+
+// store the end date
+byte "EndDate"
+txna ApplicationArgs 1
+btoi
+app_global_put
+
+// store fund goal
+byte "Goal"
+txna ApplicationArgs 2
+btoi
+app_global_put
+
+// store the fund receiver
+byte "Receiver"
+txna ApplicationArgs 3
+app_global_put
+
+// set the total raised to 0 and store it
+byte "Total"
+int 0
+app_global_put
+
+// store the fund close date
+byte "FundCloseDate"
+txna ApplicationArgs 4
+btoi
+app_global_put
+
+// return a success
+int 1
+return
+
+not_creation:
+// check if this is deletion transaction
+int DeleteApplication
+txn OnCompletion
+==
+bz not_deletion
+
+// NOTE: To delete the app
+// The creator must empty the escrow
+// to the fund receiver if the escrow has funds
+
+// only the creator
+// can delete the app
+byte "Creator"
+app_global_get
+txn Sender
+==
+
+// check that we are past fund close date
+global LatestTimestamp
+byte "FundCloseDate"
+app_global_get
+>=
+&&
+bz failed
+
+// if escrow balance is zero
+// let the app be deleted
+// escrow account must be passed
+// into the call as an argument
+int 0
+int 1
+balance
+==
+
+// if the balance is 0 allow the delete
+bnz finished
+
+// if the escrow is not empty then
+// there must be need two transactions
+// in a group
+global GroupSize
+int 2
+==
+
+// second tx is an payment
+gtxn 1 TypeEnum
+int pay
+==
+&&
+
+// the second payment transaction should be
+// a close out transaction to receiver
+byte "Receiver"
+app_global_get
+gtxn 1 CloseRemainderTo
+==
+&&
+
+// the amount of the payment transaction
+// should be 0
+gtxn 1 Amount
+int 0
+==
+&&
+
+// the sender of the payment transaction
+// should be the escrow account
+byte "Escrow"
+app_global_get
+gtxn 1 Sender
+==
+&&
+bz failed
+int 1
+return
+
+not_deletion:
+// check if this is update
+int UpdateApplication
+txn OnCompletion
+==
+
+bz not_update
+// verify that the creator is
+// making the call
+byte "Creator"
+app_global_get
+txn Sender
+==
+
+// the call should pass the escrow
+// address
+txn NumAppArgs
+int 1
+==
+&&
+bz failed
+
+// store the address in global state
+// this parameter should be addr:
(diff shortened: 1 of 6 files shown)


rmeena840:
@sczembor @thdailong Please review the PR.

packages/runtime/test/fixtures/escrow-account/assets/crowdFundApproval.teal
```diff
@@ -0,0 +1,470 @@
+#pragma version 4
```
thdailong, line 1: This teal file looks familiar, Where did you get these code?

rmeena840: It was in some other asset folder. I couldn't access it directly from there bcz of fixtures. So, I added this file over here.

rmeena840:
> Could you provide some description about the task you are doing. I am a bit confused about you are gonna do

Check description.

packages/runtime/test/src/logicsig.ts
```diff
@@ -1,27 +1,66 @@
 import { decodeAddress, multisigAddress, MultisigMetadata } from "algosdk";
 import { assert } from "chai";
-
+const { convert } = require("@algo-builder/algob");
```
thdailong, line 3: I think you can use `import` instead of const

packages/runtime/test/src/logicsig.ts
```diff
@@ -1,27 +1,66 @@
...
 import { AccountStore } from "../../src/account";
 import { RUNTIME_ERRORS } from "../../src/errors/errors-list";
 import { Runtime } from "../../src/runtime";
 import { useFixture } from "../helpers/integration";
 import { expectRuntimeError } from "../helpers/runtime-errors";
+const { types } = require("@algo-builder/web");
```
thdailong, line 9: same

packages/runtime/test/src/logicsig.ts
```diff
@@ -50,6 +89,35 @@ describe("Logic Signature", () => {
 		assert.equal(result, false);
 	});
 
+	it("should handle contract lsig (escrow account) verification correctly with empty smart contract parmas", () => {
```
thdailong, line 92:
```suggestion
	it ("Should handle contract lsig (escrow account) verification correctly with empty smart contract params", () => {
```

packages/runtime/test/src/logicsig.ts
```diff
@@ -50,6 +89,35 @@ describe("Logic Signature", () => {
...
+		const lsig = runtime.loadLogic(programName, {});
+
+		let result = lsig.lsig.verify(decodeAddress(lsig.address()).publicKey);
+		assert.equal(result, true);
+
+		result = lsig.lsig.verify(johnPk);
+		assert.equal(result, false);
+	});
+
+	it("should handle contract lsig (crowd fund escrow account) verification correctly with non-empty smart contract parmas", () => {
```
thdailong, line 103:
```suggestion
	it ("Should handle contract lsig (crowd fund escrow account) verification correctly with non-empty smart contract params", () => {
```

packages/runtime/test/src/logicsig.ts
```diff
@@ -1,27 +1,66 @@
...
+	let appDefinition: any;
+	let applicationId: any
 
 	before(() => {
+		goal = 0.01e6;
+		crowdfundApprovalFileName = "crowdFundApproval.teal";
+		crowdFundClearFileName = "crowdFundClear.teal";
 		john = new AccountStore(10);
-		bob = new AccountStore(100);
+		bob = new AccountStore(10e6);
```
thdailong, line 37: Why do you set bob's account balance from 100 to 10e6?

packages/runtime/test/src/logicsig.ts
```diff
@@ -1,27 +1,66 @@
...
 	let bob: AccountStore;
 	let runtime: Runtime;
 	let johnPk: Uint8Array;
+	let goal: number;
+	let crowdfundApprovalFileName: string;
+	let crowdFundClearFileName: string;
+	let now: Date;
+	let beginDate: Date;
+	let endDate: Date;
+	let fundCloseDate: Date;
```
thdailong, line 27: you should declare variables like `goal`, `now`, `beginDate`, `endDate`, `fundCloseDate` in a smaller describe

thdailong (commented):
I recommend you to put those 2 `it` into a describe under `describe ("Logic Signature"`, that describe can be like `Load logic for teal program`

packages/runtime/test/src/logicsig.ts
rmeena840: more balance is needed

packages/runtime/test/src/logicsig.ts
rmeena840: why so??? I can see in other tests we have used the same code format case.

rmeena840:
> I recommend you to put those 2 `it` into a describe under `describe ("Logic Signature"`, that describe can be like `Load logic for teal program`

Done @thdailong

packages/runtime/test/src/logicsig.ts
thdailong: Well those variable are not use globally but only use within `crowdfundingApp`
You should put those here only if you use globally

thdailong (commented):
Sorry I let misunderstand my idea. I would like to have the `describe("Load logic for teal program")` in the `describe("Logic Signature")`
Cause your crow funding app use logic signature

packages/runtime/test/src/logicsig.ts
rmeena840: Done

packages/runtime/test/src/logicsig.ts
rmeena840: DOne

packages/runtime/test/src/logicsig.ts
rmeena840: Done

rmeena840: We need them global bcz it will be initialized before running each unit test. This is standard practice while writing unit test case.

rmeena840:
> Sorry I let misunderstand my idea. I would like to have the `describe("Load logic for teal program")` in the `describe("Logic Signature")` Cause your crow funding app use logic signature

Done @thdailong

packages/runtime/test/src/logicsig.ts
```diff
@@ -1,27 +1,66 @@
...
 import { useFixture } from "../helpers/integration";
 import { expectRuntimeError } from "../helpers/runtime-errors";
+import { types } from "@algo-builder/web";
 
 const programName = "escrow.teal";
 const multiSigProg = "sample-asc.teal";
+const crowdFundEscrow = "crowdFundEscrow.teal";
 
-describe("Logic Signature", () => {
+describe("Load logic for teal program", () => {
```
thdailong, line 15: I would prefer change back to `Logic Signature` since many test case here use logic signature

thdailong (approved):
Pre approving!
Now this one looks good to me but you also check my comment one more time

packages/runtime/test/src/logicsig.ts
rmeena840: done

PR #798, opened 2022-10-02
fix typo (or wrong implementation) in 'runtime/src/ctx.ts'

PabloLION:
I think this implementation doesn't return the same result as algod sandbox (not tested yet) If so, these related tests also needs a change.
https://github.com/scale-it/algo-builder/blob/2bcef8f611b349dfb8dc3542ed2f0a129a0c405c/packages/runtime/test/src/interpreter/opcode-list.ts#L2989-L3011

In both [pyteal doc](https://developer.algorand.org/docs/get-details/dapps/avm/teal/opcodes/?from_query=CallerApplicationID#global-f) and [algorand doc](https://developer.algorand.org/docs/get-details/dapps/avm/teal/opcodes/?from_query=CallerApplicationID#global-f) says that `global CallerApplicationID` returns 
> The application ID of the application that called this application. 0 if this application is at the top-level. Application mode only.

But the current implementation of `global CallerApplicationID` seems to return a same `uint64` as the `global CurrentApplicationID`

## TODO

- [ ] test it with sandbox private network


thdailong:
This implementation should be right since `global CallerApplicationID` would return uint64
While `global CurrentApplicationAddress` return bytes which is address.
Is there anything wrong?

PabloLION:
@thdailong Sorry that I made typo. I was trying to say `global CurrentApplicationID` instead of `global CurrentApplicationAddress`. I've edited the description.

PabloLION:
Then I'm making an edge case here. I'll try to offer a MRE next week and reopen this. I close it for now.

PabloLION:
I've finished the MRE here 

https://github.com/PabloLION/algo-builder-example/blob/f0ab2b3d50b7b1653f16438b67989f2fb0c13d80/test/caller-app-id.test.ts#L68-L69

run `yarn` then `yarn algob test` and you'll see that the `expect`s (should fail) passes 

---

During which, I noticed that 
1. the `yarn run algob init --typescript` is missing a parameter of path, like `yarn run algob init --typescript .`
2. the `ts-mocha` (required by `yarn algob test`) was not installed by `yarn run algob init --typescript.`

Should I make a new fix-typo PR on this?

thdailong:
Thanks for your suggestion. I am on the way find out why this error exist.

PabloLION:
@thdailong So the current version is at least incorrect👍. With the version of my PR, my MRE worked as intended (tho the test in this repo fails). 

I tried to add a commit with `resolutions` overriding the `@algo-builder/runtime`, to use the version of my PR. But I can only build it locally and then use a local path in the `resolutions`. I don't know how this should be done. For now I guess you'd have to try it yourself🥲.

PR #800, opened 2022-10-04
Added error checks in examples

Megha-Dev-19:


examples/ref-templates/assets/htlc.py
```diff
@@ -41,16 +42,16 @@ def htlc(ARG_RCV,
...
     # Next, check that the Amount of algos transferred is 0. This is for the same reason as
     # above: we only allow transactions that close out this account completely, which
     # having a non-zero-address CloseRemainderTo will handle for us.
     amount_check = Txn.amount() == Int(0)
 
-    # Always verify that the RekeyTo property of any transaction is set to the ZeroAddress
+    # Always verify that the RekeyTo property of any transaction is set to the receiver addr
     # unless the contract is specifically involved ina rekeying operation.
-    rekey_check = Txn.rekey_to() == Global.zero_address()
+    rekey_check = Txn.rekey_to() == Txn.receiver()
```
sczembor, line 54: why do we change this? I think it should stay global zero address

sczembor:
I don't think these errors has anything to do with the changes you've made

examples/ref-templates/assets/htlc.py
Megha-Dev-19: But as you said on discord 
*I think the receiver should be set to the same address as close reminder to. And in the smart contract we should check if receiver === the same address that was passed to the smart contract when deploying as close reminder to*

Megha-Dev-19:
@sczembor the htlc-pyteal example script is working on my local but it is failing on CI. could you please confirm if it is failing on your local as well.?
Thanks

examples/ref-templates/scripts/dynamic-fee.js
```diff
@@ -15,17 +15,25 @@ async function run(runtimeEnv, deployer) {
...
+	await deployer
+		.fundLsigByFile(
+			contractName,
+			{ funder: masterAccount, fundingMicroAlgo: 100000000 },
+			{ closeRemainderTo: masterAccount.addr },
+			scInitParam
+		)
+		.catch((error) => {
+			throw error;
+		});
```
ratik21, line 27: I don't follow. The error throwing functionality is already there in algob (and more verbose). Here we're just catching the thrown error, and re-throwing it. No?

examples/ref-templates/assets/htlc.py
Megha-Dev-19: corrected it

packages/algob/src/internal/deployer.ts
```diff
@@ -526,7 +526,12 @@ class DeployerBasicMode {
 				break;
 			}
 			case wtypes.SignType.LogicSignature: {
-				signer.lsig.lsig.args = signer.args ?? [];
+				if (signer?.lsig?.lsig?.args) {
+					signer.lsig.lsig.args = signer.args ?? [];
+				} else {
+					(signer.lsig as any).args = signer.args ?? []; // args property didn't exist in earlier version of API (for reference: see the multisig example)
```
sczembor, line 532: we should avoid as any whenever possible

packages/algob/src/lib/tx.ts
```diff
@@ -100,7 +100,12 @@ function signTransaction(txn: Transaction, signer: wtypes.Sign): Uint8Array {
 			return txn.signTxn(signer.fromAccount.sk);
 		}
 		case wtypes.SignType.LogicSignature: {
-			signer.lsig.lsig.args = signer.args ?? [];
+			if (signer?.lsig?.lsig?.args) {
+				signer.lsig.lsig.args = signer.args ?? [];
+			} else {
+				(signer.lsig as any).args = signer.args ?? []; // args property didn't exist in earlier version of API (for reference: see the multisig example)
```
sczembor, line 106: same here

examples/unique-nft-asa/scripts/0-deploy-app.js
```diff
@@ -8,27 +8,36 @@ async function run(runtimeEnv, deployer) {
...
-			metaType: types.MetaType.FILE,
-			approvalProgramFilename: "nft-app-approval.py",
-			clearProgramFilename: "nft-app-clear.py",
-			localInts: 1, // p
-			localBytes: 1, // creator
-		},
-		{}
-	);
+	const nftAppInfo = await deployer
+		.deployApp(
```
sczembor, line 12: formatting

examples/trampoline/scripts/deployApp.js
```diff
@@ -21,25 +22,28 @@ async function run(runtimeEnv, deployer) {
...
-			globalInts: 0,
-			globalBytes: 0,
-		},
-		{}
-	); //execute create App Transction
+	await deployer
+		.deployApp(
+			john,
+			{
+				appName: "proxy_trampoline",
```
sczembor, line 31: formatting

sczembor (commented):
there are some problems with formatting.

Megha-Dev-19:
> there are some problems with formatting.

yeah I checked the formatting, and it is right, the change is because we have another set of curly brackets (for try and catch) now which affected the whole function

packages/algob/src/internal/deployer.ts
Megha-Dev-19: Actually it is necessary since the new type of LogicSigAccount type consist of args, due to which it was giving me a type error which casting it to any

examples/asa/scripts/2-gold-asc.js
```diff
@@ -1,35 +1,41 @@
...
 
 	// save Smart Signature by name & fund the account
-	const ascInfoContract = await deployer.mkContractLsig(
-		"Gold_C_Lsig",
-		"2-gold-contract-asc.teal",
-		{}
-	);
+	const ascInfoContract = await deployer
+		.mkContractLsig("Gold_C_Lsig", "2-gold-contract-asc.teal", {})
+		.catch((error) => {
```
sczembor, line 18: why do we catch it and imidietly throw it again here?

examples/asa/scripts/transfer/gold-delegated-lsig.js
```diff
@@ -54,7 +62,11 @@ async function run(runtimeEnv, deployer) {
 
 	// Transaction FAIL - rejected by lsig because amount is not <= 100
 	txnParam.amountMicroAlgos = 580;
-	await tryExecuteTx(deployer, txnParam);
+	try {
```
sczembor, line 65: will the console.error appear in GitHub CI?

examples/bond/scripts/run/common/common.js
```diff
@@ -21,6 +21,16 @@ const asaDef = {
 
 const tokenMap = new Map();
 
+async function tryExecuteTx(deployer, txnParams) {
```
sczembor, line 24: we need a description for the function

examples/crowdfunding/scripts/common/common.js
```diff
@@ -0,0 +1,13 @@
+async function tryExecuteTx(deployer, txnParams) {
```
sczembor, line 1: why do we have another tryExecuteTx? Could we make the use of the other function? maybe we can move it to utils and reuse it everywhere?

examples/permissioned-token/scripts/4-setup-new-permissions-app.js
```diff
@@ -12,20 +12,24 @@ async function setupNewPermissionsApp(runtimeEnv, deployer) {
...
-			approvalProgramFilename: "permissions_new.teal", // new permissions contract
-			clearProgramFilename: "clear_state_program.py",
-			localInts: 0,
-			localBytes: 0,
-			globalInts: 0,
-			globalBytes: 0,
-		},
-		{}
-	);
+	const newPermissionsAppInfo = await deployer
```
sczembor, line 15: maybe we could create something similar for the deployApp as tryExecuteTx? I am not sure however since this will make it more disconnected from the way we should always use it

sczembor (commented):
It looks good, however I think we should rethink it. Maybe it would be better to throw the error inside of deployer? I am not sure about creating more wrappers for deployer since the examples are supposed to show real use cases that user can just adopt in their projects.

examples/asa/scripts/2-gold-asc.js
Megha-Dev-19: because else the tests won't fail in case there is an error, it might just log (depending upon what we pass in catch)

examples/asa/scripts/transfer/gold-delegated-lsig.js
Megha-Dev-19: I think so, it should

examples/bond/scripts/run/common/common.js
Megha-Dev-19: will add it

examples/crowdfunding/scripts/common/common.js
Megha-Dev-19: Yeah I thought the same, but we should keep the examples independent of each other since if a use copies an example folder to use in their own code, they should easily do it rather than also copying an outside folder (which they will release after the code throws an error that it is missing)

examples/bond/scripts/deploy.js
```diff
@@ -40,22 +46,26 @@ async function run(runtimeEnv, deployer) {
...
+				clearProgramFilename: "bond-dapp-clear.py",
+				localInts: 1,
+				localBytes: 1,
+				globalInts: 8,
+				globalBytes: 15,
+				appArgs: appArgs,
+			},
+			{},
+			placeholderParam
+		)
```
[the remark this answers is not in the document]
Megha-Dev-19: > It looks good, however I think we should rethink it. Maybe it would be better to throw the error inside of deployer? I am not sure about creating more wrappers for deployer since the examples are supposed to show real use cases that user can just adopt in their projects.

The better solution to his can be to directly throw the error in executeTx method, like I did with web pacakge, then in the examples we only need to catch it in places where the Transaction is SUPPOSED TO FAIL, and all other places if it failed it will throw an error

Megha-Dev-19:
> It looks good, however I think we should rethink it. Maybe it would be better to throw the error inside of deployer? I am not sure about creating more wrappers for deployer since the examples are supposed to show real use cases that user can just adopt in their projects.

The better solution to this can be to directly throw the error in executeTx method, like I did with web package, then in the examples we only need to catch it in places where the Transaction is SUPPOSED TO FAIL, and all other places if it failed it will throw an error

examples/permissioned-token/scripts/4-setup-new-permissions-app.js
Megha-Dev-19: I think you feel the issue more since we have multiple examples to handle, but for the user they might use these method once or twice, so it would make sense to have within the same file

examples/bond/scripts/deploy.js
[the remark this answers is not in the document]
Megha-Dev-19: can do that, but user usually deploys the app once or twice in real use case, so I thought it's better to keep it this way
What do you think?

examples/asa/scripts/2-gold-asc.js
sczembor: hm I wa thinking that if the error is thrown inside of `mkContractLsig` and we do not catch it it will behave exactly the same

examples/asa/scripts/transfer/gold-delegated-lsig.js
sczembor: I mean if it will make the CI fail

examples/asa/scripts/2-gold-asc.js
Megha-Dev-19: It should but we don't throw error in `mkContractLsig `

examples/asa/scripts/transfer/gold-delegated-lsig.js
Megha-Dev-19: yes it will, the logs will be in red color and few other things
but since we are using the throw error, it is almost same

examples/bond/scripts/run/common/common.js
```diff
@@ -21,6 +21,22 @@ const asaDef = {
 
 const tokenMap = new Map();
 
+/**
+ * Execute single transaction or group of transactions (atomic transaction)
+ * @param deployer deployer instance
+ * @param transactions transaction parameters,  atomic transaction parameters
+ * @returns TxnReceipt which includes confirmed txn response along with txID
+ */
+async function tryExecuteTx(deployer, txnParams) {
```
vuvoth, line 30: Just a bit wonder why we need tryExecuteTx here. because we deployer.executeTx also return error...

examples/bond/scripts/run/createBuyback.js
```diff
@@ -17,7 +17,11 @@ exports.createBuyback = async function (deployer, managerAcc, n) {
 		TMPL_APP_MANAGER: managerAcc.addr,
 		TMPL_BOND: bondToken,
 	};
-	const buybackLsig = await deployer.loadLogicByFile("buyback-lsig.py", scInitParam);
+	const buybackLsig = await deployer
+		.loadLogicByFile("buyback-lsig.py", scInitParam)
+		.catch((error) => {
```
vuvoth, line 22: I think this will auto throw error. we don't need catch like this I think. Please check again

vuvoth (approved):

examples/bond/scripts/run/common/common.js
Megha-Dev-19: it returns the error, it doesn't throw it, we need throw to stop the script

examples/bond/scripts/run/createBuyback.js
Megha-Dev-19: yes it will throw, I made the changes some places but didn't in some, will remove it

examples/asa/scripts/2-gold-asc.js
```diff
@@ -38,7 +44,9 @@ async function run(runtimeEnv, deployer) {
 	const lsig = await deployer.getLsig("Gold_C_Lsig");
 	const goldAsset = deployer.asa.get("gold");
 	const goldAssetID = goldAsset.assetIndex;
-	await deployer.optInLsigToASA(goldAssetID, lsig, { totalFee: 1000 });
+	await deployer.optInLsigToASA(goldAssetID, lsig, { totalFee: 1000 }).catch((error) => {
```
sczembor, line 47: Hm, it will not fail in the CI if the error is thrown in deployer maybe we should just do assert.notThrow? It looks a bit strange the construction here

examples/asa/scripts/3-contract-owned-asa.js
```diff
@@ -7,44 +7,54 @@
...
+		});
 
 	console.log(appInfo);
 
 	// Get Statless Account Address
-	await deployer.mkContractLsig("StateLessASALsig", "5-contract-asa-stateless.py", {
-		APP_ID: appInfo.appID,
-	});
+	await deployer
+		.mkContractLsig("StateLessASALsig", "5-contract-asa-stateless.py", {
```
sczembor, line 47:
```suggestion
		.mkContractLsig("StatelessASALsig", "5-contract-asa-stateless.py", {
```

sczembor (approved):
Approve. I left few comments. I am still not sure about the `.catch.throw` construction. Maybe we can achieve the same results with different approach. Lets discuss it tomorrow during the review

packages/algob/src/lib/tx.ts
Megha-Dev-19: Actually it is necessary since the new type of LogicSigAccount type consist of args, due to which it was giving me a type error which casting it to any

PR #803, opened 2022-10-10
Feature: MyAlgo Wallet Multisig

Megha-Dev-19:
Added support for multisig for MyAlgo Wallet

### DIFF
--- a/packages/web/src/lib/web-mode.ts
+++ b/packages/web/src/lib/web-mode.ts
@@ -261,7 +261,10 @@ export class WebMode {
 					return { txn: txn, msig: msig.mparams }; // multi singature
 				}
 				default:
-					return { txn: txn, authAddr: execParams[txnId].fromAccount?.addr || execParams[txnId].fromAccountAddr }; // set signer
+					return {
+						txn: txn,
+						authAddr: execParams[txnId].fromAccount?.addr || execParams[txnId].fromAccountAddr
+					}; // set signer
 			}
 		});
 		// checks if any sign txn exists else it throws error of empty signers array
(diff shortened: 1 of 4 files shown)


packages/web/src/lib/web-mode.ts
```diff
@@ -261,7 +261,10 @@ export class WebMode {
 					return { txn: txn, msig: msig.mparams }; // multi singature
 				}
 				default:
-					return { txn: txn, authAddr: execParams[txnId].fromAccount?.addr }; // set signer
+					return {
+						txn: txn,
+						authAddr: execParams[txnId].fromAccount?.addr || execParams[txnId].fromAccountAddr
+					}; // set signer
```
ratik21, line 267: Is it possible to add a test?

sczembor (commented):
Lets first merge the PR with the mock for myAlgo wallet and then add test to this one.

packages/web/src/lib/web-mode.ts
Megha-Dev-19: we already have a test for algosigner webmode, the common conventions is to use `fromAccount: {addr:"",sk:""}`, but sometimes user also use `fromAccountAddr`, thus added it

rmeena840 (approved):

packages/web/src/lib/web-mode.ts
Megha-Dev-19: Nah, I don't think so because the wallet actually wait for the cosigner signature, and I don't think it is possible with unit test, it can done from the wallet UX only

rmeena840:
@Megha-Dev-19 Update changelog.md

ratik21 (approved):

Megha-Dev-19:
> Lets first merge the PR with the mock for myAlgo wallet and then add test to this one.

As I said I don't think we can add the test for multisig using unit test because the wallet actually wait for the cosigner signature, and I don't think it is possible with unit test, it can done from the wallet UX only

PR #806, opened 2022-10-10
pipeline fix

sczembor:
Closes #x

## Proposed Changes

- update workflows to additionally install `poetry` package
- update `Pipefile.lock`

## TODO

- [ ] update change log


.github/workflows/01-main.yaml
```diff
@@ -15,7 +15,7 @@ on:
 
 jobs:
   tests:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
```
rmeena840, line 18: Was this causing the issue?

rmeena840 (approved):
Approved

.github/workflows/01-main.yaml
sczembor: LMK what do you think and lets merge it

Megha-Dev-19 (approved):
Thanks for the fixes

PR #807, opened 2022-10-10
MyAlgo Wallet test cases

Megha-Dev-19:
Added test cases for MyAlgoWallet webmode

### DIFF
--- a/packages/web/test/mocks/myalgowallet-mock.ts
+++ b/packages/web/test/mocks/myalgowallet-mock.ts
@@ -1,0 +1,91 @@
+import type {
+    Options,
+    Accounts,
+    Address,
+    AlgorandTxn,
+    Base64,
+    ConnectionSettings,
+    EncodedTransaction,
+    SignedTx,
+    SignTransactionOptions,
+} from "@randlabs/myalgo-connect";
+
+export default class MyAlgoConnectMock {
+
+    /**
+     * @param {Options} options Override default popup options.
+     */
+    constructor(options?: Options) {
+
+    };
+
+    /**
+     * @async
+     * @description Receives user's accounts from MyAlgo.
+     * @param {ConnectionSettings} [settings] Connection settings
+     * @returns Returns an array of Algorand addresses.
+     */
+    connect(settings?: ConnectionSettings): Promise<Accounts[]> {
+        return new Promise((resolve, reject) => {
+            return resolve([{
+                address: "",
+                name: ""
+            }]);
+        });
+    };
+
+    /**
+     * @async
+     * @description Sign an Algorand Transaction.
+     * @param transaction Expect a valid Algorand transaction
+     * @param signOptions Sign transactions options object.
+     * @returns Returns signed transaction
+     */
+    signTransaction(transaction: AlgorandTxn | EncodedTransaction, signOptions?: SignTransactionOptions): Promise<SignedTx>;
+
+    /**
+     * @async
+     * @description Sign an Algorand Transaction.
+     * @param transaction Expect a valid Algorand transaction array.
+     * @param signOptions Sign transactions options object.
+     * @returns Returns signed an array of signed transactions.
+     */
+    signTransaction(transaction: (AlgorandTxn | EncodedTransaction)[], signOptions?: SignTransactionOptions): Promise<SignedTx[]>;
+
+    signTransaction(transaction: AlgorandTxn | EncodedTransaction | (AlgorandTxn | EncodedTransaction)[], signOptions?: SignTransactionOptions): Promise<SignedTx | SignedTx[]> {
+        return new Promise((resolve, reject) => {
+            return resolve([{
+                blob: new Uint8Array(),
+                txID: ""
+            }]);
+        });
+    }
+
+
+    /**
+     * @async
+     * @description Sign a teal program
+     * @param logic Teal program
+     * @param address Signer Address
+     * @returns Returns signed teal
+     */
+    signLogicSig(logic: Uint8Array | Base64, address: Address): Promise<Uint8Array> {
+        return new Promise((resolve, reject) => {
+            return resolve(new Uint8Array());
+        });
+    };
+
+    /**
+     * @async
+     * @description Creates a signature the data that can later be verified by the contract through the ed25519verify opcode
+     * @param data Arbitrary data to sign
+     * @param contractAddress Contract address/TEAL program hash
+     * @param address Signer Address
+     * @returns Returns the data signature
+     */
+    tealSign(data: Uint8Array | Base64, contractAddress: Address, address: Address): Promise<Uint8Array> {
+        return new Promise((resolve, reject) => {
+            return resolve(new Uint8Array());
+        });
+    };
+}
--- a/packages/web/test/src/lib/myalgowallet-mode.ts
+++ b/packages/web/test/src/lib/myalgowallet-mode.ts
@@ -25,7 +25,7 @@ describe("Webmode - MyAlgo Wallet test cases ", () => {
 
 	});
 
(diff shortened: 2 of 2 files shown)


packages/web/test/mocks/myalgowallet-mock.ts
```diff
@@ -0,0 +1,91 @@
...
+
+    };
+
+    /**
+     * @async
+     * @description Receives user's accounts from MyAlgo.
+     * @param {ConnectionSettings} [settings] Connection settings
+     * @returns Returns an array of Algorand addresses.
+     */
+    connect(settings?: ConnectionSettings): Promise<Accounts[]> {
```
rmeena840, line 28: You are returning empty. Y???

packages/web/test/mocks/myalgowallet-mock.ts
```diff
@@ -0,0 +1,91 @@
...
+    SignedTx,
+    SignTransactionOptions,
+} from "@randlabs/myalgo-connect";
+
+export default class MyAlgoConnectMock {
+
+    /**
+     * @param {Options} options Override default popup options.
+     */
+    constructor(options?: Options) {
```
rmeena840, line 18: Why to even create it if you don't wanna use it?

packages/web/test/src/lib/myalgowallet-mode.ts
```diff
@@ -0,0 +1,60 @@
...
+
+	beforeEach(() => {
+		MyAlgoConnect
+		sender = algosdk.generateAccount();
+		receiver = algosdk.generateAccount();
+		connector = new MyAlgoWalletSession(walletURL, new MyAlgoConnectMock());
+
+	});
+
+	it("Should executeTx without throwing an error", async () => {
```
rmeena840, line 31: Don't use the arrow method. Use normal functions.
Reason: https://mochajs.org/#arrow-functions

packages/web/test/src/lib/myalgowallet-mode.ts
```diff
@@ -0,0 +1,60 @@
...
+		const txnParams: types.AlgoTransferParam = {
+			type: types.TransactionType.TransferAlgo,
+			sign: types.SignType.SecretKey,
+			fromAccount: sender,
+			toAccountAddr: receiver.addr,
+			amountMicroAlgos: 10000n,
+			payFlags: {},
+		};
+
+		assert.doesNotThrow(async () => {
```
rmeena840, line 41: U should also check for deducted and transferred balance.

packages/web/test/src/lib/myalgowallet-mode.ts
```diff
@@ -0,0 +1,60 @@
...
+	it("Should signTx without throwing an error", async () => {
+		const txnParams: types.AlgoTransferParam = {
+			type: types.TransactionType.TransferAlgo,
+			sign: types.SignType.SecretKey,
+			fromAccount: sender,
+			toAccountAddr: receiver.addr,
+			amountMicroAlgos: 10000n,
+			payFlags: {},
+		}
+		assert.doesNotThrow(async () => {
```
rmeena840, line 55: Same here

rmeena840 (changes requested):
Reviewed

packages/web/test/mocks/myalgowallet-mock.ts
Megha-Dev-19: we need to initialize it, not necessarily to return any values

packages/web/test/mocks/myalgowallet-mock.ts
Megha-Dev-19: we need to create a mock to be used as connector to connect to wallet, thus we need to create a dummy class for it

packages/web/test/src/lib/myalgowallet-mode.ts
Megha-Dev-19: but I m not using any context in any of the cases, so it shouldn't be a problem

packages/web/test/mocks/myalgowallet-mock.ts
ratik21: Then how about changing the function name to `init(..)`?

packages/web/test/mocks/myalgowallet-mock.ts
```diff
@@ -0,0 +1,91 @@
...
+    /**
+     * @async
+     * @description Sign an Algorand Transaction.
+     * @param transaction Expect a valid Algorand transaction array.
+     * @param signOptions Sign transactions options object.
+     * @returns Returns signed an array of signed transactions.
+     */
+    signTransaction(transaction: (AlgorandTxn | EncodedTransaction)[], signOptions?: SignTransactionOptions): Promise<SignedTx[]>;
+
+    signTransaction(transaction: AlgorandTxn | EncodedTransaction | (AlgorandTxn | EncodedTransaction)[], signOptions?: SignTransactionOptions): Promise<SignedTx | SignedTx[]> {
```
ratik21, line 55: Not sure if i follow. 2 declarations?

packages/web/test/src/lib/myalgowallet-mode.ts
```diff
@@ -0,0 +1,111 @@
...
+	let algodClient: algosdk.Algodv2;
+	const walletURL: HttpNetworkConfig = {
+		token: "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
+		server: "http://localhost",
+		port: 4001,
+	}
+	algodClient = algoexplorerAlgod(walletURL);
+
+	beforeEach(() => {
+		MyAlgoConnect
```
ratik21, line 23: what's this line doing?

packages/web/test/src/lib/myalgowallet-mode.ts
rmeena840: Yes, but we have to use convert it to normal methods. It's needed.

packages/web/test/mocks/myalgowallet-mock.ts
Megha-Dev-19: yeah there official docs also have the two similar method declarations since the parameters type is different and return type is also different

packages/web/test/mocks/myalgowallet-mock.ts
Megha-Dev-19: actually can't rename the method since it needs to be in sync with the MyAlgoWallet class, since it is mock it needs to be a copy of it

packages/web/test/src/lib/myalgowallet-mode.ts
Megha-Dev-19: but why is that needed, we are already checking if this throw an error, and if it doesn't it mean it worked, also I made it by checking the already existing algosigner webmode tests, in there we do the same thing, and contacted Vu if we need more cases but he asked me to keep it simple

packages/web/test/src/lib/myalgowallet-mode.ts
Megha-Dev-19: oops I was checking the connection locally, have to remove it, Thanks

packages/web/test/src/lib/myalgowallet-mode.ts
rmeena840: It's always a good idea to have a strong unit test.

packages/web/test/mocks/myalgowallet-mock.ts
ratik21: alright

ratik21 (approved):

rmeena840 (approved):

PR #808, opened 2022-10-11
Wallet Connect test cases 

Megha-Dev-19:
Added test cases for wallet connect webmode

### DIFF
--- a/packages/web/test/src/lib/walletconnect-mode.ts
+++ b/packages/web/test/src/lib/walletconnect-mode.ts
@@ -2,7 +2,7 @@ import WalletConnect from "@walletconnect/client";
 import algosdk, { Account, Transaction } from "algosdk";
 import assert from "assert";
 import QRCodeModal from "algorand-walletconnect-qrcode-modal";
-import { types, WallectConnectSession } from "../../../src";
+import { testnetURL, types, WallectConnectSession } from "../../../src";
 import { algoexplorerAlgod, getSuggestedParams } from "../../../src/lib/api";
 import { HttpNetworkConfig } from "../../../src/types";
 
@@ -13,9 +13,9 @@ describe("Webmode - Wallet Connect test cases ", function () {
 
     let algodClient: algosdk.Algodv2;
     const walletURL: HttpNetworkConfig = {
-        token: "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
-        server: "http://localhost",
-        port: 4001,
+        token: "",
+        server: testnetURL,
+        port: "",
     }
     algodClient = algoexplorerAlgod(walletURL);
 
@@ -34,7 +34,7 @@ describe("Webmode - Wallet Connect test cases ", function () {
             sign: types.SignType.SecretKey,
             fromAccount: sender,
             toAccountAddr: receiver.addr,
-            amountMicroAlgos: 10000n,
+            amountMicroAlgos: 1e6,
             payFlags: {},
         };
         assert.doesNotThrow(async () => {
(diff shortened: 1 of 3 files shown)


packages/web/test/src/lib/walletconnect-mode.ts
```diff
@@ -0,0 +1,111 @@
...
+    this.beforeEach(function () {
+        sender = algosdk.generateAccount();
+        receiver = algosdk.generateAccount();
+        connector = new WallectConnectSession(walletURL, new WalletConnect({
+            bridge: "https://bridge.walletconnect.org",
+            qrcodeModal: QRCodeModal,
+        }));
+    });
+
+    it("Should executeTx without throwing an error", function () {
```
rmeena840, line 31:
```suggestion
    it("Should executeTx not throw an error", function () {
```

packages/web/test/src/lib/walletconnect-mode.ts
```diff
@@ -0,0 +1,111 @@
...
+        server: testnetURL,
+        port: "",
+    }
+    algodClient = algoexplorerAlgod(walletURL);
+
+    this.beforeEach(function () {
+        sender = algosdk.generateAccount();
+        receiver = algosdk.generateAccount();
+        connector = new WallectConnectSession(walletURL, new WalletConnect({
+            bridge: "https://bridge.walletconnect.org",
```
rmeena840, line 26: If it's already declared as constant then please use it else create one.

packages/web/test/src/lib/walletconnect-mode.ts
```diff
@@ -0,0 +1,111 @@
...
+                type: types.TransactionType.TransferAlgo,
+                sign: types.SignType.SecretKey,
+                fromAccount: sender,
+                toAccountAddr: receiver.addr,
+                amountMicroAlgos: 1e6,
+                payFlags: {},
+            };
+
+            const txnParams = await getSuggestedParams(algodClient)
+            const transactions: Transaction[] = connector.makeTx([execParams], txnParams);
```
rmeena840, line 74: Let's also check receiver's fund.

rmeena840: and sender's fund too

packages/web/test/src/lib/walletconnect-mode.ts
```diff
@@ -0,0 +1,111 @@
...
+                sign: types.SignType.SecretKey,
+                fromAccount: sender,
+                toAccountAddr: receiver.addr,
+                amountMicroAlgos: 1e6,
+                payFlags: {},
+            };
+            const txnParams = await getSuggestedParams(algodClient)
+            assert.doesNotThrow(async () => {
+                await connector.makeAndSignTx([execParams], txnParams);
+            });
```
rmeena840, line 92: same here

packages/web/test/src/lib/walletconnect-mode.ts
```diff
@@ -0,0 +1,111 @@
...
+                type: types.TransactionType.TransferAlgo,
+                sign: types.SignType.SecretKey,
+                fromAccount: sender,
+                toAccountAddr: receiver.addr,
+                amountMicroAlgos: 1e6,
+                payFlags: {},
+            };
+            const txnParams = await getSuggestedParams(algodClient)
+            const signedTx = await connector.makeAndSignTx([execParams], txnParams);
+            assert.doesNotThrow(async () => {
```
rmeena840, line 106: same here

rmeena840 (changes requested):
Reviewed.

packages/web/test/src/lib/walletconnect-mode.ts
Megha-Dev-19: I think without throwing makes more sense

packages/web/test/src/lib/walletconnect-mode.ts
Megha-Dev-19: it is not declared as constant, and it shouldn't be declared one as we create the instance in `beforeEach`

packages/web/test/src/lib/walletconnect-mode.ts
rmeena840: Think about it once before merge.

ratik21 (approved):
lgtm

packages/web/test/src/lib/walletconnect-mode.ts
Megha-Dev-19: updated it with `Should run executeTx function without throwing an error`

PR #809, opened 2022-10-11
Fix opcode global CallerApplicationID

thdailong:
Closes #x

## Proposed Changes

- Fix opcode global CallerApplicationID which previously, we have implemented it wrong way
## TODO

- [x] update change log

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,7 +62,7 @@ from `algosdk` and sends it to the network.
 - Fix `gitxna 1 Logs 0` opcode. Previously any attempt to use this opcode would result in a "Not supported" error.
 - Fix `TxParams.noteb64` encoding - should use base64 decoder rather than TextEncoder.
 - Fix `ed25519verify` opcode implementation. Previously the signature was only checked against the data not the concatenation of "ProgData"||program||data. Additionally test scenarios was added to check the correct implementation. 
-- Fix `global CallerApplicationID` opcode. Previously the opcode get the wrong caller application ID. 
+- Fix `global CallerApplicationID` opcode. Previously the opcode get the wrong caller application ID[798](https://github.com/scale-it/algo-builder/pull/798). 
 
 ### Breaking Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,7 @@ from `algosdk` and sends it to the network.
 - Fix `gitxna 1 Logs 0` opcode. Previously any attempt to use this opcode would result in a "Not supported" error.
 - Fix `TxParams.noteb64` encoding - should use base64 decoder rather than TextEncoder.
 - Fix `ed25519verify` opcode implementation. Previously the signature was only checked against the data not the concatenation of "ProgData"||program||data. Additionally test scenarios was added to check the correct implementation. 
+- Fix `global CallerApplicationID` opcode. Previously the opcode get the wrong caller application ID. 
 
 ### Breaking Changes
 
--- a/packages/runtime/src/ctx.ts
+++ b/packages/runtime/src/ctx.ts
@@ -215,8 +215,8 @@ export class Ctx implements Context {
 
 	getCallerApplicationID(): number {
 		let callerApplicationID = 0;
-		if (this.innerTxAppIDCallStack.length > 0) {
-			callerApplicationID = this.innerTxAppIDCallStack[this.innerTxAppIDCallStack.length - 1];
+		if (this.innerTxAppIDCallStack.length > 1) {
+			callerApplicationID = this.innerTxAppIDCallStack[this.innerTxAppIDCallStack.length - 2];
 		}
 		return callerApplicationID;
 	}
--- a/packages/runtime/src/ctx.ts
+++ b/packages/runtime/src/ctx.ts
@@ -215,8 +215,8 @@ export class Ctx implements Context {
 
 	getCallerApplicationID(): number {
 		let callerApplicationID = 0;
-		// when 
-		// the top value in the `innerTxAppIDCallStack` is the own appID not the caller appID(we need to get the second top one)
+		// when the top value in the `innerTxAppIDCallStack` is the own appID
+		// not the caller appID(we need to get the second top one)
 		if (this.innerTxAppIDCallStack.length > 1) {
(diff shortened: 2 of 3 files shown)


packages/runtime/src/ctx.ts
```diff
@@ -215,8 +215,9 @@ export class Ctx implements Context {
 
 	getCallerApplicationID(): number {
 		let callerApplicationID = 0;
-		if (this.innerTxAppIDCallStack.length > 0) {
-			callerApplicationID = this.innerTxAppIDCallStack[this.innerTxAppIDCallStack.length - 1];
+		// the top value in the `innerTxAppIDCallStack` is the own appID not the caller appID
```
sczembor, line 218: lets add more description

sczembor (commented):
there is one test failing. Please take a look at the comments I left

CHANGELOG.md
```diff
@@ -62,6 +62,7 @@ from `algosdk` and sends it to the network.
 - Fix `gitxna 1 Logs 0` opcode. Previously any attempt to use this opcode would result in a "Not supported" error.
 - Fix `TxParams.noteb64` encoding - should use base64 decoder rather than TextEncoder.
 - Fix `ed25519verify` opcode implementation. Previously the signature was only checked against the data not the concatenation of "ProgData"||program||data. Additionally test scenarios was added to check the correct implementation. 
+- Fix `global CallerApplicationID` opcode. Previously the opcode get the wrong caller application ID.
```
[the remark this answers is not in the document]
PabloLION: It gives the current application ID. See https://github.com/scale-it/algo-builder/pull/798

packages/runtime/src/ctx.ts
```diff
@@ -215,8 +215,9 @@ export class Ctx implements Context {
...
+		if (this.innerTxAppIDCallStack.length > 1) {
```
PabloLION, line 219: It seems you are using my code from https://github.com/scale-it/algo-builder/pull/798/files but not with my commit nor adding me as a co-author. This is not very fair.

packages/runtime/src/ctx.ts
```diff
@@ -215,8 +215,9 @@ export class Ctx implements Context {
...
+			callerApplicationID = this.innerTxAppIDCallStack[this.innerTxAppIDCallStack.length - 2];
```
PabloLION, line 220: Good job on this new PR. Though it seems that you are using my commit from https://github.com/scale-it/algo-builder/pull/798/files but not adding commit in that branch, nor adding me as a co-author. TBH it feels not very fair to me.

PabloLION: Actually, I think my fix was also not so perfect, tho it gives the wanted result. 
And it looks correct to me what you said here in https://github.com/scale-it/algo-builder/pull/798#issuecomment-1268772011.

packages/runtime/test/src/interpreter/opcode-list.ts
```diff
@@ -3051,10 +3051,10 @@ describe("Teal Opcodes", function () {
 			});
 
 			it("Tealv6: CallerApplicationID", function () {
-				// caller app id = 2
+				// caller app id = 1
 				const op = new Global(["CallerApplicationID"], 1, interpreter);
 				op.execute(stack);
-				assert.equal(2n, stack.pop());
+				assert.equal(1n, stack.pop());
```
PabloLION, line 3057: The problem is that the current application id is not pushed to the `innerTxAppIDCallStack`. Getting the second last element makes more sense when `currentAppId` was pushed to the call stack, or otherwise we should still use the last one and remove the action of `push`.
Here, the last element of `interpreter.runtime.ctx.innerTxAppIDCallStack = [1, 2];` was not coherent with `CurrentAppID`. Something is still wrong. (tho tests all pass)

PabloLION (commented):
Seems to me that more changes are needed...

packages/runtime/test/src/interpreter/opcode-list.ts
thdailong: Actually this one is for unit test. Since you see that, we set up the inner transaction application call ID in [here](https://github.com/scale-it/algo-builder/pull/809/files#diff-0bd11169d40e3be81fef88fe8a8e4cf0795255b19d1f4d44d615ea872a1e0e90R3039). So the comment here https://github.com/scale-it/algo-builder/pull/798#issuecomment-1268772011 is where I misunderstand you

CHANGELOG.md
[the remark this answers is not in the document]
thdailong: Let me update the changelog

packages/runtime/src/ctx.ts
thdailong: Oh sorry for making this unfair to you! I am not really familiar with git but I will make add your commit to this branch and make you a co-author

PabloLION: It's ok 😃 I already have the "contributor" chip. You can do it the way you like!

packages/runtime/test/src/interpreter/opcode-list.ts
PabloLION: Oh, ok. That's what I was thinking. Then your fix is what I was trying (but failed) to do. Congrats on finishing it 👍

PabloLION: I'm not the creator not have write access in this repo. So you'd have to resolve these conversations by yourself @thdailong

packages/runtime/src/ctx.ts
```diff
@@ -215,8 +215,10 @@ export class Ctx implements Context {
 
 	getCallerApplicationID(): number {
 		let callerApplicationID = 0;
-		if (this.innerTxAppIDCallStack.length > 0) {
-			callerApplicationID = this.innerTxAppIDCallStack[this.innerTxAppIDCallStack.length - 1];
+		// when 
+		// the top value in the `innerTxAppIDCallStack` is the own appID not the caller appID(we need to get the second top one)
```
sczembor, line 219:
```suggestion
		// when the top value in the `innerTxAppIDCallStack` is the own appID
		// not the caller appID(we need to get the second top one)
```

sczembor (approved):
Good job, please check the comment before merging

PR #811, opened 2022-10-13
Update `sendAndWait` and `waitForConfirmation` methods

Megha-Dev-19:
Updated return type of `waitForConfirmation` and `sendAndWait` in `algob` and `web` packages to make it compatible with `executeTx`

### DIFF
--- a/packages/algob/src/internal/deployer.ts
+++ b/packages/algob/src/internal/deployer.ts
@@ -127,7 +127,7 @@ class DeployerBasicMode {
 		return this.algoOp.algodClient;
 	}
 
-	async waitForConfirmation(txId: string): Promise<ConfirmedTxInfo> {
+	async waitForConfirmation(txId: string): Promise<TxnReceipt> {
 		return await this.algoOp.waitForConfirmation(txId);
 	}
 
@@ -274,7 +274,7 @@ class DeployerBasicMode {
 	sendAndWait(
 		rawTxns: Uint8Array | Uint8Array[],
 		waitRounds = wtypes.WAIT_ROUNDS
-	): Promise<ConfirmedTxInfo> {
+	): Promise<TxnReceipt> {
 		return this.algoOp.sendAndWait(rawTxns, waitRounds);
 	}
 
--- a/packages/algob/src/internal/deployer.ts
+++ b/packages/algob/src/internal/deployer.ts
@@ -32,7 +32,6 @@ import type {
 	ASCCache,
 	CheckpointFunctions,
 	CheckpointRepo,
-	ConfirmedTxInfo,
 	Deployer,
 	FundASCFlags,
 	LogicSig,
@@ -270,6 +269,7 @@ class DeployerBasicMode {
 	 * Send signed transaction to network and wait for confirmation
 	 * @param rawTxns Signed Transaction(s)
 	 * @param waitRounds number of rounds to wait for transaction to be confirmed - default is 10
+	 * @returns TxnReceipt which includes confirmed txn response along with txID
 	 */
 	sendAndWait(
 		rawTxns: Uint8Array | Uint8Array[],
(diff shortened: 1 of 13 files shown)


rmeena840 (approved):

packages/algob/src/internal/deployer.ts
```diff
@@ -274,7 +274,7 @@ class DeployerBasicMode {
 	sendAndWait(
```
sczembor, line 274: lets update the method description about the return type

packages/algob/src/lib/algo-operator.ts
```diff
@@ -140,7 +140,7 @@ export class AlgoOperatorImpl implements AlgoOperator {
 	async sendAndWait(
 		rawTxns: Uint8Array | Uint8Array[],
```
sczembor, line 141: same here

packages/algob/src/lib/algo-operator.ts
```diff
@@ -151,13 +151,14 @@ export class AlgoOperatorImpl implements AlgoOperator {
...
 		waitRounds = wtypes.WAIT_ROUNDS
-	): Promise<ConfirmedTxInfo> {
+	): Promise<TxnReceipt> {
 		const pendingInfo = await algosdk.waitForConfirmation(this.algodClient, txId, waitRounds);
 		if (pendingInfo["pool-error"]) {
 			throw new Error(`Transaction Pool Error: ${pendingInfo["pool-error"] as string}`);
 		}
 		if (pendingInfo[confirmedRound] !== null && pendingInfo[confirmedRound] > 0) {
-			return pendingInfo as ConfirmedTxInfo;
+			const data: TxnReceipt = { txID: txId, ...pendingInfo as ConfirmedTxInfo }
```
sczembor, line 160: can just return the created object? there is no need to create a variable for that if we just return it

packages/algob/src/types.ts
```diff
@@ -492,7 +492,7 @@ export interface Deployer {
 	 * Send signed transaction to network and wait for confirmation
 	 * @param rawTxns Signed Transaction(s)
 	 */
-	sendAndWait: (rawTxns: Uint8Array | Uint8Array[]) => Promise<ConfirmedTxInfo>;
+	sendAndWait: (rawTxns: Uint8Array | Uint8Array[]) => Promise<TxnReceipt>;
```
sczembor, line 495: lets update the description

packages/algob/test/mocks/tx.ts
```diff
@@ -6,7 +6,7 @@ import algosdk, {
 	SuggestedParams,
 } from "algosdk";
 
-import { ConfirmedTxInfo } from "../../src/types";
+import { ConfirmedTxInfo, TxnReceipt } from "../../src/types";
```
sczembor, line 9: do we need the ConfirmedTxInfo import if we dont use it anymore?

sczembor (changes requested):
Please update the function descriptions about the return type. I also think we should add a simple unit test for all the updated functions if the testing mock is available

packages/algob/src/internal/deployer.ts
Megha-Dev-19: sure

packages/algob/src/lib/algo-operator.ts
Megha-Dev-19: okay

packages/algob/test/mocks/tx.ts
Megha-Dev-19: some methods do use it

Megha-Dev-19: I will check if we can remove those as well

packages/algob/src/lib/algo-operator.ts
Megha-Dev-19: yeah we can do that, will update

packages/algob/src/lib/algo-operator.ts
sczembor: this hasn't been fixed

packages/algob/src/lib/tx.ts
```diff
@@ -335,7 +335,7 @@ export async function executeTx(
 export async function executeSignedTxnFromFile(
 	deployer: Deployer,
 	fileName: string
```
sczembor, line 337: lets update the description

sczembor (changes requested):
There are still some missing descriptions

packages/algob/src/lib/tx.ts
Megha-Dev-19: done

packages/algob/test/mocks/tx.ts
```diff
@@ -104,14 +104,15 @@ export const TXN_OBJ: algosdk.EncodedTransaction = {
...
+export const mockTxnReceipt: TxnReceipt = {
+	txID: "1",
 	"confirmed-round": 1,
 	"asset-index": 1,
 	"application-index": 1,
 	txn: {
 		txn: TXN_OBJ,
 	},
-};
+}
```
sczembor, line 115:
```suggestion
};
```

sczembor (approved):
utACK. Check the comment before merging

PR #813, opened 2022-10-18
Web Package Methods Error Added

Megha-Dev-19:
Added throw error to web-mode methods so they fail on UI as well if there occur an error

### DIFF
--- a/packages/web/src/lib/myalgowallet-mode.ts
+++ b/packages/web/src/lib/myalgowallet-mode.ts
@@ -85,6 +85,7 @@ export class MyAlgoWalletSession {
 			}
 		} catch (err) {
 			error(err);
+			throw error
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


packages/web/src/lib/myalgowallet-mode.ts
```diff
@@ -85,6 +85,7 @@ export class MyAlgoWalletSession {
 			}
 		} catch (err) {
 			error(err);
+			throw err
```
rmeena840, line 88: What does `error(err)` does then?

packages/web/src/lib/myalgowallet-mode.ts
```diff
@@ -148,72 +162,90 @@ export class MyAlgoWalletSession {
...
+			const toBeSignedTxns: TransactionInGroup[] = execParams.map(
+				(txn: ExecParams, index: number) => {
+					return txn.sign === SignType.LogicSignature
+						? { txn: txns[index], shouldSign: false } // logic signature
+						: { txn: txns[index], shouldSign: true }; // to be signed
+				}
+			);
+			// only shouldSign txn are to be signed, algowallet doesn't accept lsig ones
+			const nonLsigTxn = toBeSignedTxns.filter((txn) => txn.shouldSign);
+			if (nonLsigTxn.length > 0) {
```
rmeena840, line 225:
```suggestion
			if (nonLsigTxn.length) {
```

rmeena840 (commented):
Left comment

packages/web/src/lib/myalgowallet-mode.ts
Megha-Dev-19: it logs the error (`error` is our internal logger)

rmeena840 (approved):

PR #814, opened 2022-10-18
implementation of vrf_verify

sczembor:
Closes #x

## Proposed Changes

- Add test scenario for vrf_verify opcode 
- Add implementation of vrf_opcode

This is a bit problematic since right now the opcode only support the `vrf` implemented by Algorand. The library I used uses different vrf and this will not work. I am not sure how to approach it

## TODO

- [x] update change log

### DIFF
--- a/packages/runtime/src/interpreter/opcode-list.ts
+++ b/packages/runtime/src/interpreter/opcode-list.ts
@@ -8,14 +8,13 @@ import algosdk, {
 	encodeUint64,
 	getApplicationAddress,
 	isValidAddress,
-	makeApplicationCallTxnFromObject,
 	modelsv2,
 	verifyBytes,
 } from "algosdk";
 import chalk from "chalk";
 import { ec as EC } from "elliptic";
 import { Hasher, Message, sha256 } from "js-sha256";
-import { sha512_256 } from "js-sha512";
+import { sha512, sha512_256 } from "js-sha512";
 import JSONbig from "json-bigint";
 import cloneDeep from "lodash.clonedeep";
 import { Keccak, SHA3 } from "sha3";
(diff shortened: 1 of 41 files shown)


packages/runtime/src/interpreter/opcode-list.ts
```diff
@@ -4353,10 +4353,10 @@ export class ITxnSubmit extends Op {
...
 				types.isExecParams(txnParam)
 					? {
-						sig: Buffer.alloc(5),
-						sgnr: Buffer.from(algosdk.decodeAddress(contractAddress).publicKey),
-						txn: webTx.mkTransaction(txnParam, mockSuggestedParams(txnParam.payFlags, 1)),
-					}
+							sig: Buffer.alloc(5),
+							sgnr: Buffer.from(algosdk.decodeAddress(contractAddress).publicKey),
+							txn: webTx.mkTransaction(txnParam, mockSuggestedParams(txnParam.payFlags, 1)),
+					  }
```
Megha-Dev-19, line 4359: formatting is updated?

packages/runtime/test/src/interpreter/opcode-list.ts
```diff
@@ -7563,4 +7564,69 @@ describe("Teal Opcodes", function () {
...
+			assert.equal(cost, 5700);
+		})
+		it("Should throw error when field = VrfStandard", function(){
+			const op = new VrfVerify(["VrfStandard"], 1);
+			stack.push(publicKey);
+			stack.push(proof);
+			stack.push(message);
+			expectRuntimeError(() => op.execute(stack), RUNTIME_ERRORS.TEAL.UNSUPPORTED_VRF_STANDARD);
+		})
+		it("Should throw error when unknown field", function(){
```
Megha-Dev-19, line 7628:
```suggestion
		it("Should throw error when field is unknown", function(){
```

packages/runtime/src/interpreter/opcode-list.ts
sczembor: it was wrong from the beginning

Megha-Dev-19 (approved):
It looks fine to me, but I m not much aware about the vrf_verify, so you can ask @robert-zaremba or @vuvoth to final review it

PR #817, opened 2022-10-20
Feature: Create algob.config.ts for typescript project.

rmeena840:
Added support for to generate the `algob.config.ts` file generation via algob (`yarn run algob . --typescript`) and read the algob.config.ts` file, if present in root directory.


docs/guide/README.md
```diff
@@ -191,7 +191,7 @@ algob -h deploy
 You can write your scripts and tests in TS. To initialize a new typescript project add `--typescript` to the init flag. Example, if you follow the [Create an algob project](#create-an-algob-project) section then:
 
 ```shell
-yarn run algob init --typescript
+yarn run algob init <path-where-to-create> --typescript
```
thdailong, line 194: Should we add some example to make more clearly?

packages/algob/sample-project/ts/algob.config.ts
```diff
@@ -0,0 +1,89 @@
...
+	+ using binary secret key
+	+ using KMD daemon
+	+ loading from a file
+	+ loading from an environment variable
+	+ ...
+*/
+
+// ## ACCOUNTS USING mnemonic ##
+const { mkAccounts, algodCredentialsFromEnv } = require("@algo-builder/algob");
+let accounts = mkAccounts([
```
thdailong, line 21:
```suggestion
const accounts = mkAccounts([
```

packages/algob/sample-project/ts/algob.config.ts
```diff
@@ -0,0 +1,89 @@
...
+// ## ACCOUNTS loaded from a FILE ##
+// const { loadAccountsFromFileSync } = require("@algo-builder/algob");
+// const accFromFile = loadAccountsFromFileSync("assets/accounts_generated.yaml");
+// accounts = accounts.concat(accFromFile);
+
+/// ## Load accounts from KMD ##
+/// Please check https://github.com/scale-it/algo-builder/blob/master/docs/guide/algob-config.md#network-credentials for more details and more methods.
+//   let kmdCred = KMDCredentialsFromEnv();
+
+let defaultCfg = {
```
thdailong, line 41: Here I think you can make `defaultCfg` as type `Network` or something like that

thdailong (commented):
Have you try to run project with algob.config.ts yet?

docs/guide/README.md
sczembor: agree

packages/algob/sample-project/ts/algob.config.ts
```diff
@@ -0,0 +1,89 @@
...
+*/
+
+// ## ACCOUNTS USING mnemonic ##
+const { mkAccounts, algodCredentialsFromEnv } = require("@algo-builder/algob");
+let accounts = mkAccounts([
+	{
+		// This account is created using `make setup-master-account` command from our
+		// `/infrastructure` directory. It already has many ALGOs
+		name: "master",
+		addr: "WWYNX3TKQYVEREVSW6QQP3SXSFOCE3SKUSEIVJ7YAGUPEACNI5UGI4DZCE",
```
sczembor, line 26: I think this is already somewhere in the project maybe we can move it to a separate file an export it so we have access to it from the rest of the project instead of copy pasting this every time?

sczembor (commented):
left few comments

packages/algob/sample-project/ts/algob.config.ts
rmeena840: We can't do that. On running `yarn run algob . --typescript`, `sample-project` folder content is copied as it is. Whatever the files present inside it will be used to initialize the project via `algob`.

rmeena840:
> Have you try to run project with algob.config.ts yet?

Yes. It works well and unit tests are also passing.

packages/algob/sample-project/ts/algob.config.ts
rmeena840: We can't import this type outside the `sample-project` folder.

docs/guide/README.md
rmeena840: Done. Thanks

rmeena840:
@thdailong @sczembor Ready for another review.

packages/algob/sample-project/ts/algob.config.ts
sczembor: can't we do import {accountConfig} or sth like that?

docs/guide/README.md
```diff
@@ -191,7 +191,15 @@ algob -h deploy
...
+```
+#### Example on `algob init`
+
+You can use below command to initialize the typescript project in `sample-project` folder.
+If `sample-project` folder is not present then it will create one for you.
```
sczembor, line 199: lets add info that we can also do --javascirpt. I think it works the same for both right?

rmeena840: If you don't add --typescript then the default is javascript. --typescript will generate typescript files other javascript files.

packages/algob/sample-project/ts/algob.config.ts
rmeena840: No, we can't do that. `sample-project` will be copied as it is but not the files outside this project folder.

docs/guide/README.md
sczembor: and what happens if you add --javascirpt? works the same as you would add any flag?

sczembor: if we can do it lets add information something like that `Two flags can be used --javascirpt --typescirpt. If no flag was passed the default one is javascirpt...`

rmeena840: I have updated the doc.

sczembor (approved):
good job

PR #824, opened 2022-10-27
Feature: Add --javascript flag to `algo init` task

rmeena840:
Using --javascript with `algob init`, now doesn't throws any error.
Task: https://www.pivotaltracker.com/story/show/183626726

### DIFF
--- a/packages/runtime/src/types.ts
+++ b/packages/runtime/src/types.ts
@@ -7,21 +7,6 @@ import algosdk, {
 	modelsv2,
 	Transaction,
 } from "algosdk";
-
-import {
-	Add,
-	Addr,
-	Arg,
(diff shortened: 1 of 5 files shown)


packages/runtime/src/types.ts
```diff
@@ -8,20 +8,6 @@ import algosdk, {
...
-	Byte,
-	Bytec,
-	Bytecblock,
-	Div,
-	Int,
-	Len,
-	Mul,
-	Pragma,
-	Sub,
-} from "./interpreter/opcode-list";
```
Megha-Dev-19, line 24: why are we removing it?

Megha-Dev-19 (approved):

PR #827, opened 2022-10-28
Web Package tests error fix

Megha-Dev-19:
Fixed the error of `Network request error. Received status 400: empty txgroup`, by actually executing the txn in tests since the methods in webmode require real values


packages/web/test/mocks/myalgowallet-mock.ts
```diff
@@ -54,10 +48,17 @@ export default class MyAlgoConnectMock {
...
     signTransaction(transaction: AlgorandTxn | EncodedTransaction | (AlgorandTxn | EncodedTransaction)[], signOptions?: SignTransactionOptions): Promise<SignedTx | SignedTx[]> {
         return new Promise((resolve, reject) => {
-            return resolve([{
-                blob: new Uint8Array(),
-                txID: ""
-            }]);
+            if (Array.isArray(transaction)) {
+                const signedTransaction = []
+                for (const txn of transaction) {
+                    const decodedtransaction = decodeUnsignedTransaction(txn as unknown as Uint8Array);
```
rmeena840, line 54: can we directly set it to Uint8Array?

packages/web/test/mocks/myalgowallet-mock.ts
```diff
@@ -54,10 +48,17 @@ export default class MyAlgoConnectMock {
...
+                    signedTransaction.push(algosdk.signTransaction(decodedtransaction, senderAccount.sk))
+                }
+                return resolve(signedTransaction)
+            } else {
+                const decodedtransaction = decodeUnsignedTransaction(transaction as unknown as Uint8Array);
```
rmeena840, line 59: same

rmeena840 (approved):
left comment

packages/web/test/mocks/myalgowallet-mock.ts
Megha-Dev-19: do you mean `txn as Uint8Array`?
no we can't since the reader reads the txn with another type (their can be multiple values of it as per the parameter type)

packages/web/test/mocks/tx.ts
```diff
@@ -99,4 +109,14 @@ export const mockSuggestedParam: SuggestedParams = {
...
-};
\ No newline at end of file
+};
+
+export const txnReceiptMock = {
+	txID: "1",
+	"confirmed-round": 1,
+	"asset-index": 1,
+	"application-index": 1,
+	txn: {
```
rmeena840, line 119: This doesn't lookd good. Usally, you shouldn't repeat the root nodes.

packages/web/test/mocks/myalgowallet-mock.ts
Megha-Dev-19: let me see if it is possible, i remember it showed me some error

Megha-Dev-19: oh yes, it worked, thanks

packages/web/test/mocks/myalgowallet-mock.ts
Megha-Dev-19: done

packages/web/test/mocks/tx.ts
Megha-Dev-19: I think we don't require them now, will remove it

PR #828, opened 2022-11-01
Improve error handle when we don't install pyteal, or miss package

rmeena840:
Proper display of error message when no module is found.
Task: https://www.pivotaltracker.com/story/show/182269993


packages/runtime/src/lib/pycompile-op.ts
```diff
@@ -1,5 +1,5 @@
 import { BuilderError, ERRORS } from "@algo-builder/web";
-import { spawnSync, SpawnSyncReturns } from "child_process";
+import { spawnSync, SpawnSyncReturns, exec } from "child_process";
```
sczembor, line 2: where do we use it?

rmeena840: In the same file inside `validatePythonModule` method.

rmeena840:
@sczembor Ready for another review.

packages/runtime/src/lib/pycompile-op.ts
```diff
@@ -90,6 +92,18 @@ export class PyCompileOp {
 		return spawnSync("python3", [filePath, scInitParam], { encoding: "utf8" });
 	}
 
+	/**
+	 * Description: This method checks if given module is installed or not.
```
sczembor, line 96: add information about the exception here `if given module is installed, otherwise throws exception...`

sczembor:
```suggestion
	 * Method checks if given module is installed or not. exception info...
```

sczembor (approved):
Please update the description before merging as well as change log

PR #837, opened 2022-11-07
Foreign app access

sczembor:
Closes #x

## Proposed Changes

- Add additional test scenarios for foreign app account access
- Allow to access foreign app account access bases on the appID

## TODO

- [x] update change log

### DIFF
--- a/packages/runtime/src/interpreter/interpreter.ts
+++ b/packages/runtime/src/interpreter/interpreter.ts
@@ -230,7 +230,7 @@ export class Interpreter {
 	
 	/**
 	 * Queries accesible accounts by smart contract. It can be either accounts defined
-	 * in accounts field or accounts associated with apps decalred in foreingApps array
+	 * in accounts field or accounts associated with apps declared in foreingApps array
 	 * @param accountRef index of account or app ID
 	 * @param line line number
 	 * @param create create flag
--- a/packages/runtime/src/interpreter/interpreter.ts
+++ b/packages/runtime/src/interpreter/interpreter.ts
@@ -55,7 +55,7 @@ export class Interpreter {
 	length: number; // total length of 'compiled' TEAL code
 	// local stores for a transaction.
 	bytecblock: Uint8Array[];
-	intcblock: BigInt[];
+	intcblock: bigint[];
 	scratch: StackElem[];
 	// TEAL parsed code - instantiated during the execution phase.
 	instructions: Op[];
@@ -136,14 +136,14 @@ export class Interpreter {
 	 * When `create` flag is false we will create new account and add it to context.
 	 * @param accountPk public key of account
 	 * @param line line number in TEAL file
-	 * @param create create flag
+	 * @param createFlag create flag
 	 * @param immutable allow to access foreign application account or not(false), default is true
 	 * https://developer.algorand.org/articles/introducing-algorand-virtual-machine-avm-09-release/
 	 */
 	private _getAccountFromAddr(
 		accountPk: Uint8Array,
 		line: number,
-		create: boolean,
+		createFlag: boolean,
 		immutable: boolean
 	): AccountStoreI {
 		const txAccounts = this.runtime.ctx.tx.apat; // tx.Accounts array
@@ -183,7 +183,7 @@ export class Interpreter {
 				) !== undefined)
 		) {
 			const address = encodeAddress(pkBuffer);
-			const account = create
+			const account = createFlag
 				? this.createAccountIfAbsent(address)
 				: this.runtime.ctx.state.accounts.get(address);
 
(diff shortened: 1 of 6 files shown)


packages/runtime/src/interpreter/interpreter.ts
```diff
@@ -215,31 +215,50 @@ export class Interpreter {
...
+			return this._getAccountFromReference(accountRef, line, create);
 		} else {
 			return this._getAccountFromAddr(accountRef, line, create, immutable);
 		}
 
+	}
+	
+	/**
+	 * Queries accesible accounts by smart contract. It can be either accounts defined
+	 * in accounts field or accounts associated with apps decalred in foreingApps array
```
Megha-Dev-19, line 233:
```suggestion
	 * in accounts field or accounts associated with apps declared in foreingApps array
```

packages/runtime/src/interpreter/interpreter.ts
```diff
@@ -215,31 +215,50 @@ export class Interpreter {
...
+	 * @param accountRef index of account or app ID
+	 * @param line line number
+	 * @param create create flag
```
Megha-Dev-19, line 236: may be we can rename it to `createFlag`.?

packages/runtime/test/src/interpreter/interpreter.ts
```diff
@@ -198,6 +198,7 @@ describe("Interpreter", function () {
 				setUpInterpreter(7, 1e9);
 			});
 			it("Should not throw an error when accessing account that are in transaction's account field", function () {
+				// in which exact part are we accesing the account???
```
Megha-Dev-19, line 201: should it be there..?

sczembor: you are right, I just could not figure out what is going on here and left it there to ask the rest of the team. @thdailong could you provide some info about it?

packages/runtime/src/interpreter/interpreter.ts
sczembor: that's a good idea, I will update it

packages/runtime/src/interpreter/interpreter.ts
```diff
@@ -215,31 +215,50 @@ export class Interpreter {
...
-					? this.createAccountIfAbsent(address)
-					: this.runtime.ctx.state.accounts.get(address);
-			}
+
+		if (accountRef === 0n) {
+			address = encodeAddress(this.runtime.ctx.tx.snd);
+			account = this.runtime.ctx.state.accounts.get(address);
+			return this.runtime.assertAccountDefined(address, account, line);
+		} else if (typeof accountRef === "bigint") {
+			return this._getAccountFromReference(accountRef, line, create);
```
rmeena840, line 224: You are revering the condition? Please check if this is really needed.

rmeena840 (commented):
left comment

packages/runtime/src/interpreter/interpreter.ts
sczembor: I am not sure which part are you referring to

packages/runtime/src/interpreter/interpreter.ts
```diff
@@ -55,7 +55,7 @@ export class Interpreter {
 	length: number; // total length of 'compiled' TEAL code
 	// local stores for a transaction.
 	bytecblock: Uint8Array[];
-	intcblock: BigInt[];
+	intcblock: bigint[];
```
rmeena840, line 58: I don't think we have bigint in small case. Please check this.

rmeena840 (commented):
left comments

packages/runtime/src/interpreter/interpreter.ts
rmeena840: `typeof accountRef === "bigint"` when this is true we were doing something but we are doing the opposite.

packages/runtime/src/interpreter/interpreter.ts
sczembor: bigint small case is the type and BigInt is the constructor of the type

sczembor: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt

packages/runtime/src/interpreter/interpreter.ts
sczembor: we are doing the same it was just moved to a separate method

packages/runtime/src/interpreter/interpreter.ts
rmeena840: ok

rmeena840 (approved):
Approved

PR #838, opened 2022-11-08
Fix pre commit

sczembor:
Closes #x

## Proposed Changes

- added additional flag to `git diff`
- moved the script to be run as `yarn run check:only`

## TODO

- [x] update change log


packages/runtime/test/src/interpreter/opcode-list.ts
```diff
@@ -5739,8 +5740,8 @@ describe("Teal Opcodes", function () {
 				assert.deepEqual(
 					stack.pop(),
 					new Uint8Array([
-						94, 220, 186, 137, 254, 220, 186, 137, 254, 220, 186, 137, 254, 220, 186, 137,
-						254, 220, 186, 137, 254, 220, 186, 137, 254, 220, 186, 137, 254, 220, 186, 137,
+						94, 220, 186, 137, 254, 220, 186, 137, 254, 220, 186, 137, 254, 220, 186, 137, 254,
+						220, 186, 137, 254, 220, 186, 137, 254, 220, 186, 137, 254, 220, 186, 137,
```
robert-zaremba, line 5744: Why this was wrong?

robert-zaremba (approved):
utAck

packages/runtime/test/src/interpreter/opcode-list.ts
sczembor: this is formatting introduced by the `yarn lint-staged`

PR #839, opened 2022-11-10
Allow runtime to get account from algob.config.js

thdailong:
Closes #x

## Proposed Changes

- Allow runtime to get account from algob.config.js

## TODO

- [x] update change log


packages/runtime/src/runtime.ts
```diff
@@ -422,6 +423,22 @@ export class Runtime {
 		this.store.accounts.set(feeSink.address, feeSink);
 	}
 
+	/**
+	 * Set up account from deployer to runtime
```
sczembor, line 427: I am not sure what we are doing here. Could you explain?

packages/algob/sample-project/js/test/sample-test.js
```diff
@@ -36,6 +37,10 @@ describe("Sample Test", function () {
 		assert.equal(fundReceiver.balance(), minBalance);
 		assert.equal(master.balance(), masterBalance);
 
+		// In case you want to get account from algob.config.js with network default
```
sczembor, line 40: I am not sure if its a good place to put the description at

CHANGELOG.md
```diff
@@ -77,6 +77,8 @@ Features, Bug Fixes, API Breaking, Deprecated, Infrastructure, Template Updates
 - Added blocks to `Runtime`. It simulates the block generation by using radnom bytes generator as the first seed. The following seeds are MD5 hash of the seed from the previous block.
 - Added support for `secp256r1` curve to `ecdsa_verify` and `ecdsa_pk_decompress` opcodes.
 - Added support for `FirstValidTime` field for transactions opcode.
+- Allow `Runtime` to get account from `algob.config.js`
```
sczembor, line 80: Lets add a little bit more explanation of the feature

sczembor (commented):
left few comments

packages/runtime/src/runtime.ts
thdailong: In runtime, I do not see we have method that adding account to the store so I create this method so as to help me adding account from the config to the store

packages/algob/sample-project/js/test/sample-test.js
thdailong: I think I put the wrong position. I will put them in the `this.beforeEach
`

packages/algob/sample-project/js/test/sample-test.js
```diff
@@ -21,6 +21,10 @@ describe("Sample Test", function () {
 		master = new AccountStore(masterBalance);
 		fundReceiver = new AccountStore(minBalance);
 		runtime = new Runtime([master, fundReceiver]);
+		// ## ACCOUNTS loaded from algob.config.js file ##
```
rmeena840, line 24: You can remove them , if not needed.

packages/runtime/src/runtime.ts
```diff
@@ -422,6 +423,18 @@ export class Runtime {
 		this.store.accounts.set(feeSink.address, feeSink);
 	}
 
+	/**
+	 * Adding accounts to Store
```
rmeena840, line 427:
```suggestion
	 * Add accounts to the Store
```

packages/runtime/src/runtime.ts
```diff
@@ -422,6 +423,18 @@ export class Runtime {
...
+	 * @param accounts: array of accounts
+	 */
```
rmeena840, line 429: Please add balance param description as well.

rmeena840 (dismissed):
Reviewed

packages/algob/sample-project/js/test/sample-test.js
thdailong: Ah I need this one for example how can we get account from algob.config.js in case user need

packages/algob/sample-project/js/test/sample-test.js
```diff
@@ -21,6 +21,10 @@ describe("Sample Test", function () {
...
+		// const cfg = require("../algob.config");
+		// runtime.addAccounts(cfg.networks.default.accounts, 1000000);
```
sczembor, line 26: please remove the comments if not needed

packages/runtime/src/runtime.ts
```diff
@@ -422,6 +423,19 @@ export class Runtime {
 		this.store.accounts.set(feeSink.address, feeSink);
 	}
 
+	/**
+	 * Add accounts to the Store
+	 * @param accounts: array of accounts
+	 * @param balance: balance for accounts
+	 */
+	addAccounts(accounts: Account[], balance: number | bigint): void {
```
sczembor, line 431:
```suggestion
	addAccountsToStore(accounts: Account[], balance: number | bigint): void {
```

sczembor: also why accounts is an array and balance is just a single value?

sczembor (commented):
lets add a test for the new method

packages/algob/sample-project/js/test/sample-test.js
thdailong: But shouldn't we need comments to tell user how to get account from algob.config.js?

packages/runtime/src/runtime.ts
```diff
@@ -422,6 +426,30 @@ export class Runtime {
 		this.store.accounts.set(feeSink.address, feeSink);
 	}
 
+	/**
+	 * Add accounts from config file to the Store
+	 * @param network: the network of accounts to add
+	 * @param balance: balance for accounts
+	 */
+	loadAccountsFromConfig(network = "default", balance?: number): void {
```
sczembor, line 434: lets add a const for `default`

packages/runtime/src/runtime.ts
```diff
@@ -422,6 +426,30 @@ export class Runtime {
...
+		const find = findupSync([JS_CONFIG_FILENAME, TS_CONFIG_FILENAME]);
+		if (!find) {
+			throw new RuntimeError(RUNTIME_ERRORS.GENERAL.CONFIG_FILE_NOT_FOUND);
+		}
+		// eslint-disable-next-line @typescript-eslint/no-var-requires
+		const config = require(find);
+
+		console.log(config);
+		console.log(config.networks[network]);
```
sczembor, line 443: lets remove the console.logs

packages/runtime/src/runtime.ts
```diff
@@ -422,6 +426,30 @@ export class Runtime {
...
+		const accounts = config.networks[network].accounts;
```
sczembor, line 444: what happens if the network user provided is not defined?

sczembor (commented):
left few comments

packages/runtime/src/runtime.ts
thdailong: Just add new error check if network exist or not

packages/runtime/src/errors/errors-list.ts
```diff
@@ -652,6 +652,24 @@ const runtimeGeneralErrors = {
...
+	CONFIG_FILE_NOT_FOUND: {
+		number: 1325,
+		message: "Config file is not found",
+		title: "Config find not found",
+		description: "The config file in project is not found",
+	},
+	NETWORK_ACCOUNT_NOT_FOUND: {
+		number: 1326,
+		message: "Accounts in network %network% is not found",
+		title: "Accounts in network  is not found",
```
Megha-Dev-19, line 664:
```suggestion
		title: "Accounts is not found in network",
```

packages/runtime/src/errors/errors-list.ts
```diff
@@ -652,6 +652,24 @@ const runtimeGeneralErrors = {
 		title: "Invalid round",
 		description: "round %round is not available. It's outside [%firstAvail-%lastAvail]",
 	},
+	CONFIG_FILE_NOT_FOUND: {
+		number: 1325,
+		message: "Config file is not found",
+		title: "Config find not found",
```
Megha-Dev-19, line 658:
```suggestion
		title: "Config file not found",
```

packages/algob/sample-project/js/test/sample-test.js
sczembor: so lets add a comment above this part `This is an example to retrieve account from... if the user wants to use it uncomment the following section...`

packages/runtime/src/runtime.ts
```diff
@@ -422,6 +427,37 @@ export class Runtime {
...
+	 * Add accounts from config file to the Store
+	 * @param network: the network of accounts to add
+	 * @param balance: balance for accounts
+	 */
+	loadAccountsFromConfig(network = NETWORK_DEFAULT, balance?: number): void {
+		const find = findupSync([JS_CONFIG_FILENAME, TS_CONFIG_FILENAME]);
+		if (!find) {
+			throw new RuntimeError(RUNTIME_ERRORS.GENERAL.CONFIG_FILE_NOT_FOUND);
+		}
+		// eslint-disable-next-line @typescript-eslint/no-var-requires
```
sczembor, line 440: can we do it by `import as` instead of the require?

sczembor (approved):
look good. Please take a look at the comments I left before merging

packages/runtime/src/runtime.ts
thdailong: `import as` should be use on the top of the file right. so using `require` here would be better than `import as`

PR #840, opened 2022-11-10
Update: Use enum instead of constant strings for transaction fields

rmeena840:
Use enum instead of constant strings for transaction fields


packages/runtime/src/interpreter/opcode-list.ts
```diff
@@ -1835,7 +1836,7 @@ export class Global extends Op {
 
 	/**
 	 * Stores global field to query as string
-	 * @param args Expected arguments: [field] // Ex: ["GroupSize"]
+	 * @param args Expected arguments: [field] // Ex: [TxFieldEnum.GroupSize]
```
sczembor, line 1839: What happens if we pass the string to a constructor? will it fail?

rmeena840: Yes.

packages/runtime/src/interpreter/opcode-list.ts
```diff
@@ -4989,11 +4986,11 @@ export class Base64Decode extends Op {
 		assertLen(args.length, 1, line);
 		const argument = args[0];
 		switch (argument) {
-			case "URLEncoding": {
+			case TxFieldEnum.URLEncoding: {
```
sczembor, line 4989: That's not a Tx field. We should have a separate enum for json encodings

packages/runtime/src/lib/constants.ts
```diff
@@ -536,36 +536,121 @@ export enum vrfVerifyFieldTypes {
 }
 
 export enum TxFieldEnum {
+	Amount = "Amount",
+	Accounts = "Accounts",
+	AcctBalance = "AcctBalance",
+	AcctMinBalance = "AcctMinBalance",
```
sczembor, line 542: This should only have txFields. Some of them are just opcode parameters which are not part of a tx. Lets move it to a separate enums

packages/runtime/src/interpreter/opcode-list.ts
sczembor: I think we should allow it. Some users might be using it and this will be a breaking change.

sczembor (changes requested):
The enum should only store values of the tx fields. The rest should be moved to a separate enums

packages/runtime/src/interpreter/opcode-list.ts
rmeena840: Sorry, are you talking the args param?
No, it won't break. The string is still allowed. We have an enum of fields in string so it will work.

rmeena840: Yes, it still accepts string in the constructor.

packages/runtime/src/interpreter/opcode-list.ts
rmeena840: Done

packages/runtime/src/lib/constants.ts
rmeena840: Done

rmeena840:
@sczembor We still allow string in the constructor. It won't be a breaking change.
I have moved the opcode fields to enum. Please check

packages/runtime/src/lib/constants.ts
```diff
@@ -307,22 +307,94 @@ AssetParamMap[6] = { ...AssetParamMap[5] };
...
+	NumApplications = "NumApplications",
+	NumApprovalProgramPages = "NumApprovalProgramPages",
+	NumClearStateProgramPages = "NumClearStateProgramPages",
+	NumLogs = "NumLogs",
+	OnCompletion = "OnCompletion",
+	TxID = "TxID",
+}
+
+export enum TxnaField {
+	Accounts = "Accounts",
```
sczembor, line 373: I think these can also be used with regular txn/itxn opcodes

packages/runtime/src/lib/constants.ts
```diff
@@ -307,22 +307,94 @@ AssetParamMap[6] = { ...AssetParamMap[5] };
 AssetParamMap[7] = { ...AssetParamMap[6] };
 AssetParamMap[8] = { ...AssetParamMap[7] };
 
+export enum AssetParamGetField {
```
sczembor, line 310: lets add links to some reference for all the enums

packages/runtime/src/lib/constants.ts
```diff
@@ -307,22 +307,94 @@ AssetParamMap[6] = { ...AssetParamMap[5] };
...
+	AssetURL = "AssetURL",
+	AssetMetadataHash = "AssetMetadataHash",
+	AssetManager = "AssetManager",
+	AssetReserve = "AssetReserve",
+	AssetFreeze = "AssetFreeze",
+	AssetClawback = "AssetClawback",
+	AssetCreator = "AssetCreator",
+}
+
+export enum AppParamField {
```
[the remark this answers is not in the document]
sczembor: also should we move it to web?

packages/runtime/src/lib/constants.ts
rmeena840: Done

packages/runtime/src/lib/constants.ts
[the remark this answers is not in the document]
rmeena840: Done

packages/runtime/src/lib/constants.ts
rmeena840: Done

sczembor (approved):
utACK

PR #841, opened 2022-11-14
Update:  Batches in packages tests

rmeena840:


.github/workflows/01-main.yaml
```diff
@@ -47,5 +47,77 @@ jobs:
...
+      - name: Setup Node
+        uses: actions/setup-node@v2-beta
+        with:
+          node-version: "16"
+          cache: "yarn"
+      - name: Yarn version check
+        run: yarn --version
+      - name: Install dependencies
+        run: yarn install
+      - name: Execute format
```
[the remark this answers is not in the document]
rmeena840: It also checks for linter as well, which is needed.

sczembor (commented):
its ok but I think we should build the entire project once and run all the checks and then just execute the tests in batches. Is it possible or every batch is running in isolated environment?

rmeena840:
> its ok but I think we should build the entire project once and run all the checks and then just execute the tests in batches. Is it possible or every batch is running in isolated environment?

Every batch is running in isolated env.

.github/workflows/01-main.yaml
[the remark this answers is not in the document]
sczembor: If its possible lets move it to a one script and just invoke it for every batch so we do not repeat the same code multiple times

rmeena840:
@sczembor Ready for another review.

.github/workflows/reusable-workflow.yaml
```diff
@@ -0,0 +1,45 @@
...
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Setup Node
+        uses: actions/setup-node@v2-beta
+        with:
+          node-version: "16"
+          cache: "yarn"
+      - name: Yarn version check
```
sczembor, line 23:
```suggestion
      - name: Check Yarn version
```

.github/workflows/reusable-workflow.yaml
```diff
@@ -0,0 +1,45 @@
...
+        with:
+          python-version: "3.10"
+          cache: "pipenv"
+      - name: Install poetry
+        run: pip install -U poetry
+      - name: Install pipenv
+        run: python -m pip install pipenv
+      - name: Install python dependencies
+        run: pipenv sync
+      - name: Run test
```
sczembor, line 44:
```suggestion
      - name: Run tests
```

.github/workflows/reusable-workflow.yaml
```diff
@@ -0,0 +1,45 @@
+name: Reusable workflow example
```
sczembor, line 1: why is it call example?

sczembor (commented):
left few comments

.github/workflows/reusable-workflow.yaml
rmeena840, line 1:
```suggestion
name: Reusable workflow
```

rmeena840: Corrected it.

sczembor (approved):
Good job! Please update the update the changelog before merging

sczembor:
You did not update the change log

PR #842, opened 2022-11-14
Update `appendSignMultisigTransaction` 

Megha-Dev-19:
`appendSignMultisigTransaction` now expects `Encoded MultiSignature Transaction` as first argument and returns an object containing a blob key encoded in base64.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ Features, Bug Fixes, API Breaking, Deprecated, Infrastructure, Template Updates
 ### Breaking Changes
 
 - The method `ProduceBlock` has been renamed to `ProduceBlocks(numberOfBlocks=1)` and now accepts optional parameter that allows user to specify the number of blocks that will be produced.
+- `appendSignMultisigTransaction` now expects `Encoded MultiSignature Transaction` as first argument and returns an object containing a blob key encoded in base64.
 
 #### TEALv8
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,7 @@ Features, Bug Fixes, API Breaking, Deprecated, Infrastructure, Template Updates
 ### Breaking Changes
 
 - The method `ProduceBlock` has been renamed to `ProduceBlocks(numberOfBlocks=1)` and now accepts optional parameter that allows user to specify the number of blocks that will be produced.
-- `appendSignMultisigTransaction` now expects `Encoded MultiSignature Transaction` as first argument and returns an object containing a blob key encoded in base64.
+- `appendSignMultisigTransaction` now expects `EncodedSignedTransaction` as first argument and returns an object containing a blob key encoded in base64.
 
 #### TEALv8
 
--- a/packages/web/src/lib/web-mode.ts
+++ b/packages/web/src/lib/web-mode.ts
@@ -1,5 +1,6 @@
 import algosdk, {
 	EncodedSignedTransaction,
+	encodeObj,
 	SignedTransaction,
 	SuggestedParams,
 	Transaction,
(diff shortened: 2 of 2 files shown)


packages/web/src/lib/web-mode.ts
```diff
@@ -184,62 +185,49 @@ export class WebMode {
...
-				const txnToBeSign = algosdk.Transaction.from_obj_for_encoding(partialTxn.txn);
-				const txnToBeSign_Uint8Array = algosdk.encodeObj(txnToBeSign.get_obj_for_encoding());
-				const txnToBeSign_Base64 =
-					this.algoSigner.encoding.msgpackToBase64(txnToBeSign_Uint8Array);
-
-				const mparams = partialTxn.msig as algosdk.EncodedMultisig;
-				const addrs = mparams.subsig.map((signData) => {
-					return algosdk.encodeAddress(signData.pk);
-				});
+			const encodedTxn = this.algoSigner.encoding.msgpackToBase64(encodeObj(txn.txn))
```
thdailong, line 197: Shouldn't we need to check whether the txn input should be a multisig transaction first?

packages/web/src/lib/web-mode.ts
```diff
@@ -184,62 +185,49 @@ export class WebMode {
...
+					txn: encodedTxn,
+					msig: multisigParams,
+					signers: signers,
+				},
+			]);
+
+			const signedTxnJson = signedTxn[0] as JsonPayload;
+			let combineBlob = this.algoSigner.encoding.base64ToMsgpack(signedTxnJson.blob as string)
+			// multiple signatures
+			if (txn.msig?.subsig.findIndex((item) => item.s?.length) !== -1) {
```
thdailong, line 222: From here, I think you can throw error like `Can not sign due to signer is not in the msig address`

thdailong (commented):
Just leave some comments. The code is good but I have some concern

packages/web/src/lib/web-mode.ts
Megha-Dev-19: the input parameter is already updated from `WalletTransaction` to `EncodedSignedTransaction`

packages/web/src/lib/web-mode.ts
Megha-Dev-19: why this error? 
this code is to combine the blob (if the transaction provided is partially signed not completely)
and even if it is signed wholly, it throws error by itself

packages/web/src/lib/web-mode.ts
thdailong: However, in type [EncodedSignedTransaction](https://algorand.github.io/js-algorand-sdk/interfaces/EncodedSignedTransaction.html), `msig` is an optional parameter so you still need to check it

packages/web/src/lib/web-mode.ts
thdailong: Yes it should throw error but you can define the content of the error you want to throw so that user would know better what is the error instead of some throwing error can cause user to confuse

Megha-Dev-19: I don't think this is the error which is appropriate here, because we are not really checking for the addr here, and also there can be the case of rekeyed account

packages/web/src/lib/web-mode.ts
Megha-Dev-19: oh okay, will update

packages/web/src/lib/web-mode.ts
```diff
@@ -184,62 +185,52 @@ export class WebMode {
...
-				const txnToBeSign_Uint8Array = algosdk.encodeObj(txnToBeSign.get_obj_for_encoding());
-				const txnToBeSign_Base64 =
-					this.algoSigner.encoding.msgpackToBase64(txnToBeSign_Uint8Array);
-
-				const mparams = partialTxn.msig as algosdk.EncodedMultisig;
-				const addrs = mparams.subsig.map((signData) => {
-					return algosdk.encodeAddress(signData.pk);
-				});
+			if (!txn.msig) {
+				throw new Error("Current transaction is not a Multisign Transaction.")
```
rmeena840, line 198:
```suggestion
				throw new Error("Current transaction is not a Multisig Transaction.")
```

rmeena840 (changes requested):
Reviewed

thdailong (approved):
Now it seems better to me!

CHANGELOG.md
```diff
@@ -31,6 +31,7 @@ Features, Bug Fixes, API Breaking, Deprecated, Infrastructure, Template Updates
 ### Breaking Changes
 
 - The method `ProduceBlock` has been renamed to `ProduceBlocks(numberOfBlocks=1)` and now accepts optional parameter that allows user to specify the number of blocks that will be produced.
+- `appendSignMultisigTransaction` now expects `Encoded MultiSignature Transaction` as first argument and returns an object containing a blob key encoded in base64.
```
rmeena840, line 34:
```suggestion
- `appendSignMultisigTransaction` now expects `EncodedSignedTransaction` as first argument and returns an object containing a blob key encoded in base64.
```

rmeena840: It would be good to keep exact parameter type.

packages/web/src/lib/web-mode.ts
```diff
@@ -184,62 +185,52 @@ export class WebMode {
...
+			const version = mparams.v;
+			const threshold = mparams.thr;
+			const addr = mparams.subsig.map((signData) => {
+				return algosdk.encodeAddress(signData.pk)
+			});
+
+			const multisigParams = {
+				version: version,
+				threshold: threshold,
+				addrs: addr,
```
rmeena840, line 211: typo error. `address`

rmeena840 (commented):
Reviewed

packages/web/src/lib/web-mode.ts
Megha-Dev-19: .? Algosigner expects the multisig params with these keys only
https://github.com/scale-it/algo-builder/blob/5581a5663dc304650fd83e0bfbadc9ad743578c9/packages/web/src/algo-signer-types.ts#L31

rmeena840: ok

rmeena840 (approved):

PR #844, opened 2022-11-15
Update `connectToMyAlgo`

Megha-Dev-19:
`allowMultipleAccounts` parameter added to `connectToMyAlgo` to give user the flexibility to allow multi accounts login using MyAlgo Wallet

### DIFF
--- a/packages/web/src/lib/myalgowallet-mode.ts
+++ b/packages/web/src/lib/myalgowallet-mode.ts
@@ -89,12 +89,12 @@ export class MyAlgoWalletSession {
 		}
 	}
 
-	// https://connect.myalgo.com/docs/interactive-examples/Connect
 	/**
 	 * @async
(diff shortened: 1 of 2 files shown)


packages/web/src/lib/myalgowallet-mode.ts
```diff
@@ -90,10 +90,16 @@ export class MyAlgoWalletSession {
 	}
 
 	// https://connect.myalgo.com/docs/interactive-examples/Connect
```
rmeena840, line 92: You can add this link in the description you have added. Something like `for more info visit: <url>`

rmeena840 (approved):
Approved. See comments before merge.

packages/web/src/lib/myalgowallet-mode.ts
Megha-Dev-19: oh yes, thanks will do that

PR #850, opened 2022-11-23
Feature: Algob cli format for passing the arguments in script.

rmeena840:
New format for passing the arguments in algob cli:
`yarn algob run scripts/0-gold-asa.js --arg '{"firstname":"Jesper","surname":"Aaberg","phone":["555-0100",   "555-0120"]}'`

### DIFF
--- a/packages/algob/src/builtin-tasks/run.ts
+++ b/packages/algob/src/builtin-tasks/run.ts
@@ -133,7 +133,9 @@ async function executeRunTask(
 	const logDebugTag = "algob:tasks:run";
 	let scriptName;
 	if (arg && !isValidJsonString(arg)) {
-		throw new BuilderError(ERRORS.BUILTIN_TASKS.RUN_ARGUMENT_INVALID);
+		throw new BuilderError(ERRORS.BUILTIN_TASKS.RUN_ARGUMENT_INVALID, {
+			jsonString: arg
+		});
 	}
 	if (script && script.length) {
 		// get script from script array, first element should be script
--- a/packages/web/src/errors/errors-list.ts
+++ b/packages/web/src/errors/errors-list.ts
@@ -675,9 +675,9 @@ Use 'deployer.isDefined(name)' to check if the name is already used.
 	},
 	RUN_ARGUMENT_INVALID: {
 		number: 618,
-		message: "The JSON string passed %jsonString% is invalid.",
+		message: "Invalid JSON string: %jsonString%",
 		title: "Run argument is invalid.",
-		description: "The JSON string passed is invalid.",
+		description: "Invalid JSON string: %jsonString%",
 	},
 };
 


sczembor (commented):
Let's mention the changes in guide, also we should create an example/test for it

packages/web/src/errors/errors-list.ts
```diff
@@ -667,6 +667,12 @@ Use 'deployer.isDefined(name)' to check if the name is already used.
 Use 'deployer.isDefined(name)' to check if the name is already used.
 `,
 	},
+	RUN_FILE_NOT_FOUND_WITH_SUGGESTION: {
+		number: 617,
+		message: "Script not found. Please check the format: yarn algob run script.js --args arg1 arg2 arg3",
```
rmeena840, line 672:
```suggestion
		message: `Script not found. Please check the format: yarn algob run script.js --arg '<JSON String>' ,
```

packages/web/src/errors/errors-list.ts
```diff
@@ -667,6 +667,12 @@ Use 'deployer.isDefined(name)' to check if the name is already used.
...
+		title: "Run file not found.",
+		description: `Script not found. Please check the format: yarn algob run script.js --args arg1 arg2 arg3`,
```
rmeena840, line 674:
```suggestion
		description: `Script not found. Please check the format: yarn algob run script.js --arg '<JSON String>' ,
```

rmeena840:
> example/test

I have created a task for this. https://www.pivotaltracker.com/story/show/183908501

thdailong (commented):
I am bit curious if we catch error  the argument user passed in is a correct json format or not?

rmeena840:
> I am bit curious if we catch error the argument user passed in is a correct json format or not?

I have handled this case.

packages/algob/test/builtin-tasks/run.ts
```diff
@@ -31,30 +31,62 @@ describe("Run task", function () {
...
+			ERRORS.BUILTIN_TASKS.RUN_FILE_NOT_FOUND_WITH_SUGGESTION,
+		);
+	});
+
+	it("Should throw error when a valid JSON string is not passed as argument in the script", async function () {
+		await expectBuilderErrorAsync(
+			async () =>
+				await this.env.run(TASK_RUN, {
+					script: ["./scripts/async-script.js"],
+					arg: "{name: usersname}"
```
sczembor, line 84: why is it invalid json?

sczembor (commented):
lets add a simple example with the argument and very basic script where we use an argument from cli

packages/algob/test/builtin-tasks/run.ts
rmeena840: It should be like `{"name": "username"}`

packages/web/src/errors/errors-list.ts
```diff
@@ -667,6 +667,18 @@ Use 'deployer.isDefined(name)' to check if the name is already used.
...
 	},
+	RUN_FILE_NOT_FOUND_WITH_SUGGESTION: {
+		number: 617,
+		message: "Script not found. Please check the format: yarn algob run script.js --arg '<JSON String>'",
+		title: "Run file not found.",
+		description: "Script not found. Please check the format: yarn algob run script.js --arg '<JSON String>'",
+	},
+	RUN_ARGUMENT_INVALID: {
+		number: 618,
+		message: "The JSON string passed is invalid.",
```
thdailong, line 678: You could add parameter to this message so that user can have better understand
```suggestion
		message: "The JSON string passed %jsonString% is invalid.",
```

thdailong (commented):
Will it throw error if we run `yarn algob run script.js '<JSON String>'`?

rmeena840:
> Will it throw error if we run `yarn algob run script.js '<JSON String>'`?

Yes, that's the cli format.

thdailong:
> > Will it throw error if we run `yarn algob run script.js '<JSON String>'`?
> 
> Yes, that's the cli format. @thdailong

No, my means is that passing like this `yarn algob run script.js '<JSON String>'` without `--arg` flag would throw error or not

rmeena840:
> 

Yes, it will throw error.

rmeena840:
> > > Will it throw error if we run `yarn algob run script.js '<JSON String>'`?
> > 
> > 
> > Yes, that's the cli format. @thdailong
> 
> No, my means is that passing like this `yarn algob run script.js '<JSON String>'` without `--arg` flag would throw error or not

@thdailong Yes, it will throw error.

thdailong (approved):
Check the comments before merge! You should make sure Stan also approve this one too!

packages/algob/src/builtin-tasks/run.ts
```diff
@@ -112,38 +111,63 @@ async function runSortedScripts(
...
+			arg,
+			(_cpData: CheckpointRepo, _relativeScriptPath: string) => { }, // eslint-disable-line @typescript-eslint/no-empty-function
+			true,
+			logDebugTag,
+			false,
+			deployerCfg
+		);
+	} else {
+		throw new BuilderError(ERRORS.BUILTIN_TASKS.RUN_FILE_NOT_FOUND_WITH_SUGGESTION);
+	}
```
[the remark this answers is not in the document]
rmeena840: The format is correct.

rmeena840:
@sczembor Please review the PR.

robert-zaremba (approved):

PR #854, opened 2022-11-25
Rename sign msig transaction

robert-zaremba:
Closes #x

## Proposed Changes

renamed `web.signMsigTransaction` to `web.signMsigTx` to make it "compatible" with other functions.


Megha-Dev-19 (approved):
Thanks for the update

PR #859, opened 2022-12-07
Add interface Executor

thdailong:
Closes #x

## Proposed Changes

- Add interface Executor which can work in all 3 packages
-
-

## TODO

- [ ] update change log

### DIFF
--- a/packages/web/src/lib/executor.ts
+++ b/packages/web/src/lib/executor.ts
@@ -1,0 +1,58 @@
+import algosdk, { Account, SignedTransaction, Transaction } from "algosdk";
+
+import { ExecParams, Sign, TransactionAndSign, TxnReceipt } from "../types";
+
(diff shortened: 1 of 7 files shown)


packages/web/src/lib/executor.ts
```diff
@@ -0,0 +1,58 @@
+import algosdk, { Account, SignedTransaction, Transaction } from "algosdk";
+
+import { ExecParams, Sign, TransactionAndSign, TxnReceipt } from "../types";
+
+export interface Executor {
+	/**
+	 * Execute single transaction or group of transactions (atomic transaction)
```
robert-zaremba, line 7:
```suggestion
	 * Executes (creates, signs and sends) a single transaction or group of transactions (atomic transaction)
```

robert-zaremba (approved):

